PR #55, opened 2026-04-15
feat(engine): scope dedup measurement to Rocky-managed tables

hugocorreia90:
## Summary

- `rocky compact --measure-dedup` now defaults to scanning only tables that correspond to Rocky-managed models (replication target tables or transformation model targets), instead of all non-system tables in the warehouse
- A new `--all-tables` flag preserves the previous behavior for users who want to scan everything
- Adds a `scope` field (`"managed"` or `"all"`) to the `CompactDedupOutput` JSON schema so downstream consumers can see which mode was used

## Details

For **replication pipelines**, managed tables are resolved by running source discovery through the configured adapter and mapping each discovered source table through the schema-pattern + template resolution to produce target `TableRef`s.

For **transformation pipelines**, managed tables are loaded from the models directory and their sidecar TOML target configs (catalog, schema, table).

The `enumerate_tables()` warehouse scan still runs, but the result is filtered (via case-insensitive match) to only include tables present in the resolved managed set. This approach avoids changing the SQL query itself while ensuring the measurement stays focused on tables Rocky actually manages.

## Test plan

- [x] `cargo clippy --all-targets -- -D warnings` passes clean
- [x] `cargo test` full suite passes (1700+ tests, 0 failures)
- [x] Existing E2E test updated to pass `all_tables=true` (test creates tables manually without Rocky models)
- [x] New E2E test `compute_measure_dedup_scoped_to_managed_tables` validates that a transformation pipeline with 2 models scans only those 2 tables (not the 3rd unmanaged table), and that `--all-tables` scans all 3
- [x] Schema + Pydantic + TypeScript bindings regenerated via codegen

### DIFF
--- a/editors/vscode/src/types/generated/compact_dedup.ts
+++ b/editors/vscode/src/types/generated/compact_dedup.ts
@@ -42,6 +42,10 @@ export interface CompactDedupOutput {
    */
   raw: DedupSummary;
   /**
+   * Table scope used for the measurement: - `"managed"` — only tables that correspond to Rocky-managed models. - `"all"` — all non-system tables in the warehouse (legacy behavior, activated by `--all-tables`).
+   */
+  scope: string;
+  /**
    * Dedup computed after excluding Rocky-owned metadata columns. The gap between `raw` and `semantic` is "dedup we'd unlock if we ignored Rocky's own per-row metadata."
    */
   semantic: DedupSummary;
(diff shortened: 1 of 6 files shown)


hugocorreia90:
Closing — will re-implement on top of #54 (blake3) to avoid merge conflicts

PR #128, opened 2026-04-17
feat(engine): graceful Ctrl-C handling in rocky run

hugocorreia90:
## Summary

First Ctrl-C on `rocky run` now triggers graceful cleanup instead of a hard bail. Watermarks for completed tables are flushed, in-flight tables are marked `TableStatus::Interrupted` in the state store, partial JSON is emitted, and the binary exits with code 130. A second Ctrl-C hard-exits via a spawned watcher so the user has an out if cleanup itself hangs.

## Behaviour

| Phase | Pre-PR | Post-PR |
|---|---|---|
| 1st SIGINT during `rocky run` | `main.rs` outer `select!` → `anyhow::bail!`, exit 1, no state update | In-process handler flushes watermarks, writes `Interrupted` rows, returns sentinel → exit 130 |
| 2nd SIGINT | N/A | Spawned watcher calls `process::exit(130)` directly |
| `rocky run --resume-latest` after interrupt | Re-does completed tables (no `Success` barrier) | Skips completed tables, retries `Interrupted` ones |
| JSON consumers | `"tables_copied": N` only | `"interrupted": true` + `TableStatus::Interrupted` rows |

## What moved

- **`rocky_core::state::TableStatus::Interrupted`** + **`rocky_core::incremental::PartitionStatus::Interrupted`** — new enum variants for the state store. Both enums were non-exhaustive in match sites, so the additions drop in without breaking callers.
- **`rocky_cli::commands::Interrupted`** — `thiserror::Error` sentinel returned via `anyhow::Error::new(Interrupted)`. `main.rs` downcasts to it and calls `process::exit(130)`.
- **`RunOutput.interrupted: bool`** — always serialised. Pydantic + TypeScript regenerated via `just codegen`.
- **Internal `tokio::select!` in `run.rs`** — wraps both `semaphore.acquire_owned()` in the spawn loop and `join_set.join_next()` in the drain loop. First SIGINT arms a hard-exit watcher, sets `interrupted = true`, stops issuing new work. Pinned `ctrl_c()` future is shared across both loops; the `if !interrupted` guard prevents the already-resolved future from being polled twice.
- **`main.rs`** — outer `tokio::select!` + `shutdown_signal()` removed for the `Run` command specifically. Internal handler owns the signal so cleanup can run. SIGTERM for `run` falls through to the default Rust handling (follow-up if Kubernetes shutdown graceful paths are needed).

## Schema cascade

Ran `just codegen`. Regenerated:
- `schemas/run.schema.json`
- `integrations/dagster/src/dagster_rocky/types_generated/run_schema.py`
- `editors/vscode/src/types/generated/run.ts`

All committed. `codegen-drift` CI should pass.

## Known limitation (out of scope, filed as follow-up)

A SIGINT mid-`ALTER TABLE` (schema drift evolution) can leave the target half-evolved — the in-flight statement can't be rolled back through the generic adapter path. Handling this requires schema-drift-aware cancellation (detect mid-evolve state on next run and either resume or roll forward). Not blocking this PR.

## Test plan

- [x] `cargo test -p rocky-core --lib` — 920/920 pass (enum additions clean).
- [x] `cargo test -p rocky-cli` — 141/141 pass.
- [x] `cargo clippy -p rocky-core -p rocky-cli -p rocky-server` — clean.
- [x] `cargo check --workspace` — clean.
- [x] `just codegen` — bindings regenerated, diff is confined to `run` artefacts.
- [ ] **Manual**: start a long-running `rocky run` (playground POC with a handful of tables + artificial delay via a slow dispatcher).
  - Hit Ctrl-C after first table completes → exit 130, `rocky state` shows N `Success` + M `Interrupted`.
  - `rocky run --resume-latest` against the same state → only touches `Interrupted` tables, skips `Success`.
  - Hit Ctrl-C twice rapidly → process exits immediately (exit 130 via hard-exit watcher).
- [ ] **Follow-up**: automated e2e test (`engine/crates/rocky-cli/tests/e2e_cancel.rs`) — deferred; it requires a slow mock dispatcher harness that doesn't exist yet. Manual verification in this PR; the harness + test can land as a separate testing-infra PR.

### DIFF
--- a/editors/vscode/src/types/generated/run.ts
+++ b/editors/vscode/src/types/generated/run.ts
@@ -81,6 +81,10 @@ export interface RunOutput {
   excluded_tables: ExcludedTableOutput[];
   execution: ExecutionSummary;
   filter: string;
+  /**
+   * `true` when the run was cancelled by a SIGINT (Ctrl-C). Surfaced so orchestrators can distinguish "user interrupted" from "run failed". Tables that hadn't reached `Success` or `Failed` at interrupt time are recorded as `TableStatus::Interrupted` in the state store. Always serialised (even when `false`) so consumers don't have to treat its absence specially.
+   */
+  interrupted: boolean;
   materializations: MaterializationOutput[];
   metrics?: MetricsSnapshot | null;
   /**
--- a/engine/crates/rocky-cli/src/commands/mod.rs
+++ b/engine/crates/rocky-cli/src/commands/mod.rs
@@ -78,7 +78,7 @@ pub use optimize::run_optimize;
 pub use plan::plan;
 pub use playground::{run_playground, run_playground_with_template};
 pub use profile_storage::run_profile_storage;
-pub use run::{PartitionRunOptions, run};
+pub use run::{Interrupted, PartitionRunOptions, run};
 pub use run_dag_exec::run_with_dag;
 pub use seed::run_seed;
 pub use serve::run_serve;
--- a/engine/crates/rocky-cli/src/commands/run.rs
+++ b/engine/crates/rocky-cli/src/commands/run.rs
@@ -75,6 +75,33 @@ struct TableError {
     task_index: Option<usize>,
 }
 
+/// Sentinel error signalling that `rocky run` was cancelled by a SIGINT
+/// (first Ctrl-C). The outer binary (`rocky/src/main.rs`) downcasts this
+/// off the `anyhow::Error` chain and maps it to exit code 130 so callers
+/// (shell, CI, orchestrators) can distinguish "user cancelled" from a
+/// generic run failure. A second SIGINT bypasses this path entirely and
+/// calls `std::process::exit(130)` directly.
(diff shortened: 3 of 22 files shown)


hugocorreia90:
Follow-up commit 3c68e8b addresses two issues flagged during review:

1. **SIGTERM regression fix.** Removing the outer `shutdown_signal()` select in `main.rs` dropped SIGTERM handling (K8s pod eviction path). Restored inside `run.rs` via a `#[cfg(unix)]`-gated branch in both select blocks, matching the SIGINT cleanup exactly. Renamed the helper + `Interrupted` message to "shutdown signal" since either source reaches this path now.

2. **Stale dagster fixtures.** `RunOutput.interrupted` is a required field and the committed `fixtures_generated/*run*.json` files predated it, so `dagster-ci` would reject them. Regenerated via `just regen-fixtures`; 13 fixture files updated with `"interrupted": false`.

PR #304, opened 2026-04-30
feat(rocky-cli): add --verbose flag to rocky doctor

### LINKED ISSUE #21: Add --verbose flag to rocky doctor
`rocky doctor` runs health checks but currently outputs a summary. A `--verbose` flag should show detailed information for each check.

**Expected behavior:**
```bash
rocky doctor           # summary: ✓ config, ✓ adapter, ✓ state, ✓ credentials
rocky doctor --verbose # full details: config path, adapter version, state file size, credential type, etc.
```

**Where to look:**
- Doctor command: `engine/crates/rocky-cli/src/commands/doctor.rs`

mvanhorn:
## Summary

Adds a `--verbose` flag to `rocky doctor` (per #21). When passed, the human-readable output prints additional per-check context (config path, state DB path and size, adapter type and credential signal, pipeline names, and the configured state backend) on indented lines beneath each check. The default (non-verbose) output is unchanged, and the JSON output stays backwards-compatible because the new `HealthCheck.details` field is `#[serde(skip_serializing_if = "Vec::is_empty", default)]` and only ever populated when `--verbose` is set.

## Subproject(s) touched

- [x] `integrations/dagster/` (Python package, codegen only)
- [x] `editors/vscode/` (TypeScript extension, codegen only)
- [x] `schemas/` (canonical CLI JSON schema)

## Changes

In the engine, the `Doctor` command variant in `engine/rocky/src/main.rs` gains a `verbose: bool` field that's plumbed through to `commands::doctor`. The `doctor` function in `engine/crates/rocky-cli/src/commands/doctor.rs` takes that `verbose` flag as a new last parameter, and `HealthCheck` grows an optional `details: Vec<(String, String)>` field that each check populates when verbose is true. The human-readable rendering prints the details inline under each check; the JSON branch is unaffected because the new field serde-skips empty vectors. Each check populates the kind of detail the reporter cares about for that check: the `config` check adds the resolved config path, `state` adds the DB path and on-disk size in bytes, `adapters` adds an entry per adapter for type and credential signal (one of token, oauth_client, api_key, or none), `pipelines` lists each configured pipeline name, and the `state_sync` and `state_rw` checks add the backend identifier.

The schema and binding regen comes from running `just codegen` after the engine change. That updates `schemas/doctor.schema.json`, `editors/vscode/src/types/generated/doctor.ts`, and `integrations/dagster/src/dagster_rocky/types_generated/doctor_schema.py` to declare the new optional `details` field. Without that step the codegen-drift CI check would fire because the rocky JSON contract is the source of truth for the dagster and VS Code consumers; with it, both consumers see the new field as optional and existing payloads stay valid. The `npm install` step that ran during codegen also bumped `editors/vscode/package-lock.json` from 1.9.1 to 1.11.0, which is unrelated to this PR; I reverted that file so the only diff here is the doctor surface area.

## Test Plan

From `engine/`:

- `cargo build --package rocky-cli --package rocky` — clean.
- `cargo clippy --package rocky-cli --package rocky --all-targets -- -D warnings` — clean.
- `cargo fmt --check` — clean.
- `cargo test --package rocky-cli` — 304 passed, 0 failed.

Codegen (from repo root): `just codegen` ran cleanly. The three regenerated files in this PR are the only meaningful diff.

The change is additive. Existing scripts that read `rocky doctor --output json` continue to work because `HealthCheck.details` only appears in the payload when `--verbose` is also set, and the field is optional in the regenerated TS/Pydantic types.

## Checklist

- [x] If this is a CLI JSON schema or DSL syntax change, all consumers (`integrations/dagster/`, `editors/vscode/`) are updated in this same PR

### Engine (`engine/`)

- [x] Codegen-only change. The regenerated `doctor_schema.py` adds an optional `details: list[list[str]] | None = None` field.

- [x] Codegen-only change. The regenerated `doctor.ts` adds an optional `details?: [string, string][]` field.

Closes #21

### DIFF
--- a/editors/vscode/src/types/generated/doctor.ts
+++ b/editors/vscode/src/types/generated/doctor.ts
@@ -24,6 +24,7 @@ export interface DoctorOutput {
  * A single health check result.
  */
 export interface HealthCheck {
+  details?: [string, string][];
   duration_ms: number;
   message: string;
   name: string;
--- a/engine/crates/rocky-cli/src/commands/doctor.rs
+++ b/engine/crates/rocky-cli/src/commands/doctor.rs
@@ -21,6 +21,8 @@ pub struct HealthCheck {
     pub status: HealthStatus,
     pub message: String,
     pub duration_ms: u64,
+    #[serde(skip_serializing_if = "Vec::is_empty", default)]
+    pub details: Vec<(String, String)>,
 }
 
 /// Doctor output structure.
@@ -38,6 +40,7 @@ pub async fn doctor(
     state_path: &Path,
     output_json: bool,
     check_filter: Option<&str>,
+    verbose: bool,
 ) -> Result<()> {
     let mut checks: Vec<HealthCheck> = Vec::new();
     let mut suggestions: Vec<String> = Vec::new();
@@ -45,6 +48,11 @@ pub async fn doctor(
     // 1. Config validation
     if should_run("config", check_filter) {
         let start = Instant::now();
+        let details = if verbose {
+            vec![("path".into(), config_path.display().to_string())]
+        } else {
+            Vec::new()
+        };
         match rocky_core::config::load_rocky_config(config_path) {
             Ok(_) => {
                 checks.push(HealthCheck {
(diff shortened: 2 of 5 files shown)


hugocorreia90 (approved):
Thank you @mvanhorn ! ❤️

PR #321, opened 2026-05-01
fix(engine,dagster): validate archive model name + mark excluded_tables optional

hugocorreia90:
## Summary

Two bugs surfaced by an engine + dagster audit:

1. **`rocky archive` SQL injection (HIGH)** — `generate_archive_sql` in `engine/crates/rocky-cli/src/commands/archive.rs` interpolated the unvalidated `--model` argument (and per-table FQNs from `--catalog` scope resolution) directly into `DELETE FROM {target}` and `VACUUM {target}`, violating engine/CLAUDE.md's "never use `format!()` to build SQL with untrusted input" rule. Each dot-separated segment is now validated through `rocky_sql::validation::validate_identifier` before interpolation; the function returns `Result`. Added `test_generate_archive_sql_rejects_injection`.

2. **`DiscoverOutput.excluded_tables` / `RunOutput.excluded_tables` schema mismatch (MEDIUM)** — both fields had `#[serde(skip_serializing_if = "Vec::is_empty")]` (CLI omits the field when empty) but no `#[serde(default)]`, so the generated JSON schemas marked them required. The autogenerated Pydantic `DiscoverOutput`/`RunOutput` would refuse real CLI output any time no tables were filtered. Added `#[serde(default)]` and cascaded the change through `schemas/`, the dagster Pydantic models in `types_generated/`, and the vscode TypeScript interfaces in `editors/vscode/src/types/generated/`. (Cascaded by hand because the sandbox has no crates.io network; `just codegen` produces the same diff.)

## Test plan

- [x] Dagster: `uv run pytest -x` → 500 passed
- [ ] Engine: `cargo test -p rocky-cli` (could not run in the sandbox — no crates.io network; changes are surgical and the new injection test mirrors the existing `test_format_table_ref_rejects_injection` in `rocky-sql`)
- [ ] CI `codegen-drift` should be green: hand-edited schemas/Pydantic/TS match what `just codegen` would emit from the new `#[serde(default)]`

## What was deliberately not fixed

- **Regex pattern interpolation in `regex_match_predicate` (Databricks/Snowflake/DuckDB/BigQuery)** — the trait doc in `rocky-core/src/traits.rs:444` documents callers must pre-validate via `tests::validate_regex_pattern`, which rejects `'`, `` ` ``, `;`. Both internal callers (`tests.rs:473`, `quarantine.rs:360`) do this. Adding redundant validation in 4 dialects would mask future contract violations rather than catch them at the boundary.
- **`RESOURCE_DOES_NOT_EXIST` in Databricks `is_transient`** — borderline; placement looks deliberate for Unity Catalog metadata propagation after `CREATE TABLE/SCHEMA/CATALOG`. Worth a product call; not flipping unilaterally.

### DIFF
--- a/editors/vscode/src/types/generated/discover.ts
+++ b/editors/vscode/src/types/generated/discover.ts
@@ -17,7 +17,7 @@ export interface DiscoverOutput {
   /**
    * Tables filtered out of `sources` because they were reported by the discovery adapter but do not exist in the source warehouse. Same shape as `RunOutput.excluded_tables` so consumers can use one parser. Empty when nothing was filtered.
    */
-  excluded_tables: ExcludedTableOutput[];
+  excluded_tables?: ExcludedTableOutput[];
   /**
    * Sources the discovery adapter attempted to fetch metadata for and failed (transient HTTP error, timeout, rate-limit budget exhausted, auth blip). Their absence from `sources` does NOT mean they were removed upstream — consumers diffing against a prior run must treat failed sources as "unknown state, do not delete." Empty when discovery completed cleanly. See FR-014.
    */
--- a/editors/vscode/src/types/generated/run.ts
+++ b/editors/vscode/src/types/generated/run.ts
@@ -92,7 +92,7 @@ export interface RunOutput {
   /**
    * Tables that the discovery adapter reported as enabled but that do not exist in the source warehouse (e.g. Fivetran has them configured but hasn't synced them, or they carry the `do_not_alter__` broken-table marker). Surfaced so orchestrators can flag the gap in their UIs instead of silently dropping the assets.
    */
-  excluded_tables: ExcludedTableOutput[];
+  excluded_tables?: ExcludedTableOutput[];
   execution: ExecutionSummary;
   filter: string;
   /**
--- a/engine/crates/rocky-cli/src/commands/archive.rs
+++ b/engine/crates/rocky-cli/src/commands/archive.rs
@@ -3,7 +3,8 @@
 use std::collections::BTreeMap;
 use std::path::Path;
 
-use anyhow::Result;
+use anyhow::{Context, Result};
+use rocky_sql::validation::validate_identifier;
 
 use crate::output::{ArchiveOutput, ArchiveTableEntry, ArchiveTotals, NamedStatement, print_json};
 use crate::scope::resolve_managed_tables_in_catalog;
@@ -19,7 +20,7 @@ pub fn run_archive(
 ) -> Result<()> {
     // Parse the older_than duration (e.g., "90d", "6m", "1y")
     let days = parse_duration_days(older_than)?;
-    let statements = generate_archive_sql(model, days);
+    let statements = generate_archive_sql(model, days)?;
 
     if output_json {
         let typed_statements: Vec<NamedStatement> = statements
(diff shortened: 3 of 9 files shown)


hugocorreia90:
## Review

Both fixes are well-scoped, surgical, and the test coverage on the injection path mirrors the existing `validate_identifier` pattern. Approving in spirit but blocking on one parity gap.

### Blocker — `compact.rs` has the same SQL injection

`engine/crates/rocky-cli/src/commands/compact.rs:1114` `generate_compact_sql(model: &str, ...)` is the exact pattern this PR fixes in `archive.rs`:

```rust
format!("OPTIMIZE {model} WHERE true\n  -- target file size: {target_size_mb}MB"),
format!("VACUUM {model} RETAIN 168 HOURS"),
```

Same `--model` CLI arg, same `--catalog`-scoped FQN feed (#315 added the catalog scope to compact and archive together — the catalog-resolution path in `scope::resolve_managed_tables_in_catalog` lowercases but doesn't validate identifiers, which the PR body itself flags as the archive vector). Compact already imports `validate_identifier` at line 555, so the project's own pattern is to validate before format — the OPTIMIZE/VACUUM path bypasses it.

Recommend folding the same `validate_identifier`-per-segment fix into this PR rather than opening a follow-up. Shipping a SQL-injection fix that leaves a sibling injection on the same audit's blast radius is awkward, and the diff mirrors archive's almost line-for-line (~10 lines).

There's also `compact.rs:1014` `format!("SELECT * FROM {table_name} LIMIT {CALIBRATE_SAMPLE_ROWS}")` on the calibrate-bytes path. `table_name` comes from `pick_sample_tables` → `DedupStats.{top_dedup_pairs, per_table}` → ultimately `rocky.toml` pipeline config (same trust boundary as archive's catalog scope). Worth validating each segment for defense-in-depth at the same time.

### Non-blocking nits (one-liners)

- **Pydantic shape `list | None = None`**: from the Rust side `#[serde(default)]` on `Vec<T>` always yields an empty Vec, never `None`. Confirmed no current consumer hits this — the soft-swap `DiscoverResult.excluded_tables` has its own `= []` default. Follow-up: add `#[schemars(default)]` so the codegen contract matches the Rust semantics.
- **~12 other `Vec` fields in `output.rs`** carry `skip_serializing_if = "Vec::is_empty"` without `#[serde(default)]` — same asymmetry, may or may not be biting. Worth a follow-up sweep but the PR is correctly scoped to the two bugs the audit surfaced.
- **BigQuery hyphenated catalogs** (e.g. project IDs like `rocky-sandbox-hc-test-63874`) would fail `validate_identifier` on the catalog segment. Archive isn't supported on BigQuery today so not a regression, but if archive lands on BQ later, the catalog segment will need `validate_gcp_project_id` while schema/table stay on `validate_identifier`. Forward-compat note.

The "What was deliberately not fixed" section is good — `regex_match_predicate`'s contract-at-the-boundary stance is sound, no need to revisit.

hugocorreia90:
Pushed [784de4b](https://github.com/rocky-data/rocky/commit/784de4b) addressing the blocker — `generate_compact_sql` now validates each dot-separated segment of `model` via `validate_identifier` and returns `Result`; both callers (`run_compact` + `run_compact_catalog`) propagate. The `--measure-dedup --calibrate-bytes` path validates `sample_tables` upfront before the `SELECT * FROM {table_name}` interpolation. Added `test_generate_compact_sql_rejects_injection` mirroring the archive test.

Locally verified: `cargo test -p rocky-cli --lib` → 320 passed (including PR 321's previously-unrun archive injection tests), `cargo clippy -p rocky-cli -- -D warnings` clean, `cargo fmt --check` clean.

Non-blocking nits left for follow-up as agreed in the review.

PR #391, opened 2026-05-04
fix(engine/rocky-snowflake): narrow WAREHOUSE transient check

hobostay:
## Summary
- The Snowflake connector's `is_transient` function uses `msg.contains("WAREHOUSE")` to decide whether to retry failed statements. This is overly broad — it matches permanent configuration errors like "WAREHOUSE not found" or "Invalid WAREHOUSE name", causing unnecessary retries on errors that will never succeed.
- Narrowed to `"WAREHOUSE IS STARTING"` to match only the transient case where a warehouse is being resumed from suspend. This aligns with the Databricks connector which already uses this specific substring.

## Test plan
- [ ] Verify existing unit tests in `engine/crates/rocky-snowflake/` still pass (`cargo test -p rocky-snowflake`)
- [ ] Confirm that `"WAREHOUSE IS STARTING"` still triggers retry behavior
- [ ] Confirm that `"WAREHOUSE not found"` / `"Invalid WAREHOUSE"` are no longer treated as transient

### DIFF
--- a/engine/crates/rocky-snowflake/src/connector.rs
+++ b/engine/crates/rocky-snowflake/src/connector.rs
@@ -507,7 +507,7 @@ fn is_transient(err: &ConnectorError) -> bool {
             let msg = message.to_uppercase();
             msg.contains("THROTTL")
                 || msg.contains("TEMPORARILY_UNAVAILABLE")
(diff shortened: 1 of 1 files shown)


hugocorreia90:
Thank you!

PR #392, opened 2026-05-04
fix(engine,vscode,docs): unbreak rocky import-dbt --output clash

hobostay:
## Summary
- The `importDbt` command in the VS Code extension passes `--output` twice: once for the user-selected output directory and once with value `"json"`. Since `--output` expects a directory path, the second value overwrites the first, causing the imported models to be written to a directory literally named `json` instead of the user's chosen path.
- The JSON output format should be requested via the global `--json` flag (as used by all other commands in the extension), not `--output json`.

## Test plan
- [ ] Run the "Rocky: Import dbt project" command from the VS Code command palette
- [ ] Select a dbt project, specify a custom output directory (e.g., "my-models"), and run the import
- [ ] Verify models are written to the specified directory, not to a directory named "json"


hugocorreia90 (approved):
Thank you!

PR #393, opened 2026-05-04
fix(vscode): use relative path as tree/test item ID to prevent duplicates

hobostay:
## Summary
- `ModelTreeItem` (models view) and `modelIdFor()` (test explorer) both use `path.basename(uri.fsPath, ".rocky")` as the tree/test item ID. When two `.rocky` files in different directories share the same basename (e.g., `models/staging/customer.rocky` and `models/marts/customer.rocky`), VS Code deduplicates by ID and silently drops one item.
- Changed both to use `vscode.workspace.asRelativePath(uri)` which is unique per file within a workspace.

## Test plan
- [ ] Create two `.rocky` files with the same name in different directories under `models/` (e.g., `models/staging/customer.rocky` and `models/marts/customer.rocky`)
- [ ] Verify both appear in the Models tree view
- [ ] Verify both appear in the Test Explorer panel

### DIFF
--- a/editors/vscode/src/testExplorer.ts
+++ b/editors/vscode/src/testExplorer.ts
@@ -1,3 +1,4 @@
+import * as path from "path";
 import * as vscode from "vscode";
 import { runRockyJson, RockyCliError } from "./rockyCli";
 import type { TestResult } from "./types/rockyJson";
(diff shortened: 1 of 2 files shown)


hugocorreia90 (approved):
Thanks again! ❤️

PR #619, opened 2026-05-20
fix(engine): Snowflake replication-merge column threading + dialect parity

hugocorreia90:
## Summary

Replication's `strategy = "merge"` was unusable against Snowflake. The runner dispatched `ColumnSelection::All` to SQL-gen, which emitted `UPDATE SET *`; Databricks / BigQuery / DuckDB accept that shorthand, Snowflake rejects it. Transformation users had an escape hatch via the model TOML sidecar's `update_columns`; replication had none.

**Approach (Option B):** resolve `ColumnSelection::All` at runtime against the discovered source schema BEFORE the IR reaches SQL-gen. `process_table` already awaits `describe_table(&source_table)` for drift detection — `source_cols` is in scope and free; no new config surface, no extra round-trip. Picked over Option A (`update_columns` on `ReplicationPipelineConfig`) because the resolution is mechanical from data the runner already holds, and a new config field would push the resolution onto the user.

Live-verification surfaced two adjacent Snowflake-MERGE bugs the original spec didn't call out — without them, the column-threading fix alone doesn't unblock Snowflake replication-merge. Both are receipts from `ROCKY_TEST`:

1. **`INSERT *` rejected.** Snowflake's `WHEN NOT MATCHED` branch needs `INSERT (col1, ...) VALUES (s.col1, ...)`. The dialect comment claimed "explicit column list" but emitted `INSERT *`. Snowflake error: `syntax error line 7 at position 29 unexpected '*'`.
2. **Unquoted identifiers folded to UPPER.** Source tables created with lowercase double-quoted column names (the common case for Fivetran-loaded raw tables) don't resolve under `t.id = s.id` because Snowflake folds the unquoted `id` to `ID`. Snowflake error: `invalid identifier 'ID'`. Fix mirrors `format_table_ref`'s existing `quote_identifier` policy — every MERGE-clause identifier is double-quoted.

Split into two scoped commits so each layer is independently bisectable:
- `1e32d853` — dialect: `INSERT (...) VALUES (...)` + quoted identifiers
- `d9efbbaa` — runner: `resolve_merge_update_columns` + live `#[ignore]` test

## Why the existing transformation escape-hatch wasn't enough

`update_columns` on a model sidecar TOML doesn't apply to replication pipelines — replication has no per-table model files, so users would need a new config surface to provide the column list. Resolving from the runner-held source schema avoids that ask entirely.

## Test plan

- [x] `cargo fmt --all` clean
- [x] `cargo clippy --all-targets --all-features -p rocky-snowflake -p rocky-core -p rocky-cli -- -D warnings` clean
- [x] `cargo test -p rocky-snowflake -p rocky-core -p rocky-cli` — 554 + 144 cli/snowflake unit tests pass; 3 new `resolve_merge_update_columns_*` tests added
- [x] `cargo test -p rocky-cli --test ir_golden` — 2 snapshots updated (`03-replication-merge/snowflake.sql`, `05-transformation-merge/snowflake.sql`) reflect the new quoted-INSERT form
- [x] `cargo build --workspace` clean
- [x] `just codegen` — zero drift (no `*Output` struct touched)
- [x] **Live-verify** against Snowflake `ROCKY_TEST` (statement handle `01c48118-0001-c951-0002-6a7a0006d302`):
  - `live_merge_with_explicit_columns_is_idempotent` — RED on `main` (Snowflake rejects `INSERT *`); GREEN on this branch. 3 back-to-back merges leave target row count stable at 3.
  - `live_merge_with_update_set_star_is_rejected_by_snowflake` — pins Snowflake's parser-level rejection of the broken pre-fix SQL.

### DIFF
--- a/engine/crates/rocky-cli/src/commands/run.rs
+++ b/engine/crates/rocky-cli/src/commands/run.rs
@@ -4544,7 +4544,13 @@ fn accumulate_bytes(acc: Option<u64>, next: Option<u64>) -> Option<u64> {
 ///   reading keys from `resolved_merge_keys()` (`merge_keys` falls back to
 ///   `merge_keys_fallback`). `merge` requires keys; absence is treated as
 ///   a defensive `anyhow!` because `validate_replication_strategies` should
-///   have rejected it at config load time.
+///   have rejected it at config load time. The `ColumnSelection::All`
+///   placeholder is replaced by the explicit source-schema column list in
+///   `process_table` BEFORE the IR reaches SQL-gen — Snowflake's MERGE
+///   rejects `UPDATE SET *`, so threading discovered columns is required for
+///   adapter parity. This function stays warehouse-free; the resolution
+///   happens at the call site where `describe_table(&source_table)` is
+///   already awaited for drift detection.
 /// Anything else → `FullRefresh` (backwards-compatible fallback; see follow-up
 ///   note in PR body about tightening unknown strategy values).
 ///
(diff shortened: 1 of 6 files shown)


hugocorreia90:
Force-pushed a rebase onto current main (was forked from 7b4c4410 before PRs #617 + #618 merged). Pre-rebase diff vs main showed those PRs' files as deletions, which would have reverted them at squash-merge time. Post-rebase diff is just the 6 files this fix actually touches. Build + lib tests green; live `#[ignore]` test logic unchanged.

PR #620, opened 2026-05-20
feat(rocky-server): AI LSP code actions for W004 + .rocky DSL parse errors

hugocorreia90:
## Summary

Extends the deferred AI code-action surface (introduced for `E010` / `E013`) to two more diagnostic codes, sharing the existing `code_action_resolve` deferred-resolve mechanism. `code_action_resolve` now dispatches on the `kind` discriminator of the JSON payload so each arm carries its own typed payload while sharing the resolve channel — the contract arm's wire shape is unchanged.

| Diagnostic | Arm | Payload variant | Suggested-fix mechanism |
|---|---|---|---|
| `W004` | Unresolved classification tag on a column | `AiClassificationActionData` (tag, column, model, `rocky.toml` URI + text) | LLM emits a `[mask.<tag>]` TOML block; resolve appends it to the end of `rocky.toml`. |
| `E028` (new) | `.rocky` DSL parse failure | `AiDslActionData` (parse-error message, raw `.rocky` source, file URI) | LLM emits a corrected `.rocky` pipeline, grounded in the DSL grammar surface pinned in the system prompt; resolve replaces the file body. |

### Prerequisite wiring (folded in)

Two LSP-side prerequisites had to land in the same change to satisfy the production-handler wiring requirement:

- **W004:** `recompile()` previously left `mask` + `allow_unmasked` empty, so the compiler's `check_classification_tags` was a no-op in the LSP path and W004 never reached `code_action`. The handler now loads `rocky.toml` from `<models_dir>.parent()` (the same root the schema-cache loader uses) and threads the values into `CompilerConfig`.

- **E028:** `ProjectError::RockyParse` aborts the compile pipeline before any `Diagnostic` is built. `recompile()` now re-parses each `.rocky` file in `models_dir` on a compile failure and publishes a synthetic LSP diagnostic with code `E028` so `code_action` has something to hook. The diagnostic span is derived from `rocky_lang::ParseError`'s byte offset.

### vscode cascade

**Follow-up, not bundled.** The LSP client passes `CodeAction.data` through verbatim (per the tower-lsp / vscode-languageclient default), so no client-side dispatch change is needed for either arm. The contract arm in #374 also shipped engine-only and works in vscode unchanged.

## Test plan

- [x] `cargo test -p rocky-server -p rocky-cli -p rocky-compiler -p rocky-lang` (76 server tests passing; +15 new)
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo fmt -- --check`
- [x] `just codegen` — no schema drift
- [x] `parse_classification_message` round-trips the compiler's W004 message shape against `check_classification_tags`
- [x] `w004_fires_through_compile_when_rocky_toml_declares_no_mask` — end-to-end test that drives the production `compile` pipeline with mask threading and asserts W004 surfaces with the expected message shape
- [x] `e028_synthetic_diagnostic_from_broken_rocky_file` — exercises `rocky_lang::parse` + `lsp_range_from_parse_error` against a real parse failure
- [x] Per-arm `code_action` emission tests (`ai_classification_action_carries_payload`, `ai_dsl_action_carries_payload`) and guard tests (skip on unparseable message; skip on non-`.rocky` URI)
- [x] Per-arm `code_action_resolve` disabled-without-API-key tests covering the AI fallback path with no live calls

No live AI-service calls in CI — every resolver short-circuits on the missing `ANTHROPIC_API_KEY` and the disabled reason is asserted.

### DIFF
--- a/engine/crates/rocky-server/src/lsp.rs
+++ b/engine/crates/rocky-server/src/lsp.rs
@@ -397,16 +397,26 @@ impl RockyLsp {
         )
         .await;
 
+        // Load `rocky.toml` so the W004 classification-tag check fires in
+        // the LSP. Without `mask` + `allow_unmasked` populated, the
+        // compiler's `check_classification_tags` is a no-op and the IDE
+        // never sees the warning. Project root = `models_dir.parent()`
+        // (the same assumption the schema-cache loader makes above).
+        // Missing or unreadable `rocky.toml` falls back to empty defaults.
+        let (mask, allow_unmasked) = dir_path
+            .parent()
+            .map(|root| root.join("rocky.toml"))
+            .and_then(|toml_path| rocky_core::config::load_rocky_config(&toml_path).ok())
+            .map(|c| (c.mask, c.classifications.allow_unmasked))
+            .unwrap_or_default();
+
         let config = CompilerConfig {
-            models_dir: dir_path,
+            models_dir: dir_path.clone(),
             contracts_dir: None,
             source_schemas,
             source_column_info: HashMap::new(),
-            // W004 is gated on a loaded RockyConfig; the LSP init path
-            // doesn't currently hold one, so leave the defaults empty.
-            // Follow-up: thread `rocky.toml` through initialize_params
-            // so IDEs surface unresolved-classification warnings live.
-            ..Default::default()
+            mask,
+            allow_unmasked,
         };
 
         match rocky_compiler::compile::compile(&config) {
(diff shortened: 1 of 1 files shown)


hugocorreia90:
Force-pushed a rebase onto current main (was forked from 7b4c4410 before PRs #617 + #618 merged). Pre-rebase diff vs main showed those PRs' files as deletions, which would have reverted them at squash-merge time. Post-rebase diff is just `engine/crates/rocky-server/src/lsp.rs`. Build + 76 rocky-server tests green.

PR #630, opened 2026-05-22
perf(engine/rocky-lang): expose parallel parse_all helper for corpus parsing

hugocorreia90:
## Summary

Adds a public `rocky_lang::parse_all` helper backed by rayon's work-stealing pool for parsing a corpus of `.rocky` sources in parallel. Centralises the par-iter strategy for read-only corpus callers (LSP, AI, future incremental front-ends) so the heuristics live in one place; the pre-existing parallel parse-and-lower loop in `rocky-compiler::project::load_rocky_models` is unchanged.

- New `pub fn parse_all<S: AsRef<str> + Sync>(&[S]) -> Vec<Result<RockyFile, ParseError>>` in `rocky-lang`. Preserves input order. Errors are not short-circuited — every source is attempted so callers can surface multiple parse failures in one pass.
- Falls back to a sequential pass below `PAR_THRESHOLD = 32` (mirrors `rocky-compiler/typecheck.rs::with_min_len(32)`). `.rocky` parsing is ~1-2 µs per typical model on an M-series core, so small corpora otherwise regress under `par_iter`.
- New criterion bench at `engine/crates/rocky-lang/benches/parse_corpus.rs` measures `parse_all` against a sequential `.iter().map(parse)` baseline at N=10/50/200 — acts as a regression guard.
- Lifts `rayon` to `[workspace.dependencies]` so the existing `rayon = "1"` decls in `rocky-core` and `rocky-compiler` flip to `rayon = { workspace = true }`.

## Benchmark results

Local on an M-series macOS host, criterion `parse_corpus` group, post-threshold tuning:

| N   | Sequential | Parallel | Speedup |
|-----|------------|----------|---------|
| 10  | ~18.6 µs   | ~18.5 µs | wash (sequential fallback) |
| 50  | ~91 µs     | ~90.7 µs | wash (threshold boundary)  |
| 200 | ~371 µs    | ~235 µs  | **~1.58x**                 |

## Notes for reviewers

- The compiler's `load_rocky_models` keeps its own `par_iter` because it interleaves disk I/O + lowering per file; threading those through `parse_all` would force two passes. The new helper targets corpus callers that only need parse results.
- `rocky-compiler::typecheck.rs` already used `par_iter().with_min_len(32)`; the new helper uses the same threshold for consistency.
- Doctest + 3 new unit tests in `parser.rs` cover empty input, error-non-short-circuit semantics, and a 64-element corpus that actually exercises `par_iter`.

## Test plan

- [x] `cargo test --workspace` — all green locally (two timing-flake retries unrelated to this change)
- [x] `cargo clippy --all-targets --all-features --workspace -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] `cargo bench --bench parse_corpus -p rocky-lang` — numbers above
- [ ] CI: `engine-ci.yml` (clippy + fmt + test)


hugocorreia90:
Closing this spike — its purpose was to measure whether parallelizing the `.rocky` parse loop yields a measurable win. The spike answered the question:

- **Yes, `par_iter` gives ~1.58x at N=200** (wash at smaller N; `PAR_THRESHOLD=32` required to avoid regression at small corpora).
- **But** `rocky-compiler::project::load_rocky_models` is already `par_iter`-parallelized — the production path Rocky users actually hit was already getting that win.
- The new `parse_all` helper sits with no production caller. Shipping a public API before a real consumer informs the shape risks fossilizing the wrong contract.

**Revival trigger:** a real LSP / AI consumer that needs bulk-parse — e.g., LSP "find references across workspace", AI corpus tools ("summarize this project across all models"), or a `rocky check-all` style CLI verb. Branch `perf/rayon-parse-corpus` preserved; the bench fixture + numbers are durable spike artifacts even if the helper itself never ships in this form.

PR #810, opened 2026-06-02
docs: refresh competitive positioning for dbt Core v2.0 / Fusion

hugocorreia90:
## What

On 2026-06-01 dbt Labs open-sourced the Fusion runtime as **dbt Core v2.0** (Rust, Apache 2.0, alpha). This refreshes Rocky's competitive copy to match current reality and removes several claims that had gone stale — or were wrong in Rocky's favor — across the README, docs, and the benchmark comparison.

## Key corrections

- **dbt status** — "public beta" → the Fusion runtime is now dbt Core v2.0 (Apache 2.0); Fusion is the recommended distribution, extending Core with SQL comprehension.
- **Static analysis** — Fusion runs it by default (baseline / warn-only); only the build-failing `strict` mode is opt-in. (Earlier copy said "opt-in", which was wrong.)
- **Column-level lineage** — not in the OSS dbt Core CLI (Fusion static analysis or the paid Catalog only); fixed the `dbt-core` "Yes" cells.
- **`dbt clone`** — shipped in dbt Core 1.6 and in Fusion since 2025; dropped the "coming soon" claim. The differentiator is Rocky's first-class named-branch primitive, not the absence of clone.
- **Cross-project contracts** — dbt Mesh ships them in OSS (the seamless cross-project `ref` is Enterprise-gated). Rocky's contracts are intra-project, so "cross-team contracts" is no longer claimed as a Rocky differentiator (matches `architecture-of-trust.md`).
- **SQLMesh** — it analyzes statically via SQLGlot (not "runtime-ish") and ships virtual environments / plan-apply / column-level lineage; those are conceded, and the comparison now leads on the enforcement plane (OSS governance, build-failing budgets).
- **Cost** — scoped honestly: per-model cost on every run in the free CLI, warehouse-billed on BigQuery and a duration × DBU-rate estimate elsewhere; only the build-failing budget gate is absent from all dbt distributions (dbt's Cost Insights is paid and visibility-only).
- **Replay / branches** — aligned to `architecture-of-trust.md`: schema-prefix branches and an inspectable, content-addressed run record ship today; warehouse-native clones and run re-execution are follow-ups. Updated the README adapter table and the roadmap "Shipped" section to match.
- **Tables** — added a `dbt-core` (1.x) vs `dbt-fusion` (v2.0 runtime + Fusion) column note; refreshed open-source / manifest / orchestration cells (incl. a current `dagster-sqlmesh` entry). The standalone benchmark file keeps its April snapshot date behind a partial-refresh banner rather than asserting fresh currency.

## Notes for review

- Verified against current dbt docs (supported-features, static-analysis, model-contracts, Cost Insights) and the engine source for the cost-fidelity claim.
- **Brand wording:** the hero value-lists now read `content-addressed run records` instead of `replay`, since `rocky replay` inspects/verifies a record today and re-execution is on the roadmap. Isolated in its own commit so it's easy to revert if the original word is preferred.
- Deeper re-lead of the hero value-prop sentence (it still leads with branches / column lineage, now industry-common) intentionally left out as a brand call.
- Docs site builds clean (83 pages).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hugocorreia90:
## Post-open review round

Ran a comprehensive review of the full diff plus a repo-wide sweep. It caught and fixed:

- **Fusion strict-mode gating (factual correction).** dbt's per-mode static-analysis feature table shows column-level lineage *and* data-type/function-signature checking are `strict`-only; the default `baseline` mode is lighter and warn-only. The earlier "on by default" wording was wrong — corrected across the comparison page, introduction, engine README, and architecture-of-trust, and the CLL table cell now reads "strict mode".
- **Incomplete propagation.** The refresh hadn't reached the docs homepage, the canonical glossary `Replay`/`Content-addressed` definitions, the examples table, or the benchmarks page (which explained the Rust Fusion engine's overhead via "Python"). All now consistent with the rest of the copy.
- **Cost fidelity.** Named Databricks/Snowflake + DuckDB = 0 instead of "elsewhere", and aligned the architecture-of-trust cost section, which had been the stale outlier claiming cost is "BigQuery-only in practice".

Scope grew 7 → 11 files as a result.

### Deferred (out of dbt-positioning scope, pre-existing)
- Diagnostic-code range disagrees across files (`E001–E026` vs `E001–E027`; the engine actually reaches `E033`). Unrelated to this refresh — worth a separate cleanup pass.
- Two internal Rust doc-comments (`output.rs`, `apply.rs`) still say "deterministic replay". Verified they do **not** propagate into the generated TS / JSON / Pydantic bindings, so this is internal-only and safe to defer.

PR #836, opened 2026-06-04
feat(engine/rocky-core): recognize BigQuery type names in contract type-matching

hugocorreia90:
`warehouse_type_to_rocky` (used by the load contract gate to compare a landed column's warehouse type against the contract's declared type) had no arms for BigQuery's scalar type names. BQ `DESCRIBE` returns `INT64` / `FLOAT64` / `BYTES` / `DATETIME` / `BIGNUMERIC`, all of which normalized to `Unknown` — and an `Unknown` landed type matches any contract type. So the gate's **type** check was a silent no-op on BigQuery (presence + nullability were still enforced).

Add the BQ arms so the gate type-checks precisely on BigQuery:
- `INT64` → `Int64`
- `FLOAT64` → `Float64`
- `BYTES` → `Binary`
- `DATETIME` → `TimestampNtz` (BQ `DATETIME` is timezone-naive)
- `BIGNUMERIC(p,s)` → `Decimal { .. }` (joins the existing DECIMAL/NUMERIC/NUMBER prefix chain)

**Tested:** `test_warehouse_type_to_rocky` extended with INT64/FLOAT64/BYTES/DATETIME/BIGNUMERIC assertions; the full rocky-core contract suite (23 tests) was *run*, not just compiled. fmt + clippy `--all-targets` clean. The DESCRIBE input strings are the exact names confirmed live in #833.

**Note:** the full end-to-end BQ contract-gate live-verify is blocked by a separate reachability bug — `load_with_contract_gate` strict-validates the target catalog as `[a-zA-Z0-9_]+`, which rejects hyphenated BigQuery project IDs before the gate runs. The non-gate `rocky load` path doesn't (it relies on the dialect's backtick quoting), so plain BQ loads already work. Relaxing the gate's catalog check to match the non-gate path is the fix; handled separately.

### DIFF
--- a/engine/crates/rocky-core/src/contracts.rs
+++ b/engine/crates/rocky-core/src/contracts.rs
@@ -268,25 +268,29 @@ pub fn warehouse_type_to_rocky(warehouse_type: &str) -> RockyType {
     match upper.as_str() {
         "BOOLEAN" | "BOOL" => RockyType::Boolean,
         "TINYINT" | "BYTE" | "SMALLINT" | "SHORT" | "INT" | "INTEGER" => RockyType::Int32,
-        "BIGINT" | "LONG" => RockyType::Int64,
+        // `INT64` / `FLOAT64` / `BYTES` / `DATETIME` are BigQuery's scalar names.
+        "BIGINT" | "LONG" | "INT64" => RockyType::Int64,
         "FLOAT" | "REAL" => RockyType::Float32,
-        "DOUBLE" | "DOUBLE PRECISION" => RockyType::Float64,
+        "DOUBLE" | "DOUBLE PRECISION" | "FLOAT64" => RockyType::Float64,
         "STRING" | "VARCHAR" | "TEXT" => RockyType::String,
-        "BINARY" => RockyType::Binary,
+        "BINARY" | "BYTES" => RockyType::Binary,
         "DATE" => RockyType::Date,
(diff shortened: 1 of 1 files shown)


hugocorreia90:
**Live-verified end-to-end on BigQuery** (sandbox, EU) on a combined tree of #836 (BQ type arms) + #837 (gate catalog fix). Path: local CSV → staging → `DESCRIBE` → contract validate → promote/drop.

**Pass-probe** — contract `id BIGINT, name STRING`; `id` lands as `INT64`:
- Gate *ran* (hyphenated catalog `…-63874` accepted), staging loaded 3 rows, `"contract": {"passed": true}`, promoted → `contract_ok_probe` (verified 3 rows on BQ).

**Bad-probe** — contract `id DATE, name STRING`; `id` lands as `INT64` → exit 1:
```json
"violations": [{ "rule": "required_column_type", "column": "id",
  "message": "column 'id' has type 'INT64' (INT64), expected 'DATE' (DATE)" }]
```
- Staging loaded, `DESCRIBE` returned `INT64`, gate closed → `contract_bad_probe` **never created**, staging table dropped.

The bad-probe is the unifying check: it fails *only* if (a) the catalog fix made the gate reachable, (b) staging was loaded + described, and (c) the arms resolve `INT64`→`Int64` (≠ `Date`). Without the arms, `INT64`→`Unknown` matches anything and it would have wrongly **passed**.

PR #966, opened 2026-06-23
chore(deps): upgrade object_store 0.11 → 0.14 (supersedes #964)

hugocorreia90:
## What
Upgrade `object_store` 0.11.2 → 0.14.0. **Supersedes #964** (dependabot's version-only bump, which doesn't compile).

## Why #964 couldn't merge
object_store 0.14 moved `get`/`put`/`delete`/`head` off the `ObjectStore` trait onto a new `ObjectStoreExt` extension trait. Dependabot bumps the version but can't adapt the code, so #964 fails to compile (Clippy/Test/build all red).

## Changes
- `ObjectStoreExt` import in the 5 files calling those convenience methods.
- `Path::child` → `Path::join` (child deprecated; **byte-identical** re-encoding — verified against both vendored versions, so fivetran state-cache keys are unchanged).
- **Pinned `.with_conditional_put(S3ConditionalPut::ETagMatch)`** on the wrapper's S3 builder. Surfaced by review: object_store's S3 conditional-put default flipped `None`(→`NotImplemented`) → `ETagMatch`(→`If-None-Match: *`) between 0.11 and 0.14, so the wrapper's `put_if_not_exists` was *latently broken on real S3* under 0.11 (it only worked through `run_content_addressed.rs`'s explicit ETagMatch). Pinning makes the concurrent-writer guarantee explicit + version-independent. The active ETagMatch create path is byte-identical across versions.

## Verified (compile-time + InMemory)
- `clippy --all-targets --all-features -D warnings` clean; affected-crate tests green (**2202 passed, 0 failed**); rustfmt clean.
- Conditional-put semantics source-proven behavior-preserving (active path identical; rocky matches `Error::AlreadyExists`).
- The dual-provider crypto-panic fix (`install_default(aws_lc_rs)` before any TLS) is intact in both binaries.

## ⚠️ DRAFT — needs a live S3 round-trip before merge
object_store 0.14 swaps its **network transport**: reqwest 0.12→0.13 (sole-consumed by object_store), direct `aws-lc-rs`, adds `rustls-platform-verifier`, bumps `quick-xml` 0.37→0.40 (ListObjects/multipart XML). These run **only against a real S3/GCS/Azure endpoint** — no unit test opens a socket (all InMemory), so compile + the 2202 green tests **cannot certify the new TLS/XML stack**. Same blind spot as the prior dual-provider TLS panic that green CI missed.

**Required before merge:** one live S3 `put` + `list` + `get` round-trip (env-var AWS creds + `ROCKY_TEST_S3_BUCKET`/`ROCKY_TEST_S3_PREFIX`). `list` is load-bearing — the only path exercising quick-xml 0.40; any op forces the new cert-verification handshake.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -669,15 +669,6 @@ dependencies = [
 
 [[package]]
 name = "block-buffer"
-version = "0.10.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3078c7629b62d3f0439517fa394996acacc5cbc91c5a20d8c658e77abd503a71"
-dependencies = [
- "generic-array",
-]
-
-[[package]]
-name = "block-buffer"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d2f6c7dbe95a6ed67ad9f18e57daf93a2f034c524b99fd2b76d18fdfeb6660aa"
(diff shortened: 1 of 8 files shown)


hugocorreia90:
Rebased onto 1.55.0 `main` (no source conflicts; `Cargo.lock` regenerated to resolve object_store 0.14 against the 1.55.0 graph). Merging on compile + InMemory verification per owner decision. **The live-S3 round-trip gate was deferred** — the `gold-engine-dev` AWS creds were expired at merge time, so object_store 0.14's new TLS/XML network stack is NOT certified against real S3 here. Recommend exercising an S3-touching path (content-addressed write / fivetran state-cache / `uniform_writer_live`) on the next deploy or whenever creds refresh. build / clippy / fmt / InMemory tests / codegen all green.

PR #1024, opened 2026-07-03
feat(engine): skip-unchanged source pruning for replication (prune_unchanged)

hugocorreia90:
## Problem

On a replication pipeline, `rocky run` executes work for **every** selected table on **every** run — the copy plus the `row_count` and `column_match` checks — even when the source table hasn't changed since the last successful copy. On high-cadence sources this per-table fixed overhead dominates cost: a run's expense is proportional to *how many tables exist*, not to *what changed*. Changing the per-table strategy doesn't fix it — `merge` still pays the per-table round-trips on an empty delta.

## What this adds

Opt-in **`[pipeline] prune_unchanged`** pruning. Before copying a table, the runner asks the adapter for a cheap source change-marker and compares it to the marker recorded at the last successful copy. If they match, the source is unchanged and the table is skipped entirely — no copy, no drift check, no data checks.

- **The marker** comes from a new `WarehouseAdapter::source_change_marker`. On Databricks it's a `DESCRIBE DETAIL`-derived `lastModified|numFiles|sizeInBytes`. Adapters without a cheap signal return `None`, in which case the runner always copies and never prunes. (`information_schema.tables.last_altered` was rejected — it advances on DDL only, not data writes; `DESCRIBE DETAIL.lastModified` advances on every Delta commit. Verified on a live workspace.)
- **A pruned table emits no materialization** and is recorded on `excluded_tables` with reason `unchanged_since_last_copy`. It reuses the existing free-form `reason` field, so there's no run-output schema change (no codegen cascade on the run output). Dagster's `satisfy_empty_outputs` already stamps the 0-row continuity signal for a selected-but-unmaterialized key, so a downstream consolidated asset isn't wholesale-skipped.

## Correctness over savings

- The marker is recorded **only on a successful copy**, and captured at the decision point (not re-read after the copy). So a failed run, or a source write that lands mid-copy, leaves the recorded marker behind and the table re-copies next run rather than skipping a change.
- A missing or non-Delta source yields no marker → the table always copies.
- A metadata-fetch error is treated as "no marker" (copy), so a transient failure never causes a false skip.
- An OPTIMIZE/VACUUM commit advances the marker with no logical data change → the table copies. That's the safe direction.
- Retries never prune — a table that failed attempt 1 is force-copied.

## Controls

- `prune_unchanged` defaults to **off** (opt in per pipeline — silently skipping copies is a behavior change).
- `rocky run --no-prune` forces a full pass regardless of config (e.g. after a manual target-side mutation).

## Storage

A new redb `source_markers` table holds the per-table last-copied marker, synced to remote state like watermarks so a distributed run reads a consistent value.

## Verification

- New unit test for the state round-trip (round-trips, overwrites, absent → `None`).
- New live Databricks test (`source_marker_live.rs`, `#[ignore]`-gated): the marker advances on a data write, stays identical when the source is untouched, and is absent for a missing table — the three properties the prune decision relies on. Verified green against a live workspace.
- Full engine unit suite green; `fmt` + `clippy --all-targets -D warnings` clean. No run-output schema change (only the config schema regenerated for the new `prune_unchanged` key).


hugocorreia90:
Follow-up (Phase 2b, separate PR): the Dagster integration needs pruned-table handling so the feature's continuity + check-preservation promises hold end-to-end — (1) don't re-emit a pruned table's declared checks as WARN placeholders (prior results should stand), (2) distinguish the pruned 0-row stamp reason from genuinely-absent, and (3) warn when a run reports `unchanged_since_last_copy` exclusions but `satisfy_empty_outputs` is off. This engine PR is safe to land ahead of it since `prune_unchanged` defaults off.

PR #1113, opened 2026-07-15
docs: add diagram-first "Rocky vs dbt, Visually" page

hugocorreia90:
## What

A new getting-started page, **Rocky vs dbt, Visually** (`/getting-started/rocky-vs-dbt/`), that explains the Rocky/dbt difference in plain language with ASCII diagrams, sitting just before the Feature Comparison in the sidebar (order 4.4).

The Feature Comparison page is deliberately dense and factual; this page is the intuition on-ramp for readers who want the mental model before the tables.

## Content notes

- Qualifies "dbt" the same way the comparison page does (Core 1.x vs Fusion vs platform) and contrasts against dbt Core first.
- Explicitly does **not** claim type-checking or column-level lineage as differentiators vs Fusion; the Fusion section leads with the enforcement plane (budgets that fail the build, drift gates, Apache-2.0 governance, portability lint, run record).
- The drift example links to the Schema Drift concept page and the comparison table rather than overstating behavior.
- Cross-links added from the Feature Comparison intro and the Introduction's "How Rocky compares" section.

## Verification

- `npm run build` passes (96 pages); the new page renders at `dist/getting-started/rocky-vs-dbt/`.
- All internal links and the `#schema-drift-detection` anchor verified against the built HTML.


hugocorreia90:
Follow-up commit 455bdd1f corrects factual claims flagged by an adversarial review of the initial version:

- **dbt Core was undersold.** The page now acknowledges parse-time `ref()` resolution (broken refs fail before running), enforced model contracts (dbt >=1.5), and `dbt build` interleaving tests with builds. The contrast is now scoped to what dbt Core genuinely lacks: comprehension of the SQL inside each model.
- **Rocky was oversold.** Drift is now correctly scoped to replication runs (source table vs maintained target, verified in the engine's run path) and the example is a column *type change*, not a rename — rename/drop handling isn't wired in production. Budget breaches default to `warn` and only fail the run with `on_breach = "error"`; replication quality checks are surfaced advisorily in typed output rather than being unconditional gates. Both are now stated accurately.
- Smaller fixes: typed-JSON claim scoped to commands supporting `--output json`; compiler framing no longer implies complete type knowledge (`Unknown` is a real outcome).

Docs build re-verified (96 pages).

hugocorreia90:
Second correction pass (43204820), this time on the drift claims, after tracing the full replication drift path in `run.rs::process_table`:

- **The drift section no longer promises unconditional resolution.** The diagram now shows all three real outcomes: safe widenings `ALTER`ed in place (rows preserved), incompatible changes dropping and re-copying the target from the current source (automatic and logged — and explicitly *not lossless* for targets that accumulate history the source has pruned), and the opt-in drift-governance gate refusing non-additive changes as a require-review failure with the target untouched.
- **`concepts/schema-drift.md` was stale and contradicted the code** (and the new page): its closing caveat claimed the safe-widening `ALTER` path isn't wired and only two actions are surfaced. The run path executes the widening and surfaces `alter_column_types`, `drop_and_recreate`, and `add_columns` — verified at the `AlterColumnTypes` branch and its `DriftActionOutput` construction. The caveat is replaced with the accurate action list, the new-columns bullet now reflects the `ALTER TABLE ADD COLUMN` behavior, and the governance gate (`auto_apply_additive_drift` + `schema_change.additive` grant) is documented.

Docs build re-verified (96 pages).

hugocorreia90:
Third correction pass (6549410a): the drift decision tree used **`INT → STRING` as the incompatible/rebuild example, but that change is actually on the default safe-widening allowlist** (`drift.rs::default_is_safe_type_widening` — "any numeric to STRING" — used by Databricks/DuckDB/Trino), so the engine ALTERs it in place rather than rebuilding. The example is now `BIGINT → INT` narrowing, which is unsafe on every dialect, and the prose notes the safe set is decided per warehouse dialect by what its `ALTER COLUMN` supports.

Also brought `concepts/schema-drift.md`'s allowlist table in line with the code: added the numeric → `STRING` row and documented the dialect overrides — Snowflake accepts only `NUMBER(p,s)` precision + `VARCHAR` length widening (integer types canonicalize to `NUMBER(38,0)` in `DESCRIBE TABLE`, so integer widening never surfaces as drift there), and BigQuery accepts only `INT64 → NUMERIC → BIGNUMERIC` (numeric → `STRING` is not assignable and falls through to full refresh).

Docs build re-verified (96 pages).

hugocorreia90:
Fourth correction pass (see latest commit): the allowlist documentation I added last round attributed the default table to Databricks/DuckDB/Trino as if the ALTER executed there. Verified adapter by adapter: only Snowflake and BigQuery override the allowlist/SQL form to match their warehouse; **on Databricks the default `ALTER COLUMN ... TYPE` is rejected by Delta** (type-widening table feature never enabled by Rocky; numeric → `STRING` never supported) **and on Trino the ANSI form isn't valid syntax** — so a classified-safe widening on those warehouses fails the table run with the warehouse error instead of evolving in place. DuckDB's default path I verified live (both `INT → BIGINT` and `INT → STRING` execute, rows preserved).

Docs now say exactly that: `schema-drift.md` gains a caution block for the Databricks/Trino execution gaps, and the comparison page scopes in-place evolution to "where the warehouse supports that ALTER" with the outcome set stated honestly (evolve, rebuild, refuse, or loud error — never silent).

The engine gap itself is now tracked as #1115 with per-adapter findings and suggested fixes (Databricks/Trino overrides + optional fallback-to-full-refresh on failed ALTER).

Docs build re-verified (96 pages).

hugocorreia90:
Fifth correction pass (33b3a107): the round-four summary said drift is "always detected — never a silent divergence", which is categorical. With the opt-in `prune_unchanged` optimization, a table whose source reports an unchanged change-marker is skipped entirely — copy, drift check, and data checks (`run.rs::process_table`) — so drift on such a table (e.g. an out-of-band change to the *target*) isn't re-evaluated until the source changes again.

Both pages now scope the claim to every table the run copies, with the prune skip called out explicitly. Docs build re-verified (96 pages).

hugocorreia90:
Sixth pass (f7cdfe81): the round-five wording still contained one categorical claim — "every table the run copies gets the drift check first." It doesn't hold under metadata errors: `process_table` swallows `describe_table` failures (`unwrap_or_default`), so an unreadable **target** is treated as absent and rebuilt with no check, and an unreadable **source** yields an empty column list — no drift result — and the copy proceeds unchecked that run.

Both pages now state the true scope: the check runs as part of a table's copy when the target exists, and the two bypasses are named explicitly (the opt-in `prune_unchanged` skip, and `DESCRIBE` failure degradation). Docs build re-verified (96 pages).

hugocorreia90:
Also pushed 9c25b709 — a proactive hardening pass from my own adversarial re-read of the final text, fixing the remaining absolutes before they get flagged: "tracks how **every column** flows" (the compiler tracks unresolvable types as `Unknown`), "classifies **every mismatch**" (removed source columns are ignored by `detect_drift`, so it's scoped to column-*type* mismatches), "content-addressed record of **every run**" (artifact content-addressing requires the S3-backed state store — it writes no artifacts on a creds-free DuckDB setup), the dbt bullet's "tests catch them **after** the model is built" (dbt 1.8+ unit tests run pre-build), and the schema-drift intro's unqualified "handled with ALTER TABLE (preserving data)" (now scoped to warehouses that support the change, consistent with the #1115 caution block).

Docs build re-verified (96 pages).

hugocorreia90:
a83ec7fa: the content-addressing caveat from the previous commit named the wrong subsystem — it said artifact content-addressing "requires the S3-backed state store", but the `[state]` backend only holds watermarks and the run record. The actual mechanism is the per-model object-store `storage_prefix` (the `_delta_log/` + Parquet prefix from [concepts/content-addressed](/concepts/content-addressed/)). The parenthetical now says the run record is always kept while artifact content-addressing applies to models with a `storage_prefix` configured, and links the concept page. Docs build re-verified (96 pages).

hugocorreia90:
One more (see latest commit): "the run record itself is always kept" was an unprovable absolute — it can't hold across crashed runs or unwritable state paths. The parenthetical now just states where run metadata lands (the state store's run history) versus what artifact content-addressing requires (a model-level `storage_prefix`), with no universal quantifier. Docs build re-verified (96 pages).

hugocorreia90:
Latest commit closes a full quantifier sweep I ran over both pages (grep for always/never/every/all/exactly/none, each hit audited against code or vetted docs). Two survived as attackable and are fixed: "matching **exactly** what their ALTER COLUMN accepts" → "aligned with" (BigQuery's `SET DATA TYPE` also accepts →`FLOAT64` coercions the override omits — the override is a deliberate subset, not an exact match), and "records checks and per-model cost **on every run**" → "in the run output". The remaining quantifiers all check out: dbt parse-time ref resolution, the `--output json` codegen contract, the per-model-cost phrasing (verbatim from the vetted Feature Comparison), the drift bypass list, and the Delta/Trino gap statements. Docs build re-verified (96 pages).

PR #1123, opened 2026-07-16
fix(sdk): reject unsupported server metrics options

### LINKED ISSUE #1122: Python SDK silently ignores metrics options in server mode
## Description

`RockyClient.metrics()` accepts `trend`, `column`, and `alerts`, but silently ignores all three when the client is configured with `server_url`.

In server mode, the method returns immediately after requesting the default-only endpoint:

```text
GET /api/v1/models/{model}/metrics
```

The endpoint intentionally mirrors default `rocky metrics <model>` behavior, so this issue is about the SDK silently accepting and discarding requested options—not about requiring the server endpoint to support them.

## Steps to Reproduce

```python
from unittest.mock import Mock

from rocky_sdk import RockyClient

body = (
    '{"version":"1","command":"metrics","model":"orders",'
    '"snapshots":[],"count":0}'
)

server_client = RockyClient(server_url="http://localhost:8080")
server_client._http_get = Mock(return_value=body)
server_client.metrics(
    "orders",
    trend=True,
    column="email",
    alerts=True,
)

cli_client = RockyClient()
cli_client._version_checked = True
cli_client.run_cli = Mock(return_value=body)
cli_client.metrics(
    "orders",
    trend=True,
    column="email",
    alerts=True,
)

print(server_client._http_get.call_args.args[0])
print(cli_client.run_cli.call_args.args[0])
```

Output:

```text
/api/v1/models/orders/metrics
['metrics', 'orders', '--trend', '--column', 'email', '--alerts']
```

The same server path is requested even when any or all non-default options are supplied.

## Expected Behavior

The SDK should not silently discard caller intent. It should either:

- preserve the requested options when calling the server, if the endpoint supports them; or
- reject non-default options in server mode with a clear error before making the request.

The second option is the smallest correctness fix with the current default-only endpoint.

## Actual Behavior

The call succeeds and returns default metrics. `trend=True`, `column="email"`, and `alerts=True` have no effect and produce no warning or error.

This is especially easy to miss in monitoring code because the response remains valid while omitting requested trend, column, and alert data.

## Version

- `rocky-sdk==0.8.0`
- Also reproduced on `main` at `7a03e647f8336d93fcc3f748d82d00c6a02037a4`

## Platform

- macOS 26.5.2, arm64
- Python 3.14.4

## Relevant Code

- [SDK early return before option handling](https://github.com/rocky-data/rocky/blob/7a03e647f8336d93fcc3f748d82d00c6a02037a4/sdk/python/src/rocky_sdk/client.py#L1146-L1168)
- [Server endpoint is explicitly default-only](https://github.com/rocky-data/rocky/blob/7a03e647f8336d93fcc3f748d82d00c6a02037a4/engine/crates/rocky-cli/src/api.rs#L756-L770)

## Suggested Regression Test

Add a focused SDK test asserting that server-mode `metrics(trend=True, column="email", alerts=True)` cannot silently fall back to default metrics. With the narrow fix, the test should assert an explicit unsupported-options error and that no HTTP request is made.

mattfaltyn:
## Summary

Prevent `RockyClient.metrics()` from silently discarding `trend`, `column`, and `alerts`
when `server_url` selects the default-only HTTP endpoint.

Closes #1122

## Subproject(s) touched

- [x] Cross-project (public docs)

## Changes

- Reject CLI-only metrics options with `ValueError` before making an HTTP request.
- Preserve the existing default server-mode request to `GET /api/v1/models/<model>/metrics`.
- Cover the default path and each unsupported option with focused SDK tests.
- Document the constraint in the public method reference and SDK changelog.

## Test Plan

- [x] `uv run python -m pytest -q` — 191 passed
- [x] `uv run python -m ruff check src/ tests/ examples/`
- [x] `uv run python -m ruff format --check src/ tests/ examples/`
- [x] `npm run build` in `docs/`
- [x] All GitHub Actions checks pass

## Checklist

- [x] Commit message follows conventional commits and is scoped to the SDK.
- [x] This is not a CLI JSON schema or DSL syntax change; consumer regeneration is not required.

### DIFF
--- a/docs/src/content/docs/dagster/resource.md
+++ b/docs/src/content/docs/dagster/resource.md
@@ -248,7 +248,9 @@ Retrieve pipeline run history. Returns `ModelHistoryResult` when filtered to a s
 
 ### `metrics(model, *, trend=False, column=None, alerts=False) -> MetricsResult`
 
-Retrieve quality metrics for a model. When `server_url` is configured, fetches from the HTTP API instead.
+Retrieve quality metrics for a model. When `server_url` is configured, fetches from the HTTP API
+instead. The HTTP endpoint supports default metrics only; passing `trend`, `column`, or `alerts`
+raises `ValueError` rather than silently ignoring the option.
 
 **Wraps**: `rocky metrics <model> --output json` or `GET /api/v1/models/<model>/metrics`
 
--- a/sdk/python/CHANGELOG.md
+++ b/sdk/python/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`RockyClient.metrics()` no longer silently ignores options in server mode.** Passing
+  CLI-only `trend`, `column`, or `alerts` options when `server_url` is set now raises a
+  clear `ValueError` before making an HTTP request. (#1122)
+
 ## [0.8.0] — 2026-07-14
 
 ### Fixed
--- a/sdk/python/src/rocky_sdk/client.py
+++ b/sdk/python/src/rocky_sdk/client.py
@@ -1151,7 +1151,13 @@ class RockyClient:
         column: str | None = None,
         alerts: bool = False,
     ) -> MetricsResult:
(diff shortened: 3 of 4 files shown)


hugocorreia90 (approved):
Thank you!!

PR #1129, opened 2026-07-16
fix(sdk): pass models directory to compliance

### LINKED ISSUE #1128: RockyClient.compliance() ignores the configured models directory
## Description

`RockyClient` accepts a custom `models_dir`, but `RockyClient.compliance()` does not pass it to `rocky compliance`. The engine therefore scans its default `models/` directory instead of the project configured on the client.

When the default directory is absent, the engine treats it as an empty model set, so the SDK can return a report with zero classified columns and zero exceptions even though the configured models directory contains classified columns.

## Steps to reproduce

```python
import json

from rocky_sdk import RockyClient

payload = json.dumps(
    {
        "version": "1.64.0",
        "command": "compliance",
        "summary": {
            "total_classified": 0,
            "total_exceptions": 0,
            "total_masked": 0,
        },
        "per_column": [],
        "exceptions": [],
    }
)

seen = []
client = RockyClient(models_dir="custom-models")
client.run_cli = lambda args, **kwargs: seen.append(args) or payload
client.compliance(env="prod")

assert seen[-1] == [
    "compliance",
    "--output",
    "json",
    "--models",
    "custom-models",
    "--env",
    "prod",
]
```

On `rocky-sdk 0.8.1`, the final assertion fails because the captured arguments are:

```python
["compliance", "--output", "json", "--env", "prod"]
```

The neighboring `compile()` method correctly emits `--models custom-models`.

## Expected behavior

`RockyClient.compliance()` should pass `--models <models_dir>` so the governance report covers the project configured on the client.

## Actual behavior

The generated command arguments omit `--models`, so the engine scans the default directory and can produce a false zero-exception report for custom-layout projects.

## Environment

- `rocky-sdk`: 0.8.1
- Rocky engine: 1.64.0
- Python: 3.13.5
- OS / architecture: macOS 26.5.2 arm64

## Suggested fix

Include `--models self.models_dir` in the argument list built by `RockyClient.compliance()` and add focused argv regression coverage.

mattfaltyn:
## Summary

Fix `RockyClient.compliance()` so it forwards the client's configured `models_dir` to the Rocky CLI. This prevents valid custom-layout projects from silently receiving an empty or incomplete compliance report.

Closes #1128.

## Subproject(s) touched

- [x] Cross-project (root docs)

## Changes

- Pass `--models <models_dir>` from `RockyClient.compliance()` to the engine.
- Add an SDK regression test covering a non-default models directory and `--env`.
- Update the Dagster wrapper assertion to cover the forwarded default models directory.
- Document compliance as a `models_dir`-aware command in the SDK and Dagster surfaces.
- Add an SDK changelog entry for the corrected behavior.

## User impact

Users who configure `RockyClient(models_dir=...)` or `RockyResource(models_dir=...)` now get compliance results for that directory instead of the engine's default `models/` directory.

## Root cause

The engine's `compliance` command accepts `--models`, but the Python SDK wrapper omitted it and therefore silently used the engine default.

## Test Plan

- `cargo test` from `engine/` — full workspace and doctests passed; credential-gated live tests remained ignored as annotated.
- `uv run pytest` from `sdk/python/` — 194 passed.
- `uv run ruff check .` and `uv run ruff format --check .` from `sdk/python/` — passed.
- `uv build` from `sdk/python/` — passed.
- `uv run pytest` from `integrations/dagster/` — 696 passed.
- `uv run ruff check .` and `uv run ruff format --check .` from `integrations/dagster/` — passed.
- `uv build` from `integrations/dagster/` — passed.
- `npm run test:unit` from `editors/vscode/` — 208 passed.
- `npm run build` from `docs/` under Node 22 — passed.

## Checklist

- [x] This does not change CLI JSON schemas or DSL syntax.

### Engine (`engine/`)

- [ ] `cargo test --all-targets` passes (engine code is unchanged; full-workspace `cargo test` passes)
- [ ] `cargo clippy --all-targets -- -D warnings` is clean (engine code is unchanged)
- [ ] `cargo fmt --check` passes (engine code is unchanged)

### SDK (`sdk/python/`)

- [x] `uv run pytest` passes
- [x] `uv run ruff check .` is clean
- [x] `uv run ruff format --check .` passes
- [x] `uv build` succeeds

- [x] `uv run pytest` passes
- [x] `uv run ruff check .` is clean
- [x] `uv run ruff format --check .` passes

- [ ] `npm run compile` succeeds (VS Code code is unchanged)
- [x] `npm run test:unit` passes (electron tests run in CI)
- [ ] `npm run lint` is clean (VS Code code is unchanged)

### DIFF
--- a/docs/src/content/docs/dagster/resource.md
+++ b/docs/src/content/docs/dagster/resource.md
@@ -15,7 +15,7 @@ sidebar:
 | `config_path` | `str` | `"rocky.toml"` | Path to the pipeline config file. |
 | `state_path` | `str` | `".rocky-state.redb"` | Path to the state store file. |
 | `state_namespace` | `str \| None` | `None` | Optional per-namespace state file (engine `--state-namespace`). Mutually exclusive with `state_path`: when set, `--state-namespace` is sent and `--state-path` is omitted, so independent fan-out runs don't serialize on a single writer lock. |
-| `models_dir` | `str` | `"models"` | Path to the directory containing `.rocky` model files. Used by `compile`, `lineage`, `test`, `ci`, `ai_sync`, `ai_explain`, and `ai_test`. |
+| `models_dir` | `str` | `"models"` | Path to the directory containing `.rocky` model files. Used by `compile`, `lineage`, `test`, `ci`, `compliance`, `ai_sync`, `ai_explain`, and `ai_test`. |
 | `contracts_dir` | `str \| None` | `None` | Optional directory containing contract files. Passed to `compile`, `test`, and `ci` when set. |
 | `server_url` | `str \| None` | `None` | Optional URL for a running `rocky serve` instance. When set, `compile()`, `lineage()`, and `metrics()` use the HTTP API instead of spawning a subprocess. |
 | `timeout_seconds` | `int` | `3600` | Subprocess timeout for any single CLI invocation (in seconds). |
@@ -282,6 +282,12 @@ Analyze materialization strategies and return cost optimization recommendations.
 Run health checks on the Rocky installation and configuration.
 
 **Wraps**: `rocky doctor --output json`
+
+### `compliance(*, env=None) -> ComplianceOutput`
+
+Runs the governance compliance rollup against the resource's configured `models_dir`.
+
+**Wraps**: `rocky compliance --models <models_dir> --output json [--env <env>]`
 
 ### `validate_migration(dbt_project, rocky_project=None, *, sample_size=None) -> ValidateMigrationResult`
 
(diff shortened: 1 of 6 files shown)


hugocorreia90 (approved):
Verified end-to-end. The engine's `compliance` subcommand accepts and uses `--models` (scans that dir for classification sidecars; a missing dir yields the empty-report bug this fixes), the SDK default `models_dir="models"` matches the engine default so default-layout projects are unaffected, and the change aligns `compliance()` with its sibling methods. SDK and Dagster suites pass locally (194 / 696) and dagster-ci is green on the current head. Thanks for the fix plus the docstring, method-reference, and changelog follow-ups.

PR #1141, opened 2026-07-17
fix(engine): scope DAG transformation nodes to models

### LINKED ISSUE #1140: run --dag executes each transformation pipeline once per model, duplicating data
## Description

`rocky run --dag` executes a transformation pipeline once for every model node in that pipeline. A pipeline with `N` models therefore runs all `N` models `N` times during one CLI invocation.

This is user-visible data corruption for append/incremental materializations: the command exits 0 and reports each DAG node completed once, while rows have actually been appended repeatedly.

Confirmed against both:

- current upstream `main` at `be7d3ff36f33463f16d21ffccda042cb1c9cfda4`
- the published `engine-v1.64.0` macOS arm64 artifact (`rocky 1.64.0`; downloaded asset SHA-256 `44eaa69317a8eabd8d9b0d4155b1d4371120a4b67bcfde33310116579ddbe66a`, matching the release metadata)

## Steps to Reproduce

Create `/tmp/rocky-dag-repro/models` with the following files.

`rocky.toml`:

```toml
[adapter.local]
type = "duckdb"
path = "/tmp/rocky-dag-repro/proj.duckdb"

[pipeline.silver]
type = "transformation"

[pipeline.silver.target]
adapter = "local"

[pipeline.silver.target.governance]
auto_create_catalogs = true
auto_create_schemas = true
```

`models/a.sql`:

```sql
SELECT 1 AS id, TIMESTAMP '2026-01-01 00:00:00' AS ts
```

`models/a.toml`:

```toml
name = "a"

[strategy]
type = "incremental"
timestamp_column = "ts"

[target]
catalog = "proj"
schema = "silver"
table = "a"
```

`models/b.sql`:

```sql
SELECT id, ts FROM proj.silver.a
```

`models/b.toml`:

```toml
name = "b"
depends_on = ["a"]

[strategy]
type = "incremental"
timestamp_column = "ts"

[target]
catalog = "proj"
schema = "silver"
table = "b"
```

Run once from `/tmp/rocky-dag-repro`:

```bash
rocky --config rocky.toml --output json --state-path state.redb run --dag

duckdb -csv proj.duckdb \
  "SELECT (SELECT COUNT(*) FROM proj.silver.a) AS a_rows, (SELECT COUNT(*) FROM proj.silver.b) AS b_rows;"
```

## Expected Behavior

The DAG has model nodes `a` and `b`; each model should execute once. The query should return:

```csv
a_rows,b_rows
1,1
```

A fresh control run using `rocky ... run --pipeline silver` does return `1,1`.

## Actual Behavior

The command exits 0 and prints two whole-pipeline summaries:

```text
transformation pipeline complete: 2 model(s) executed
  proj.silver.a
  proj.silver.b
transformation pipeline complete: 2 model(s) executed
  proj.silver.a
  proj.silver.b
```

The final DAG JSON nevertheless reports two completed nodes, `transformation:a` and `transformation:b`. The database contains:

```csv
a_rows,b_rows
2,3
```

The dependent count is 3 because the first pipeline execution appends one row to `b`, then the second sees two rows in `a` and appends both. Repeating the same outer command compounds the counts to `4,10`.

A separate fresh three-model chain produces three whole-pipeline summaries and the exact predicted counts `a=3, b=6, c=10`. A one-model DAG control executes once and produces one row. This rules out a DuckDB timing issue and shows that execution multiplicity equals the model-node count.

## Prior Art

- PR #1101 explicitly left “Dag-mode execution rework (per-model subprocess fan-out...)” as a documented follow-up. It did not change engine code, provide a correctness reproduction, or track a fix.
- PR #986 introduced the owning-pipeline dispatch while fixing seed execution. It documents that transformation nodes are per-model but does not scope their pipeline sub-runs to the model.

Neither PR fixes the repeated execution or has an active follow-up PR.

## Root Cause Evidence

There is a mismatch between DAG construction and dispatch:

- CLI help says each pipeline is a DAG node: [`engine/rocky/src/main.rs`](https://github.com/rocky-data/rocky/blob/be7d3ff36f33463f16d21ffccda042cb1c9cfda4/engine/rocky/src/main.rs#L742-L747).
- The unified DAG actually expands a transformation pipeline into one node per model: [`unified_dag.rs`](https://github.com/rocky-data/rocky/blob/be7d3ff36f33463f16d21ffccda042cb1c9cfda4/engine/crates/rocky-core/src/unified_dag.rs#L280-L289) and [node construction](https://github.com/rocky-data/rocky/blob/be7d3ff36f33463f16d21ffccda042cb1c9cfda4/engine/crates/rocky-core/src/unified_dag.rs#L543-L570).
- `SubRunner` has no model-name argument and explicitly invokes `run()` with no model selection: [`run_dag_exec.rs`](https://github.com/rocky-data/rocky/blob/be7d3ff36f33463f16d21ffccda042cb1c9cfda4/engine/crates/rocky-cli/src/commands/run_dag_exec.rs#L28-L99).
- Every per-model transformation node is then dispatched using only its owning pipeline: [`CliDispatcher::dispatch`](https://github.com/rocky-data/rocky/blob/be7d3ff36f33463f16d21ffccda042cb1c9cfda4/engine/crates/rocky-cli/src/commands/run_dag_exec.rs#L313-L338).

The current DuckDB DAG test contains only one transformation model, so it cannot detect the multiplicative case.

A regression test with two dependent incremental models should assert that one `run_with_dag` call leaves both target tables with exactly one row. Preserve the per-model DAG nodes and pass the node's model identity into a model-scoped transformation sub-run.

## Rocky Version

`rocky 1.64.0`; also reproduced from upstream `main@be7d3ff36f33463f16d21ffccda042cb1c9cfda4`.

## OS / Architecture

macOS 26.5.2 arm64. DuckDB CLI 1.5.2 used only to inspect resulting row counts.

## Configuration

Credential-free local DuckDB configuration shown above.

mattfaltyn:
## Summary

- dispatch each transformation DAG node with its model label instead of rerunning the owning transformation pipeline
- resolve the selected model through the owning pipeline's adapter and schema-creation policy
- keep the models directory model-scoped so replication/load DAG nodes cannot trigger transformation execution
- add DuckDB exact-once coverage plus a focused regression for the transformation-only scope boundary

## Root cause

The unified DAG correctly creates one node per transformation model, but every node called `run()` with the owning pipeline and no model filter. A two-model pipeline therefore ran all models twice, producing row counts `(2, 3)` for the dependent incremental targets instead of `(1, 1)`.

The dispatcher now supplies `Some(model_name)` only for transformation nodes and resolves that model through its owning transformation pipeline. The final review also tightened the shared sub-run seam: only those model-scoped calls receive `models_dir`; replication, load, quality, and snapshot nodes retain their existing pipeline-only behavior.

## Validation

- `cargo nextest run --all-features` — 5,238 passed, 102 skipped
- `cargo test -p rocky-cli --features duckdb run_dag_exec` — 4 passed
- `git diff --check`

Closes #1140

Prior art: #1101, #986

### DIFF
--- a/engine/crates/rocky-cli/src/commands/run.rs
+++ b/engine/crates/rocky-cli/src/commands/run.rs
@@ -1303,26 +1303,46 @@ pub async fn run(
         );
 
         let adapter_registry = AdapterRegistry::from_config(&rocky_cfg)?;
-        // Use the first transformation pipeline's target adapter, or fall back
-        // to the first replication pipeline's target adapter.
-        let target_adapter_name = rocky_cfg
-            .pipelines
-            .values()
-            .find_map(|p| match p {
+        let (target_adapter_name, auto_create_schemas) = if let Some(pipeline_name) =
+            pipeline_name_arg
+        {
+            match registry::resolve_pipeline(&rocky_cfg, Some(pipeline_name))?.1 {
                 rocky_core::config::PipelineConfig::Transformation(t) => {
-                    Some(t.target.adapter.clone())
+                    (
+                        t.target.adapter.clone(),
+                        t.target.governance.auto_create_schemas,
+                    )
                 }
-                _ => None,
-            })
-            .or_else(|| {
-                rocky_cfg.pipelines.values().find_map(|p| match p {
-                    rocky_core::config::PipelineConfig::Replication(r) => {
-                        Some(r.target.adapter.clone())
-                    }
-                    _ => None,
-                })
-            })
-            .unwrap_or_else(|| "default".to_string());
+                pipeline => anyhow::bail!(
+                    "pipeline '{pipeline_name}' is {}, not transformation (required for --model)",
+                    pipeline.pipeline_type_str()
+                ),
+            }
+        } else {
+            // Without an explicit pipeline, preserve the model-only fallback:
+            // first transformation adapter, then first replication adapter.
+            (
+                rocky_cfg
+                    .pipelines
+                    .values()
+                    .find_map(|p| match p {
+                        rocky_core::config::PipelineConfig::Transformation(t) => {
+                            Some(t.target.adapter.clone())
+                        }
+                        _ => None,
+                    })
+                    .or_else(|| {
+                        rocky_cfg.pipelines.values().find_map(|p| match p {
+                            rocky_core::config::PipelineConfig::Replication(r) => {
+                                Some(r.target.adapter.clone())
+                            }
+                            _ => None,
+                        })
+                    })
+                    .unwrap_or_else(|| "default".to_string()),
+                false,
+            )
+        };
 
         let warehouse = adapter_registry.warehouse_adapter(&target_adapter_name)?;
         // Propagate state-store open errors with context, matching the
@@ -1366,7 +1386,7 @@ pub async fn run(
             None, // model-only run has no pipeline hooks
             None,
             &schema_cache_cfg,
-            false, // model-only entry point has no pipeline governance context
+            auto_create_schemas,
             defer_opts,
             skip_gate,
             // Reuse is active iff `[reuse]` is enabled AND `--no-reuse` was
--- a/engine/crates/rocky-cli/src/commands/run_dag_exec.rs
+++ b/engine/crates/rocky-cli/src/commands/run_dag_exec.rs
@@ -57,10 +57,7 @@ fn default_sub_runner() -> SubRunner {
          model_name: Option<String>,
          skip_opts| {
             Box::pin(async move {
-                let models_dir = config_path
-                    .parent()
-                    .unwrap_or(Path::new("."))
-                    .join("models");
+                let models_dir = models_dir_for_model_scope(&config_path, model_name.as_deref());
                 let partition_opts = super::PartitionRunOptions {
                     partition: None,
                     from: None,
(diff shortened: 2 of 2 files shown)


hugocorreia90 (approved):
Thanks for the detailed report and fix — the credential-free repro and the predicted-vs-actual row counts made this easy to verify.

Traced the change end-to-end and it's correct:

- Transformation DAG nodes now dispatch model-scoped (`Some(model_name)`), so each node materializes only itself. Non-transformation nodes stay byte-identical to before (`model_name`/`models_dir` both `None`), which correctly avoids re-triggering full transformation execution through the `models_dir.is_some()` branch in `run()`.
- Model-in-isolation is sound: `execute_models` compiles the whole project while the execution filter is a separate later stage, model refs resolve to physical tables (not inlined SQL), and the DAG layer barrier materializes each upstream before any downstream node dispatches.
- Checked the two subtler seams too: the unified DAG's `infer_runtime_dependencies` shares `extract_lineage` with the compiler (so ordering is at least as complete as the pre-patch compiler-ordered path), and the skip gate is fail-safe under isolation (an upstream with no in-run verdict forces a rebuild rather than a stale skip).

Full `rocky-cli` suite + the new regression test pass locally; clippy/fmt clean.

Two minor behavior notes, both non-blocking and worth a short follow-up (which I'll open): `rocky run --model X --pipeline Y` now fails fast if Y isn't a transformation pipeline, and `rocky run --dag --skip-unchanged` is conservative for models with model-upstreams (always rebuilds — safe, never stale).

LGTM.

PR #1147, opened 2026-07-17
fix(engine/rocky-compiler): infer cast alias target types

### LINKED ISSUE #1145: Cast alias retaining the source column name inherits the pre-cast type
## Description

When a projected `CAST` keeps the source column name, Rocky assigns the output the source column's pre-cast type. This makes contract validation reject the correct post-cast type and accept the incorrect source type.

Confirmed against `main` at `9a3db24de47677285cea969f940447b40d1fd89a`.

## Steps to Reproduce

1. Provide an external source schema where `source.raw.users.id` is `BIGINT`.
2. Compile this model:

```sql
SELECT CAST(id AS STRING) AS id
FROM source.raw.users
```

3. Validate it against a contract declaring `id` as `String`.
4. Run `rocky compile --models models --contracts contracts --with-seed --output json`.

## Expected Behavior

The model output `id` is inferred as `String`, so a `String` contract passes and an `Int64` contract fails with `E011`.

## Actual Behavior

The output is inferred as `Int64`. A correct `String` contract fails with:

```text
E011: column 'id' type mismatch: contract expects String, got Int64
```

Conversely, an incorrect `Int64` contract passes.

## Rocky Version

`rocky 1.64.0`

## OS / Architecture

`Darwin 25.5.0 arm64`

## Root Cause

Lineage correctly marks casts as `Unknown` pending expression inference, but the production fallback resolves that unknown by the output alias. When the alias matches the source column (`id`), the fallback restores the source `Int64` instead of reading the cast target `STRING`.

mattfaltyn:
## Summary

Fix cast outputs that retain their source-column name being assigned the pre-cast source type.

Closes #1145.

## Why

Lineage intentionally leaves cast output types as `Unknown` until expression inference. The production fallback then resolved unknown columns by their output alias, so `CAST(id AS STRING) AS id` looked up source `id` and restored `Int64`. This rejected the correct `String` contract while accepting an incorrect `Int64` contract.

## Subproject(s) touched

## Changes

- Remove output-alias-based source type restoration from production type checking.
- Parse the SQL cast target for lineage-confirmed casts whose input schema is known.
- Preserve lineage-derived nullability and keep unknown-input casts conservative.
- Add a production-path regression asserting the inferred `String` type and the `E011` rejection of an `Int64` contract.

## Impact

Contracts now validate against the post-cast type when the cast output keeps the source name. Unknown-source behavior remains unchanged, and there are no CLI JSON, DSL, dependency, or generated-code changes.

## Test Plan

- `cargo test -p rocky-compiler` - 409 tests passed (387 unit + 22 integration).
- `cargo test -p rocky-compiler test_cast_alias_uses_cast_type_not_source_type -- --nocapture` - passed.
- Public CLI reproduction: the correct `String` contract passes; the incorrect `Int64` contract fails with `E011` (`got String`).
- `cargo clippy -p rocky-compiler --all-targets --all-features --no-deps -- -D clippy::all` - passed.
- `cargo fmt --all -- --check` and `git diff --check` - passed.
- GitHub Actions on `3e53718a`: `engine-ci` Test, Clippy, Format, Adapter boundary, Release-build smoke; `codegen-drift`; and both credential-containment checks - passed.

## Checklist

- [x] No CLI JSON schema or DSL syntax change

### Engine (`engine/`)

- [x] Test suite is green (`engine-ci / Test`, `cargo nextest run --all-features`)
- [x] Clippy is clean (`engine-ci / Clippy`, `cargo clippy --all-targets --all-features -- -D warnings`)

### DIFF
--- a/engine/crates/rocky-compiler/src/typecheck.rs
+++ b/engine/crates/rocky-compiler/src/typecheck.rs
@@ -443,36 +443,7 @@ fn compute_model_typecheck(
         ReferenceMap::default()
     };
 
-    // Step 1: Build type scope from upstream models
-    let mut scope = TypeScope::new();
-
-    for upstream_name in &model_schema.upstream {
-        if let Some(upstream_cols) = typed_models.get(upstream_name) {
-            for col in upstream_cols {
-                scope.columns.insert(
-                    CiKey::owned(col.name.clone()),
-                    (col.data_type.clone(), col.nullable),
-                );
-            }
-        }
-    }
-    // Also inject from external sources via lineage edges
-    for edge in graph.edges_targeting(model_name) {
-        if let Some(source_cols) = typed_models.get(&*edge.source.model) {
-            let col = col_index
-                .get(&*edge.source.model)
-                .and_then(|idx| idx.get(&*edge.source.column))
-                .map(|&i| &source_cols[i]);
-            if let Some(col) = col {
-                scope.columns.insert(
-                    CiKey::owned(edge.source.column.to_string()),
-                    (col.data_type.clone(), col.nullable),
-                );
-            }
-        }
-    }
-
-    // Step 2: Lineage-based type propagation
+    // Step 1: Lineage-based type propagation
     let mut typed_cols: Vec<TypedColumn> = Vec::with_capacity(model_schema.columns.len());
 
     for col_def in &model_schema.columns {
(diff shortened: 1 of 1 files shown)


hugocorreia90 (approved):
Reviewed and verified against the source, not just the PR body:

- **Fix logic is sound.** The `input_is_known` gate only ever narrows when a type is applied — it can't produce a wrong type. The applied type is the deterministic cast target from `sql_type_to_rocky`, and lineage-derived nullability is preserved (non-null source → non-null cast), which is correct for a plain CAST.
- **Panic audit of newly-live code.** This wires `infer_select_types` and its full transitive inference tree onto arbitrary user SQL for the first time (the old path never parsed SQL). `.ok()` catches `Err` but not panics, and the whole body walks every projection/CTE. Scanned all callees — no reachable panic on parseable SQL: total match with `_` fallback in `sql_type_to_rocky`, `arg_types[1]` guarded by `>= 2`, `CompoundIdentifier` indexing guarded by `parts.len() >= 2`, arg walkers use `.first()`/`.iter()`.
- **Regression test is non-vacuous.** Ran it against pre-fix base — fails with `Int64` vs `String`, the exact bug. It genuinely guards the regression.
- CI green on all checks; merges cleanly (no intervening `typecheck.rs` changes on main).

Non-blocking follow-ups will be addressed in a separate PR: a test for the unknown-input conservative branch, an in-code note that removing the alias lookup intentionally also stops (unsound) enforcement of expression-aliased-to-source-name columns, and a tracked issue for pre-existing TRY_CAST/SAFE_CAST nullability.

PR #1151, opened 2026-07-17
Fix quoted time-window placeholder SQL generation

### LINKED ISSUE #1150: Quoted time-window placeholders compile to invalid SQL
## Summary

Time-interval models that quote `@start_date` or `@end_date` pass compiler validation, but SQL generation adds another pair of single quotes. The result is syntactically invalid SQL.

Rocky's own cross-dialect golden fixture currently captures this malformed output.

## Reproduction

On `main` at `749ebeb170221dc39c8cbead2530c11ebaf3a0fe`:

```sh
duckdb :memory: < engine/crates/rocky-cli/tests/ir-golden/09-transformation-time-interval-bootstrap/duckdb.sql
```

DuckDB exits with:

```text
Parser Error: syntax error at or near "1900"
... event_date >= ''1900-01-01 00:00:00'' ...
```

The fixture's model SQL contains already-quoted placeholders:

```sql
WHERE event_date >= '@start_date'
  AND event_date < '@end_date'
```

Generated SQL contains double single quotes:

```sql
WHERE event_date >= ''1900-01-01 00:00:00''
  AND event_date < ''1900-01-01 00:00:00''
```

The same malformed form is present in the checked-in BigQuery, Databricks, DuckDB, and Snowflake goldens for this fixture.

## Expected behavior

Either:

1. quoted placeholders should generate valid single-quoted timestamp literals, or
2. the compiler should reject quoted placeholders with a clear diagnostic explaining that placeholders must be bare.

The documented bare form should continue to work:

```sql
WHERE event_date >= @start_date
  AND event_date < @end_date
```

## Root cause

- `contains_placeholder` in `engine/crates/rocky-compiler/src/typecheck.rs` scans raw bytes, so placeholder-looking text inside string literals satisfies validation.
- `substitute_partition_placeholders` in `engine/crates/rocky-core/src/sql_gen.rs` performs unconditional string replacement and always inserts a quoted timestamp literal.
- The golden tests compare generated strings but do not parse or execute the SQL, allowing the malformed snapshots to pass.

This affects both normal partition generation and the bootstrap path because both use the same substitution helper.

## Suggested fix and coverage

Make placeholder handling quote- and token-boundary-aware, or reject placeholders inside quoted literals during compilation. Add coverage for:

- bare placeholders;
- already-single-quoted placeholders;
- placeholder-like text inside unrelated string literals or identifiers;
- both normal and bootstrap partition generation;
- parsing or executing the generated DuckDB golden SQL so syntactically invalid snapshots cannot pass unnoticed.

Verified with DuckDB 1.5.2.

mattfaltyn:
## Summary

- preserve already-single-quoted time-window placeholders when substituting partition timestamps
- keep the documented bare `@start_date` and `@end_date` form unchanged
- correct the BigQuery, Databricks, DuckDB, and Snowflake bootstrap SQL snapshots

## Root cause

`substitute_partition_placeholders` always inserted a quoted timestamp literal. When a model already wrote `'@start_date'` or `'@end_date'`, the surrounding quotes remained and Rocky generated invalid doubled quotes such as `''1900-01-01 00:00:00''`.

The shared substitution helper now normalizes the exactly quoted form before replacing bare placeholders, covering both normal partition generation and bootstrap SQL.

## Impact

Quoted time-window placeholders now generate valid SQL instead of failing at execution time. Existing models using the documented bare placeholders retain the same output.

## Validation

- `cargo test -p rocky-core` (1,640 unit tests passed; integration and doc tests passed)
- `cargo test -p rocky-cli --test ir_golden` (4 passed)
- `cargo clippy -p rocky-core --lib --no-deps -- -D warnings`
- executed the corrected DuckDB bootstrap golden against an in-memory catalog and source table

The workspace-wide clippy command is locally blocked on Rust 1.93.1 by pre-existing `unused_assignments` diagnostics in the miette-derived `rocky-sql` and `rocky-lang` error types; the latest canonical `main` Engine CI run is green.

Closes #1150

### DIFF
--- a/engine/crates/rocky-cli/tests/ir-golden/09-transformation-time-interval-bootstrap/bigquery.sql
+++ b/engine/crates/rocky-cli/tests/ir-golden/09-transformation-time-interval-bootstrap/bigquery.sql
@@ -1,2 +1,2 @@
 CREATE OR REPLACE TABLE `tgtwarehouse`.`marts__demo`.`events_daily` AS
-SELECT event_date, COUNT(*) AS event_count FROM tgtwarehouse.raw__demo.events WHERE event_date >= ''1900-01-01 00:00:00'' AND event_date < ''1900-01-01 00:00:00'' GROUP BY 1
+SELECT event_date, COUNT(*) AS event_count FROM tgtwarehouse.raw__demo.events WHERE event_date >= '1900-01-01 00:00:00' AND event_date < '1900-01-01 00:00:00' GROUP BY 1
--- a/engine/crates/rocky-cli/tests/ir-golden/09-transformation-time-interval-bootstrap/databricks.sql
+++ b/engine/crates/rocky-cli/tests/ir-golden/09-transformation-time-interval-bootstrap/databricks.sql
@@ -1,2 +1,2 @@
 CREATE OR REPLACE TABLE tgtwarehouse.marts__demo.events_daily AS
-SELECT event_date, COUNT(*) AS event_count FROM tgtwarehouse.raw__demo.events WHERE event_date >= ''1900-01-01 00:00:00'' AND event_date < ''1900-01-01 00:00:00'' GROUP BY 1
+SELECT event_date, COUNT(*) AS event_count FROM tgtwarehouse.raw__demo.events WHERE event_date >= '1900-01-01 00:00:00' AND event_date < '1900-01-01 00:00:00' GROUP BY 1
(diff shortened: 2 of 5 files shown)


hugocorreia90 (approved):
Verified: quoted-form `'@start_date'` previously generated non-executable `''…''` SQL, and the compiler accepts the quoted form (`contains_placeholder` only enforces a trailing boundary) so it is reachable. The fix's replace order (quoted-form first, then bare) is correct across bare/quoted/mixed inputs; the bare form is byte-identical to before, so no existing model regresses. Golden snapshots corrected for all four dialects and CI Test (incl. ir_golden) is green. LGTM.

PR #1153, opened 2026-07-17
fix(engine): fail closed on compare read errors

### LINKED ISSUE #1152: bug: compare passes when both target reads fail
### Description

`rocky compare` reports an overall `pass` when both the production and shadow
tables are unreadable. The command currently converts both row-count errors to
`0` and both schema-description errors to empty schemas, so two failures look
like two identical tables.

This is fail-open behavior in the validation step used before promoting a
shadow run.

### Steps to reproduce

1. Use the shadow-mode compare playground example:
   `examples/playground/pocs/06-developer-experience/04-shadow-mode-compare`.
2. Seed only the raw `orders` source in DuckDB. Do not create either
   `poc.staging__orders.orders` or
   `poc.staging__orders.orders_rocky_shadow`.
3. Run:

   ```console
   rocky -c rocky.toml -o json compare --filter source=orders
   ```

### Expected behavior

The command fails when it cannot read either comparison target. It must not
report a passing comparison without successfully reading both tables.

### Actual behavior

The command exits successfully and reports:

```json
{
  "tables_compared": 1,
  "tables_passed": 1,
  "tables_failed": 0,
  "results": [
    {
      "production_count": 0,
      "shadow_count": 0,
      "row_count_match": true,
      "schema_match": true,
      "verdict": "pass"
    }
  ],
  "overall_verdict": "pass"
}
```

The relevant fallbacks are in
`engine/crates/rocky-cli/src/commands/compare.rs`:

- row-count read failures use `unwrap_or(0)`
- schema read failures use `unwrap_or_default()`

### Rocky version

Current `main` at `749ebeb170221dc39c8cbead2530c11ebaf3a0fe`.

### Operating system

macOS

### Additional context

A control run with both comparison tables present reports the expected
`100`/`100` passing result. The defect is specific to error handling, not the
row-count comparison itself.

mattfaltyn:
## Summary

Make `rocky compare` fail closed when either production or shadow row-count
or schema reads fail.

Closes #1152.

## Subproject(s) touched

## Changes

- Preserve target-read errors while retaining the existing placeholder values
  used by structured compare output.
- Force the table verdict to `fail` whenever any required read failed, instead
  of allowing equal `0` counts and empty schemas to produce `pass`.
- Add a DuckDB regression test that discovers a source while both production
  and shadow targets are absent.
- Update the branch-compare documentation to describe missing materializations
  as failed comparisons.

## Root cause and impact

The CLI discarded row-count and schema errors with `unwrap_or(0)` and
`unwrap_or_default()`. It then evaluated `0` versus `0` and an empty schema
versus an empty schema as exact matches. That could let the validation step
used before shadow promotion report success without reading either target.

This change keeps the existing output schema and missing-target behavior, but
ensures read failures can never produce a passing verdict.

## Test Plan

- `cargo test --all-targets`
- `cargo nextest run --all-features` — 5,378 passed, 102 skipped
- `cargo test -p rocky-cli commands::compare::tests::unreadable_target_pair_fails_comparison -- --exact`
- `cargo fmt -- --check`
- `cargo clippy -p rocky-cli --all-targets --all-features --no-deps -- -D warnings -A unused-assignments -A clippy::collapsible-else-if`

The unmodified workspace-wide `cargo clippy --all-targets --all-features --
-D warnings` command is currently blocked on `main` by pre-existing
`unused_assignments` warnings in generated diagnostic structs and
`collapsible_else_if` warnings in `commands/list.rs`. The patch-scoped
deny-warnings run above is clean.

## Checklist

- [x] If this is a CLI JSON schema or DSL syntax change, all consumers (`integrations/dagster/`, `editors/vscode/`) are updated in this same PR

### Engine (`engine/`)

- [ ] `cargo clippy --all-targets -- -D warnings` is clean

- [ ] `uv run pytest` passes

- [ ] `npm run lint` is clean

### DIFF
--- a/engine/crates/rocky-cli/src/commands/branch.rs
+++ b/engine/crates/rocky-cli/src/commands/branch.rs
@@ -185,9 +185,8 @@ pub fn run_branch_list(state_path: &Path, json: bool) -> Result<()> {
 /// whose `schema_override` points at the branch's schema. Mirrors how
 /// `rocky run --branch <name>` wires the same prefix into the write path.
 ///
-/// Branches with no materialized tables yet surface as missing rows/columns
-/// on the production side of the diff rather than crashing — the underlying
-/// compare path uses `unwrap_or_default` for describe/count failures.
+/// Branches with no materialized tables yet surface as failed comparisons
+/// rather than crashing.
 pub async fn run_branch_compare(
     state_path: &Path,
     config_path: &Path,
--- a/engine/crates/rocky-cli/src/commands/compare.rs
+++ b/engine/crates/rocky-cli/src/commands/compare.rs
@@ -277,6 +277,13 @@ fn read_or_default<T: Default, E: std::fmt::Display>(
 }
 
 /// Get the row count for a target table.
+///
+/// Returns `Err` when the `COUNT(*)` result is absent or not a non-negative
+/// integer, so a garbled read fails the comparison closed instead of silently
+/// counting as zero. Parsing goes through `cell_as_u64`, which handles the
+/// JSON-integer (Trino), integer-as-string (Databricks/Snowflake/DuckDB), and
+/// integral-float cell shapes adapters return — a bare `as_str()` would drop
+/// numeric cells to `0` and let a real mismatch pass.
 async fn get_row_count(adapter: &dyn WarehouseAdapter, target: &TargetRef) -> Result<u64> {
     let table_name = target
         .validated_full_name()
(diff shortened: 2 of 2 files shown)


hugocorreia90:
Follow-up commit `ae083982`: hardened `get_row_count` to parse the `COUNT(*)` cell via the tested `cell_as_u64` helper and fail closed on an absent/unparseable cell. The prior `as_str()`-only parse silently defaulted **numeric** JSON cells to `0`, so adapters that return numeric cells (Trino) would compare `0 == 0` and let a real row-count mismatch pass — the same fail-open this PR set out to close. Also removes the `rows[0][0]` index panic on an empty result. Local pre-flight: fmt + `cargo clippy -p rocky-cli` clean, and `unreadable_target_pair_fails_comparison` passes.

hugocorreia90 (approved):
Verified the read-error fail-closed logic (existing-but-empty tables still evaluate normally; only unreadable targets force Fail) and confirmed the follow-up `ae083982` closes the numeric-cell fail-open (Trino) via the tested `cell_as_u64` helper. The prior Test failure was an unrelated flake (`job_interrupted_by_restart_is_swept_to_failed`, an HTTP-spawn test — passes locally 3/3, green on re-run). All required checks pass. LGTM.

PR #1156, opened 2026-07-18
fix(engine): degrade type drift to full refresh on Databricks and Trino

### LINKED ISSUE #1115: drift: default safe-widening allowlist advertises ALTER paths Databricks and Trino cannot execute
## Summary

\`drift.rs::default_is_safe_type_widening\` (the \`SqlDialect\` trait default) classifies integer/float/decimal widenings and any numeric/BOOLEAN → STRING as safe, and the run path then executes the trait-default ANSI statement \`ALTER TABLE x ALTER COLUMN y TYPE z\`. Only Snowflake and BigQuery override the allowlist and/or the SQL form to match what their warehouse actually accepts. Databricks, DuckDB, and Trino all inherit both defaults.

Verified state per adapter:

- **DuckDB** — works. \`ALTER TABLE t ALTER COLUMN i TYPE BIGINT\` and \`... TYPE STRING\` both execute with rows preserved (verified live on DuckDB 1.5.x).
- **Snowflake** — correct by construction: override allows only \`NUMBER(p,s)\` precision + \`VARCHAR\` length widening, matching \`ALTER COLUMN ... SET DATA TYPE\` support; live-verified by \`drift_widening_live.rs\`.
- **BigQuery** — correct by construction: override allows only \`INT64 → NUMERIC → BIGNUMERIC\` and emits the required \`SET DATA TYPE\` form.
- **Databricks** — broken: Delta tables reject \`ALTER TABLE ... ALTER COLUMN ... TYPE\` for type changes unless the \`delta.enableTypeWidening\` table feature is enabled (Rocky never sets it), and numeric → \`STRING\` is not a supported widening at all. A drift classified \`AlterColumnTypes\` therefore fails the table's run with the warehouse error (\`run.rs\` propagates the \`execute_statement\` error; there is no fallback to full refresh).
- **Trino** — broken: the ANSI \`ALTER COLUMN y TYPE z\` form is not Trino syntax (Trino requires \`ALTER COLUMN y SET DATA TYPE z\`), so any classified-safe widening fails.

## Impact

On Databricks (the wedge warehouse) and Trino, source type drift that the engine classifies as a safe widening fails the table run instead of evolving the target in place. Loud failure, not corruption — but the graduated-evolution behavior the docs describe is only real on DuckDB / Snowflake / BigQuery today.

## Suggested fix

1. \`rocky-databricks\`: either override \`is_safe_type_widening\` to return \`false\` (drop to full refresh, correct today) or set \`delta.enableTypeWidening\` at CREATE and scope the allowlist to Delta's supported widenings (no numeric → STRING).
2. \`rocky-trino\`: override \`alter_column_type_sql\` to the \`SET DATA TYPE\` form and scope the allowlist to what the connector supports.
3. Consider a run-path fallback: on a failed widening \`ALTER\`, degrade to \`DropAndRecreate\` instead of failing the table.
4. Pair any adapter change with live \`#[ignore]\` widening tests mirroring \`rocky-snowflake/tests/drift_widening_live.rs\`.

Found while accuracy-reviewing the drift documentation in #1113.

hugocorreia90:
## Summary

Fixes #1115. `drift::default_is_safe_type_widening` (the `SqlDialect` trait default) classifies integer/float/decimal widenings and numeric/BOOLEAN → STRING as safe, and the run path then executes the ANSI `ALTER TABLE … ALTER COLUMN … TYPE …` form. **Databricks and Trino both inherited that default but cannot execute it:**

- **Databricks (Delta):** rejects `ALTER … ALTER COLUMN … TYPE` for a type change unless the `delta.enableTypeWidening` table feature is enabled (Rocky never sets it), and numeric → STRING isn't a supported Delta widening at all.
- **Trino:** the ANSI `ALTER COLUMN y TYPE z` form isn't valid Trino syntax (Trino requires `… SET DATA TYPE …`), and the accepted in-place changes are connector-specific and narrow.

So source type drift the engine classified as a *safe widening* **failed the table run** on both warehouses (loud failure, `run.rs` propagates the `execute_statement` error) instead of evolving the target in place.

## Change

Override `is_safe_type_widening` to return `false` in the Databricks and Trino dialects, so type drift degrades to a full refresh (`DropAndRecreate`) — the same path other unsafe drifts already take, which both warehouses execute correctly. This is the precise answer given current adapter state: *zero* in-place `ALTER COLUMN TYPE` widenings work on either warehouse today, so "no widening is ALTER-safe" is literally true. It replaces a guaranteed loud failure with correct (if blunter) evolution — no working behavior is lost.

- **DuckDB** keeps inheriting the default (it executes the ANSI ALTER — verified live per #1115), so it still evolves in place.
- **Snowflake / BigQuery** keep their existing type-system-scoped overrides.
- Corrects the stale default-impl doc comment (it claimed Snowflake shares the default; Snowflake overrides).

## Tests

Asserted at the `detect_drift` **decision level** (not just the predicate) with the real dialects:

- **Databricks / Trino:** `INT → BIGINT` and numeric → `STRING` both degrade to `DropAndRecreate`.
- **DuckDB (guard):** the same `INT → BIGINT` still resolves to `AlterColumnTypes` — pins the one adapter that works so a future edit to the default allowlist can't silently regress it.

## Scope / follow-up

This resolves the titled bug: the engine no longer advertises `ALTER` paths these warehouses can't execute. **Graceful in-place evolution** (Databricks `delta.enableTypeWidening` + a Delta-scoped allowlist; Trino `SET DATA TYPE` + a connector-scoped allowlist, live-verified via the `trino-conformance` harness) is tracked as a follow-up — it needs live Databricks/Trino verification to get the per-warehouse allowlists right. The run-path fallback (catch a failed `ALTER`, silently full-refresh) was **consciously rejected**: it would mask real errors (e.g. permissions) and make an expensive/destructive operation happen by accident; the up-front per-adapter decision is safer and more predictable.

### DIFF
--- a/engine/crates/rocky-core/src/drift.rs
+++ b/engine/crates/rocky-core/src/drift.rs
@@ -93,14 +93,18 @@ pub fn generate_drop_table_sql(
     Ok(dialect.drop_table_sql(&ref_str))
 }
 
-/// Default body of [`SqlDialect::is_safe_type_widening`] — Databricks /
-/// Spark / DuckDB / Snowflake share enough type-name conventions that
-/// a single allowlist covers them.
+/// Default body of [`SqlDialect::is_safe_type_widening`] — the
+/// Spark/ANSI type-name allowlist (`INT → BIGINT`, `DECIMAL(p,s)`
+/// precision, `VARCHAR(n)` length, numeric → `STRING`).
 ///
-/// Dialects with different type spellings (BigQuery `INT64` /
-/// `NUMERIC` / `FLOAT64` / `BIGNUMERIC`) override
-/// [`SqlDialect::is_safe_type_widening`] in their dialect impl rather
-/// than bolting names onto this list — keeping the global default
+/// Only DuckDB uses this default today: it executes the emitted
+/// `ALTER COLUMN … TYPE` widenings as-is. Every other adapter overrides
+/// [`SqlDialect::is_safe_type_widening`] in its dialect impl —
+/// Snowflake (`NUMBER`/`VARCHAR` precision) and BigQuery (`INT64` /
+/// `NUMERIC` / `BIGNUMERIC`) scope the allowlist to their own type
+/// system, while Databricks and Trino override to `false` because they
+/// can't execute the ANSI `ALTER COLUMN … TYPE` form and degrade type
+/// drift to a full refresh instead (#1115). Keeping the global default
 /// dialect-neutral (no BQ types here) avoids the asymmetry of having
 /// e.g. both `BIGINT → STRING` and `INT64 → STRING` for what is
 /// semantically the same conversion.
--- a/engine/crates/rocky-databricks/src/dialect.rs
+++ b/engine/crates/rocky-databricks/src/dialect.rs
@@ -252,6 +252,21 @@ impl SqlDialect for DatabricksSqlDialect {
             .join(", ");
         Ok(format!("xxhash64({arg_list})"))
     }
+
+    fn is_safe_type_widening(&self, _source_type: &str, _target_type: &str) -> bool {
+        // Delta tables reject `ALTER TABLE … ALTER COLUMN … TYPE …` for a type
+        // change unless the `delta.enableTypeWidening` table feature is enabled
+        // (Rocky never sets it), and numeric → STRING is not a supported Delta
+        // widening at all. So every widening the shared default classifies as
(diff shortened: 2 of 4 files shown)


hugocorreia90:
Graceful in-place type evolution (Databricks `enableTypeWidening` + scoped allowlist; Trino `SET DATA TYPE` + connector-scoped allowlist, live-verified) is tracked as a follow-up: #1157.

PR #1160, opened 2026-07-18
fix(engine/rocky-cli): reject unknown model filter

### LINKED ISSUE #1159: rocky run --model succeeds when the model does not exist
## Description

`rocky run --model <name>` exits successfully when `<name>` does not exist in the compiled project. The JSON response reports `"status": "Success"` with no failures or materializations.

This is a false-green result for the model-scoped path used by automation and Dagster asset materialization: a typo or stale model name can leave data unbuilt while the caller sees a successful run.

## Steps to Reproduce

1. Create a DuckDB transformation project containing one model named `known_model`.
2. Run:

   ```console
   rocky -c rocky.toml -o json run --model does_not_exist
   ```

3. Observe the exit status and JSON output.

## Expected Behavior

Rocky rejects the unknown model with a non-zero exit and a clear `model not found` error.

## Actual Behavior

The command exits `0` and reports:

```json
{
  "status": "Success",
  "tables_failed": 0,
  "materializations": []
}
```

The behavior reproduces consistently. As a control, replacing `does_not_exist` with `known_model` materializes the expected table.

## Root Cause

The model name is applied only as a per-layer filter in `execute_models`. No post-compilation validation verifies that the explicit filter matched a project model, so every execution layer becomes empty and the empty run derives `Success`.

## Rocky Version

`rocky 1.64.0`, reproduced on canonical `main` at `8b988548ef7c761ce42fa3e4af89fb3e500d42fc`.

## OS / Architecture

macOS / arm64

mattfaltyn:
## Summary

Reject `rocky run --model <name>` when `<name>` is absent from the compiled transformation project.

Previously the explicit model name was used only as a per-layer filter. A typo filtered every layer to an empty set, then the empty run exited `0` with `"status": "Success"` and no materializations. That false-green result could leave a Dagster asset or scripted model run stale while automation reported success.

## Subproject(s) touched

- [x] `engine/` (`rocky-cli`)

## Changes

- validate an explicit model filter immediately after project compilation
- return a clear `model not found` error before model execution
- add a focused DuckDB regression test for the empty-success failure mode

## Test Plan

- [x] `cargo test -p rocky-cli --features duckdb --lib unknown_model_filter_is_an_error_not_empty_success`
- [x] `cargo fmt -- --check`
- [x] `cargo build -p rocky`
- [x] Reproduced the fixed CLI behavior: unknown model exits `1` with JSON `status: "Failure"` and the specific error
- [x] Control: known model still exits `0` and materializes `poc.main.known_model`
- [x] `cargo nextest run --all-features` (5,420 passed)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` (GitHub CI, Rust 1.97.1)

## Checklist

- [x] Conventional commit with engine/crate scope
- [x] No `Co-Authored-By` trailer
- [x] No CLI JSON schema shape change; codegen is not required

Closes #1159

### DIFF
--- a/engine/crates/rocky-cli/src/commands/run.rs
+++ b/engine/crates/rocky-cli/src/commands/run.rs
@@ -6288,6 +6288,13 @@ pub(crate) async fn execute_models(
         }
     };
 
+    if let Some(name) = model_name_filter {
+        anyhow::ensure!(
+            compile_result.project.model(name).is_some(),
+            "model '{name}' not found (no transformation model with that name)"
+        );
+    }
+
     // Per-model surrogate-key specs — loaded ONCE here (execute-from-owned-value,
     // #1) so the SAME owned map feeds BOTH the TOCTOU fingerprint below AND
     // execution (via `ExecutionContext`), with no post-verification reload that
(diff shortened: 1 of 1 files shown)


hugocorreia90 (approved):
Reviewed + Codex red-team pass (both clean).

- Validation uses the exact same `config.name` identity as the per-layer execution filter, so a legitimate `--model` can never be wrongly rejected and a rejected name is provably one that would have filtered every layer to empty (the #1159 false-green).
- Only the `--model` caller passes a non-None filter; replication/backfill/local paths pass None (backfill has its own missing-model rejection).
- The `ensure!` Err is caught by the existing handler → `tables_failed` bump + JSON `status: Failure` + non-zero exit; persistence/finalization still run.
- A model that loads but fails typecheck stays in the project, so real compile errors aren't masked.
- Consistent with the existing `api.rs` model-not-found idiom. CI green.

PR #1164, opened 2026-07-18
fix(engine/rocky-cli): reject unknown tick pipeline

### LINKED ISSUE #1163: rocky tick --pipeline silently succeeds for unknown pipelines
## Description

`rocky tick --pipeline <name>` silently succeeds when `<name>` is not present
in the configuration. The tick exits `0` and reports zero evaluated, executed,
or skipped pipelines.

For unattended cron/systemd/CI scheduling, a typo can therefore stop all
intended work indefinitely without an error or actionable skip reason.

## Steps to reproduce

Using Rocky `1.65.0` and this minimal configuration:

```toml
[adapter.db]
type = "duckdb"

[pipeline.raw]
type = "transformation"

[pipeline.raw.target]
adapter = "db"

[pipeline.raw.schedule]
cron = "0 3 * * *"
```

Run:

```bash
rocky --config rocky.toml --output json tick \
  --pipeline typo \
  --now 2026-07-18T03:00:00Z
```

## Expected behavior

The command should fail before opening scheduler state or spawning work, with
the established pipeline-resolution error, for example:

```text
pipeline 'typo' not found in config. Did you mean 'raw'?
```

## Actual behavior

The command exits `0` and returns an empty successful tick:

```json
{
  "evaluated": [],
  "executed": [],
  "skipped": [],
  "counts": {
    "evaluated": 0,
    "executed": 0,
    "skipped": 0
  }
}
```

The behavior was reproduced twice. Replacing `--pipeline typo` with
`--pipeline raw` evaluates one pipeline, confirming the filter is the cause.

## Root cause

The tick path passes `pipeline_filter` to `resolve_all_schedules`, which uses
it only as a predicate. No match produces an empty schedule map rather than an
input error. Other pipeline-selecting commands already use
`registry::resolve_pipeline` to reject unknown names and provide suggestions.

## Rocky version

`rocky 1.65.0`

## OS / architecture

macOS 26.5.2 arm64

mattfaltyn:
## Summary

- reject `rocky tick --pipeline <name>` when `<name>` is not configured
- reuse the existing pipeline resolver, including its typo suggestion
- prove the error happens before scheduler state is created

## Why

An unknown pipeline filter currently produces an empty successful tick:
exit `0`, zero evaluated pipelines, and no skip or error entry. A typo in an
unattended cron/systemd/CI command can therefore stop scheduled work silently.

## Root cause

The tick wrapper passed the optional name directly to
`resolve_all_schedules`, where it was used only as a filter predicate. A name
that matched nothing became an empty schedule map. The wrapper now validates
explicit names through the same `registry::resolve_pipeline` helper used by
other CLI commands before any state I/O.

Closes #1163.

## Test plan

- `cargo test -p rocky-cli --lib unknown_pipeline_filter_fails_before_state_open`
- `cargo test -p rocky-cli --lib commands::tick::tests`
- `git diff --check`

Local Rust 1.93.1 clippy is blocked by existing warnings in unchanged
`rocky-lang`, `rocky-sql`, and `rocky-cli/src/commands/list.rs` code. The
canonical PR CI remains the readiness gate.

### DIFF
--- a/engine/crates/rocky-cli/src/commands/tick.rs
+++ b/engine/crates/rocky-cli/src/commands/tick.rs
@@ -47,9 +47,10 @@ const VERSION: &str = env!("CARGO_PKG_VERSION");
 ///
 /// # Errors
 ///
-/// Returns an error (exit 1) when the config cannot be loaded, the state store
-/// cannot be opened, or the tick itself faults (state I/O, lock I/O, an `after`
-/// cycle). Every such case runs nothing — fail closed.
+/// Returns an error (exit 1) when the config cannot be loaded, the selected
+/// pipeline does not exist, the state store cannot be opened, or the tick itself
+/// faults (state I/O, lock I/O, an `after` cycle). Every such case runs nothing
+/// — fail closed.
 pub async fn run_tick(
     config_path: &Path,
     state_path: &Path,
@@ -71,6 +72,9 @@ pub async fn run_tick(
     // Config parse error ⇒ exit 1, nothing runs.
     let config = load_rocky_config(config_path)
         .with_context(|| format!("failed to load config from {}", config_path.display()))?;
+    if let Some(name) = pipeline.as_deref() {
+        crate::registry::resolve_pipeline(&config, Some(name))?;
+    }
 
     // The only wall-clock read in the reconciler. Everything downstream takes
     // this instant as a parameter.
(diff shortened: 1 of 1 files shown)


hugocorreia90 (approved):
Approving — clean, minimal, correctly-scoped fix for #1163.

**What I verified**

- **Scoping is right.** The guard runs only for an explicit `--pipeline` name (`if let Some(name) = pipeline.as_deref()`). Validating the `None` case through `resolve_pipeline` would have errored on the 0-or-many-pipelines path and broken tick-everything — the PR correctly avoids that.
- **Precise superset check.** `registry::resolve_pipeline` looks up the same `config.pipelines` map that `resolve_all_schedules` filters against as a name predicate (`reconcile.rs:517-522`). So it rejects exactly the unknown names that previously fell through to an empty schedule map (the silent-success bug) and never a name tick could legitimately evaluate. A defined-but-*unscheduled* pipeline still passes and yields an empty tick — unchanged, and appropriately out of scope for this issue.
- **Fails before any state I/O.** The check sits immediately after config load, ahead of `create_dir_all` and `tick_once`, so it's fail-closed. The `assert!(!state_path.exists())` in the new test locks that invariant in.
- **Reuses the canonical helper**, inheriting the "Did you mean?" suggestion; the test wisely asserts only the stable `pipeline 'typo' not found in config` substring, not the fuzzy suggestion.
- **`# Errors` doc updated** to list the new exit-1 condition.
- **No JSON-schema cascade** — no `*Output` struct touched, and the codegen-drift check is green.

CI is fully green (Test, Clippy, Format, codegen, adapter boundary, credential containment, release smoke) and the branch merges clean. LGTM.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1166, opened 2026-07-18
fix(engine): align model plan scope with apply

### LINKED ISSUE #1165: `plan --model` advertises replication SQL and unrelated models that `apply` skips
## Description

`rocky plan --model <name>` persists the selector for `apply`, but its public
preview ignores that selector. The plan advertises replication SQL and reports
every compiled model in `models` / `execution_layers`, while `rocky apply`
correctly skips replication and executes only the named model.

This makes the documented plan/apply review boundary inaccurate even though
both commands succeed.

## Steps to Reproduce

1. Create a DuckDB project with:
   - one replication source table,
   - two compiled models named `known` and `other`,
   - `auto_create_schemas = true`.
2. Run:

   ```console
   rocky --config rocky.toml --output json plan --model known
   ```

3. Observe that `statements` contains replication DDL/copy SQL and that
   `models` and `execution_layers` contain both `known` and `other`.
4. Apply the returned plan ID:

   ```console
   rocky --config rocky.toml --output json apply <plan-id>
   ```

5. Inspect the target database.

Reproduced twice from a clean checkout of `main` at
`b08e9972adc85b0446e7d7157b1a5b3714ef2fa7`.

## Expected Behavior

The plan preview should match the persisted execution intent:

- no replication statements for `--model`,
- transformation SQL only for the selected model,
- `models: ["known"]`,
- execution layers containing only `known`.

## Actual Behavior

Both commands exit successfully. The plan advertises replication and both
models, but apply reports `tables_copied: 0`, materializes only `known`, does
not materialize `other`, and does not create the advertised replication target.

A control using bare `rocky plan` followed by `apply` executes the same
advertised replication SQL successfully.

## Rocky Version

`rocky 1.65.0`; affected code is unchanged between the latest release and
current `main`.

## OS / Architecture

macOS 26.5.2 arm64

## Configuration

Credential-free local DuckDB adapters; one discovery-backed replication
pipeline and two full-refresh models. No secrets or external services are
required.

mattfaltyn:
## Summary

Make `rocky plan --model <name>` describe the same execution scope that
`rocky apply` will run.

Closes #1165.

## Subproject(s) touched

## Changes

- Reuse the existing transformation SQL preview for model-scoped plans instead
  of returning unrelated replication statements.
- Persist and report only the selected model and its execution layer.
- Reject an unknown model during preview instead of producing an empty or
  misleading plan.
- Add a real-binary DuckDB regression proving the previewed, persisted, and
  applied scopes match.

## Why

The model selector was persisted for apply-time execution but ignored by the
live SQL preview and informational model metadata. A valid plan therefore
advertised replication SQL and unrelated models that a successful apply
silently skipped, undermining the plan/apply review boundary.

## Impact

Operators and CI can now trust that a model-scoped plan's SQL, `models`, and
`execution_layers` describe what apply will execute. The JSON schema is
unchanged.

## Test Plan

- `cargo test -p rocky --test plan_model_scope`
- `cargo test -p rocky-cli --lib commands::plan`
- `cargo test -p rocky-cli -p rocky --lib --bins --tests`
- `cargo fmt -- --check`
- `cargo clippy -p rocky-cli -p rocky --all-targets --all-features -- -D warnings -A unused-assignments -A clippy::collapsible-else-if`

The two clippy allowances cover existing Rust 1.93 warnings outside this diff;
all other warnings remain denied.

## Checklist

- [x] Commit messages follow conventional commits
- [x] CLI JSON schema shape is unchanged
- [x] Engine tests pass
- [x] Clippy is clean apart from explicitly identified upstream baseline lints

### DIFF
--- a/engine/CHANGELOG.md
+++ b/engine/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- `rocky plan --model <name>` now previews only the selected model and reports the same model scope that `rocky apply` executes, instead of advertising skipped replication SQL and unrelated models. (#1165)
+
 ## [1.65.0] - 2026-07-18
 
 ### Security
--- a/engine/crates/rocky-cli/src/commands/plan.rs
+++ b/engine/crates/rocky-cli/src/commands/plan.rs
@@ -18,9 +18,9 @@ use crate::registry;
 use super::run::PartitionRunOptions;
 use super::{filter_table_matches, matches_filter, parse_filter};
 
-/// Bundle of `rocky plan` flags that are not consumed by the SQL-generation
-/// preview but are persisted into `RunPlan` so `rocky apply <plan-id>` can
-/// honour them. Mirrors the flag surface of `rocky run`.
+/// Bundle of `rocky plan` execution flags persisted into `RunPlan` so
+/// `rocky apply <plan-id>` can honour them. Mirrors the flag surface of
+/// `rocky run`; `model` also scopes the SQL preview and model metadata.
 #[derive(Debug, Default, Clone)]
 pub struct PlanRunOptions {
     pub model: Option<String>,
@@ -397,6 +397,16 @@ pub async fn plan(
         .models_dir
         .clone()
         .unwrap_or_else(|| models_dir.clone());
+    if let Some(model) = run_options.model.as_deref() {
+        anyhow::ensure!(
+            blueprint_models_dir.exists(),
+            "models directory '{}' not found (required for --model)",
+            blueprint_models_dir.display()
+        );
+        output.statements =
+            plan_preview_output(Some(config_path), &blueprint_models_dir, Some(model), env)?
+                .statements;
+    }
     let mut run_plan_persisted = false;
     if blueprint_models_dir.exists() {
         match build_and_persist_run_plan(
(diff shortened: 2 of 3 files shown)


hugocorreia90 (approved):
Reviewed and red-teamed (Codex adversarial pass; all findings verified against a real DuckDB binary rather than by code-trace alone).

**The fix is sound.** `plan --model <name>` now scopes `statements` (transformation SQL for the named model only — no replication SQL), `models`, and `execution_layers` to exactly what `apply` executes. Confirmed empirically:
- `--model known` → 1 statement (`fixture.main.known`), `models=["known"]`, `execution_layers=[["known"]]`, `plan_kind=run`.
- `--model nonexistent` (project with models) → errors `model 'nonexistent' not found` — the "reject unknown model" intent holds where it matters.
- `plan` (no `--model`) → unchanged.

`models`/`execution_layers` are informational (apply executes off the persisted `model` selector, `apply.rs:664`), so hard-coding them to the selected model is reporting-alignment, not a behavior change. The real-binary regression test proves preview == persisted == applied scope, including the negative (that `other` and the replication target are not materialized).

**Red-team edges (both non-blocking, tracked in #1171):**
1. A run-plan-persist failure while `--model` is set can pair transformation statements with a replication `plan_id`. Reproducible only by forcing a path-specific write failure; the swallow-and-fall-back-to-replication root cause is pre-existing.
2. (to confirm) `--models <custom-dir>` at a zero-model dir with no conventional `models/` may bypass the guard and degrade similarly.

Neither is a common-path regression. Approving and merging.

Issue #1120, opened 2026-07-15
Governed-apply-on-remote-state integrity: config-snapshot + state-transaction + CAS (freeze kill-switch hardening)


hugocorreia90:
## Summary

The agent-policy **freeze kill-switch** and, more broadly, the **governed-apply-on-remote-state integrity plane** need an architectural pass. PR #1117 hardened the mechanical surface of `#1089-S1` (fail-closed seam sync, crash-safe lock-serialized local-only-preserving download, cross-pod freeze *enforcement* on the governed paths, mutation durability, single config snapshot for the policy *decision*). But a **fully tamper-proof cross-pod freeze cannot be reached by incremental fixes** — a 7-round adversarial review kept surfacing the same *classes* of finding in new call sites. Three residual limitations are documented in-code and were consciously landed with caveats; this issue tracks the redesign that closes them.

## The three residual limitations (documented in-code in #1117)

1. **Config-swap TOCTOU.** The governed `run` / `promote` / `propose` *execution* paths re-read `rocky.toml`, and `config_policy_identity` binds only `[adapter]` + `[pipeline]` — not `[policy]` / `[state]`. An attacker with local filesystem write access can swap `[policy]`/`[state]` on disk **between the apply gate and execution** to bypass the freeze/policy for those paths. (Doc: `config_policy_identity` and `GovernedRunContext::verify_routing_identity` in `engine/crates/rocky-cli/src/commands/apply.rs`.)
2. **No compare-and-swap.** Remote state sync is whole-file, last-writer-wins. A `run` whose start-download preceded a freeze can overwrite that freeze on its end-upload — a cross-pod lost update. A reliable *kill switch* cannot be last-writer-wins. (Doc: seam sync in `policy.rs`.)
3. **Tiered cache coherence.** On a `tiered` backend, a freeze written to the durable tier (S3) can be shadowed by a stale Valkey cache on the *read* side; a Valkey upload failure is swallowed even under `on_upload_failure = "fail"`. (Doc: `state_sync.rs`.)

Related, surfaced by the same review and only partially closed: replication has **no `verify_after` check substrate** (a governed replication apply carrying `verify_after` is currently *refused* rather than verified), and every governed mutation path needs a *consistent* download-before / upload-after transaction (restore/backfill/verify-after-custody were fixed one-by-one in #1117 — the pattern wants a single abstraction).

## Proposed architecture

[...]

PR #1170, opened 2026-07-18
feat(engine): WP-01 PR-D — compare-and-swap state writes (REFUSE-class finalize)

hugocorreia90:
## What

Delivers the **compare-and-swap** capability for remote `[state]` object writes — the last residual (residual 2) of #1120. A run whose start-download preceded a concurrent commit could previously overwrite it on end-upload (last-writer-wins); under `[state] concurrency_control = "cas"` the losing run now fail-closes instead of erasing the winner.

Residuals 1 (config-swap TOCTOU) and 3 (freeze-marker coherence) are already closed (PR-B threading; PR-F markers — see #1168). The freeze kill-switch itself is durable via PR-F markers; **this PR hardens the `state.redb` blob** against silent overwrite.

## Changes

**Primitives (`rocky-core/object_store.rs`)** — `Generation`, `PutIfMatchOutcome`, `put_if_match` (`PutMode::Create` for bootstrap, `PutMode::Update`/`If-Match` for steady-state; never an unconditional fall-through), `upload_file_if_match`, and `download_file_capturing_version` (base captured from the SAME GET — TOCTOU-free). `object_store` 0.14 already pins `S3ConditionalPut::ETagMatch`, so no client re-config and no redb schema bump.

**Session (`rocky-core/state_sync.rs`)** —
- `[state] concurrency_control = "off" | "cas"` (default `off` = today's unconditional put, **byte-identical**).
- Under `cas` on s3/gcs: `acquire` captures the base generation, `finalize` conditionally commits; a lost race → `CasConflict` (fail-closed, never `Skip`-swallowed, never retried as transient).
- Periodic uploader disabled under `cas` (finalize-only CAS is correct when nothing bumps the generation mid-run).
- Non-object-store backends auto-downgrade `cas` → unconditional + warn.

## Scope (one WP-01 slice)

REFUSE-class (run / backfill / load) whole-blob finalize CAS. **Follow-ups** (ADR-CONCURRENCY): the RETRY-class ledger seams (policy-freeze row, budget pair, gc/restore external-proof), tiered D5 cache coherence, and the shared-advancing-base periodic CAS write.

## Verification ladder

- **InMemory** — primitive tests (`put_if_match` create / update / stale-conflict; download-captured base) + a session-level cross-pod conflict test.
- **MinIO (real S3 wire path)** — ran `live_put_if_match_cas_roundtrip` against a local MinIO: bootstrap create-if-absent, matching-base commit, stale-base conflict, and download-captured base all verified. This drives the exact `AmazonS3Builder` / `PutMode::Update` / `If-Match` / `PutResult → generation` path InMemory can't.
- **Live AWS** — the same `#[ignore]` test is the formal gate (`ROCKY_TEST_S3_*` + the AWS env credential chain; also runs against MinIO via `AWS_ENDPOINT`). I couldn't run it against AWS this session — the sandbox IAM user lacks `s3:CreateBucket` / `ListAllMyBuckets` and I found no writable bucket. **Point me at a bucket + prefix I can write under and I'll run the AWS rung.**

`clippy --all-targets --all-features`, `fmt --check`, and the full `rocky-core` suite are green. Codegen regenerated (`concurrency_control` in schema + Pydantic + TS + openapi).

## Closure

Even a complete PR-D does not *close* #1120: RD-002 closure needs the operational rollout (deploy fleet-wide, set `concurrency_control = "cas"` on live prefixes, verify, then drop the interim one-writer-per-prefix serialization). This PR delivers the capability; #1120 stays open.

## Note (RD-006 / WP-02)

Refuse-on-conflict leaves a warehouse a run already wrote ahead of committed state; for append strategies an operator re-run can duplicate rows. Pre-exists (crash-between-write-and-commit), but refuse-on-conflict adds a trigger — tracked separately, not solved here.

Refs #1120

### DIFF
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -5205,6 +5205,25 @@
           }
         ]
       },
+      "ConcurrencyControl": {
+        "description": "Concurrency control for remote `[state]` object writes.\n\nGuards the cross-pod lost update where two runs sharing one `[state]` prefix race: one downloads the state, the other commits, and the first's end-of-run upload silently overwrites the second (last-writer-wins). See [`StateConfig::concurrency_control`].",
+        "oneOf": [
+          {
+            "description": "Unconditional last-writer-wins upload (default) — byte-identical to the pre-CAS behaviour. Correct for single-writer-per-prefix deployments (one run at a time, orchestrator-serialized).",
+            "enum": [
+              "off"
+            ],
+            "type": "string"
+          },
+          {
+            "description": "Compare-and-swap: the end-of-run upload is conditional on the remote object still carrying the generation the run downloaded. A run that lost a cross-pod race fail-closes (nonzero exit) instead of erasing the winner. Requires a backend with a conditional-write object tier (`s3` / `gcs`); auto-downgrades to `off` (with a warn) on other backends.",
+            "enum": [
+              "cas"
+            ],
+            "type": "string"
+          }
+        ]
+      },
       "ConcurrencyMode": {
         "anyOf": [
           {
@@ -15643,6 +15662,7 @@
             ],
             "default": {
               "backend": "local",
+              "concurrency_control": "off",
               "freeze_marker_writes": false,
               "gcs_bucket": null,
               "gcs_prefix": null,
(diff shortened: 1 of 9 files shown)


hugocorreia90:
**Live-AWS gate: GREEN.** Ran `live_put_if_match_cas_roundtrip` against a real AWS S3 bucket (`us-east-1`): bootstrap create-if-absent, a second-bootstrap 412, matching-base `If-Match` commit, stale-base 412 `PreconditionFailed` conflict, and the download-captured base all verified against real AWS — the top rung above the MinIO run. CAS behaves as designed on genuine S3.

Verification ladder is now complete: InMemory → MinIO (real S3 code path) → **AWS**. The `#[ignore]` test stays in the tree as the reproducible gate.

PR #1174, opened 2026-07-18
fix(engine): fail fast on unknown model in rocky plan --model

hugocorreia90:
## Summary

Follow-up to #1166. Make `rocky plan --model <name>` fail fast with a clear
"model not found" error — matching what `rocky apply` reports — when the named
model does not exist, instead of a confusing governance error or a misleading
empty preview.

Relates to #1165.

## Subproject(s) touched

## What was wrong (post-#1166)

#1166 correctly scoped `plan --model` for the common case (a real model in a
populated `models/` dir). Two residual gaps remained for the *unknown/zero-model*
case:

1. **Confusing error.** `rocky plan --model typo` in a project whose `models/`
   directory compiles to zero models failed at the governance-action preview
   with `failed to compute governance action preview`, because that step runs
   before the model selector was validated.
2. **Silent, misleading preview (the dangerous one).** With a `--models`
   override pointing at a model-less directory (and no conventional `models/`),
   the governance preview was skipped, `plan_preview_output` returned an *empty*
   statement list, and `plan()` fell through to persist a **full-replication**
   plan. Applying that plan id then ran replication across every discovered
   table — work the empty preview never advertised.

Reachability, stated honestly: (2) is not a routine invocation — it needs a
`--models` override to a zero-model directory. The common case in (1) already
errored (non-silently), just with the wrong message.

## Changes

- Move `--model` validation ahead of the governance/budget preview in `plan()`,
  so an unknown model (including a zero-model project) fails fast with the same
  `model 'X' not found (no transformation model with that name)` message
  `rocky run`/`rocky apply` emit — before any plan is persisted.
- Make `plan_preview_output` return a typed `ModelNotFound` error for a `--model`
  filter against a zero-model project (both the `NoModels` and empty-project
  paths), instead of returning an empty preview.
- `rocky mcp` `plan_preview` now classifies an unknown model as `model_not_found`
  (with its "list the models, retry" hint) rather than the generic
  `compile_failed` bucket, by downcasting the typed error.

No CLI JSON output struct or schema changed; the `ToolErrorCode::ModelNotFound`
variant already existed. `ModelNotFound`'s `Display` is byte-identical to the
run/apply message, so CLI output is unchanged.

## Test Plan

- `cargo test -p rocky-cli --lib commands::plan` (33 pass, incl. new
  `plan_preview_filter_on_empty_models_dir_errors`)
- `cargo test -p rocky --test plan_model_scope` (adds real-binary
  `model_plan_on_empty_models_dir_errors`: plan fails, persists no plan)
- `cargo test -p rocky-mcp --lib plan_preview_unknown_model` (asserts
  `ModelNotFound` taxonomy)
- `cargo clippy -p rocky-cli -p rocky-mcp -p rocky --all-targets --all-features -- -D warnings`
- `cargo fmt -- --check`

## Out of scope (pre-existing, non-silent)

- `rocky plan` (no `--model`) against an empty conventional `models/` still fails
  at the governance preview; that behavior is unchanged and orthogonal to model
  scoping.

### DIFF
--- a/engine/CHANGELOG.md
+++ b/engine/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `rocky plan --model <name>` now previews only the selected model and reports the same model scope that `rocky apply` executes, instead of advertising skipped replication SQL and unrelated models. (#1165)
+- `rocky plan --model <name>` now fails fast with a "model not found" error when the named model does not exist — including a project that compiles to zero models — matching what `rocky apply` reports. Previously this surfaced as a confusing governance-preview failure, or (with a `--models` override pointing at a model-less directory) as an empty preview that masked a persisted full-replication plan `rocky apply` would then execute. The `rocky mcp` `plan_preview` tool now classifies an unknown model as `model_not_found` rather than `compile_failed`. (#1165)
 
 ## [1.65.0] - 2026-07-18
 
--- a/engine/crates/rocky-cli/src/commands/mod.rs
+++ b/engine/crates/rocky-cli/src/commands/mod.rs
@@ -145,8 +145,8 @@ pub use lsp::run_lsp;
 pub use metrics::{metrics_output, run_metrics};
 pub use optimize::{optimize_output, run_optimize};
 pub use plan::{
-    PlanRunOptions, compute_embedded_capabilities, plan, plan_preview_output, plan_promote,
-    populate_governance_actions,
+    ModelNotFound, PlanRunOptions, compute_embedded_capabilities, plan, plan_preview_output,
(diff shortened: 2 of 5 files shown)


hugocorreia90:
Superseded. While this was in review, #1172 landed on `main` and comprehensively fixed the silent `--model` → replication degradation (edges: `--models` zero-model dir, run-plan persist failure, `--model`+`--dag`, and a TOCTOU guard) — more thoroughly than this PR's approach, and it deliberately left `plan_preview_output` untouched.

The only confirmed item #1172 did not cover is the MCP error-taxonomy residual: `rocky mcp`'s `plan_preview` still reports an unknown/typo'd model as `compile_failed` instead of `model_not_found`. Closing this in favour of a fresh, minimal branch scoped to just that fix (off current `main`, conflict-free with #1172).

PR #1180, opened 2026-07-19
fix(engine/rocky-cli): refresh targets without watermarks

### LINKED ISSUE #1179: Incremental replication duplicates all rows when watermark state is missing
### Description

When an incremental target table already exists but its persisted watermark is
missing, Rocky keeps the incremental strategy and emits an `INSERT` without a
real lower bound. The dialect layer substitutes an epoch sentinel, so the
entire source is appended to the existing target. The run exits successfully
and reports an incremental materialization even though every source row was
duplicated.

This can happen after local state loss, a fresh worker/pod starts against an
existing warehouse, or a remote-state schema/key change makes the previous
watermark unavailable.

### Steps to reproduce

1. Run the shipped DuckDB POC pipeline so its incremental target contains 500
   rows.
2. Remove or move Rocky's state database while leaving the target table in
   place.
3. Run the same pipeline again.
4. Compare total and distinct target rows.

The reproduction is deterministic:

```text
initial run: 500 total / 500 distinct
after state loss and one rerun: 1000 total / 500 distinct
after another rerun: 1500 total / 500 distinct
```

Each rerun exits with status 0 and reports strategy `incremental`.

### Expected behavior

An incremental or microbatch materialization with no prior watermark should
use a full refresh, replacing the existing target and rebuilding the
watermark. The target should remain at 500 total / 500 distinct rows and the
materialization metadata should report `full_refresh`.

If the watermark store cannot be read, the run should fail rather than
silently treating the error as a missing watermark and appending a full-source
scan.

### Root cause

`process_table` decides whether to full-refresh only from target-table
existence. It then converts both a missing watermark and a watermark read
failure to `None`, but retains the incremental strategy. SQL generation maps
that `None` to the dialect's epoch sentinel, producing a full-source `INSERT`
into the existing target.

### Suggested fix

For incremental and microbatch strategies:

- present watermark: continue incrementally;
- absent watermark: select full refresh;
- watermark read failure: propagate the error.

Add a dispatch-level regression test that starts with an existing target and
an empty state store, then verifies replacement semantics, a `full_refresh`
outcome, and a newly deferred watermark.

### Environment

- Rocky 1.66.0
- `main` at `851592df002134a6b1c706c10813857e2ba27070`
- macOS arm64
- DuckDB 1.5.2

mattfaltyn:
## Summary

Prevent incremental and microbatch replication from appending a full source snapshot when a persisted watermark is missing.

Closes #1179.

## Subproject(s) touched

## Changes

- Resolve watermark-dependent strategy state before prune and drift handling.
- Force a replacement full refresh when an existing incremental or microbatch target has no watermark.
- Fail before target mutation when the state read fails instead of silently falling back to an epoch-bounded append.
- Add an in-memory DuckDB regression that verifies recovery does not duplicate rows and the following run returns to incremental behavior.

## Root cause and impact

An existing target selected the incremental append path even when its watermark had been lost. The absent value became a 1970 lower bound, so every source row was appended to the existing target. Repeating the run after state loss silently multiplied all rows while still exiting successfully.

## Test Plan

- [x] `CARGO_BUILD_JOBS=4 RUSTFLAGS=-Dwarnings cargo +1.97.1 nextest run --all-features`
- [x] `CARGO_BUILD_JOBS=4 RUSTFLAGS=-Dwarnings cargo +1.97.1 clippy --all-targets --all-features -- -D warnings`
- [x] `cargo +1.97.1 fmt -- --check`
- [x] `cargo test -p rocky-cli --lib commands::export_schemas`
- [x] `python3 -m unittest discover -s .github/tests -p 'test_*.py' -v` (100 passed)
- [x] `RUSTUP_TOOLCHAIN=1.97.1 just codegen`
- [x] `RUSTUP_TOOLCHAIN=1.97.1 just regen-fixtures`
- [x] Generated schemas, OpenAPI, SDK types, editor types, and Dagster fixtures are drift-clean.
- [x] Default-feature release binary passed the tiered-cache construction smoke condition.

## Checklist

- [x] No CLI JSON schema or DSL syntax changed.

### Engine (`engine/`)

- [x] Full CI-equivalent nextest suite passes with all features.
- [x] `cargo clippy --all-targets --all-features -- -D warnings` is clean.
- [x] `cargo fmt --check` passes.

### DIFF
--- a/engine/crates/rocky-cli/src/commands/run.rs
+++ b/engine/crates/rocky-cli/src/commands/run.rs
@@ -9470,44 +9470,75 @@ fn resolve_merge_update_columns(
 /// `MAX(<timestamp_column>) FROM <target>` — the post-execute watermark for the
 /// replication path. Reads the **target** table so the recorded value reflects
 /// exactly what was loaded (closing the source-side TOCTOU described on
-/// [`resolve_new_watermark`]). Returns `None` on NULL / empty / failure so the
-/// caller can keep the prior watermark.
+/// [`resolve_new_watermark`]). Returns `Ok(None)` only when the aggregate is
+/// NULL (an empty target). Query, shape, and parse failures propagate so they
+/// cannot be mistaken for "no progress" and persisted as a valid epoch
+/// watermark.
 async fn query_target_max_timestamp(
     warehouse: &dyn WarehouseAdapter,
     dialect: &dyn rocky_core::traits::SqlDialect,
     target: &TableRef,
     timestamp_column: &str,
-) -> Option<chrono::DateTime<Utc>> {
-    if rocky_sql::validation::validate_identifier(timestamp_column).is_err() {
-        return None;
-    }
+) -> Result<Option<chrono::DateTime<Utc>>> {
+    rocky_sql::validation::validate_identifier(timestamp_column)
+        .with_context(|| format!("invalid watermark column '{timestamp_column}'"))?;
     let target_ref = dialect
         .format_table_ref(&target.catalog, &target.schema, &target.table)
-        .ok()?;
+        .map_err(anyhow::Error::from)
+        .with_context(|| {
+            format!(
+                "could not format target table {} for watermark query",
+                target.full_name()
+            )
+        })?;
     let sql = format!("SELECT MAX({timestamp_column}) FROM {target_ref}");
 
-    let result = match warehouse.execute_query(&sql).await {
-        Ok(r) => r,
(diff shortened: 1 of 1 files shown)


hugocorreia90 (approved):
Verified missing-watermark replacement semantics and fail-closed handling for post-refresh watermark query failures. Targeted watermark tests, direct helper tests, formatting, and all-targets/all-features Clippy pass.

PR #1184, opened 2026-07-20
fix(engine): refuse unsafe raw dbt incrementals

### LINKED ISSUE #1183: import-dbt --no-manifest strips incremental filters and duplicates rows
## Description

`rocky import-dbt --no-manifest` recognizes a standard dbt
`{% if is_incremental() %}` block, removes that block from the emitted SQL,
and emits a Rocky `incremental` sidecar. Transformation incrementals
intentionally execute `INSERT INTO <target> <model SQL>` without adding a
watermark predicate, so the generated model appends the entire source on every
rerun.

The import reports zero warnings and zero failures, and the generated project
passes both `rocky compile` and `rocky validate`.

When the dbt model also declares `{{ config(materialized='incremental') }}`, the
detected timestamp column is not applied: the emitted sidecar uses the fallback
`timestamp_column = "updated_at"` even when the removed predicate used
`event_time`.

## Steps to reproduce

Create a minimal dbt project with this model:

```sql
{{ config(materialized='incremental') }}

SELECT *
FROM (
    VALUES
        (1, TIMESTAMP '2026-01-01 00:00:00'),
        (2, TIMESTAMP '2026-01-02 00:00:00'),
        (3, TIMESTAMP '2026-01-03 00:00:00')
) AS source_events(id, event_time)
{% if is_incremental() %}
WHERE event_time > (SELECT MAX(event_time) FROM {{ this }})
{% endif %}
```

Then run:

```bash
rocky import-dbt \
  --dbt-project ./dbt-project \
  --no-manifest \
  --output-dir ./rocky-out \
  --output json

cd rocky-out
rocky compile --output json
rocky validate --output json
rocky run --pipeline default --output json
duckdb warehouse.duckdb \
  "SELECT COUNT(*), COUNT(DISTINCT id) FROM warehouse.main.events;"
```

Repeat the final `rocky run` and query twice.

## Expected behavior

The importer must preserve bounded incremental semantics or fail loudly rather
than emit a runnable incremental model after deleting its only row filter. With
an unchanged source, repeated runs should remain at `3 total / 3 distinct`.

## Actual behavior

All commands exit successfully. `import-dbt` reports `warnings: 0` and
`failed: 0`; compile reports `has_errors: false`; validation reports
`valid: true`; every run reports `status: "Success"`.

```text
first run:  3 total / 3 distinct
second run: 6 total / 3 distinct
third run:  9 total / 3 distinct
```

The emitted SQL contains no `WHERE`, and the sidecar names `updated_at`
instead of the detected `event_time`.

## Root cause

- `import_single_model` calls `detect_is_incremental`, which removes the
  incremental branch before conversion.
- The detected timestamp only replaces a `FullRefresh` strategy; a normal
  `config(materialized='incremental')` already maps to
  `Incremental { timestamp_column: "updated_at" }`.
- Transformation SQL generation deliberately emits a bare insert and expects
  model SQL to own filtering.

The existing importer test verifies that the Jinja block disappears but does
not execute the emitted model twice.

## Resolution

#1184 implements the fail-closed option on every raw SQL path
(`--no-manifest` and manifest nodes without `compiled_code`). A model whose raw
Jinja invokes `is_incremental()` is reported as `imported: 0`, `failed: 1`,
with its reason in `failed_details`, and no model SQL or sidecar is emitted.
Detection covers the standard guard, compound `if` / `elif` conditions,
whitespace inside the call, and indirect `{% set %}` forms.

The refusal also applies when the model has a `unique_key`. A full-source merge
is idempotent by key, but it is not equivalent to the bounded dbt query: it can
rewrite historical rows or recompute nondeterministic expressions.

The command retains its existing partial-import exit contract, so callers must
inspect `failed` / `failed_details` even when the process exits successfully.
Compiled manifest models remain supported after the user verifies that the
compiled SQL retains the intended predicate and is valid for Rocky's initial
target state.

## Suggested validation

Assert that both append-only and keyed raw models are rejected with
`imported: 0`, `failed: 1`, an actionable `failed_details` reason, and no model
artifacts. Cover standard, compound, and indirect `is_incremental()` Jinja
forms on both raw paths. A repeated-run row-count test is only applicable if a
future implementation can preserve dbt's bootstrap and incremental branches
rather than rejecting the conversion.

## Environment

- Rocky `main@9c502d1aabd6f79ef8942076e913c8a8ad17edc9`
- Rocky 1.66.0
- macOS 26.5.2 arm64
- DuckDB 1.5.2

mattfaltyn:
## Summary

Fail closed whenever a raw dbt model still contains Jinja that invokes
`is_incremental()`. Rocky cannot preserve dbt's
false-on-bootstrap/true-on-existing-target distinction from raw SQL, so the
model is now reported under `failed_details` and no model SQL or sidecar is
emitted.

This applies to both append-only models and models with a `unique_key`: a
full-source merge is idempotent by key, but it is not semantically equivalent
to the bounded dbt query.

Closes #1183.

## Root cause

The raw importer called `detect_is_incremental`, which removed the guarded
predicate before conversion. For append-only incremental materialization,
Rocky then reused the remaining unfiltered query for every `INSERT`,
duplicating the full source on each run. That detector only recognized the
exact `{% if is_incremental() %}` spelling; compound conditions could instead
fall through the generic Jinja fallback, which applies conditional bodies
unconditionally.

Allowing the same rewrite for a keyed merge avoids duplicate keys, but still
changes which source rows are evaluated and can rewrite historical rows or
recompute nondeterministic expressions. Raw guards therefore need one
strategy-independent, fail-closed rule.

## Changes

- Refuse any raw Jinja statement or expression that invokes
  `is_incremental()` on both raw SQL paths: `--no-manifest` and manifest nodes
  without `compiled_code`. This covers compound `if` / `elif` conditions,
  whitespace inside the call, and indirect `{% set %}` forms.
- Apply the refusal before strategy inference or Jinja conversion, including
  keyed merge models.
- Add compiler coverage for append and keyed models on both raw paths.
- Add a real-binary regression proving the CLI reports the model failure and
  emits no SQL or sidecar artifacts.
- Correct the migration guide, command reference, changelog, and generated
  migration notes to describe the actual semantics and remediation.

The CLI keeps its existing partial-import contract: individual model failures
are returned in `failed` / `failed_details` while the process exits
successfully. Compiled manifest models are unchanged; users should still verify
that compiled SQL retains the intended predicate and is valid for Rocky's
initial target state.

## Validation

- `cargo test -p rocky-compiler` — 394 unit tests and 22 integration tests
  passed; doc tests passed (1 ignored).
- `cargo test -p rocky --test import_dbt_incremental -- --nocapture` — passed.
- `cargo test -p rocky-cli --test import_dbt_emit -- --nocapture` — passed.
- `cargo fmt --all -- --check` and `git diff --check` — passed.
- `npm run build` in `docs/` — passed; 98 pages built.
- GitHub Actions — all 8 required checks passed.
- Patch-specific Clippy passed:
  - `cargo clippy -p rocky-compiler --all-targets --all-features -- -D warnings -A unused-assignments`
  - `cargo clippy -p rocky --test import_dbt_incremental -- -D warnings -A unused-assignments -A clippy::collapsible-else-if`

The unmodified strict Clippy command is currently blocked locally by
pre-existing Rust 1.93 `unused_assignments` diagnostics from miette-derived
structs and a `collapsible_else_if` diagnostic in
`rocky-cli/src/commands/list.rs`; none are in this diff.

### DIFF
--- a/docs/src/content/docs/guides/migrate-from-dbt.md
+++ b/docs/src/content/docs/guides/migrate-from-dbt.md
@@ -290,7 +290,6 @@ What the importer translates cleanly:
 - `{{ var('name') }}` / `{{ var('name', default) }}` → an `@var(name)` / `@var(name, default)` run-variable marker left in the emitted SQL, resolved at run time by `rocky run --var name=value` (see [Handle unsupported Jinja](#3-handle-unsupported-jinja))
 - dbt **tags** (node- and folder-level) → the sidecar `[tags]` block (`<tag> = "true"`)
 - `{{ this }}` → the model's own fully-qualified `catalog.schema.table`
-- `is_incremental()` branches → stripped (Rocky derives the watermark filter from `[strategy]`)
 - **dbt generic tests** (`unique`, `not_null`, `accepted_values`, `relationships`) are translated column-by-column to `[[tests]]` blocks — including the *configured* forms that carry `severity:` (a `warn` becomes a Rocky warning, not a hard error) and `where:` (a row filter) (see [Generic test mapping](#generic-test-mapping) below)
 - model-level **`dbt_utils.unique_combination_of_columns`** → a Rocky `composite` uniqueness `[[tests]]` block over the same column tuple (the columns come from the test config, so no model schema is needed)
 - Top-level `dbt_project.yml`, used to detect project name and seeds path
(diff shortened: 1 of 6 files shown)


hugocorreia90 (dismissed):
Approved at e06c959b. The reported unsafe raw incremental behavior and the common compound, negated, whitespace, keyed, and no-manifest variants are fixed and verified. Non-blocking alias hardening and stale example cleanup are tracked in #1194.

hugocorreia90 (approved):
Re-approved at 4f23cc6a after resolving the additive CHANGELOG conflict with main. The only merge resolution retained both independent Unreleased/Fixed bullets. Non-blocking follow-up remains tracked in #1194.

PR #1200, opened 2026-07-21
feat(engine/rocky-compiler): validate merge unique_key columns exist in model output

hugocorreia90:
## The gap

`unique_key` was never validated against a model's output schema. A typo compiled clean and only failed later, when the warehouse rejected the generated `MERGE ... ON` clause.

The neighbouring `time_column` field already has four dedicated diagnostics (E020–E023). `unique_key` had none. This closes that asymmetry.

## Scope — read this first

**This covers `merge`. It does not cover snapshots.**

Worth stating plainly, because SCD2 snapshots are the most painful place to hit this bug and a reviewer could reasonably assume this PR catches it. It does not:

| Surface | Covered | Why |
|---|---|---|
| `StrategyConfig::Merge { unique_key }` | Yes | Lives on `Model.config.strategy`, same place as `TimeInterval { time_column }`. Fully compiler-visible with a typed output schema in hand. |
| `SnapshotPipelineConfig.unique_key` | **No** | A `rocky.toml` `[pipeline.*]` block, not a compiled model. `rocky-compiler` has no reference to `SnapshotPipelineConfig` at all, and `ModelIr::snapshot()` is built at run time in `run_local.rs` with `typed_columns: Vec::new()`. There is no output schema to validate against at compile time. |
| Replication `merge_keys` | **No** | Same reason — pipeline config, not a compiled model. |

Catching the snapshot case needs validation at config-load or run time against the source table's schema, which is a different mechanism in a different crate. Flagging it rather than forcing a check the compiler cannot soundly make.

Per the three spellings (`unique_key` on merge, `unique_key` on snapshot, `merge_keys` on replication): deliberately **not** unified here. That is a separate design decision.

## The check

E036, the merge-key sibling of E020. `check_merge_strategy` mirrors `check_time_interval_strategy` in structure, message voice, and test style, and runs from the same step-5 dispatch against the same typed column list.

One diagnostic per missing key, so a composite `unique_key` reports every typo in one compile instead of one per fix-recompile round.

**Skipped when the output schema is incomplete** — a `SELECT *` over an upstream with no known schema. There `typed_cols` doesn't describe the real output, so every key would look missing. This mirrors the existing `type_known` guard the time_interval checks apply before drawing type conclusions. Models with an explicit column list still validate normally even when their source types are unknown, because the key names are matched against the model's own SELECT list.

## Error, not warning — and the behavior change

**This is an error, and it can fail a build that previously compiled.** Calling that out explicitly rather than burying it.

The reasoning:

- A `unique_key` naming a column the model doesn't output is not a style issue. The generated `MERGE ... ON` clause references a column that does not exist, so the warehouse rejects it every time. There is no configuration where this works.
- So the compile-time error does not break any *working* pipeline. It relocates an already-certain runtime failure to compile time, which is the entire point.
- E020 — the identical existence check for `time_column` — is an error. A warning here would be inconsistent with its direct sibling.
- The only path to a false positive is the incomplete-schema case, which is explicitly guarded above.

Net: a project that starts failing on this PR was already broken; it just hadn't run the merge yet.

## Also in here

The E020–E026 doc comments in the diagnostic registry had drifted and described the **wrong condition for all seven codes** — e.g. `E020` was documented as "Missing `@start_date` placeholder" when the code emits it for "`time_column` not in model output schema". The accurate mapping already lived on `check_time_interval_strategy`; the registry now agrees with the code that emits it. Verified each of the seven against its emit site. Docs-only, no behavior change, but it's why the new E036 entry reads coherently next to its neighbours.

Docs tables in `concepts/compiler.md` and `advanced/failure-modes.md` updated for E036.

## Codegen

Not required, and verified rather than assumed: `Diagnostic.code` is a free-form `Arc<str>`, not an enumerated schema type. No `*Output` struct changed, and no existing code (`E035`) appears anywhere in `schemas/`, `types_generated/`, or the vscode generated types.

## Gates

All run locally from the worktree root:

| Gate | Result |
|---|---|
| `cargo test` | 3730 passed, 0 failed (exit 0) |
| `cargo clippy --all-targets --all-features` | clean, no warnings |
| `cargo fmt --check` | clean |

Also ran `cargo doc --no-deps -p rocky-compiler` to confirm the new intra-doc link resolves. It does; the 4 pre-existing unresolved-link warnings are unrelated and unchanged.

## Adjacent, not done

`update_columns` on the same `Merge` struct has the identical failure mode (a typo lands in the `MERGE ... SET` clause) and is not validated. Left out to keep this PR to the reported gap; it would be a small follow-up reusing `check_merge_strategy`.


hugocorreia90:
## False positive on `SELECT *` models — fixed in b4253bef

E036 as first written breaks valid builds. Minimal repro:

```sql
-- models/ingest_orders.sql
SELECT * FROM poc.raw.orders
```
```toml
# models/ingest_orders.toml
[strategy]
type = "merge"
unique_key = ["order_id"]
```

```
Error E036: unique_key 'order_id' is not in the output schema of model 'ingest_orders'
  suggestion: Available columns:
```

The empty `Available columns:` is the tell — the check ran against zero columns.

### Root cause

The guard asked *"does any upstream model have an unknown schema?"*:

```rust
let schema_incomplete = model_schema.has_star
    && model_schema.upstream.iter()
        .any(|up| typed_models.get(up).is_none_or(Vec::is_empty));
```

`ModelSchema::upstream` holds **only models, never raw sources** (it is populated from the DAG's `depends_on`). A root model reading a raw table has an empty `upstream`, so `.any()` is vacuously false, `schema_known` came out `true`, and the existence check ran against an empty `typed_cols`.

It was under-inclusive in a second way too: `SELECT * FROM known_model JOIN raw.orders ...` has a known upstream, so the guard was false — yet the star still pulls columns from a non-model source. Confirmed, reporting `Available columns: customer_id` with the raw side missing.

This is not a narrow edge case. `source_schemas` is seeded from an optional, TTL'd cache written by prior `run` / `discover --with-schemas`, so an unenumerable leaf schema is the normal cold state — a fresh clone, CI, or a first compile all hit it.

### Fix

Gate on `!model_schema.has_star`.

**Tradeoff, stated plainly:** this drops E036 coverage for every `SELECT *` model, including `SELECT * FROM single_fully_known_model` where the output *is* enumerable. That is deliberate:

- E036 is an **Error**. A false positive breaks a valid build; a missed typo is caught later by the warehouse rejecting the `MERGE ... ON`. The costs are not symmetric.
- On a star model the output set is not knowable at compile time in general, so a typo there arguably cannot be *proven* anyway.
- Explicit column lists — the common case, and the only place a typo is provable — keep full coverage.

The precise alternative (a `star_fully_expanded` flag computed in `build_semantic_graph`, where the knowledge actually lives) was considered and rejected for now: `ModelSchema` is `Serialize`/`Deserialize`, is built by struct literal in `lsp.rs` and `catalog.rs`, and `has_star` already surfaces in the catalog Arrow schema and `api.rs`. That is real serialization surface to buy back a minority shape. Worth revisiting separately if star models turn out to need the coverage.

`schema_incomplete` is retained unchanged for its original W002 role.

### Tests

The existing unit tests hand `schema_known` to `check_merge_strategy` directly, so **none of them could catch a caller that computes it wrong** — including `test_e036_skipped_when_schema_unknown`, which passes `false` literally and was green on the buggy head. They are correct in isolation, so they are kept as-is; nothing asserted the buggy behavior.

Four tests added through `typecheck_project_with_models`, which *derives* the flag:

| Test | Before | After |
|---|---|---|
| `test_e036_not_fired_for_select_star_over_non_model_source` | **FAILED** | ok |
| `test_e036_not_fired_for_select_star_joining_non_model_source` | **FAILED** | ok |
| `test_e036_still_fires_end_to_end_for_explicit_columns` | ok | ok |
| `test_e036_not_fired_end_to_end_for_correct_explicit_key` | ok | ok |

Both false-positive tests were verified failing against the previous head before the fix went in. The true-positive case is unweakened — verified through the real CLI as well:

```
Error E036: unique_key 'ordr_id' is not in the output schema of model 'ingest_orders'
  suggestion: Available columns: order_id, amount
```

Gates: `cargo test` (full engine workspace, exit 0), `cargo clippy --all-targets --all-features` (clean), `cargo fmt --check` (clean).

### Sibling finding: E020 has the same latent false positive — not fixed here

`check_time_interval_strategy` does the identical existence check with **no** `schema_known` guard at all. Confirmed reproducible on this branch:

```sql
SELECT * FROM poc.raw.events WHERE ts >= @start_date AND ts < @end_date
```
```
Error E020: time_column 'ts' is not in the output schema of model 'ingest_events'
  suggestion: Available columns:
```

Deliberately **out of scope** for this PR:

- It is **pre-existing and shipping**, not a regression. `check_time_interval_strategy` is untouched by this branch, so its behavior is identical to `origin/main`.
- It is **not the same one-line guard**. The function interleaves schema-dependent checks (E020/E021/E022/E025) with schema-independent ones (E023 identifier validation, E024/W003 placeholders, `first_partition`). A blanket early return would silently disable the placeholder checks, so gating it correctly is surgical.
- The existing mitigation pattern differs: E020 false positives have historically been fixed by *enumerating more* in `build_semantic_graph` (derived tables, `derived_sources`, `source_schemas`) rather than by guarding — see `test_e020_not_fired_for_select_star_over_derived_table`. Which of the two strategies E020 should adopt is a design call worth making on its own.

Suggest a follow-up issue rather than widening this PR.

hugocorreia90:
## Second false-positive class: case-sensitive key matching — fixed in 4f0deb09

Reproduced against the previous head:

```sql
SELECT order_id, amount FROM poc.raw.orders
```
```toml
unique_key = ["ORDER_ID"]
```
```
Error E036: unique_key 'ORDER_ID' is not in the output schema of model 'ingest_orders'
  suggestion: Available columns: order_id, amount
```

### I agree with going case-insensitive, and the deciding evidence is stronger than the convention argument

I checked the conventions and they hold — adapters use `eq_ignore_ascii_case` (`rocky-databricks/src/connector.rs:817,877`, `rocky-snowflake/src/loader.rs:327`, `rocky-databricks/src/loader.rs:195`), while the compiler matches exactly (`contracts.rs:121`, E020's `c.name == *time_column`). One caveat on that precedent: the adapter sites all match *warehouse-returned result metadata* (`num_affected_rows`, `rows_loaded`, `sizeInBytes`), not user-declared model columns, so it is suggestive rather than decisive.

What actually settles it is what happens to the key downstream:

- `unique_key` is constrained to `^[a-zA-Z0-9_]+$` by `validate_identifier` (`rocky-sql/src/validation.rs:6`). **A merge key can never be a quoted identifier.**
- The dialects interpolate it verbatim as a bare unquoted name — `format!("t.{k} = s.{k}")` (`rocky-cli/src/commands/plan.rs:769`, and the `MERGE INTO` builders in the duckdb/snowflake/databricks dialects).

So the key always reaches the warehouse unquoted, where resolution is case-insensitive on every supported target. A case-only mismatch *always* resolves at run time, which means there is no true positive in that class to lose — the asymmetry argument from the first fix applies with no residual cost. Consistency with E010's exact match is worth less than not breaking valid builds.

### Fix

`eq_ignore_ascii_case`. Case folding is the **only** relaxation — this is not fuzzy matching:

- `ORDERID` vs `order_id` → still errors (separator difference, not case)
- `order_idd` vs `order_id` → still errors (genuine typo)

### On the `Available columns:` suggestion path

Checked — **no issue, no change needed.** Both suggestion sites (E036 and E020) join `typed_cols` names verbatim without any matching, so there is nothing case-sensitive to get wrong. It was already printing correctly (`Available columns: order_id, amount`) even in the failing repro.

### Quoted-identifier boundary — asked about, and it is narrower than expected

The pathological case cannot arise through a quoted key, because `validate_identifier` forbids quotes outright. What remains is a model projecting **two columns differing only in case** (`Order_ID` and `order_id`): `unique_key = ["ORDER_ID"]` now matches both and E036 stays quiet.

I judged that acceptable and documented it rather than handling it. E036 is an existence check, and under case-insensitive resolution the column does exist. Such a projection is already ambiguous on every case-insensitive warehouse regardless of E036 — the ambiguity belongs to the projection, and diagnosing it is a duplicate-column check's job, not this one. Noted in the doc comment.

### Tests

| Test | Before | After |
|---|---|---|
| `test_e036_not_fired_for_case_mismatched_unique_key` (end-to-end) | **FAILED** | ok |
| `test_e036_case_insensitive_match_both_directions` | **FAILED** | ok |
| `test_e036_case_insensitivity_does_not_swallow_near_miss_typos` | ok | ok |
| `test_e036_case_insensitive_only_ignores_case_not_separators` | ok | ok |

Both false-positive tests verified failing on the previous head first. The two negative controls were green throughout — they exist to catch a fix that over-corrects into fuzzy matching. All 12 E036 tests now pass. True positive re-verified through the real CLI: `unique_key 'ordr_id' is not in the output schema`.

A doc-comment section on `check_merge_strategy` records why E036 matches case-insensitively while E010/E020 nearby do not, so the inconsistency reads as deliberate.

### E010 / E020 case exposure — report only, not fixed

Both have it. Confirmed empirically on this branch, not just by reading:

```
# time_column = "TS" against SELECT ts, id
Error E020: time_column 'TS' is not in the output schema of model 'ev'

# [rules] required = ["ORDER_ID"] against SELECT order_id, amount
Error E010: required column 'ORDER_ID' missing from model output
```

Left alone deliberately — both are pre-existing and shipping, same handling as the E020 `SELECT *` finding above. Worth flagging that they are probably **not** the same call as E036:

- **E020** looks like the same false-positive class, but `time_column` is not interpolated into generated SQL the way `unique_key` is — it drives `@start_date`/`@end_date` substitution while the user writes the column reference in their own SQL. Whether the case variant is genuinely harmless at run time should be confirmed before changing it, rather than assumed by analogy.
- **E010** is arguably different in kind. A contract's `required` list is a governance assertion about schema shape, not an identifier destined for SQL, so exact matching may well be intended there. I would not change it without a deliberate decision.

Suggest folding these into the same follow-up issue as the E020 `SELECT *` finding.

Gates: `cargo test` (full engine workspace, exit 0), `cargo clippy --all-targets --all-features` (clean), `cargo fmt --check` (clean).

hugocorreia90:
Adversarial review found a third false-positive class and an unsoundness in the case fix. Both are addressed, and the diagnostic is downgraded to a warning.

## 1. Schema completeness cannot be inferred from `!has_star`

`SELECT (order_id) FROM poc.raw.orders` with `unique_key = ["order_id"]` was rejected:

```
E036: unique_key 'order_id' is not in the output schema
  suggestion: Available columns:      <- empty
```

The generated `MERGE` runs fine. `(order_id)` parses as `Expr::Nested`, which the lineage extractor does not trace; the item is unnamed so there is no alias fallback, and it is not a wildcard so `has_star` stays false. The gate read `!has_star` as "schema fully known" and ran an absence check against zero columns.

A `typed_cols.is_empty()` guard would not have been enough — `SELECT (order_id), amount` produced `Available columns: amount`: non-empty and still incomplete.

Completeness is now something the extractor **reports**, not something the consumer infers:

- `LineageResult::unresolved_projections: usize` counts projection items that did not yield a correctly named output column. That covers both the items lineage cannot trace at all (`(order_id)`, a literal) **and** the ones it traces but cannot name — `SELECT UPPER(name)` records an entry called `name`, which is the right answer for lineage and the wrong one for "what does this model output?". Aliased items (`expr AS name`) always resolve, so the common explicit-projection model keeps full coverage.
- The count propagates onto `ModelSchema`, and `ModelSchema::schema_is_complete()` is `!has_star && unresolved_projections == 0`. `has_star` is now one input to completeness rather than a stand-in for it.
- `columns` is unchanged — the signal is purely additive, so lineage edges and blast-radius behaviour are untouched.

Both fields are `#[serde(default)]`, so anything cached by an older build still deserializes.

## 2. Case-insensitive matching is unsound on Snowflake

The earlier justification ("the key always reaches the warehouse unquoted") traced only the generic path. Per-adapter survey of `merge_into`:

| Adapter | Renders the key as | Case behaviour |
|---|---|---|
| Databricks | `t.{key} = s.{key}` — bare | resolves case-insensitively |
| DuckDB | `t.{key} = s.{key}` — bare | resolves case-insensitively |
| BigQuery | `target.{key} = source.{key}` — bare | resolves case-insensitively |
| Snowflake | `t.{q} = s.{q}` — **double-quoted** via `quote_identifier` | case-**sensitive** |
| Trino | n/a — returns `not_supported`; merge is rejected at validate time | n/a |

Snowflake quotes deliberately (it folds unquoted identifiers to uppercase), so `unique_key = ["ORDER_ID"]` against a projected `order_id` passes the check and then fails at run time there.

**Kept case-insensitive matching, documented the gap.** Reasoning:

- Exact matching trades a Snowflake-only false negative for a false positive on the other four adapters.
- Deciding per dialect is not possible at this layer — I checked: neither `CompilerConfig` nor `typecheck_project_with_models` carries an adapter, and the LSP and `rocky-ai` callers legitimately have none configured. Plumbing one through to serve a single warning is not worth it.
- The typo this check exists to catch is still caught; only the case-variant subset escapes, and only on Snowflake.

The function's doc comment now carries the survey table and states the limitation plainly instead of arguing it away. Closing it belongs with a dialect-aware validation pass that has the adapter in hand.

## 3. `E036` → `W006`

Three false-positive classes before shipping is a base rate, not a run of bad luck. The check is only as sound as a best-effort schema enumeration recovered from lineage; failing a build on a soft signal is the wrong trade. `W006` was the next free code in the typecheck band. Docs updated: the row now sits in the warning table in `concepts/compiler.md`, and `advanced/failure-modes.md` is back to its original state (that page's table is error-severity failure modes, and a warning is not one). Zero `E036` references remain repo-wide. `E036` was never released.

The E020–E026 rustdoc correction from earlier in this PR is untouched.

## Tests

Both new repros were verified **failing on the previous branch head** before the fix:

```
test test_w006_not_fired_for_parenthesised_projection ... FAILED
  E036: unique_key 'order_id' is not in the output schema  (Available columns: )
test test_w006_not_fired_for_partially_parenthesised_projection ... FAILED
  E036: unique_key 'order_id' is not in the output schema  (Available columns: amount)
```

Added, all green after:

- `test_w006_not_fired_for_parenthesised_projection` — repro 1
- `test_w006_not_fired_for_partially_parenthesised_projection` — repro 2, the case an emptiness guard misses
- `test_w006_not_fired_for_unnamed_computed_projection` — the general form; guards against a fix that only special-cases `Expr::Nested`
- `test_w006_still_fires_when_expressions_are_aliased` — aliasing restores completeness, so the gate does not silently switch the check off for every model that computes something
- `test_w006_qualified_identifier_projection_stays_complete` — `o.order_id` must not count as unresolved, or the check would go dark on most joined models
- `test_w006_is_a_warning_not_an_error` — pins the severity to the code prefix
- 6 in `rocky-sql` covering `unresolved_projections` directly, 3 in `semantic.rs` covering `schema_is_complete()` through `build_semantic_graph`

Preserved: the true positive (explicit columns, typo'd key), `SELECT *` over a raw source, `SELECT *` joining a raw source, and case-mismatch — all still behave as before.

Per the earlier review note, every behavioural test runs through `typecheck_project_with_models` rather than hand-passing the flag. The two prior false positives were both a *caller* computing the flag wrong, which a hand-injected test can never catch.

## Gates

- `cargo test` (full engine workspace) — green, no failures, no `state_sync` flake this run
- `cargo clippy --all-targets --all-features` — clean
- `cargo fmt --check` — clean

No `*Output` struct touched, so no `just codegen` cascade: `ModelSchema` is a compiler internal and does not appear in `output.rs` or any exported schema.

PR #1201, opened 2026-07-21
fix(engine/rocky-cli): gate both promote entrypoints on runtime principal (+ gc/restore honesty)

hugocorreia90:
## ⚠️ NOT a documentation-only PR — it now contains a policy-enforcement behavior change

This PR started as user-facing-text honesty corrections plus characterization pins. It now **also carries a policy-enforcement behavior change** (owner-approved) on the promote path. Please do not review it as doc-only. The two *gc* defects below (A and B) remain pinned-but-unfixed; the promote-gate asymmetry is separate and **is fixed here**.

---

## Behavior change — both promote entrypoints now gate identically

The same agent-authored promote plan was enforced differently depending on entrypoint:

- `rocky apply <promote-plan>` passed `plan.enforcement_principal(runtime)` to `gate_promote_plan` — correctly ignoring the tamperable stored stamp and deciding on **runtime + kind**.
- `rocky branch promote --plan` passed `plan.resolved_principal()` — honoring the stored `principal` stamp, which is **not** integrity-protected (`compute_plan_id` digests only `{kind, payload}`).

Net effect: an agent-authored plan was **denied** via branch-promote but **allowed** via apply under a human runtime, and the branch path trusted a forgeable field.

**Fix:** `run_branch_promote_from_plan` now takes a `runtime_principal` and passes `plan.enforcement_principal(runtime_principal)` to the shared gate — the exact expression apply uses. Its `rocky branch promote --plan` call site resolves the principal via the same `resolve_cli_principal(cli.principal)` the apply arm uses. Both entrypoints now reach the identical verdict: an agent runner is gated as agent regardless of the plan file, a human runner vouches.

New test `promote_gate_is_identical_across_entrypoints` drives the real `gate_promote_plan` → `evaluate_apply_policy_with_policy` through **both** entrypoints on one agent-stamped plan: agent runtime → both deny, human runtime → both allow. The pre-existing `branch_promote_from_plan_gates_a_denied_agent_promote` now passes an explicit `Agent` runtime — its denial previously relied on the stored stamp reaching `Agent`, not on the runtime.

Docs corrected to the post-fix reality: `gate_promote_plan`'s doc, the `--principal` help, `build_promote_plan_inner`'s comment, and the branch-promote comments no longer frame enforcement as keying off the authoring/stored identity, and the misleading `main.rs` "skipping gates" comment is corrected (the approval/breaking-change gates already ran, but the policy gate **is** enforced).

## Also in this round

- **Production-faithful gc/restore pin.** `gc_admits_multi_input_recipe_that_restore_refuses_known_gap` previously hand-fed `build_candidate` a fabricated classification (`verdict: "replayable"`, `inputs: Vec::new()`) that production's `classify_model` could never compute for a 2-upstream artifact. It now seeds real `ArtifactRecord`s for both upstream hashes so the real `classify_model` naturally yields `replayable` with two **resolvable** inputs, and asserts that faithfulness before proving the gc-admits / restore-refuses asymmetry. `refcount == 1` and the six-checks-pass assertions are retained.
- **Surviving stale plan-id comment fixed.** `plan_store.rs`'s `principal_is_outside_the_plan_id_digest` doc no longer opens with the false "the authoring principal is bound into the plan_id integrity digest" claim (which contradicted the corrected paragraph beneath it). It now states the principal is stored separately and deliberately kept out of the `{kind, payload}` digest — exactly why enforcement never trusts the stored stamp.

---

## Original scope (still in this PR, still accurate) — gc/restore restorability honesty

`rocky gc` told users that every eviction is "always restorable from the recorded recipe". That is not true once the recipe has recorded upstreams:

- gc's `check_recipe_recorded` passes any artifact with provenance and a `strong` input closure. Its own detail string says "every upstream is a content hash", explicitly contemplating multiple upstreams.
- `rocky restore` refuses on `!prov.upstreams.is_empty()` — restoring a multi-input recipe needs DAG re-derivation from the recorded upstream bytes, which is a later phase, and it refuses rather than substituting current data.

The predicate is `!is_empty()`, so restore accepts only recipes with **zero** recorded upstreams. gc's eviction set is therefore strictly larger than restore's recovery set. Nothing is destroyed (the tombstone is durable and records the full recipe, and physical deletion is a hard error today), but the recovery route users were pointed at does not work for these artifacts — they have to re-run the pipeline instead.

Corrected everywhere the claim appeared:

| Location | What it was |
|---|---|
| `engine/crates/rocky-cli/src/commands/gc.rs` — module doc | "an evicted cache entry is always restorable" |
| `engine/crates/rocky-cli/src/commands/gc.rs` — apply notes | "so an evicted cache entry is always restorable" |
| `engine/crates/rocky-cli/src/commands/gc.rs` — plan notes (x2) | "(always restorable from the recorded recipe)" / "Every eviction is restorable" |
| `engine/crates/rocky-cli/src/commands/gc.rs` — `physical_delete` hard error | "(always restorable from the recorded recipe)" |
| `engine/crates/rocky-core/src/config.rs` — `GcConfig` doc | "the eviction of record, always restorable from the recorded recipe" |

The `GcConfig` doc comment feeds `JsonSchema`, so it cascades. `just codegen` regenerated `schemas/rocky_project.schema.json`, `docs/public/openapi.json`, `editors/vscode/schemas/rocky-project.schema.json`, `editors/vscode/src/types/generated/rocky_project.ts`, and `sdk/python/src/rocky_sdk/types_generated/rocky_project_schema.py`. No `*Output` struct changed, so no fixture regen was needed.

## Defect A — gc's eviction set exceeds restore's recovery set

**Not fixed** (owner call between two mutually exclusive repairs). Pinned by `gc_admits_multi_input_recipe_that_restore_refuses_known_gap` in `commands/restore.rs`. It builds one real content-addressed artifact whose recipe has two content-hashed upstreams, then drives **both** sides against it: `check_recipe_recorded` passes and the full six-check `build_candidate` admits it, while `restore_one` returns `Refused` with an upstream-mentioning reason. As of this round the pin's classification is production-faithful (see "Also in this round").

## Defect B — gc plans evaluate as `human`, and humans are fail-open

`PlanKind::Gc` falls into the `_ => Human` arm of `default_principal_for_kind`, and an unmatched rule under `Human` returns a hardcoded `Allow`, so a governor's `deny agent gc` rule can never fire. **Not fixed** — routing `PlanKind::Gc` to `Agent` changes the effective policy outcome for deployments on a surface public since 1.0, so it needs a deprecation path. Pinned by `gc_kind_resolves_to_human_so_agent_rules_never_fire_known_gap` (`plan_store.rs`) and `deny_agent_gc_does_not_fire_for_human_principal_known_gap` (`policy.rs`).

> Note: Defect B (gc → human) is unrelated to the promote-gate behavior change above. The promote path already routes correctly; the bug there was the *branch* entrypoint trusting the stored stamp instead of the runtime principal.

## Out of scope, deliberately

No change to gc's eviction predicate, gc's principal resolution (Defect B), gc/human policy evaluation, multi-input restore, or the physical-delete hard error. The behavior change in this PR is confined to the **promote** enforcement path — `run_branch_promote_from_plan` now enforces on the runtime principal exactly as `run_apply_promote_plan` already did. `resolved_principal()` remains the advisory accessor used by `rocky review` / `rocky audit` (unchanged).

## Gates

- `cargo test` — all packages green (`rocky-cli --lib` 1186 passed, `rocky-core --lib` 1682 passed in isolation, `rocky` bin green). The full-workspace parallel run tripped the pre-existing redb `state_sync` page-manager flake under load; it passes cleanly in isolation and is unrelated to this change (rocky-core is untouched).
- `cargo clippy --all-targets --all-features` — clean.
- `cargo fmt --check` — clean.
- No `*Output`/schema doc changed this round, so no `just codegen` / fixture regen was needed. `package-lock.json` / `uv.lock` untouched.


hugocorreia90:
## Correction: the replacement text made a second inaccurate claim

The first pass on this branch removed the false "always restorable from the recorded recipe" promise, but the text it introduced said the evicted artifact **"is recovered by re-running its pipeline"** and that **"nothing is lost"**. That is also wrong, and `bc752844` corrects it.

**Why re-running a pipeline is not recovery.** A content-addressed artifact's identity is its whole-output blake3 — `restore` verifies the rebuild reproduces those exact bytes, and has a `RestoreOutcome::HashMismatch` branch for when it doesn't. `check_replayable` admits only deterministic recipes, so same-inputs implies same-bytes. But a re-run reads **current** upstreams, not the recorded ones. If any upstream has moved on, the re-run produces different bytes: a new artifact at the same logical path, not the evicted one. That is precisely why multi-input restore is described as needing DAG re-derivation *from the recorded upstream bytes*.

The contradiction was direct: `restore.rs` refuses multi-input restore with the reason **"Refused rather than substituting current data"**, while the docs pointed the user at substituting current data and called the result recovery.

**What the text says now** (canonical `GcConfig` doc, `rocky-core/src/config.rs`):

> `rocky restore` rebuilds an evicted artifact from its tombstone when the recorded recipe is non-partitioned, content-addressed, and reads no recorded upstreams. A recipe with ANY recorded upstream cannot be restored today (multi-input DAG re-derivation is a later phase). Re-running the pipeline is not an equivalent: it recomputes the model from *current* upstreams and need not reproduce the evicted bytes, so if those upstreams have moved on the exact bytes are unrecoverable until multi-input restore lands. What the tombstone durably retains is the recipe and the eviction record, not a guarantee that the exact bytes remain recoverable.

**Locations corrected**

| File | What |
|---|---|
| `rocky-core/src/config.rs` | `GcConfig` doc — canonical text, feeds the codegen cascade |
| `rocky-cli/src/commands/gc.rs` | module doc (dropped "nothing is lost") |
| `rocky-cli/src/commands/gc.rs` | `gc_plan_notes()` — plan/apply operator note |
| `rocky-cli/src/commands/gc.rs` | apply-notes `KNOWN LIMITATION` string |
| `rocky-cli/src/commands/gc.rs` | `physical_delete = true` bail message |
| `rocky-cli/src/commands/restore.rs` | known-gap test doc — recovery-framing clause only |

Regenerated via `just codegen`: `schemas/rocky_project.schema.json`, `docs/public/openapi.json`, `editors/vscode/schemas/rocky-project.schema.json`, `editors/vscode/src/types/generated/rocky_project.ts`, `sdk/python/src/rocky_sdk/types_generated/rocky_project_schema.py`. No `package-lock.json` drift this pass.

**Scope unchanged.** Documentation and characterization tests only. gc's eviction predicate, principal resolution, policy evaluation, and restore's refusal are untouched; the three `*_known_gap` tests keep their assertions — only the recovery-framing prose inside one doc comment changed.

Gates from the worktree root: `cargo test` all green (0 failures; the intermittent redb `state_sync` panic did not reproduce), `cargo clippy --all-targets --all-features` clean, `cargo fmt --check` clean.

hugocorreia90:
Corrected the review findings. This PR stays documentation + characterization tests only; no gc/restore/policy behavior changed.

**A — policy governability (the actively-wrong claim).** The assertion that `deny agent gc` "never fires" and that reclamation is ungovernable is false. `PersistedPlan::enforcement_principal(runtime)` = `most_restrictive(runtime, default_principal_for_kind(kind))`, and `most_restrictive` returns `Agent` if either side is `Agent`, so `ROCKY_PRINCIPAL=agent rocky apply <gc-plan>` enforces as **agent** and the rule fires (the gc apply gate calls `enforcement_principal` at `commands/gc.rs`). The real, narrow behavior is deliberate design — a *human* applying a gc plan enforces as `human` (the "human vouches" model documented on `enforcement_principal`).
  - Removed the false evaluator-only pin in `rocky-core/src/policy.rs`. It hardcoded `Human` and never invoked kind routing, so it would not have failed if `Gc` were mapped to `Agent`; and `rocky-core` cannot depend on `rocky-cli`, so it physically cannot exercise the real enforcement path.
  - Reframed the `plan_store.rs` pin from a `_known_gap` to `gc_plan_enforces_as_agent_for_an_agent_applier`, asserting through `enforcement_principal` in both directions: agent applier → `Agent` (deny fires), human applier → `Human` (vouches). This pins the true, governable behavior.

**B — honesty sweep completion.** `derivable` is now stated as an eligibility verdict (a recipe was recorded and bound to the artifact's bytes), not a rebuild guarantee. Corrected: output-schema doc (`output.rs`, codegen source), gc report note, gc plan/apply notes, `[gc]` config reference (`config.rs`, codegen source), CLI help (`main.rs`), `README.md`, and `docs/.../governance-reclamation.md` (which also carried a stale "restore not yet shipped" line and a physical-delete claim).

**C — check count.** `build_candidate` applies **six** checks; the docs said five and omitted the exact-output binding check (`recipe_produces_output`). Fixed the count and enumeration everywhere it appears (module doc, `GcReportOutput`/`GcCandidateOutput`/`GcCheckOutput` schema docs). The unit test at `gc.rs:1959` stays at `count() == 5` — that is correct arithmetic (six checks, one fails, five pass).

**D — factual errors.** (1) The tombstone does not hold the recipe — it holds a `(run_id, model_name, file_path)` pointer plus identity hashes; the canonical `ModelIr` and upstreams live in the referenced `ProvenanceRecord`. Corrected every "retains the full recipe" site. (2) The supported recipe shape (non-partitioned, content-addressed, no recorded upstreams) is now stated as **necessary, not sufficient** — a supported recipe can still refuse (missing provenance binding, undeserializable IR, unreachable storage, a hash that no longer reproduces, a path outside the prefix, or a lost reinstatement race).

**E — internal contradiction.** `state.rs` said the physical delete happens after eviction while the apply path performs none. Corrected to state eviction is ledger-only today (with the ordering kept as guidance for when physical reclamation lands). Text-only; the apply path is unchanged.

**CHANGELOG.** Added an `[Unreleased]` correcting entry. For the shipped 1.64.0 `rocky restore` entry I struck only the false "closes the reclaim→replay loop that makes gc eviction safe" clause with an inline correction, rather than rewriting the historical record — the entry is otherwise accurate.

**Gates.** `cargo test` green on a clean run (126 suites, 0 failed); the two initial failures were the known redb `state_sync`/schema-cache parallel-load flakes and passed isolated + on rerun. `cargo clippy --all-targets --all-features` clean. `cargo fmt --check` clean. `just codegen` is a confirmed no-op from HEAD (codegen-drift should pass); regenerated schemas + Pydantic/TypeScript bindings are committed.

hugocorreia90:
## Third-pass honesty sweep — now complete across the whole worktree

The prior two passes fixed the sites they touched but left equivalent false claims standing. This pass grepped the **entire** worktree for the concepts (not literal strings) and corrected every surviving instance. Every claim was re-verified against the implementation.

### Class 1 — Restorability / rebuildability / physical deletion

Sites from the confirmed list:
- `engine/crates/rocky-cli/src/commands/gc.rs` — module header ("can Rocky prove it can rebuild … derivable bytes are reclaimable"), the apply-surface bullet ("best-effort physical object-store delete"), the per-candidate detail ("the recipe rebuilds these bytes") and its mismatch branch, the check-2 rationale doc.
- `engine/rocky/src/main.rs` — `restore` help ("a gc tombstone claims 'these bytes can be rebuilt'; restore is the proof").
- `engine/crates/rocky-cli/src/output.rs` — `GcPlanOutput.notes` doc, `GcApplyOutput.notes` doc (both "records everything restore needs to rebuild it hash-exact"), and the `physical_reclaimed` field doc (which implied physical deletion happens).
- `engine/crates/rocky-cli/src/plan_store.rs` — `PlanKind::Gc` doc ("rebuildable bit-exact … best-effort physical object-store delete").
- `engine/README.md` — "provably-rebuildable artifacts".
- `docs/src/content/docs/getting-started/roadmap.md` — "recorded recipe demonstrably reproduces them".

**New sites the whole-worktree grep found (beyond the list):**
- `engine/crates/rocky-cli/src/commands/gc.rs` — the **Reachability** section referenced an `ObjectStoreEvictor` "driven on the sandbox". That type **does not exist anywhere in the tree** and no physical delete runs — corrected to state eviction is ledger-only (also removes a broken intra-doc link).
- `engine/crates/rocky-core/src/state.rs` — the `TOMBSTONES` const doc ("bytes Rocky proved it could rebuild"), the `TombstoneRecord` struct doc ("captures everything restore needs … how to rebuild it", and `file_path` "to physically reclaim"), and `record_tombstone`'s doc (described flipping `physical_reclaimed` "after a best-effort physical delete succeeds" — a method with **zero callers**; reworded as a reserved future seam).
- `engine/crates/rocky-cli/src/commands/restore.rs` — the binding-check refusal string ("no longer proves the recipe rebuilds these exact bytes" → "no longer binds the recorded recipe to these exact bytes").
- `integrations/dagster/tests/scenarios.py` — the hand-authored gc-apply parser scenario fabricated `physical_reclaimed: True` / `physical_status: "reclaimed"` and a `"physical reclamation is object-store-only"` note — an outcome the engine never emits. Made faithful (always `false` / "not attempted …"). The parser-routing test asserts none of these fields, so it stays green.

The consistent framing now: `derivable` is a *recorded-and-bound eligibility verdict*, not a rebuild guarantee; `rocky restore` covers only recipes that read **no** recorded upstreams (a multi-input recipe is refused), and even a supported recipe can still refuse; eviction is **ledger-only** and `[gc] physical_delete = true` is a hard error.

### Class 2 — Principal / author-identity

- `engine/rocky/src/main.rs` — global `--principal` help ("a later apply evaluates them against the identity that authored them").

**New sites the grep found (beyond the list):**
- `engine/crates/rocky-cli/src/commands/plan.rs` — the `RunPlanOptions.principal` field doc **and** the propose-time stamp comment ("evaluates the plan against the identity that authored it"). Kept the capability-embed clause (that *is* read at apply); killed only the identity clause.
- `engine/rocky/src/main.rs` — the **restore** handler comment ("deny agent restore fires on an agent-run restore") mirrored the same false model as the gc handler comment.
- `engine/crates/rocky-cli/src/commands/gc.rs` — the `run_gc_plan` principal doc.

True model everywhere now: apply enforces `most_restrictive(apply-time runtime principal, plan-kind default)` and **never reads the stored stamp** (advisory/display only). An agent applier of a gc plan is gated as agent (a `deny agent gc` rule fires); a human applier vouches; a machine-composed kind (ai-authored / backfill) stays agent even under a human applier.

### Class 3 — "Five checks" → six

- `engine/crates/rocky-cli/src/output.rs` — `GcCandidateOutput` ("five printed eligibility checks") and `GcPlan` ("all five fail-closed checks").
- `docs/src/content/docs/reference/commands/governance-reclamation.md` — the enumeration listed only five; added the omitted **exact-output binding** check (`recipe_produces_output`), matching the six in `build_candidate`.

### CHANGELOG decisions (per entry)

- **1.64.0 (`rocky restore` entry):** the opening still carried the false universal ("a gc tombstone claims 'these bytes can be rebuilt'; restore is the proof") even though a later clause was already struck — **rewrote the opening** to the honest framing (as instructed).
- **1.63.0 (`rocky gc` deletion entry):** materially misleading ("evicts only artifacts it can still prove are rebuildable … before a best-effort physical delete") — **struck with an inline `Correction (see #1201):` note** rather than rewriting the history, because the "best-effort physical delete" describes a path that was never implemented.
- **1.60.0 (`rocky gc --dry-run` entry):** "provably-rebuildable artifacts" — **inline `Correction (see #1201):` note** for the same reason.
- Older entries: left as history.
- The `[Unreleased]` #1201 entry was expanded to cover all three classes (it previously described only restorability + five-vs-six) and to note the new denial test.

### Denial-path test (finding #2)

The reframed `plan_store::gc_plan_enforces_as_agent_for_an_agent_applier` only called `enforcement_principal`, pinning the principal *combination* but never the *claimed denial*. Added `commands::apply::tests::gc_deny_agent_rule_fires_for_agent_applier_not_human`, which:

- writes a real `[policy]` block carrying a `deny agent gc` rule;
- builds the `{model → PolicyCapability::Gc}` touched set the gc apply gate constructs;
- drives **`evaluate_apply_policy`** — the exact function the gc apply gate calls — with the principal `PersistedPlan{kind: Gc}.enforcement_principal(runtime)` produces;
- asserts an **agent** applier → `PolicyGate::Deny`, a **human** applier → `PolicyGate::Allow`.

The human assertion is explicitly documented as "the agent-scoped rule does not fire", not "a human gc is ungated" — the unconditional review-marker gate is separate and still applies.

### Gates + CI

- `cargo fmt --check` — pass.
- `cargo clippy --all-targets --all-features` — clean.
- `cargo test` — all pass **except** `state_sync::tests::upload_delete_table_error_fails_closed_no_unfiltered_upload`, the pre-existing redb parallel-load flake (panics inside redb's `page_manager`, in a test this PR does not touch). It **passes in isolation** — not chased.
- `just codegen` — ran; the regenerated `schemas/` + Pydantic + TS diffs are **description-only** (no shape change), so no `regen-fixtures` needed. No `package-lock.json` / `uv.lock` drift committed.
- New denial test + full `test_types.py` (dagster) — green; dagster `ruff format --check` — clean.

hugocorreia90:
Round-5 corrections pushed as `4ece344e`. Documentation + tests only; no behavior change to gc's eviction predicate, principal resolution, policy evaluation, or restore's write/ledger logic. Per finding:

**F1 — denial test pinned a proxy; backwards digest comment**
- `commands/apply.rs`: the gc-denial test now drives `evaluate_apply_policy_with_policy` (the exact function the gc apply gate calls at `gc.rs:1629`) and passes the result of `gc.enforcement_principal(runtime)` — the exact expression the gate evaluates at its call site. The plan's stored stamp is set to `Some(Human)` while the runtime applier is `Agent`, so a Deny proves the stored stamp is ignored and runtime+kind decides. Asserts agent-runtime -> Deny, human-runtime -> Allow.
  - Now exercises: `enforcement_principal`'s body (ignores the stamp) + `evaluate_apply_policy_with_policy` on that principal.
  - Still does NOT cover (documented in the test): that `gc.rs:1632` computes the principal via `enforcement_principal(runtime)` rather than the stamp-reading `resolved_principal()`, that `runtime_principal` is the apply-time `--principal`/`ROCKY_PRINCIPAL`, and that the gate builds `touched` as one `Gc` capability per model. A regression swapping `enforcement_principal` for `resolved_principal` at the call site would stay green here — named explicitly rather than claimed.
- `plan_store.rs`: rewrote the false/backwards comment — the authoring principal is **not** in the `plan_id` digest (`compute_plan_id` hashes only `{kind, payload}`); the stamp is deliberately tamperable, which is why enforcement ignores it.

**F2 — multi-input test didn't characterize gc admission**
- `commands/restore.rs`: the characterization test now drives the gc side through `build_candidate` (the full six-check admission), asserting all six checks pass and `derivable == true` for the multi-input recipe, with `refcount == 1` asserted first (the `unreferenced` check admits only 1). `build_candidate` widened to `pub(crate)` for the test. So the test proves the actual admission set includes a multi-input recipe `restore_one` then refuses — a new production check rejecting upstream-bearing recipes would flip `derivable` to false and fail it.
  - Now exercises: `build_candidate` (all six checks: recipe_recorded, exact-output binding, replayable, unreferenced, policy, age) + `restore_one`'s refusal.
  - Still can't cover: the plan->approve->apply wiring around `build_candidate` (the test calls the admission function directly).
- Corrected the invert-direction comment: narrowing gc to reject recorded upstreams breaks the `build_candidate` admission assertion (gc stops admitting) while `restore_one` keeps refusing; only building multi-input restore makes `restore_one` stop refusing.
- Aligned the CHANGELOG "asymmetry is pinned" claim to what the test now proves.

**F3 — "refused = nothing written" was false about runtime**
- Corrected `output.rs` (`RestoreRefusedOutput`, `RestoreApplyOutput.refused`), `restore.rs` `restore_one` doc, and the CHANGELOG to the accurate statement: a refusal means the ledger was not reinstated (no restore of record); verified content-addressed bytes may already be written and are safe to reuse on re-run. The genuinely pre-write clauses (re-derivation failure, hash mismatch) are kept as true, not mangled.

**F4 — physical-reclamation claims**
- `output.rs`: `physical_status` schema doc now matches the actual longer emitted string ("...future protocol-aware VACUUM work; the tombstone + retired ledger row is the eviction of record").
- `integrations/dagster/tests/scenarios.py`: fixture now emits the real string, not the never-emitted short one.
- `sdk/python/tests/test_client.py`: fixture no longer fabricates `physical_reclaimed: true` / `physical_status: "reclaimed"` — now `false` + the real status (the test asserts only evicted_count/model_name/bytes_evicted, so no assertion change needed).
- `gc.rs` `gc_plan_notes` "before deleting it" and the apply section title "physical reclaim" corrected to ledger-only eviction, no byte deletion.
- `state.rs` `TombstoneRecord.physical_reclaimed` doc: corrected the deferred/failed-delete + reclaiming-sweep description — the flag is always `false` by construction; `[gc] physical_delete = true` hard-fails; there is no sweep.
- Sweep-found equivalents also fixed: `review.rs` "before deleting" and `gc.rs:1486` "before deleting anything".

**F5 — module-header check numbering**
- `gc.rs` module doc: replayability renumbered check 2 -> 3, refcount check 3 -> 4, and the omitted exact-output binding check added as check 2; noted checks 5 (policy) and 6 (age) round out the six.

**F6 — promote carried the same false author-identity model**
- `plan.rs` `build_promote_plan_inner`: corrected the doc — the stored stamp is advisory; a later `rocky apply` enforces `most_restrictive(apply-time runtime principal, default_principal_for_kind(Promote)=Human)`, so an agent applier is gated as agent regardless of who authored, and a human applier resolves to human. (`main.rs` Apply/Gc handlers and the global `--principal` help were already corrected in earlier rounds.)

**Whole-worktree semantic sweep** — ran all enumerated terms (`derivable`, `reclaim`, `rebuild`, `restorable`, `reproduce`, `provably`, `physical delet`, `before deleting`, `nothing...written`, `authored`, `five check`, `closes the loop`, `always restorable`). No surviving false claim. Remaining hits are legitimate and left in place: `ReclaimVerdict`/`Reclaimable` (the Delta-log `remove` proof, not byte deletion), `derivable` used as the defined eligibility verdict, restore's "rebuild" (its actual job), the ground-truth-correct "re-running need not reproduce the evicted bytes", the genuinely pre-write "nothing was written" runtime strings, state-layer "nothing written" (ledger transaction rollback), and "AI-authored plan" (provenance description, not an enforcement claim).

**Gates** (worktree root): `cargo test` (full workspace) green — 0 failures, the pre-existing redb `state_sync` parallel-load flake did not trip; `cargo clippy --all-targets --all-features` clean; `cargo fmt --check` clean. `just codegen` regenerated the `gc_apply` + `restore_apply` schemas, Pydantic, and TypeScript bindings (committed); no `package-lock.json`/`uv.lock` drift (reverted an unrelated `dagster/uv.lock` version re-resolve). SDK: `ruff format --check` clean, 53 pytest passed. Dagster: `ruff format --check` clean, 696 pytest passed.

hugocorreia90:
### Round 6 — promote-gate behavior change + three corrections

This round makes the PR **no longer doc-only**. Four changes (see updated PR description):

**1. Behavior change (`fix:`) — both promote entrypoints gate identically.**
`run_branch_promote_from_plan` now takes a `runtime_principal` and passes `plan.enforcement_principal(runtime_principal)` to `gate_promote_plan` instead of the stamp-reading `plan.resolved_principal()`. Its `rocky branch promote --plan` call site (`main.rs`) resolves it via the same `resolve_cli_principal(cli.principal)` the apply arm uses (that function was reachable in `main.rs`; no re-implementation of `ROCKY_PRINCIPAL` parsing). Both entrypoints now decide on runtime + kind and ignore the tamperable stored stamp. New parity test `promote_gate_is_identical_across_entrypoints` drives the real gate through both entrypoints (agent → both deny, human → both allow). The one pre-existing test that called the branch entrypoint (`branch_promote_from_plan_gates_a_denied_agent_promote`) was updated to pass an explicit `Agent` runtime — its denial previously came from the stored stamp, not a runtime.

**2. Test faithfulness (`test:`) — multi-input gc/restore pin.**
`gc_admits_multi_input_recipe_that_restore_refuses_known_gap` now seeds real `ArtifactRecord`s for both upstream hashes and drives the real `classify_model` (the exact verdict `rocky gc` reuses), which naturally yields `replayable` with two **resolvable** inputs — instead of the old hand-fabricated `inputs: Vec::new()` that production could never produce. Faithfulness is asserted before the asymmetry (six checks pass / `restore_one` refuses). `refcount == 1` and six-checks-pass retained.

**3. Doc (`docs:`) — surviving stale plan-id comment.**
`plan_store.rs`'s `principal_is_outside_the_plan_id_digest` doc no longer opens with the false "principal is bound into the plan_id digest" claim; it now matches `compute_plan_id`'s `{kind, payload}`-only digest.

**4. Doc — promote docs made truthful.**
`gate_promote_plan`'s doc, the `--principal` help, `build_promote_plan_inner`'s comment, and the branch-promote comments now say both promote entrypoints enforce identically on runtime + kind and ignore the stored stamp. The `main.rs` "skipping gates" comment is corrected (approval/breaking-change gates already ran at plan time, but the policy gate is enforced). No remaining promote/apply doc claims enforcement against the authoring/stored identity.

**Gates:** `cargo fmt --check` clean · `cargo clippy --all-targets --all-features` clean · `cargo test` all packages green (`rocky-cli --lib` 1186, `rocky-core --lib` 1682 isolated, `rocky` bin green). The full-workspace parallel run tripped the pre-existing redb `state_sync` flake, which passes in isolation and is unrelated (rocky-core untouched). No schema/`*Output` doc changed → no `just codegen` / fixture regen.

hugocorreia90:
Superseded by a clean split. This branch tangled a documentation-honesty sweep, a promote-enforcement behavior change, characterization tests, and seven rounds of corrections into one incoherent PR (its CHANGELOG even claimed 'documentation only' while carrying a behavior change).

Replacing it with focused PRs:
- gc/restore documentation honesty + characterization tests → a fresh docs-only PR (the verified corrections from here, re-applied cleanly).
- the promote-enforcement fix → its own PR alongside the compact/archive enforcement-parity work.

Closing rather than force-rebasing to keep history clean. Branch retained for reference.

PR #1207, opened 2026-07-22
fix(engine): make bootstrap table creation non-replacing

### LINKED ISSUE #1206: bug: target metadata failures can trigger destructive bootstrap replacement
## Description

Rocky treats any target `DESCRIBE` failure as if the target does not exist. Replication and transformation bootstrap paths then emit replacing DDL, so a transient metadata, permission, or connectivity failure can turn an incremental run into destructive replacement of a live target.

The same mistaken-absence path can pre-drop a Trino target before bootstrap.

## Steps to reproduce

1. Create an incremental source and target with different sentinel rows.
2. Persist a watermark for the target.
3. Make the target metadata probe fail while the source metadata probe succeeds.
4. Run the replication model.

## Expected behavior

The run fails closed and leaves the existing target unchanged.

## Actual behavior

The failed target probe is represented as an empty column list. Rocky concludes that the target is absent and enters the full-refresh/bootstrap path, where `CREATE OR REPLACE TABLE AS` can erase the live target.

## Version

`main` at `7b1cbf7d3a29ff40b7742b07a2fcff1a600d568f`

## Environment

Adapter-independent. Reproduced with DuckDB on macOS; replacing bootstrap DDL is also generated for BigQuery, Databricks, and Snowflake, while Trino uses a pre-drop.

## Additional context

Bootstrap creation should use plain `CREATE ... AS`, preserving intentional replacement only when metadata confirms that the target exists. Plain creation also closes the existence-check/create race because an unexpectedly existing target causes an error instead of replacement.

mattfaltyn:
## Summary

- make replication, transformation, and lakehouse bootstrap CTAS non-replacing
- preserve intentional full-refresh replacement only when metadata confirms that the target exists, including Trino pre-drop behavior
- fall back to per-table metadata probes when a batch prefetch is partial
- add a DuckDB regression test and update per-dialect bootstrap goldens

## Problem

A target `DESCRIBE` error is represented as an empty column list. Rocky then treats the target as absent and enters its bootstrap/full-refresh path. Because bootstrap used replacing DDL (or a Trino pre-drop), a transient metadata, permission, or connectivity failure could erase a live incremental target.

Bootstrap now uses plain `CREATE ... AS`. If the target really is absent, behavior is unchanged; if the probe missed a live target or the check/create race is lost, creation fails without modifying existing data. Confirmed targets still use the existing replacement path for intentional full refreshes.

Fixes #1206

## Test plan

- `cargo nextest run --all-features` — 5,494 passed, 107 skipped
- `cargo test -p rocky-core bootstrap` — passed
- `cargo test -p rocky-cli target_describe_failure_does_not_replace_existing_replication_target --features duckdb` — passed
- `cargo test -p rocky-cli --test ir_golden ir_sql_per_dialect --features duckdb` — passed
- `cargo fmt --all -- --check` — passed
- `git diff --check` — passed

## Existing upstream check failures

With Rust 1.93, exact workspace Clippy (`cargo clippy --all-targets --all-features -- -D warnings`) fails only in untouched files, including generated diagnostic `unused_assignments`, Trino test `clippy::duplicated_attributes`, and existing `list.rs` `clippy::collapsible_else_if` warnings. The adapter-boundary script likewise reports 26 existing imports, all outside this patch. This change does not add a new lint or boundary violation.


hugocorreia90 (approved):
Approving. Reviewed the fail-closed bootstrap change across the replication, transformation, time-interval, and lakehouse paths: the non-replacing CTAS preserves intentional full-refresh (target_exists-gated) while making a misprobed bootstrap fail closed rather than erase a live target.

Adversarial review surfaced one sibling of the same data-loss class not originally covered: the contract-gated `rocky load` promotion derived `target_exists` from a DESCRIBE probe that collapses transient failures to "absent", then used the replacing CTAS — reproducibly erasing a live target on a probe misfire. Added the fail-closed fix (`create_table_as_new`) plus a fault-injected regression test on this branch; the non-contract load paths were verified to already use plain `CREATE` / `CREATE IF NOT EXISTS`.

CI green: Test, Clippy, Format, codegen-drift, release-build smoke, and credential-containment gates all pass.

PR #1209, opened 2026-07-22
fix(engine/rocky-cli): gate compact/archive apply through the policy plane

hugocorreia90:
## What

`rocky compact apply` and `rocky archive apply` executed their warehouse SQL
without consulting the agent-policy plane, while the other seven apply kinds
already gate. This is an **enforcement-parity** fix: it makes compact/archive
apply gate under `PolicyCapability::Apply` exactly like the sibling kinds.

The gap existed on **both routes** that reach the sinks:

- the generic `rocky apply <plan>` dispatch (`run_apply_in` dropped the
  `runtime_principal` on its Compact/Archive arms), and
- the direct `rocky compact apply <plan>` / `rocky archive apply <plan>`
  aliases (which bypass `run_apply_in` entirely).

Both converge on `run_compact_apply_in` / `run_archive_apply_in`, so the gate
lives at the sink and both routes exercise it.

## How

Before the first `adapter.execute_statement`, each sink now:

1. Builds the touched-target set from the typed plan IR (`CompactPlanIr` /
   `ArchivePlanIr`), mapping a physical `catalog.schema.table` FQN back to its
   declaring model's logical name where possible.
2. Evaluates against `plan.enforcement_principal(runtime_principal)` — the
   apply-time runtime identity combined most-restrictively with the plan kind,
   never the plan's stored (tamperable) stamp — via the shared
   `evaluate_apply_policy_with_policy` + `apply_policy_gate`.
3. Handles the verdict identically to the gated arms: **Deny** refuses before
   any execution; **RequireReview** requires the `rocky review <plan> --approve`
   marker; **Allow / NotConfigured** proceed.

A single fingerprinted `rocky.toml` snapshot backs the gate, the touched-set
resolution, the dialect, and the adapter (one snapshot, no re-reads).

**Framing:** this is enforcement parity, not a hardened security boundary — the
policy plane remains cooperative (self-asserted principal). Absent a `[policy]`
block, behaviour is byte-identical to before.

## Judgment calls worth a look

- **Targetless archive.** An archive plan can carry `target_table = None`
  (`DELETE FROM *`). It has no concrete target to key a decision on, so treating
  it as an empty touched set would vacuously pass the gate. Decision: **refuse
  it outright when a `[policy]` block is configured**; with no policy configured
  its behaviour is unchanged (the wildcard flows to the warehouse as today).
- **Freeze sync.** The gate pulls the authoritative remote freeze/budget ledger
  before reading it (a cross-pod freeze must deny — fail-closed). There is **no
  upload-after half**: compact/archive mutate no durable ledger state, so the
  gate's decision rows are local audit only.
- **`require_review`** is satisfiable the usual way — `rocky review <plan>
  --approve`, then re-apply. These plans are not auto-escalated into the review
  queue the way `gc` plans are at creation time (out of scope here).

## Known limitation

The FQN → logical-name mapping is best-effort: an identifier that resolves to
neither a known model name nor a known target FQN is kept verbatim, so the
evaluator sees default attributes. A selector-free rule (`any` / bare `deny
agent apply`) still fires against those, so the deny is never lost — only
attribute-scoped (classification/layer) enrichment on the catalog route depends
on the mapping resolving.

## Tests

Cover **both routes** for each kind — the injectable sink (`*_in_with` with a
counting adapter) and the generic `run_apply_in` route:

- agent under `deny agent apply` → refused, **zero statements reach the
  adapter**;
- human → allowed, statements execute;
- a downgraded stored stamp (`human`) applied by an agent → still denied
  (enforcement uses runtime + kind);
- targetless archive → refused under policy, proceeds without policy.

## Gates

`cd engine && cargo test` (full workspace, 0 failures), `cargo clippy
--all-targets --all-features` (clean), `cargo fmt --check` (clean).


hugocorreia90:
Follow-up from adversarial review (commit `7a4bba01`).

## Blocker fixed — `rocky review --approve` now accepts compact/archive plans

The apply gate could resolve to `require_review`, but `rocky review` refused to
sign off on compact/archive: `compute_review`'s reviewable-kind allowlist
(`review.rs`) omitted `Compact` and `Archive`, so `rocky review --approve`
bailed with "review only applies to …" and the verdict could never be cleared —
the plan was stranded forever. That broke the whole parity contract (the point
is that these gate *and can be cleared* like their siblings).

Fix:
- Added `Compact` / `Archive` to the reviewable allowlist and the error text.
- Routed them through the payload-agnostic marker-only review path (renamed
  `compute_review_gc_plan` → `compute_review_marker_only`, now serving
  gc/restore/compact/archive), so they do **not** fall through to the `RunPlan`
  deserialize — their payload is a `CompactPlanIr` / `ArchivePlanIr` envelope,
  and they change no model definitions, so it is purely the human sign-off
  marker the apply gate already reads (same `<plan_id>.reviewed.json` path).

New end-to-end test per kind: an agent apply resolving to `require_review` is
refused without a marker, then **proceeds** after the real `rocky review
--approve` path (`compute_review`) writes the sign-off. Driven through the
actual review-compute function, not a hand-written marker file. Verified these
tests **fail on the previous head** (they hit the "only applies to …" bail) and
pass after the fix.

## Documented parity limitation (no code change)

When a compact/archive target FQN maps to no known project model,
`resolve_touched_apply_targets` keeps the raw FQN, so **attribute-scoped** rules
(`layer = "gold"`, `classification = "pii"`, …) do not match and the target
falls to `default_agent_effect` — the same physical-name fallback the
replication apply gate already uses for runtime-discovered targets. A
selector-free rule (`any` / bare `deny agent apply`) still fires against it, so
an unresolved target is never silently un-gated; only an explicitly permissive
`default_agent_effect = "allow"` yields auto-execution (factory default is
`require_review`). Captured in a doc comment on `resolve_touched_apply_targets`
and noted here.

## Gates
`cargo test` (rocky-cli, all pass — incl. duckdb-gated), `cargo clippy
--all-targets --all-features` (clean), `cargo fmt --check` (clean).

PR #1211, opened 2026-07-22
fix(engine/rocky-cli): gate all three promote routes on the runtime principal

hugocorreia90:
## What

Consolidates all three `rocky branch promote` routes onto a single runtime-principal policy gate, closing a promote bypass.

Three routes existed, gating differently:

| Route | Before | After |
|---|---|---|
| `rocky apply <promote-plan>` | gated on `enforcement_principal(runtime)` | unchanged |
| `rocky branch promote --plan <id>` | gated, but on the plan's stored stamp (`resolved_principal()`) | gated on `enforcement_principal(runtime)` |
| bare `rocky branch promote <name>` | **no policy gate** — reached `execute_statement` after only the approval + breaking-change checks | gated on `enforcement_principal(runtime)` |

Bare promote was the bypass: an agent-run promote that a `deny agent promote` rule (or an active freeze) would refuse could still execute, even though the deprecation notice already promised it "internally chains `plan + apply` … which runs the approval + breaking-change gates before any production data is touched." This makes that promise true and extends it to the policy plane.

All three now key on `enforcement_principal(runtime)` = `most_restrictive(runtime, kind_default)`, **not** the plan's stored `principal` stamp (which is advisory and not integrity-protected — `compute_plan_id` digests only `{kind, payload}`). An agent runner is gated as agent regardless of the plan file; a human runner vouches.

## Bare-promote delegation design

`run_branch_promote` no longer executes SQL directly. It now honors the documented plan+apply chain, reusing the seams the other two routes already use — no third hand-written policy implementation:

1. **Plan-build (approval + breaking-change gates):** `build_promote_plan_inner` — the extraction point `rocky plan promote` already reuses — runs the *same* approval and breaking-change gates as before and persists a `Promote` plan.
2. **Apply (policy gate):** `gate_promote_plan(root, config, plan_id, enforcement_principal(runtime), …)` runs the agent-policy gate before any statement, then `run_promote_apply` executes against the gate-verified config snapshot.

The happy-path **output contract is preserved**: `command: "branch promote"`, the `promoted branch 'X' (N targets)` message, and the `BranchPromoteOutput` JSON shape are unchanged (delegation goes through the `gate_promote_plan` seam, not `run_branch_promote_from_plan`, precisely so the message stays put).

To preserve the bare verb's `--skip-approval` **flag** (which `build_promote_plan_inner` previously honored only via `ROCKY_BRANCH_APPROVAL_SKIP`), `build_promote_plan_inner` gains a `skip_approval_flag` param; `rocky plan promote` passes `false`.

### Two intentional normalizations (no consumer break)

1. The bare-promote audit trail now includes a `PromotePlanCreated` event — it genuinely builds a plan now.
2. A breaking-change block now fails with a non-zero exit and the findings on **stderr** (matching `rocky plan promote`), rather than *also* printing a `success:false` JSON document to stdout before exiting. Verified this breaks no first-party caller: `rocky-sdk`'s `branch_promote` raises on any non-zero exit (it never parses breaking-blocked stdout into a result), and the exception subtype shifts from `RockyPartialFailure` to `RockyCommandError` — its **superclass** — which the Dagster resource already handles identically (`isinstance(exc, RockyCommandError)`). This normalizes bare promote to the canonical flow the deprecation notice points at.

### One parity note

`run_promote_apply` resolves its executor adapter via `resolve_pipeline(cfg, None)`, so bare `branch promote --pipeline P` now executes against the default pipeline's adapter rather than P's. This is a behavior change only for bare-with-`--pipeline` in a multi-adapter config, and it is exactly what `--plan` / `apply` already do — parity, not a new bug, and a no-op when pipelines share `adapter = "default"`.

First-party callers (`rocky-sdk`, `dagster-rocky`, VS Code) invoke the CLI verb; this internal delegation leaves their interface unchanged. The bare command is not removed (that's a future major release).

## Tests (security-adjacent — the backbone)

- **`promote_gate_is_identical_across_all_three_entrypoints`** (branch.rs): one real branch + seeded source table under a `deny agent promote { any }` policy. An **agent** runtime is refused on all three routes with **zero** SQL executed (asserts the prod table stays absent); a **human** runtime is allowed on all three and copies the branch rows into prod. Routes 1/2 share a pre-built plan; route 3 rebuilds its own — "same verdict" is over the same branch/targets.
- **`promote_gate_enforces_runtime_principal_not_stored_stamp`** (apply.rs): drives the real gate via `run_apply_promote_plan` and `run_branch_promote_from_plan`. Proves the stored stamp decides nothing in **both** directions — an agent-stamped plan is allowed under a human runtime, and a **downgraded** human-stamped plan is still denied under an agent runtime.
- **`branch_promote_from_plan_gates_a_denied_agent_promote`** (branch.rs): the route-2 regression guard, updated for the runtime-principal signature.

**Fails on pre-fix code:** empirically confirmed by simulating the bypass (gating bare promote as `Human` = ungated, like the old no-gate path) — the headline test then fails at the bare-agent-REFUSED assertion with `promoted branch 'fix-price' (1 targets)` on stdout, proving the pre-fix path executes the denied promote. The two apply-side routes' deny assertions likewise fail without the `enforcement_principal` change.

## Gates

- `cargo test -p rocky-cli`: lib **1201 passed / 0 failed / 10 ignored**; all integration test binaries green.
- `cargo clippy --all-targets --all-features`: clean (exit 0, no warnings).
- `cargo fmt --check`: clean.
- No `*Output` / schema struct touched → no `just codegen` needed.


hugocorreia90:
Addressed the review in `03129ec7`. Enforcement property is unchanged — all three routes still gate on `enforcement_principal(runtime)`; these are the two regression fixes plus the test-gap.

### BLOCKER 1 — `--pipeline` restored (both bare and `--plan`)
`run_promote_apply` hardcoded `resolve_pipeline(cfg, None)`, so bare `rocky branch promote --pipeline X` reached execution and then failed `multiple pipelines defined`. Fixed by threading the selector as a param:
- bare `run_branch_promote` → its own `pipeline_name`;
- `run_branch_promote_from_plan` → new param, sourced from the `--plan` dispatch's `pipeline.as_deref()` (so `rocky branch promote --plan <id> --pipeline X` also disambiguates);
- `run_apply_promote_plan` → `None` (`rocky apply` has no selector — pre-existing gap on a multi-pipeline config, not introduced here, documented in the code).

New test `bare_promote_pipeline_selector_disambiguates_multi_pipeline`: two-pipeline config, `--pipeline marts` promotes (rows copied), omitting `--pipeline` still errors `multiple pipelines`. Fails on the pre-fix `None`.

### BLOCKER 2 — breaking-block stdout JSON restored (delta (b) reverted)
Reverted the "match `rocky plan promote` / stderr-only" behavior — it broke the shipped POC `06-.../15-semantic-breaking-change-gate` (its `run.sh` `json.load`s the block JSON off stdout; runs in `engine-weekly.yml`). Instead of `?`-swallowing the block, `build_promote_plan_inner` now returns a typed `PromoteBlockedByBreakingChanges` whose `Display` byte-matches the old bail message. The bare verb catches it and re-emits the `success:false` block JSON to stdout (empty `targets`, findings under `breaking_changes`, leading `BreakingChangesBlocked` audit) before propagating the non-zero exit; `rocky plan promote` still `?`-propagates the identical message unchanged.

Verified end-to-end: the POC passes against a freshly-built binary (not a stale PATH `rocky`) — step 6 reports `exit 1`, `success=False`, `audit kinds: ['breaking_changes_blocked']`, `breaking_changes carried: 2`. Tests: `breaking_block_output_preserves_stdout_contract` (pins the exact fields the POC reads) and `bare_promote_breaking_block_errors_and_routes_stdout_json` (a real git column-drop routes through the typed error with a non-zero exit).

The `PromotePlanCreated` audit event (delta (a)) is kept — confirmed additive/safe in review.

### FINDING 3 — bare-route stamp-independence: pinned by composition, not a standalone test
The requested "build under one principal, apply under a different runtime" test **cannot exist for the bare route**, and here's the proof: for a `Promote` plan, `enforcement_principal(r) = most_restrictive(r, default_for_kind(Promote)) = most_restrictive(r, Human) = r`. The bare verb stamps the plan with `r` in the same call, so `resolved_principal() = r` too. Therefore `resolved_principal() ≡ enforcement_principal(r) ≡ r` on **every** bare invocation — a regression to `resolved_principal()` on that line is behavior-preserving and no black-box test through `run_branch_promote` can distinguish it (build and apply are one call with one principal; there is no way to decouple stamp from runtime).

I kept `enforcement_principal` (with a comment on the line explaining why, so nobody "simplifies" it — it would become exploitable only if the stamp ever decoupled from the runtime). The stamp-independence that *does* have teeth is pinned as a composition of existing tests: `build_promote_plan_inner_stamps_agent_principal` (bare stamps == runtime) + `promote_gate_enforces_runtime_principal_not_stored_stamp` (the shared `gate_promote_plan` enforces runtime, not stamp, in both directions — and all three routes funnel through it) + `promote_gate_is_identical_across_all_three_entrypoints` (bare reaches the gate: agent → deny, human → allow; catches hardcoded-Human / dropped-runtime regressions).

### Gates
`cargo test -p rocky-cli` (lib 1204 passed / 0 failed / 10 ignored; all integration binaries green), `cargo clippy --all-targets --all-features` (clean, 0 warnings), `cargo fmt --check` (clean), semantic-breaking-change POC (exit 0 against a freshly-built binary). No `*Output`/schema type touched → no codegen.

hugocorreia90:
Follow-up `dadc3da8`: closed two review-of-review items.

- **Per-PR stdout-write coverage.** The earlier per-PR tests asserted the block's exit code + error-routing + reconstructed content, but the actual stdout *write* was only gated by the weekly POC. Added `engine/rocky/tests/promote_breaking_block_stdout.rs` — a `CARGO_BIN_EXE_rocky` subprocess test that spawns the real binary against a git column-drop fixture and asserts a non-zero exit **and** that stdout carries the `success:false` block JSON (empty `targets`, findings, `breaking_changes_blocked` audit). The byte-level write is now gated per-PR, not weekly-only.
- **CHANGELOG wording.** Fixed a garbled sentence that read as if the `--pipeline` selector and `--skip-approval` flag 'emit success:false JSON' (only a breaking-change block does).

Gates re-run clean: `cargo fmt --check`, `cargo clippy --all-targets --all-features` (0 warnings), `cargo test -p rocky-cli -p rocky` (0 failures).

PR #1226, opened 2026-07-25
fix(rocky-cli): classify model renames by target identity

### LINKED ISSUE #1225: ci-diff misclassifies filename-inferred model renames as modifications
## Description

`rocky ci-diff` reports a filename-inferred model rename as one modified
model with no column changes, even though the rename removes the old model
identity/target and adds a new one.

Model `name` defaults to the filename stem, and `target.table` defaults to
the model name. Therefore, renaming `orders.sql` + `orders.toml` to
`purchases.sql` + `purchases.toml` without explicit `name` or
`target.table` changes `poc.marts.orders` into `poc.marts.purchases`.

The default structural report hides that removal. `--semantic` detects the
removal and addition separately, but it leaves the incorrect structural
summary in the same output.

## Steps to reproduce

1. Create a Git repository containing these files:

`models/orders.sql`

```sql
SELECT 1 AS id, 'paid' AS status
```

`models/orders.toml`

```toml
[strategy]
type = "full_refresh"

[target]
catalog = "poc"
schema = "marts"
```

2. Commit the files.
3. Rename both files without changing their contents:

```bash
git mv models/orders.sql models/purchases.sql
git mv models/orders.toml models/purchases.toml
git commit -m "refactor: rename orders model"
```

4. Confirm Git detects the renames:

```bash
git diff --name-status HEAD~1...HEAD
```

```text
R100	models/orders.sql	models/purchases.sql
R100	models/orders.toml	models/purchases.toml
```

5. Run:

```bash
rocky ci-diff HEAD~1 --models models --output json
```

## Expected behavior

The structural diff should report:

- `orders` as removed, with its columns removed.
- `purchases` as added, with its columns added.

If an explicit model `name` and target keep the resolved identity stable,
a file-only rename should remain a modification rather than being split
unconditionally.

## Actual behavior

The command exits successfully and reports:

```json
{
  "summary": {
    "total_models": 1,
    "modified": 1,
    "added": 0,
    "removed": 0
  },
  "models": [
    {
      "model_name": "purchases",
      "status": "modified"
    }
  ]
}
```

There are no `column_changes`.

With `--semantic`, the same output still contains that structural summary,
but `breaking_findings` correctly contains:

```json
[
  {
    "change": {
      "kind": "model_removed",
      "model": "poc.marts.orders"
    },
    "severity": "breaking"
  },
  {
    "change": {
      "kind": "model_added",
      "model": "poc.marts.purchases"
    },
    "severity": "info"
  }
]
```

## Root cause

`parse_name_status` recognizes the three-field Git rename record but keeps
only the new path. `classify_model_changes` maps every status other than
`A` or `D`, including `R`, to `ModelDiffStatus::Modified`. Later,
`build_diff_results` can only calculate columns for a modified model when
the same name exists in both base and head schema maps. The new stem
`purchases` has no base entry, so the column diff is empty.

The existing `parse_rename_uses_new_path` test pins the lossy parse but
does not test the resulting model classification.

## Suggested direction

Preserve both paths for rename records and compare their resolved
base/head model identities. Emit removed + added when the identity changes,
while preserving modified for a file rename whose explicit configuration
keeps the model identity stable.

Add an end-to-end regression test in
`engine/crates/rocky-cli/src/commands/ci_diff.rs` using a temporary Git
repository with the valid SQL/sidecar pair above.

## Environment

- Rocky: `1.67.0`
- Current upstream checked: `main@d75471165d0b5cdbf1167515cfcdd6523b858b75`
- OS / architecture: macOS 26.5.2 arm64
- Reproduced twice with identical output.
- Control: an in-place edit to `orders.sql` correctly reports `orders` as
  modified and includes the added column.

## Existing-history check

Searched issues, pull requests, and comments across all states for
`ci-diff`, `rename`, `model rename`, `renamed model`, `R100`,
`parse_rename_uses_new_path`, `git diff --name-status`, and structural
added/removed terminology. No issue or pull request addresses this
structural misclassification.

- #67 introduced `ci-diff` and the current rename parser.
- #508, #509, and #510 added the separate semantic classifier and promote
  gate; they do not change the default structural result.

mattfaltyn:
## Summary

- Preserve both paths from Git rename and copy records.
- Pair successfully compiled base and head models by resolved target identity while retaining their distinct schema keys.
- Keep a conservative filename fallback when either compile is unavailable, and correctly handle contract sidecars, moves outside the model root, copies, and same-stem SQL/Rocky models.

## Why

A filename-inferred rename changes both the logical model name and its default target, but structural `ci-diff` collapsed it into an empty `Modified` result instead of reporting the old model as removed and the new model as added.

## Validation

- `cargo nextest run --all-features` (5,632 passed; 107 skipped)
- `cargo test -p rocky-cli --lib commands::ci_diff::tests -- --nocapture` (29 passed)
- `cargo fmt -- --check`
- `git diff --check`
- `bash ../scripts/lint-adapter-boundary.sh`
- Real CLI reproduction: `orders` is removed and `purchases` is added; an in-place control remains modified

## Independent review

Codex Terra independently reviewed the plan and final diff under `AGENT_REVIEW.md`. It identified target-identity pairing, compile-fallback, contract-sidecar/root filtering, rename-out, sidecar-only, and same-stem SQL/Rocky risks. All were addressed with regression coverage; its final review reported no actionable correctness findings and confirmed that `ModelIr` and contract semantics are preserved.

Fixes #1225

### DIFF
--- a/engine/crates/rocky-cli/src/commands/ci_diff.rs
+++ b/engine/crates/rocky-cli/src/commands/ci_diff.rs
@@ -6,7 +6,7 @@
 //! each changed model as added, modified, or removed and generates a structured
 //! diff report in JSON and Markdown formats.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
@@ -85,13 +85,30 @@ pub(crate) fn validate_base_ref(base_ref: &str) -> Result<()> {
     Ok(())
 }
 
+/// Build a `git` invocation, optionally rooted at an explicit repository
+/// directory instead of the process's current one.
+///
+/// Production always passes `None` and inherits the process cwd, which is the
+/// long-standing behavior. Tests pass `Some(dir)` so they can drive a scratch
+/// repository without calling `set_current_dir` — cwd is process-wide, and this
+/// crate's tests run as threads in one process under `cargo test`, so mutating
+/// it races every other test that reads it (directly or through the many
+/// commands that resolve relative paths).
+fn git_in(repo_dir: Option<&Path>) -> Command {
+    let mut cmd = Command::new("git");
+    if let Some(dir) = repo_dir {
+        cmd.current_dir(dir);
+    }
+    cmd
+}
+
 /// Run `git diff --name-status` between `base_ref` and HEAD to find changed files.
 ///
 /// Uses three-dot syntax (`base...HEAD`) for merge-base semantics — this matches
 /// what CI systems care about: changes since the branch diverged from the base,
 /// not changes since the base's current tip.
-fn git_changed_files(base_ref: &str) -> Result<Vec<ChangedFile>> {
-    let output = Command::new("git")
+fn git_changed_files(base_ref: &str, repo_dir: Option<&Path>) -> Result<Vec<ChangedFile>> {
+    let output = git_in(repo_dir)
         .args(["diff", "--name-status", &format!("{base_ref}...HEAD")])
         .output()
         .context("failed to run `git diff` — is git installed and is this a git repository?")?;
@@ -103,7 +120,7 @@ fn git_changed_files(base_ref: &str) -> Result<Vec<ChangedFile>> {
             "three-dot git diff failed (exit {}), falling back to two-dot",
             output.status
         );
-        let output = Command::new("git")
+        let output = git_in(repo_dir)
             .args(["diff", "--name-status", base_ref, "HEAD"])
             .output()
             .context("failed to run `git diff` with two-dot syntax")?;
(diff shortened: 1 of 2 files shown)


hugocorreia90 (dismissed):
Thanks for this — the diagnosis in #1225 was exact, and the fix is right: keying changed models by resolved target identity is the correct model, and I verified it end-to-end against a real `git mv` repo. It also quietly fixes a second bug you didn't claim: `build_diff_results` was looking up the schema maps by **filename stem** while `typed_columns_from_compile` keys them by `config.name`, so any model whose sidecar `name` differed from its stem reported empty `column_changes` forever. Worth crediting.

Review found that the new `model_stem` helper became the single gate deciding whether `ci-diff` runs at all, and a few kinds of real model file fell through it. Because a file that classifies as nothing also skips both compiles, that silently skipped the `--semantic` breaking-change classifier too — so some changes that report a BREAKING finding on `main` today reported nothing at all.

I've pushed `c2e95645` to this branch rather than round-tripping it, since the pieces are interlocking. It changes five things:

| Case | `main` | this PR before | now |
|---|---|---|---|
| #1225 rename | 2 modified ❌ | correct ✅ | correct ✅ |
| in-place retarget | 1 modified | 2 rows, **order varies per run** ❌ | deterministic ✅ |
| `groups/*.toml` retarget | BREAKING fires | **silent, exit 0** ❌ | BREAKING fires ✅ |
| `models/rocky.sql` | reported | **invisible** ❌ | reported ✅ |
| model stem ending `.contract` | reported | **invisible** ❌ | reported ✅ |
| two models, one target | both reported | **one vanishes** ❌ | both ✅ |
| `_defaults.toml` retarget | silent ❌ | silent ❌ | BREAKING fires ✅ |

Details:

1. **Ownership resolves from the compiled model's own `file_path`** — its source file, `<stem>.toml` sidecar, and `<stem>.contract.toml` contract — rather than re-deriving a stem from the changed path. Filename inference can't distinguish `orders.contract.toml`-the-sidecar-of-`orders.contract.sql` from `orders.contract.toml`-the-contract-of-`orders.sql`.

2. **The `_defaults`/`rocky`/`Cargo` exclusion applies only to `.toml`.** Those names mark shared config as TOML, but the loaders accept any flat `.sql`/`.rocky`, so `models/rocky.sql` is an ordinary model — it compiles clean and was being dropped from the report.

3. **Both refs compile whenever a compiler-readable file under the models root moved**, not only when a model file did. A `groups/*.toml` or `_defaults.toml` edit retargets every member model while producing no structural rows. `run_ci_diff` also classifies findings *before* its empty-structural-diff early return. A change touching nothing the loaders can read still short-circuits, so a docs-only PR doesn't pay for two compiles.

4. **Colliding target identities get name-qualified keys.** `Project::from_models` rejects duplicate `config.name` but says nothing about duplicate targets, so keying purely by target let one model overwrite another and disappear.

5. **Sort ties break on the change-map key.** `model_name` stops being unique once a model's target moves (it appears twice — removed + added), so a stable sort over randomized `HashMap` iteration left the results vector — and the Markdown rendered from it — varying between runs. Hashing the `markdown` field across 10 identical runs gave two different digests.

Plus the end-to-end regression #1225 asked for: real temp git repos driven through `compute_ci_diff` with real `git mv`. Spliced onto the pre-fix classifier, the group-config and reserved-stem tests fail and the rename tests pass, so they discriminate rather than just decorate.

One follow-up worth flagging, not a blocker: the split now fires for **any** target-identity change, not just renames — so an in-place `target.schema` edit with a stable `name` reports removed+added. That's defensible (it's what `breaking_change.rs` already reports under `--semantic`), but it's broader than #1225 asked for and it's why two rows can share a `model_name`. Surfacing the resolved target on `DiffResult` would make those rows self-describing; that's a schema change, so a separate PR.

Reviewed adversarially by Codex Terra across two rounds — once on the original diff, once on the follow-up commit, where it caught three real defects in my own changes (a cwd race the new tests introduced, a tautological determinism assertion, and an over-broad compile trigger). All fixed; the remaining findings are dispositioned in the commit message.

Gates on the merged result: `cargo fmt --check`, `cargo nextest run -p rocky-cli` (1301 passed), `cargo clippy --all-targets --all-features -D warnings` (0 warnings).

hugocorreia90 (approved):
Added `bda4bbd2` after a pre-merge adversarial pass over the combined diff (both commits reviewed as one unit, since that union is what lands). No blockers surfaced; these are the four nits it did find, all in the follow-up commit's own new code:

- The human-readable block headed **"Markdown (for PR comment)"** was still rendered from the structural summary alone — so the text a reviewer actually copies omitted the breaking-change section that the JSON `markdown` field had just gained. Both now render through one helper, which is total: no findings means the structural Markdown comes back untouched rather than gaining an empty table.
- The findings section replaced the clean verdict on `models.is_empty()` while the structural renderer decides cleanliness with `DiffSummary::is_clean()`. An all-`Unchanged` rows vec satisfies one and not the other, which would leave "No data changes detected" standing above a findings table. Latent today — `compute_ci_diff` never emits `Unchanged` rows — but a predicate mismatch waiting on the first caller that does.
- Findings table cells are escaped now. Nothing validates `config.name` as a SQL identifier, so a `|` or backtick in a model name split the row it landed in.
- On the filename-stem fallback, a git copy record no longer lends its source's stem to a destination outside the models root: `C100 models/orders.sql archive/orders.sql` reported the untouched `orders` as newly **Added**. It now reports nothing. Renames keep the old-path fallback they need. (Reachable only with copy detection enabled in gitconfig plus a failed compile, but it was a behavior delta rather than pre-existing.)

Also re-gated against current `main` (`0975efff`) in a clean worktree rather than trusting the PR's base-time CI: fmt, clippy `-D warnings`, and 1320 tests all green. Schema export diffed byte-identical against the committed `schemas/`, so no codegen cascade is needed.

Issue #1228, opened 2026-07-25
Ledger-seam writers bypass compare-and-swap — a concurrent gc/policy/apply can silently erase a run's committed state


hugocorreia90:
## Summary

`concurrency_control = "cas"` protects the **end-of-run** state upload, but the *ledger-seam* writers do not go through it. They call `upload_only_fail_closed` → `upload_state` → `dispatch_upload`, never `upload_state_cas`. So a seam write is an **unconditional last-writer-wins** overwrite of the whole remote state object, even on a fully CAS-configured deployment.

Live callers today:

- `engine/crates/rocky-cli/src/commands/policy.rs:631` (freeze / unfreeze)
- `engine/crates/rocky-cli/src/commands/apply.rs:1071` (apply / restore / backfill)
- `engine/crates/rocky-cli/src/commands/gc.rs:1705` — `gc.rs:1598` already documents that its upload has no CAS

## Failure

No exotic timing required, and nothing reports an error:

1. A run `R` and a `gc` both download durable generation `G0`.
2. `R` compare-and-swap commits its state as `G1` and exits successfully.
3. `gc` then calls `upload_only_fail_closed`, which unconditionally overwrites the object with its own `G0`-derived blob.
4. **`R`'s committed watermarks and run records are gone.** No conflict is raised; `gc` exits 0.

The window is the whole duration of the seam operation, and the two processes need not know about each other. Any deployment that runs maintenance (`gc`, `compact`, `archive`) or governance (`policy freeze`, `apply`) alongside pipeline runs against a shared remote state prefix is exposed.

## Why this is the blocker, not a detail

ADR-CONCURRENCY **D1** is normative: *"CAS on every remote state write, split by writer class"* — cheap single-record ledger seams **RETRY**, full runs **REFUSE**. The seam half of that split was never implemented. Two consequences follow:

- **The lost-update finding cannot be closed by configuration.** Setting `concurrency_control = "cas"` on every pod still leaves these writers clobbering. The rollout gate in the ADR ("CAS active on every live writer") is therefore unsatisfiable as the code stands.
- **`rocky doctor` must not report the concurrency check healthy** while any writer class bypasses CAS, on any backend.

Also stale and worth fixing with this: `ConcurrencyControl`'s doc comment (`engine/crates/rocky-core/src/config.rs:540-545`) scopes CAS to "the end-of-run upload", which contradicts D1. D1 is the normative statement; the doc should be narrowed to match it.

## Scope

[...]

PR #1230, opened 2026-07-25
feat(engine): compare-and-swap the end-of-run state upload on tiered

hugocorreia90:
## Summary

Implements ADR-CONCURRENCY decision **D5 (tiered cache coherence)** and, on the back of it, enables `concurrency_control = "cas"` for the `tiered` backend — which previously force-downgraded to an unconditional last-writer-wins upload.

## What changes

The cached entry becomes **freshness-checked against the durable tier**. The generation the durable compare-and-swap committed at is framed inside the cached value in one atomic write, under a key namespace disjoint from the non-CAS path. A reader never trusts the cache on its own: it reads the durable object's current generation and accepts the cached blob only on an exact match.

Two consequences follow, and both are deliberate:

- **Invalidation is hygiene, not a correctness dependency.** A surviving stale entry carries a superseded generation and is rejected anyway — proven by a test that faults both the write and the delete so a stale entry genuinely survives, and is still rejected.
- **A sidecar generation key was rejected.** Two writers interleaving a blob write and a generation write could pair a stale blob with a current generation — an entry that validates and is wrong.

Per D5: compare-and-swap runs against the durable leg first; the cache is populated only with the committed generation on success and invalidated on conflict or failure; and reads validate the cached generation before use.

## Scope — what this does *not* do

`cas` protects the **end-of-run state upload**. The single-record ledger seams — the state writes made by `rocky policy freeze`, `gc apply`, and `apply` — still upload unconditionally on **every** backend, so a concurrent maintenance or governance command can overwrite a run's committed state without raising a conflict and still exit zero. That is tracked as **#1228** and applies equally to `s3`, `gcs`, and `tiered`.

So enabling `cas` on `tiered` reaches **parity with `s3`/`gcs`** — it does not satisfy ADR D1's writer-class contract. Until #1228 closes, keep the orchestrator-level rule of one writer per `[state]` prefix for the seam commands. `rocky doctor`'s `state_concurrency` check reports this on every remote backend.

## Trust boundary (stated, not assumed silently)

The cache check proves **freshness, not provenance** — the framed generation is not authenticated. This is recorded in the module docs as a named assumption rather than a verified property, on the grounds that nothing authenticates the durable object either: whoever can write the state object owns the state. Under `off` the cached blob is served with no check at all, so `cas` strictly tightens this trust rather than introducing it.

## Notes

- No redb schema bump — the generation is the external ETag.
- The `off` default and the rollout semantics are unchanged; `off` remains byte-identical to previous behaviour.
- The ADR's D5 status and capability-matrix row are annotated to match what actually landed. D1's normative text and its closure criterion are untouched.

## Testing

`cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` clean · `cargo nextest run --all-features` **5667 passed** · `just codegen` drift-free.

Covered: commit/conflict on tiered with exactly one winner; a faulted cache write leaving a stale entry that is still rejected; stale-generation rejection; the matching-generation fast path still serving; bootstrap conflicting rather than overwriting; backend errors propagating as errors rather than conflicts; and `s3`/`gcs` behaviour unchanged.

**One unverified assumption, stated plainly:** generation equality across PUT/HEAD/GET on real S3. A live receipt test ships `#[ignore]`d behind `ROCKY_TEST_S3_*`. If a provider formats them differently per verb, the failure is fail-safe — every validation misses and reads stay durable — but the cache's performance benefit silently disappears. An in-memory store cannot catch this.


hugocorreia90:
## Residual resolved: the generation-identity assumption is now verified

The PR shipped one unverified premise — that `Generation` is identical across `PUT` / `HEAD` / `GET` — with an `#[ignore]`d receipt test behind `ROCKY_TEST_S3_*`. That test has now been run against a real S3-compatible endpoint (MinIO, via Docker).

```
running 5 tests
test live_freeze_marker_list_create_roundtrip ... ok
OK: S3 generation identity across PUT / HEAD / GET verified
test live_generation_is_identical_across_put_head_and_get ... ok
test live_put_if_not_exists_conflict ... ok
OK: S3 CAS roundtrip (bootstrap + match + stale-conflict + download-base) verified
test live_put_if_match_cas_roundtrip ... ok
test live_round_trip ... ok

test result: ok. 5 passed; 0 failed
```

So the tiered cache's validation path works end to end against a real object store, not just the in-memory harness: the conditional-write semantics (`If-Match` / `If-None-Match`), the stale-generation conflict, and the download-base threading all behave as designed.

**Scope of the evidence, stated honestly:** this is MinIO, not AWS. It exercises Rocky's real `Generation` type through the real `object_store` code path against an S3-compatible server, which is what the assumption is about — but it is not a guarantee about AWS specifically. AWS documents ETag-based conditional writes and strong read-after-write consistency, so the assumption is well-founded there too; a run against a real AWS bucket would still be the strongest form of this receipt.

PR #1239, opened 2026-07-26
fix(engine/rocky-compiler): make decimal AVG inference conservative

### LINKED ISSUE #1238: [engine/rocky-compiler] AVG(DECIMAL) is inferred as Float64, breaking contract validation
## Description

`rocky-compiler` infers every `AVG(...)` result as `RockyType::Float64`, even when the known input type is `RockyType::Decimal`.

That concrete type is incorrect for exact-numeric inputs on supported warehouses:

- Databricks documents `AVG(DECIMAL(p, s))` as `DECIMAL(p + 4, s + 4)`.
- BigQuery documents `AVG(NUMERIC)` as `NUMERIC` and `AVG(BIGNUMERIC)` as `BIGNUMERIC`.

Because the inferred schema flows into compile-time contract validation, Rocky rejects a valid Decimal contract with `E011` and accepts a Float64 contract for a model whose warehouse output is Decimal.

This is a compiler soundness issue, not only a display/type-hint problem.

## Steps to reproduce

Against `main` at `4bf43c8c1a4a8b5403dcd0a020e399e19d3aa9f4`:

1. Construct a source schema containing:

   ```rust
   TypedColumn {
       name: "amount".into(),
       data_type: RockyType::Decimal {
           precision: 10,
           scale: 2,
       },
       nullable: false,
   }
   ```

2. Compile/typecheck a model through the normal semantic-graph path:

   ```sql
   SELECT AVG(amount) AS avg_amount
   FROM source.raw.orders
   ```

3. Validate `avg_amount` against a Decimal contract and then a Float64 contract.

The reproduction was run twice from a clean detached worktree using the public `build_semantic_graph`, `typecheck_project_with_models`, and `validate_contract` APIs:

```text
inferred=Float64, nullable=true
decimal_contract=["E011"]
float_contract=[]
```

As a control, replacing `AVG(amount)` with `SUM(amount)` correctly produces:

```text
control_sum_inferred=Decimal { precision: 10, scale: 2 }
control_sum_decimal_contract=[]
```

The existing focused test also shows that current coverage encodes the unconditional rule:

```bash
cd engine
cargo test -p rocky-compiler test_infer_expr_avg --lib
```

It passes because `test_infer_expr_avg` asserts `Float64` without providing a Decimal input.

## Expected behavior

For a known Decimal input, `AVG` should resolve to the Decimal family rather than `Float64`.

Exact precision/scale widening is warehouse-specific. If the compiler cannot represent the target dialect's rule at this stage, `RockyType::Unknown` is a safer conservative result than an incorrect concrete type.

A Decimal contract should be accepted, and a Float64 contract should be rejected. Integer and floating-point `AVG` inputs can continue to resolve as Float64.

## Actual behavior

`infer_aggregation_type` ignores `input_type`:

```rust
"AVG" => (RockyType::Float64, true),
```

The same unconditional rule is duplicated in `infer_function_type`.

The normal compiler path calls `infer_aggregation_type` for an aliased aggregate with a known upstream source type, then passes the resulting schema to contract validation.

## Affected code

- `engine/crates/rocky-compiler/src/typecheck.rs`
  - `compute_model_typecheck`
  - `infer_aggregation_type`
  - `infer_function_type`
  - `test_infer_expr_avg`
- `engine/crates/rocky-compiler/src/compile.rs`
  - `compile_project`
  - `validate_all_contracts`

## Suggested fix and regression coverage

Branch `AVG` inference on the input type:

- Decimal input: retain a Decimal-family type, with target-aware widening if available.
- Decimal input without sufficient dialect context: return `Unknown`.
- Other supported numeric inputs: preserve the current Float64 behavior where correct.

Add a normal-path inline regression test that:

1. supplies a Decimal source schema,
2. typechecks aliased `AVG(amount)`,
3. positively asserts the inferred `(RockyType, nullable)`,
4. asserts the specific contract diagnostic: Decimal accepted and Float64 rejected with `E011`.

## References

- Databricks `AVG`: https://docs.databricks.com/aws/en/sql/language-manual/functions/avg
- BigQuery `AVG`: https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate_functions#avg

mattfaltyn:
## Summary

Make `AVG` type inference conservative for exact-numeric inputs whose result type depends on the target SQL dialect.

Closes #1238.

## Subproject(s) touched

## Changes

- Infer `AVG` of a `Decimal` input as `Unknown` because the compiler layer has no adapter context and engines disagree: DuckDB returns `DOUBLE`, while Databricks (`DECIMAL(p + 4, s + 4)`) and BigQuery (`NUMERIC`/`BIGNUMERIC`) preserve an exact-numeric family.
- Keep `AVG` of **every other** input — including an argument that does not resolve — inferred as `Float64`.
- Route both aggregation-lineage and parsed-expression inference through the same rule.
- Add normal-pipeline regression coverage for inferred types, nullability, and contract validation, plus focused expression/window tests.

`ModelIr` and public CLI/DSL contracts are unchanged.

## Review update (2026-07-26)

The first revision withheld `Float64` for every input that was not an integer or a float. That `else` branch also caught `RockyType::Unknown` — an argument Rocky failed to resolve — and `validate_contract` (`rocky-compiler/src/contracts.rs:144`) skips type validation when the inferred type is `Unknown`. The withhold therefore widened into **silence at a validation gate**.

Measured before/after through the real gate path (`build_semantic_graph` → `typecheck_project_with_models` → `validate_contract`, which is what `compile.rs` calls), running identical probe code against both revisions:

| `AVG(x)` argument | `main` | first revision | this revision |
|---|---|---|---|
| `Decimal(10,2)` | `Float64` — false **E011** on a Decimal contract (the reported bug) | `Unknown` — all contracts pass | `Unknown` — all contracts pass |
| `Int64` | `Float64` — mismatches rejected | unchanged | unchanged |
| `String` | `Float64` — mismatches rejected | `Unknown` — **all pass** | `Float64` — mismatches rejected |
| unresolved (no source schema) | `Float64` — mismatches rejected | `Unknown` — **all pass** | `Float64` — mismatches rejected |
| downstream `SELECT avg_amount FROM …` | rejected | **silence propagates** | rejected |

The carve-out is now bounded to `RockyType::Decimal`, so the only behavior change versus `main` is the one #1238 reports. `test_avg_over_unresolved_input_still_validates_contracts` pins the boundary.

### Known residual

A `Decimal`-input `AVG` column is no longer contract-checked at all — a `Float64`, `String` or `Boolean` contract on it now passes silently, where `main` rejected it. This is a deliberate, bounded trade of a false reject for a missed check, and it is **not** the end state: the correct outcome is a third gate verdict (`Unverifiable`) rather than silence, which is the `ContractVerdict{Satisfied, Violated, Unverifiable}` + `UNVERIFIABLE_POLICY` design already approved for the decimal-inference work. This column should move onto that verdict when it lands.

`breaking_change.rs` is unaffected: every consumer calls `diff_project_ir(&base_ir, &head_ir)` with both sides compiled by the same binary, and since `main` maps all `AVG` to a single type, there is no type *difference* on `main` that this change can erase.

### Dialect grounding

The `DECIMAL(p + 4, s + 4)` claim is verified against Databricks' primary documentation. Snowflake and Trino are **not** covered: Snowflake's `AVG` reference documents no return type at all, so the "engines disagree" premise is unverified for 2 of Rocky's 5 warehouse adapters, and the `Float64`-for-integers half is unsourced. Settling that needs a live-execute test against the Snowflake sandbox — tracked as follow-up, not a blocker for withdrawing a known-wrong concrete type.

## Test Plan

- `cargo test -p rocky-compiler` — 419 unit tests and 22 integration tests passed; 1 doctest ignored.
- `cargo nextest run --all-features --no-fail-fast` — **5,713 passed, 108 skipped, 0 failed**, run on this branch rebased onto `main` (`89cf525e`). This includes `run_watch_reruns_on_file_change_and_exits_clean_on_sigint`, which failed in the first CI run; that failure did not reproduce and is unrelated to this diff.
- `cargo clippy --all-targets --all-features -- -D warnings` — clean **workspace-wide** on rustc 1.96.0. The earlier note about pre-existing Rust 1.93 lints in generated diagnostics and `rocky-trino/src/test_helpers.rs` does not reproduce; no allow-list narrowing is needed.

## Checklist

- [x] If this is a CLI JSON schema or DSL syntax change, all consumers (`integrations/dagster/`, `editors/vscode/`) are updated in this same PR (not applicable)

### Engine (`engine/`)

- [x] `cargo test --all-targets` passes (covered by the full all-features nextest run above)

- [ ] `uv run pytest` passes

- [ ] `npm run lint` is clean

### DIFF
--- a/engine/crates/rocky-compiler/src/typecheck.rs
+++ b/engine/crates/rocky-compiler/src/typecheck.rs
@@ -1387,12 +1387,16 @@ fn infer_aggregation_type(func: &str, input_type: &RockyType) -> (RockyType, boo
             (ty, true)
         }
         "AVG" => {
-            // Exact-numeric AVG is dialect-dependent, and this layer has no
-            // adapter context. Avoid a false concrete type (#1238).
-            let ty = if input_type.is_integer() || input_type.is_float() {
-                RockyType::Float64
-            } else {
+            // `AVG` of an exact-numeric input is dialect-dependent and this layer
+            // has no adapter context: DuckDB returns DOUBLE, Databricks returns
+            // DECIMAL(p + 4, s + 4), BigQuery preserves NUMERIC/BIGNUMERIC. Only
+            // `Decimal` degrades to `Unknown` (#1238) — every other input keeps
+            // `Float64`, so an argument Rocky simply failed to resolve does not
+            // silently drop out of contract validation.
+            let ty = if matches!(input_type, RockyType::Decimal { .. }) {
                 RockyType::Unknown
+            } else {
+                RockyType::Float64
             };
             (ty, true)
         }
@@ -2567,6 +2571,50 @@ mod tests {
         );
     }
 
+    /// The `Decimal` carve-out must not widen into "any input Rocky failed to
+    /// resolve". `validate_contract` skips type validation when the inferred
+    /// type is `Unknown`, so an over-broad withhold would turn E011 into
+    /// silence for every `AVG` in a project compiled without source schemas —
+    /// a fail-open at a gate, not conservative inference.
+    #[test]
+    fn test_avg_over_unresolved_input_still_validates_contracts() {
+        let models = vec![make_model(
+            "averages",
+            "SELECT AVG(amount) AS avg_amount FROM source.raw.orders",
+        )];
+        let project = Project::from_models(models).unwrap();
+
+        let mut external = HashMap::new();
+        external.insert("source.raw.orders".to_string(), vec![]);
+        let graph = build_semantic_graph(&project, &external).unwrap();
+
+        // No source schema for the referenced table — the `AVG` argument does
+        // not resolve, exactly as with a cold schema cache or a fresh clone.
+        let sources = HashMap::new();
+        let result = typecheck_project_with_models(&graph, &sources, None, &project.models, None);
+        let columns = &result.typed_models["averages"];
+        let avg_amount = columns.iter().find(|col| col.name == "avg_amount").unwrap();
+        assert_eq!(avg_amount.data_type, RockyType::Float64);
+
+        let contract = CompilerContract {
+            columns: vec![ContractColumn {
+                name: "avg_amount".to_string(),
+                type_name: Some("Boolean".to_string()),
+                nullable: Some(true),
+                description: None,
+            }],
+            rules: ContractRules::default(),
+        };
+        let diagnostics = validate_contract("averages", columns, &contract);
+        assert!(
+            diagnostics
+                .iter()
+                .any(|diagnostic| &*diagnostic.code == "E011"
+                    && diagnostic.message.contains("avg_amount")),
+            "an unresolved AVG argument must not silence the contract check: {diagnostics:?}"
+        );
+    }
+
     #[test]
     fn test_cast_alias_uses_cast_type_not_source_type() {
         let models = vec![make_model(
@@ -3020,13 +3068,24 @@ mod tests {
     fn test_infer_expr_avg() {
         let mut scope = TypeScope::new();
         let expr = parse_expr("AVG(x)");
-        assert_eq!(infer_expr_type(&expr, &scope).0, RockyType::Unknown);
+
+        // An argument Rocky cannot resolve keeps `Float64`. Degrading it to
+        // `Unknown` would silently disable the column's E011 contract check
+        // (`contracts.rs` skips type validation for `Unknown`), which is a
+        // fail-open at a gate rather than conservative inference.
+        assert_eq!(infer_expr_type(&expr, &scope).0, RockyType::Float64);
 
         scope
             .columns
             .insert(CiKey::owned("x".to_string()), (RockyType::Int64, false));
         assert_eq!(infer_expr_type(&expr, &scope).0, RockyType::Float64);
 
+        scope
+            .columns
+            .insert(CiKey::owned("x".to_string()), (RockyType::Float32, false));
+        assert_eq!(infer_expr_type(&expr, &scope).0, RockyType::Float64);
+
+        // Only an exact-numeric input is dialect-dependent enough to withhold.
         scope.columns.insert(
             CiKey::owned("x".to_string()),
             (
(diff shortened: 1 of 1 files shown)


hugocorreia90 (approved):
Reviewed, and pushed one commit (`ee20246d`) to bound the fix. Approving with that in.

**The bug is real and the diagnosis is right.** Verified `avg(DECIMAL(p,s))` → `DECIMAL(p+4, s+4)` against Databricks' primary docs, and reproduced the false `E011` on a Decimal contract on `main`.

**What needed fixing.** The withhold was keyed on "not integer and not float", which also catches `RockyType::Unknown` — an argument Rocky failed to resolve. `validate_contract` (`rocky-compiler/src/contracts.rs:144`) skips type validation entirely when the inferred type is `Unknown`:

```rust
if !matches && col.data_type != RockyType::Unknown {   // ← Unknown ⇒ E011 never fires
```

So the withhold widened from "don't assert a dialect-dependent type" into "stop checking this column". I measured it through the real gate path with identical probe code on both revisions: an `AVG` over an unresolved argument went from rejecting a `Boolean` contract to accepting it, and the silence propagated to downstream models through direct-select lineage. In a project compiled without source schemas that is *every* `AVG` column, not an edge case.

`AGENT_REVIEW.md` is explicit that `Unknown` is a valid inference outcome and not a defect — that rule governs **inference**. The problem here was an `Unknown` consumed by a **validation gate**, where it reads as "compatible". Conservative for one, permissive for the other.

**The fix** narrows the carve-out to `RockyType::Decimal`. Everything else keeps `Float64`, so the only behavior change versus `main` is the one #1238 reports. Added `test_avg_over_unresolved_input_still_validates_contracts` to pin that boundary; inverted `test_infer_expr_avg`, which had frozen the unresolved-argument silence as intended behavior.

**Residual, disclosed in the body rather than papered over:** a Decimal-input `AVG` column is now not contract-checked at all — any declared type passes. That is a bounded trade of a false reject for a missed check, and the right end state is an `Unverifiable` gate verdict rather than silence. This column should move onto `ContractVerdict` when that lands.

**Two things the test plan claimed that did not hold, both now corrected in the body:** the first CI run failed and cancelled at 1,211/5,704 with **zero `rocky-compiler` tests executed**, so the new tests had no CI evidence; and workspace clippy is clean on rustc 1.96.0, so the checklist item can be ticked. The `run_watch` SIGINT failure did not reproduce across a full 5,713-test local run on a rebased base.

Not blocking, worth logging: the "engines disagree" premise is grounded for DuckDB, Databricks and BigQuery but not for Snowflake or Trino — Snowflake's `AVG` reference documents no return type, and the `Float64`-for-integers half cites nothing. A live-execute test against the Snowflake sandbox is the way to settle it.

PR #1241, opened 2026-07-26
feat(engine/rocky-core): compare-and-swap the policy-freeze ledger seam

hugocorreia90:
## Summary

First slice of #1228. Adds a generation-owning `LedgerSeamSession` that compare-and-swaps the shared `state.redb` blob with bounded retries, and migrates the **policy-freeze** seam onto it.

Refs #1228 (does **not** close it — see *What remains open*).

## Subproject(s) touched

## The problem

The end-of-run writer already compare-and-swaps: `RemoteStateSession::finalize` commits against a `Generation` captured at `acquire`. Three ledger seams do not — they call `upload_only_fail_closed`, which forces upload errors to propagate and then delegates to the **unconditional** `upload_state`.

So a run R and a concurrent seam both download generation G0; R CAS-commits G1 and reports success; the seam then unconditionally overwrites the object with its G0-derived blob, erasing R's committed watermarks and run records, and exits 0. No conflict is raised.

ADR-CONCURRENCY D1 requires CAS on every shared-blob write, split by writer class: cheap ledger seams **retry**, full runs **refuse**.

## Design

The retry unit is the **full typed transition**, not a ledger delta:

```rust
attempt(fresh_store, fresh_base) -> Result<T, _>
```

Each attempt downloads the blob *and* its generation together, re-runs the transition against the winner's state, applies its ledger mutation, drops the store, and CASes. The helper returns `T` from the CAS-winning attempt — a `Result<()>` would let a command print stale first-attempt output once `gc` migrates, since a retry can turn "evicted" into "already evicted" or "held".

Three total attempts (initial + 2), jittered, counted separately from transport retries. On exhaustion: the remote winner is preserved, a typed `LedgerSeamConflict` is returned, the command exits nonzero, and no unconditional upload is ever issued.

Two properties worth calling out because they are easy to get wrong:

- **Freeze markers stay outside the retry closure.** They are create-once OR-set members: re-writing the same UUID returns `IdCollision`, and minting a fresh one would create a *second independent active freeze*. The marker is written once, before the loop; only the exact pre-constructed ledger records are replayed. If retries exhaust, the already-created marker stays engaged — over-enforcement is the monotone-safe direction.
- **Unfreeze keeps its inverse ordering** — ledger row first, then the marker.

Nothing is reconstructed across retries that embeds `Utc::now()`.

## What remains open

This is deliberately partial.

- `gc apply` and the shared `apply.rs` ledger helper still use the unconditional path. Each needs its own typed transition — `gc` is replayable only by re-running `execute_gc_apply` end to end (fresh candidate/refcount derivation and both liveness reads, since a Delta re-add legitimately flips a tombstone into a hold), and restore needs tombstone re-resolution plus a hash-exactness proof.
- `upload_only_fail_closed` is therefore **kept**, with a `TODO(#1228)`. Removing production access to it lands with the last seam PR — that structural removal, not a `seams_use_cas()` boolean, is what will make the guarantee provable.
- `rocky doctor` is untouched and still reports Warning. That is correct while the migration is partial.
- Behaviour is inert unless `[state] concurrency_control = "cas"` on a CAS-capable backend. `off` remains the default, so this changes nothing for existing deployments until they opt in.

**Scope of the guarantee:** this protects the shared `state.redb` blob from lost updates. It is *not* "all remote state writes are CAS-protected" — freeze markers, idempotency objects and doctor's read/write probe are independent keys by design and remain outside CAS. Relatedly, CAS does not close the gate-to-mutation TOCTOU: a marker landing after a seam's LIST does not advance the blob generation, so it raises no conflict. RD-006 (warehouse ahead of committed state) is unaffected and still open.

## Test Plan

Semantic-retry coverage, not conditional-put mechanics — the existing session-level conflict tests already cover the latter. All new tests drive the production `run_policy_freeze` entry point, not a replica.

- `freeze_marker_is_written_once_across_two_cas_conflicts` — arms two real precondition failures, asserts the third attempt commits, and asserts marker `Create == 1`, blob `Create == 3`, **`Unconditional == 0`**. The last assertion is what makes "never falls back to an unconditional upload" a tested property.
- `freeze_cas_exhaustion_is_nonzero_and_preserves_remote_winner` — winner preserved, nonzero exit, marker stays engaged.
- `unfreeze_marker_is_not_written_before_ledger_cas_wins` — pins the inverse ordering.
- `freeze_local_backend_records_without_remote_sync` — `Local` issues no object-store operations at all.

Gates:

- `cargo test -p rocky-core` — 1,746 passed, 1 ignored; all integration targets green.
- `cargo test -p rocky-cli --lib commands::policy` — 21 passed.
- `cargo clippy --all-targets --all-features -- -D warnings` — clean.
- `cargo fmt -- --check` — clean.

## Known residuals

- After retry exhaustion the local ledger file retains the losing replay until the next download. The remote winner is provably preserved and the command exits nonzero, and every remote-backed path downloads before reading, so this should self-heal — but that is reasoning, not a test.
- No live S3/GCS/tiered two-writer semantic race test yet. In-memory cross-pod coverage is green; a provider-backed race would additionally validate generation metadata and tiered-cache behaviour together.

## Checklist

- [x] Not a CLI JSON schema or DSL syntax change — no consumer updates needed

### Engine (`engine/`)


hugocorreia90:
**Red team: `codex review --base main` (gpt-5.6-sol, xhigh).** One P2 finding, confirmed and fixed in `ffc6aee1`.

**Finding:** this branch made two operator-facing texts factually false. Both the `state_concurrency` doctor check (`doctor.rs:710-733`) and the `ConcurrencyControl` rustdoc (`config.rs:547-555`) listed `rocky policy` among the writers that publish state unconditionally — which stopped being true the moment freeze/unfreeze moved onto `LedgerSeamSession`.

The direction of the error was fail-safe (it overstated exposure rather than understating it), but shipping a diagnostic we know misdescribes the code is the defect class this repo treats as first class, and this PR introduced it. Fixed by naming only the writers that still bypass CAS. The check deliberately stays `Warning`: `rocky gc` and `rocky apply` are enough for a concurrent command to overwrite a run's committed state, so no deployment is fully protected yet. It flips to Healthy when the last seam lands, not this one.

`schemars` exports that rustdoc verbatim as the JSON Schema `description`, so the regenerated project schema, VS Code types and OpenAPI document moved with it. The `editors/vscode/package-lock.json` churn `just codegen` produced was reverted — it was macOS `npm install` dropping `libc` metadata from linux-only optional deps, unrelated to this change and a regression for Linux CI.

**Disposition of the rest:** no other findings. Two residuals the reviewer raised are disclosed in the PR body rather than fixed here — no provider-backed live S3/GCS/tiered contention test (in-memory cross-pod coverage is green), and the local ledger retaining a losing replay after retry exhaustion (the remote winner is provably preserved and every remote-backed path downloads before reading, so it should self-heal, but that is reasoning rather than a test).

**Independence caveat, on the record:** Codex authored this code, so this pass is a second adversarial look, not an independent red team in the sense `AGENTS.md` requires. The cross-model review was mine (Claude Opus) — I verified each binding requirement against the diff, confirmed the new tests drive the production `run_policy_freeze` rather than a replica, and re-ran every gate myself. Worth noting my first gate run used `-p rocky-core` and therefore did not execute the new tests at all, which live in `rocky-cli`; they were run separately.

**Open nit, not fixed:** the `unreachable!()` after the attempt loop is unreachable by construction today, but would panic if `LEDGER_SEAM_MAX_ATTEMPTS` were ever set to `0`. Left alone deliberately — restructuring the control flow of a concurrency primitive for a cosmetic gain is the wrong trade.

hugocorreia90:
**Correction to my own fix.** Codex's stop-time review caught that `ffc6aee1` published a self-contradictory CAS contract. It was right; fixed in `aef47dad`.

Two contradictions, both introduced while fixing the previous one:

1. The rustdoc still asserted compare-and-swap applies to **"every remote state write."** It does not, and never will — freeze markers, idempotency records and the doctor read/write probe are written under their own keys. Being create-once or self-keyed they cannot lose an update, so they sit outside CAS *by design*, not as an unclosed gap.
2. Listing the `rocky policy` seam as protected without qualification implies its **markers** are CAS-committed. They are not, and must not be: a replayed marker UUID collides, and a freshly minted one would create a second independent active freeze. Markers are deliberately written once, outside the retry. Only the seam's **ledger write** is CAS-committed.

Together those read as "policy freeze is fully CAS-protected," which is false in a way that matters — it is the marker half that carries the kill switch.

Both statements are now scoped to the shared state object, with the marker carve-out named explicitly in the rustdoc and the doctor diagnostic. Regenerated artifacts follow the exported description; the `package-lock.json` churn was reverted again.

Worth recording plainly: this is the exact requirement I wrote into the implementation brief — scope every claim to the shared `state.redb` blob, never imply CAS covers the sibling keys — and I broke it myself in the commit that was fixing someone else's stale claim. Third instance in this PR's lineage of "correcting a wrong statement by publishing a subtly wrong one." The PR body's own scope section was accurate throughout; the drift was confined to the code-adjacent text I edited by hand.

hugocorreia90:
**Third red-team pass + fix chain.** The CAS-contract wording is settled; the pass surfaced one new P2, which turned out to be a real bug rather than the cosmetic residual I had written off.

**P2 — losing transition stayed locally visible after CAS exhaustion.** On three conflicts, the final attempt's mutation remained in the local ledger. I had previously reasoned this self-heals "because every remote-backed path downloads before reading." That was wrong, and I asserted it without checking: `audit.rs:47,157,907` and `brief.rs:115,160` open the store directly with no download at all. So a failed `rocky policy unfreeze` left an uncommitted unfreeze row locally while the remote blob stayed frozen, and `rocky audit` would report an all-clear on an engaged kill switch.

Fixed in `2bf49a64`: the exhaustion arm best-effort restores the remote winner, warning (never masking) if that restore fails, and returns `LedgerSeamConflict` unchanged either way. I verified the regression is a real discriminator by reverting the production hunk — the test fails on exactly the intended assertion, and passes again with it restored.

**Two follow-on defects the fix introduced, both caught by independent verification rather than self-report:**

`54bc10d1` — the accompanying negative test made the suite non-deterministic. Load-matched measurement, five runs each on the same machine: baseline `74050a86` **5/5 green**, fix commit **2 of 5 failed** — and it took an unrelated pre-existing test (`upload_state_unless_recreated_...`) down with it, while passing in isolation. Classic cross-test interference. That test was the optional extra in the brief and had stopped being cheap, so it was deleted rather than patched with sleeps or `#[ignore]`. The primary regression stays. Worth noting it had never reached CI, so this would have merged and then failed intermittently on unrelated PRs.

`f86ceb7b` — the stabilization gave the upload scratch database a unique suffix under `cfg(test)` only. That split the suite away from the code it covers, and left the collision it was working around live in production: two uploads observing the same `nanos` on a coarse clock derive the same scratch path, and that file is read back and uploaded, so a collision publishes another upload's bytes. The sequence is now unconditional — no test/production divergence, and the hazard is closed.

Determinism re-verified by me, not reported: 10 consecutive `cargo test -p rocky-core --lib` runs, 10/10 green. Plus clippy, fmt, and `rocky-cli` policy 21/21.

**Independence, on the record:** every red-team pass here was Codex reviewing code Codex wrote. Its attempts to reach an independent model were blocked by the environment, and it disclosed that each time. The cross-model check was mine, and it is what caught both follow-on defects — but I am also the author of the briefs, so this PR has not had the independent review `AGENTS.md` asks for.

hugocorreia90:
**Fourth pass — `6c1ce913`.** Codex's stop-time review flagged contradictory `cas` capability guidance, and it was right again. My two previous fixes each corrected the claim in front of me and missed the rest.

Migrating the policy seam falsified **six** statements that still described it as an unconditional writer. I had fixed two (the `ConcurrencyControl` enum rustdoc and the doctor diagnostic). The four still standing:

- the `concurrency_control` **field** doc — which framed `cas` as making a lost race "fail-close instead of silently overwriting the winner's committed state", the exact thing `gc`/`apply` still do. Schema-exported, so it reached the JSON schema, the VS Code types and the SDK Pydantic model.
- `reference/cli.md` — doctor table row listing `rocky policy` as unconditional.
- `reference/configuration.md` — "what `cas` does not yet cover".
- `concepts/state-management.md` — the ledger-seam note.
- the **unreleased** tiered-CAS changelog entry, which named `rocky policy freeze` among the seams that still write unconditionally. That one would have shipped a self-contradiction in the same release as this change.

All six now agree, and the changelog gains an entry for the migration itself.

The thing worth extracting: I found these by grepping every `#1228` reference in the repo and treating the result as the closure set, rather than by fixing the files this branch happened to touch. Both earlier rounds used the latter method and both were incomplete. A change that alters a capability has to enumerate everything that *describes* that capability, not everything it edits.

`rocky gc` and `rocky apply` remain unconditional, so every statement still says `cas` does not yet eliminate lost updates, and `rocky doctor` still warns for every remote deployment.

Gates after the change: fmt, clippy `--all-targets --all-features`, `rocky-core` lib 1746, `rocky-cli` commands 1074 — all green. Codegen regenerated; `package-lock.json` churn reverted again.

PR #1244, opened 2026-07-27
fix(engine): propagate DAG model load errors

### LINKED ISSUE #1243: run --dag reports success after model loading fails
## Description

`rocky run --dag` converts a top-level model-loading error into an empty model
list. A transformation-only project then builds an empty DAG, exits 0, and
reports zero failed nodes even though no transformations were materialized.

This can leave production tables stale while an orchestrator records a
successful run.

## Steps to Reproduce

Create a project with these files:

`rocky.toml`:

```toml
[adapter.local]
type = "duckdb"
path = "proj.duckdb"

[pipeline.silver]
type = "transformation"

[pipeline.silver.target]
adapter = "local"
```

`models/broken.sql`:

```sql
SELECT 1 AS id
```

`models/broken.toml`:

```toml
name = [
```

Run:

```bash
rocky --config rocky.toml --output json --state-path state.redb run --dag
```

## Expected Behavior

The command exits non-zero and reports the malformed model sidecar.

## Actual Behavior

The model-load error is discarded. The command returns a successful
`DagRunOutput` with `total_nodes = 0`, `failed = 0`, and no node results.

## Root Cause

`run_with_dag` uses:

```rust
super::dag::load_all_models(&models_dir).unwrap_or_default()
```

The sibling `rocky dag` command correctly propagates the same loader result
with `?`. Once the error is converted to an empty list, transformation DAG
construction adds no nodes and `DagExecutor` treats the empty result as
successful.

## Suggested Fix

Propagate the loader error from `run_with_dag` and add a CLI regression test
that asserts a malformed top-level model causes a non-zero exit.

## Rocky Version

`rocky 1.67.0`; also present on `main@e3aff608e0bdcbdf291abf89ee560d19f0c6ef98`.

## OS / Architecture

macOS 26.5.2 arm64.

## Configuration

Credential-free DuckDB configuration shown above.

mattfaltyn:
## Summary

Make `rocky run --dag` fail when top-level model loading fails instead of
silently running an empty transformation DAG and reporting success.

Closes #1243

## Subproject(s) touched

## Changes

- Propagate `load_all_models` errors from the DAG run path.
- Add a binary regression test proving that a malformed model sidecar exits
  non-zero and reports the loading failure.

## Test Plan

- [x] `cargo test -p rocky --test run_dag_json_stdout`
- [x] `CARGO_BUILD_JOBS=4 cargo nextest run --all-features` (5,721 passed)
- [x] `CARGO_BUILD_JOBS=4 cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo fmt -- --check`
- [x] `git diff --check`

## Checklist

- [x] Commit message follows conventional commits and is scoped to `engine`.
- [x] No CLI JSON schema or DSL syntax changes.

### DIFF
--- a/engine/crates/rocky-cli/src/commands/run_dag_exec.rs
+++ b/engine/crates/rocky-cli/src/commands/run_dag_exec.rs
@@ -177,7 +177,7 @@ pub async fn run_with_dag(
     let config_dir = config_path.parent().unwrap_or(Path::new("."));
     let models_dir = config_dir.join("models");
     let models = if models_dir.is_dir() {
-        super::dag::load_all_models(&models_dir).unwrap_or_default()
+        super::dag::load_all_models(&models_dir)?
     } else {
         Vec::new()
     };
--- a/engine/rocky/tests/run_dag_json_stdout.rs
+++ b/engine/rocky/tests/run_dag_json_stdout.rs
@@ -54,6 +54,18 @@ depends_on = ["ingest"]
 adapter = "default"
 "#;
 
+const TRANSFORMATION_ONLY_ROCKY_TOML: &str = r#"
+[adapter]
+type = "duckdb"
+path = "fixture.duckdb"
+
+[pipeline.transform]
+type = "transformation"
+
+[pipeline.transform.target]
+adapter = "default"
+"#;
+
 const MODEL_SQL: &str = "SELECT order_id, customer_id, amount FROM raw__orders.orders\n";
 
 const MODEL_TOML: &str = r#"
@@ -132,5 +144,39 @@ fn run_dag_json_stdout_is_a_single_json_document() {
     assert!(
         stderr.contains("Copied"),
         "expected the replication summary on stderr, got:\n{stderr}"
+    );
+}
+
+#[test]
+fn run_dag_fails_when_model_loading_fails() {
+    let tmp = tempfile::tempdir().expect("tempdir");
+    let dir = tmp.path();
+    fs::write(dir.join("rocky.toml"), TRANSFORMATION_ONLY_ROCKY_TOML).expect("write rocky.toml");
+    let models_dir = dir.join("models");
+    fs::create_dir(&models_dir).expect("mkdir models");
+    fs::write(models_dir.join("broken.sql"), "SELECT 1 AS id\n").expect("write model sql");
+    fs::write(models_dir.join("broken.toml"), "name = [\n").expect("write malformed model toml");
+
+    let out = Command::new(env!("CARGO_BIN_EXE_rocky"))
+        .arg("-c")
+        .arg(dir.join("rocky.toml"))
+        .arg("run")
+        .arg("--dag")
+        .arg("--output")
+        .arg("json")
+        .current_dir(dir)
+        .env("RUST_LOG", "error")
+        .output()
+        .expect("spawn rocky");
+
+    let stdout = String::from_utf8(out.stdout).expect("utf8 stdout");
+    let stderr = String::from_utf8(out.stderr).expect("utf8 stderr");
+    assert!(
+        !out.status.success(),
+        "malformed models must fail the DAG run\n--- stdout ---\n{stdout}\n--- stderr ---\n{stderr}"
+    );
+    assert!(
+        stderr.contains("failed to load models from"),
+        "expected the model-load error on stderr, got:\n{stderr}"
     );
 }


hugocorreia90 (commented):
The fix is correct and I verified it against a pre-fix binary rather than by reading alone. One introduced behavior change is worth a decision before merge, and three pre-existing siblings keep #1243's class partly open.

Independently red-teamed with a different model (OpenAI Codex); its findings are folded in below.

## Verification

Built this branch and compared it against a binary built from the parent commit:

- **Pre-fix, on the issue's fixture:** exit 0, `total_nodes: 0, failed: 0`, empty `nodes[]`, and *nothing on stderr*. Confirms the report exactly.
- **Post-fix:** exit 1, error names the file and the TOML parse position.
- **Mutation check:** running the new test's fixture against a pre-fix binary exits 0, so the `!out.status.success()` assertion genuinely fails without the change. The test binds.
- **Test premise:** swapping the malformed sidecar for a valid one yields `total_nodes: 1` labelled `"ok"`, so that fixture's pipeline really would have produced a node. The broken model collapsed a live pipeline to zero.
- **Happy path:** a real playground POC gives identical output (`3 nodes, 0 failed`) before and after.
- **Error contract:** on the new path stdout is empty, the error goes to stderr, exit code 1 — the same shape as the pre-existing `?` on config load two lines above, which is already after `reserve_stdout_for_json()`. Nothing downstream is perturbed: `--dag` isn't reachable from `rocky-sdk` (`client.py` exposes only the read-only `rocky dag`), so `dagster-rocky` is unaffected.

The strongest argument for the change isn't in the PR description: `apply.rs` already fail-closes the *non-dag* path on this exact invariant — "*apply reports SUCCESS without executing the planned models*" — and `apply.rs` routes `run_plan.dag` through `run_with_dag`. The DAG path was the remaining hole in a fail-closed invariant the codebase had already committed to.

## Introduced — worth a decision

`run_with_dag` loads `<config_dir>/models` unconditionally, but only transformation pipelines consume models (`add_transformation_nodes` in `unified_dag.rs`). A replication-only, quality-only, or snapshot-only project with a models directory now fails on models it would never execute.

I swept every in-repo project that has a `models/` directory next to its `rocky.toml`. Exactly **1 of 83** newly fails: `engine/examples/docs-demo`, whose `stg_customers.rocky` uses `||` for string concatenation, which the lexer tokenizes as `Pipe` (`token.rs`). That file has never parsed — it dates to the initial commit — and `rocky docs`, the example's documented usage, already fails on it today. So this PR surfaces a latent breakage rather than causing one, and the diagnostic it prints is precise.

Worth noting the asymmetry it creates. Plain `rocky run` only enters the compiled-model phase under `run_all || models_dir.is_some()`, so it never touches that directory on a replication-only project. Verified on `docs-demo` with this branch: plain `rocky run` exits 0, `rocky run --dag` exits 1. Scoping the load to "at least one transformation pipeline is configured" would remove the asymmetry and keep everything this PR fixes.

## Pre-existing, left open — #1243's class isn't fully closed

1. **Subdirectory load failures are still swallowed** (`models_loader.rs` takes subdirectories through `if let Ok(sub)`). Reproduced: a healthy project plus `models/staging/broken.toml` exits 0 with `failed: 0` and never mentions the broken model. If every model lives in a failing subdirectory, this reproduces the original zero-node success exactly.
2. **The configured models glob is still ignored.** `run_with_dag` hardcodes `<config_dir>/models`, but a transformation pipeline's `models` field is configurable (`default_models_glob` in `config.rs`), and `--models` doesn't `conflicts_with = "dag"`. A project with `models = "transforms/**"` and no `./models` still gets `total_nodes: 0` and exit 0 — the same silent success by a different route, which this one-liner never reaches.
3. **`discover_seeds(...).unwrap_or_default()`** two lines below is the same shape. A malformed seed sidecar yields an empty seed list, so no seed node and no seed→model edge is created; a model then runs against stale seed data and the DAG reports success.

## Nit

The new test captures `stdout` but only asserts on status and stderr. Since this file exists to pin the "exactly one JSON document on stdout" contract, adding `assert!(stdout.trim().is_empty())` would catch a future regression that writes a stray non-JSON line on the error path. The current implementation already emits empty stdout, so this is coverage, not a live defect.

hugocorreia90 (approved):
Approving. Verified against a pre-fix binary: the bug reproduces exactly as reported (exit 0, `total_nodes: 0`, silent stderr), the new test genuinely fails without the change, and the happy path is unchanged. The introduced strictness on a `./models` directory that non-transformation pipelines don't consume is a known, accepted trade here — it's loud and precisely diagnosed, versus a silent success that leaves warehouse tables stale. Tracking the residuals separately.

Issue #1262, opened 2026-07-27
Models below the first subdirectory level are never loaded, despite the default models/** glob


hugocorreia90:
## Description

`load_project_models` reads the top-level models directory plus **one** level of immediate subdirectories. A model at `models/a/b/` is not loaded at all — not an error, simply absent — even though the default transformation glob is `models/**`, which reads as fully recursive.

This is a third silent-drop in the same family as #1243 (fixed for the top level in #1244) and #1257 (subdirectory *errors*): here it is valid models that vanish.

## Steps to Reproduce

```
models/
  top.sql / top.toml
  a/
    lvl1.sql / lvl1.toml
    b/
      lvl2.sql / lvl2.toml
```

```bash
rocky list models --models models --output json
```

## Actual Behavior

Only `top` and `lvl1` are returned. `lvl2` is absent. Replacing `lvl2.toml` with malformed TOML changes nothing — the file is never read, so it cannot even fail.

Any command built on this loader inherits it, including `run --dag`: a project that organizes models as `models/marts/core/` silently builds nothing for that subtree and reports success.

## Expected Behavior

Either the loader walks the full tree the `models/**` glob describes, or the supported depth is documented and a deeper file is reported rather than ignored.

## Risk to weigh before recursing

Making the loader fully recursive would newly discover `.sql` / `.rocky` files that projects may keep under `models/` deliberately unbuilt (archives, scratch, vendored subtrees). Those would start materializing tables. That is a behavior change worth an explicit decision rather than a silent widening — which is why this is filed separately from #1257 rather than folded into it.

## Notes

Latent in-repo: no example or POC nests models deeper than one level, so CI is green.

hugocorreia90:
Scope finding from the independent reviews of candidate fix PR #1307: **the one-level walk is not one defect — it is a family of five, in five different scanners.** Closing this issue with the loader alone leaves four siblings, two of which turn the loader fix into a new hazard.

| # | Scanner | Depth today | Consequence if only the loader recurses |
|---|---|---|---|
| 1 | `models_loader::load_project_models` | root + 1 level | (the PR fixes this one) |
| 2 | `Project::load_models_with_db` (compile path) | **root only** | a deep model becomes a DAG node whose sub-run compiles flat → `model not found`, plus the promote-gate set mismatch — the PR's four HIGH findings |
[...]

PR #1263, opened 2026-07-27
fix(engine): make the docs-demo example parse, and pin every example's models dir

### LINKED ISSUE #1260: docs-demo example: stg_customers.rocky has never parsed (|| is not a DSL operator)
## Description

`engine/examples/docs-demo/models/stg_customers.rocky` cannot be parsed by any Rocky version. It uses `||` for string concatenation:

```
derive {
    full_name: first_name || " " || last_name,
    ...
}
```

The lexer has no `||` token — `|` is `Token::Pipe`, the pipeline separator — so the parser reports:

```
failed to parse .rocky file 'models/stg_customers.rocky':
unexpected token at offset 136: expected identifier, found Pipe
```

## Impact

- `rocky docs`, the example's own documented usage, fails on it.
- Every command that loads that models directory fails: `dag`, `list`, `compile`, `preview`, `compliance`, and since #1244 also `rocky run --dag`.
- It is the only project in a clean checkout (78 have a `models/` directory beside a `rocky.toml`) whose top-level model load fails.

The file dates to the initial commit and no test covers it, which is why this went unnoticed.

## Suggested Fix

Either rewrite the expression using a supported concatenation form, or add `||` to the DSL. The example README documents the derived `full_name` column, so the model is intentional — only the operator is unsupported.

Adding `engine/examples/` to whatever loads the playground POCs in CI would stop this class recurring.

hugocorreia90:
## Summary

`engine/examples/docs-demo/models/stg_customers.rocky` has never parsed. It uses `||` for string concatenation, which the DSL lexer tokenizes as two `Pipe` tokens — `|` is the pipeline separator and there is no `||` operator. The file dates to the initial commit, so `rocky docs`, the example's own documented usage, has always failed on it.

It surfaced only when `rocky run --dag` started propagating model-load errors (#1244). Nothing caught it because no test loads the examples.

Closes #1260

## Subproject(s) touched

## Changes

- Rewrite the derived `full_name` column as `concat(first_name, " ", last_name)`. The parser already supports function calls, so this needs no DSL change. The README documents that column, so the model itself was intentional — only the operator was unsupported.
- Add `engine/rocky/tests/examples_models_load.rs`: walk the repository for every `rocky.toml` and load the `models/` directory beside it through the CLI's own loader.

Discovery is filesystem-driven rather than a hardcoded list, so a new example is covered the moment it lands. The test also asserts a floor on how many directories it checked, so a broken repository walk fails loudly instead of passing vacuously over zero projects.

## Test Plan

- [x] `cargo test -p rocky --test examples_models_load` — 78 project models directories load.
- [x] Mutation check: with the example reverted to `||`, the test fails and names the file, the offset, and the `Pipe` token. It binds.
- [x] `cargo fmt -- --check`

## Scope note

The loader reads the top level plus one directory down, so this test cannot cover a model at `models/a/b/` — that file is not loaded at all today (#1262). The test pins what the loader actually claims to read, not more.

## Review

Red-teamed by OpenAI Codex (`gpt-5.6-sol`, `xhigh` reasoning) together with #1264 and #1266; findings and their disposition are in a comment below.

**What I am least confident about:** whether `concat(...)` is the concatenation form this DSL wants to be idiomatic. It parses and lowers today, but if `||` is meant to be supported, the real fix is the operator and this example goes back to reading naturally. I picked the change that needs no cross-subproject cascade; the operator is a separate decision.

**The most important thing I am missing:** this test only covers projects that have a `models/` directory beside a `rocky.toml`. Example projects are exercised by no other automated check — nothing runs their documented commands — so an example whose `rocky.toml` or README drifts from the CLI still ships broken. Loading models is the cheapest slice of that problem, not the whole of it.

## Checklist

### DIFF
--- a/engine/examples/docs-demo/models/stg_customers.rocky
+++ b/engine/examples/docs-demo/models/stg_customers.rocky
@@ -2,7 +2,7 @@
 from raw_customers
 where is_active == true
 derive {
-    full_name: first_name || " " || last_name,
+    full_name: concat(first_name, " ", last_name),
     days_since_signup: current_date - signup_date
 }
 select {
--- a/engine/rocky/tests/examples_models_load.rs
+++ b/engine/rocky/tests/examples_models_load.rs
@@ -1,0 +1,87 @@
+//! Every in-repo example project's models directory must load.
+//!
+//! `engine/examples/docs-demo/models/stg_customers.rocky` shipped from the
+//! initial commit with `||` for string concatenation, which the DSL lexer
+//! tokenizes as two `Pipe` tokens. Nothing caught it: no test loaded the
+//! examples, so the file never parsed and `rocky docs` — the example's own
+//! documented usage — failed on it. It only surfaced when `run --dag` started
+//! propagating model-load errors (#1244).
+//!
+//! This walks the repository for every `rocky.toml`, and loads the `models/`
+//! directory beside it through the same loader the CLI uses. Discovery is
+//! filesystem-driven rather than a hardcoded list, so a newly added example is
+//! covered the moment it lands.
+//!
+//! Scope note: the loader reads the top level plus one directory down, so a
+//! model at `models/a/b/` is not covered here because it is not loaded at all
+//! (tracked separately). This test pins what the loader actually claims to read.
+
+use std::path::{Path, PathBuf};
+
+/// Repository root: `engine/rocky` → `engine` → repo root.
+fn repo_root() -> PathBuf {
+    Path::new(env!("CARGO_MANIFEST_DIR"))
+        .join("..")
+        .join("..")
+        .canonicalize()
+        .expect("canonicalize repo root")
+}
+
+/// Every `rocky.toml` in the repo, skipping build and dependency output.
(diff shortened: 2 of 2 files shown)


hugocorreia90:
Red team: **OpenAI Codex (`gpt-5.6-sol`, `xhigh` reasoning)**, reviewing this branch alongside #1264 and #1266. Verdict for this branch: **MERGE**, no findings.

What it confirmed rather than took on trust:
- `concat(...)` parses as a function call and lowers to `CONCAT(...)`, so the rewrite is valid rather than merely parseable.
- The docs-demo mutation is top-level, so the new test definitely binds — matching the local mutation check, where reverting the example produced a failure naming the file, the offset and the `Pipe` token.
- The `>= 50` floor is not vacuous: the walk finds 78 directories. It correctly noted the floor is a coarse partial-walk guard rather than a proof the walk is complete, which is fair — it catches "found nothing", not "found half".

Findings it raised against the sibling PRs (a policy fail-open in #1266 and a glob-shape regression in #1264) do not touch this branch.

Issue #1261, opened 2026-07-27
run --dag and dag fail with "circular dependency detected involving: []" for any project with 2+ transformation pipelines


hugocorreia90:
## Description

`build_unified_dag` calls `add_transformation_nodes` once **per transformation pipeline**, passing the same flat model list each time. Every model therefore gets a node under every transformation pipeline, and the node id is `NodeId::new("transformation", model_name)` — identical across pipelines. The duplicated nodes and their duplicated intra-pipeline edges make phase computation report a cycle whose node list is empty.

The result: **the unified DAG is unusable for any project with two or more transformation pipelines**, and the error names no node.

## Steps to Reproduce

Credential-free, one model:

```toml
[adapter]
type = "duckdb"
path = "p.duckdb"

[pipeline.silver]
type = "transformation"
[pipeline.silver.target]
adapter = "default"

[pipeline.gold]
type = "transformation"
[pipeline.gold.target]
adapter = "default"
```

`models/ok.sql`: `SELECT 1 AS id`, with a sidecar naming a target.

```bash
rocky dag --models models --output json
rocky run --dag --output json
```

## Actual Behavior

Both commands:

```
Error: failed to compute execution phases

Caused by:
    circular dependency detected involving: []
```

Deleting the second transformation pipeline makes the same project succeed with `total_nodes: 1`, so the second pipeline is the trigger, not the model or the adapter config.

## Expected Behavior

Either each pipeline gets its own nodes (distinct ids, models attributed to the pipeline whose configured `models` glob matched them), or the DAG rejects the configuration with a message that names the real problem.

An empty `involving: []` cycle is a bug in its own right — a cycle error should never be reachable with no nodes to report.

## Notes

Latent in-repo: no example or POC currently declares two transformation pipelines, which is why CI is green. Fixing it properly means giving `build_unified_dag` per-pipeline model sets rather than one flat list, which also removes the ambiguity behind #1258.

hugocorreia90:
Broader than the title suggests: the trigger is **any duplicate model name reaching the DAG**, not specifically two transformation pipelines.

Reproduced on `4f37c1dd` with a single transformation pipeline and one models directory:

```
models/orders.sql        + orders.toml   (name = "orders")
models/staging/orders.sql + orders.toml  (name = "orders")
```

```
executing unified DAG  nodes=2 edges=0
Error: DAG execution failed
Caused by:
    0: DAG validation failed: circular dependency detected involving: []
[...]

PR #1264, opened 2026-07-27
fix(engine/rocky-cli): resolve run --dag models from the pipeline config

### LINKED ISSUE #1258: run --dag ignores the configured transformation models glob and --models
## Description

`run_with_dag` hardcodes `<config_dir>/models`. A transformation pipeline's model set is configurable through the `models` field (default `models/**`), and plain `rocky run` resolves that glob — the unified-DAG path does not. `--models` is also accepted alongside `--dag` and silently dropped.

## Steps to Reproduce

```toml
[pipeline.silver]
type = "transformation"
models = "transforms/**"
```

with models under `transforms/` and no `./models` directory.

```bash
rocky run --dag --output json
```

## Actual Behavior

`models_dir.is_dir()` is false, so the model list is empty, no transformation nodes are created, and the command exits 0 with `total_nodes: 0`.

This is the same silent-success shape as #1243, reached by a different route: the fix in #1244 never applies, because there is no error to propagate — the directory simply isn't there.

Separately, `rocky run --dag --models transforms` parses cleanly (the `models` arg has no `conflicts_with = "dag"`, unlike `--defer` and `--watch`) and the flag is then ignored.

## Expected Behavior

The DAG is built from each transformation pipeline's configured model set, so `run --dag` and `run` agree on which models exist.

## Suggested Fix

Resolve each transformation pipeline's configured glob in `run_with_dag` the way `run` does, and either honor `--models` under `--dag` or reject the combination at parse time.

### LINKED ISSUE #1259: run --dag swallows seed-discovery errors and can run models against stale seed data
## Description

`run_with_dag` discovers seeds with `discover_seeds(...).unwrap_or_default()` in `engine/crates/rocky-cli/src/commands/run_dag_exec.rs`. Seed discovery propagates malformed-sidecar and directory errors, but the DAG path converts them into an empty seed list — the same shape as the model-load bug fixed in #1244, one seam over.

## Impact

Seed nodes and seed to model edges are built only from that list. With a malformed seed sidecar:

- no seed node is created, so the seed is never loaded;
- `infer_runtime_dependencies` has no seed to order the model after;
- a model that reads the seed executes against whatever the previous run left in the table.

The DAG then reports success. This is the #1243 failure mode — a green run over stale data — with seeds instead of models.

## Steps to Reproduce

Add `seeds/countries.csv` with a malformed `seeds/countries.toml`, and a model that reads the seed. Run `rocky run --dag`. The seed is absent from the node list, the model runs anyway, and the command exits 0.

## Suggested Fix

Propagate the discovery error the way the model load now does. Worth doing in the same pass as #1257 so the DAG path has one consistent tolerance for discovery failures.

hugocorreia90:
## Summary

`run_with_dag` hardcoded `<config_dir>/models`. That was wrong in both directions:

- A transformation pipeline declaring `models = "transforms/**"` loaded nothing, built zero transformation nodes, and exited 0. This is #1243's silent success by a route the error propagation in #1244 could never reach — there was no error, just an absent directory.
- A project with **no** transformation pipeline was still forced to validate a `models/` directory that only transformation pipelines consume (`add_transformation_nodes`). An unrelated broken model there failed a replication-only `--dag` run that plain `rocky run` executes happily, because `run` only enters the compiled-model phase under `run_all || models_dir.is_some()`.

Closes #1258
Closes #1259

## Subproject(s) touched

## Changes

- Load the union of every transformation pipeline's declared model set, deriving each base directory from its `models` glob exactly the way `run` derives it (the prefix up to the first `**`). Directories and model names are deduplicated, since pipelines commonly share the default `models/**`.
- A project with no transformation pipeline loads nothing, so it can never fail on a directory it would not have executed. This removes the `run` vs `run --dag` asymmetry #1244 introduced.
- Propagate seed-discovery errors instead of `unwrap_or_default()`. Seed nodes and the seed→model edges that order a model after the seed it reads are built only from that list, so swallowing a malformed sidecar dropped the seed node, left the model unordered, and let it run against whatever the previous run left in the table — a green DAG over stale data.
- Point each transformation sub-run at its **own** pipeline's directory. `models_dir_for_model_scope` also hardcoded `models`, so fixing discovery alone was not enough: a pipeline declaring `models = "transforms/**"` built a correct node whose sub-run then died with `models directory '<config>/models' not found (required for --model)`. Discovery and execution have to agree.
- Error on a duplicate model name, naming both files. A transformation node is keyed by model name alone, so two models sharing a name cannot both be built — today that surfaces from the executor as `circular dependency detected involving: []` (#1261), which names nothing. Reproduced on `4f37c1dd` with `models/orders.sql` plus `models/staging/orders.sql`.
- Deduplicate model directories by canonical path, so two pipelines writing the same directory differently (`models/**` and `./models/**`) load it once rather than colliding with themselves.
- Reject `--models` with `--dag` rather than silently ignoring it, mirroring the existing `--var` guard. The DAG resolves each pipeline's own glob, so an override passed here would build the DAG from a different set of files than the caller asked for.

## Deliberately not changed

`build_unified_dag` takes one flat model list and hands every transformation pipeline the same nodes, so per-pipeline attribution is not attempted here. That indirection is its own bug — two transformation pipelines currently make both `dag` and `run --dag` fail with `circular dependency detected involving: []` (#1261) — and fixing it means changing a rocky-core signature that `dag.rs` also calls.

A missing model directory stays a no-op rather than an error, matching `run`'s existing tolerance.

## Test Plan

New `engine/rocky/tests/run_dag_model_discovery.rs`, five cases:

- [x] `custom_models_glob_is_honored` — `models = "transforms/**"` produces the `stg_orders` node **and executes it** (`completed: 1, failed: 0`). The execution assertion is load-bearing: an earlier revision of this PR passed a node-count-only version of this test while every custom-glob node failed at run time, which is how the `models_dir_for_model_scope` half of the bug was found.
- [x] `duplicate_model_names_are_reported_by_file` — the duplicate is named, and the phantom cycle does not appear.
- [x] `replication_only_run_ignores_an_unrelated_models_dir` — a broken sidecar under `models/` no longer fails a replication-only run.
- [x] `transformation_run_still_fails_on_a_broken_model` — #1244's guarantee survives the scoping change.
- [x] `malformed_seed_sidecar_fails_the_run` — seed discovery errors reach the caller.
- [x] `models_flag_with_dag_is_rejected` — the guard message fires.
- [x] Mutation check: with the two source files reverted and the new tests kept, **all five fail**; with the fix, all five pass. Every assertion binds.
- [x] `cargo test -p rocky --test run_dag_json_stdout` — #1244's own tests still pass.
- [x] `cargo fmt -- --check`

## Review

Red-teamed by OpenAI Codex (`gpt-5.6-sol`, `xhigh` reasoning) together with #1263 and the loader change; findings and their disposition are in a comment below.

**What I am least confident about:** the union-of-directories shape. It is correct when pipelines share a model set, which is every in-repo project, but for genuinely disjoint globs it still hands every pipeline every model — the same flat-list indirection behind #1261. Nothing in this repository exercises that case, so the tests cannot distinguish "correct" from "correct enough until #1261 lands." What would settle it is fixing #1261 first and giving each pipeline its own set; I took the smaller change because #1261 needs a rocky-core signature change.

**The most important thing I am missing:** a missing model directory is still silent. `models = "transfrms/**"` — a typo — resolves to a directory that does not exist, loads nothing, and reports success with `total_nodes: 0`. That is the exact shape of #1243, and this PR does not close it; it only fixes *which* directory gets consulted. Making it fatal would diverge from `run`, which treats an absent models directory as a no-op, so the two need to change together rather than `--dag` alone drifting stricter again.

## Checklist

### DIFF
--- a/engine/crates/rocky-cli/src/commands/dag.rs
+++ b/engine/crates/rocky-cli/src/commands/dag.rs
@@ -98,9 +98,17 @@ pub fn dag_output(
     // Load models from the models directory (including subdirectories).
     let models = load_all_models(models_dir)?;
 
-    // Load seeds if the directory exists.
+    // Load seeds if the directory exists. A discovery failure is propagated, not
+    // flattened to "no seeds": the seed nodes and the seed→model edges are built
+    // only from this list, so swallowing a malformed sidecar prints a DAG that
+    // is missing both — an answer that looks complete and is not.
+    let discover = |dir: &Path| -> Result<Vec<rocky_core::seeds::SeedFile>> {
+        rocky_core::seeds::discover_seeds(dir)
+            .map_err(|e| anyhow::anyhow!("{e}"))
+            .with_context(|| format!("failed to discover seeds in {}", dir.display()))
+    };
     let seeds = match seeds_dir {
-        Some(dir) if dir.exists() => rocky_core::seeds::discover_seeds(dir).unwrap_or_default(),
+        Some(dir) if dir.exists() => discover(dir)?,
         _ => {
             // Try default "seeds" relative to config dir.
             let default_seeds = config_path
@@ -108,7 +116,7 @@ pub fn dag_output(
                 .unwrap_or_else(|| Path::new("."))
                 .join("seeds");
             if default_seeds.is_dir() {
-                rocky_core::seeds::discover_seeds(&default_seeds).unwrap_or_default()
+                discover(&default_seeds)?
             } else {
                 vec![]
             }
--- a/engine/crates/rocky-cli/src/commands/run_dag_exec.rs
+++ b/engine/crates/rocky-cli/src/commands/run_dag_exec.rs
@@ -72,7 +72,13 @@ fn default_sub_runner() -> SubRunner {
          model_name: Option<String>,
          skip_opts| {
             Box::pin(async move {
-                let models_dir = models_dir_for_model_scope(&config_path, model_name.as_deref());
+                let models_dir = models_dir_for_model_scope(
+                    &config_path,
+                    &loaded.config,
+                    &pipeline_name,
+                    model_name.as_deref(),
+                )
+                .map_err(|e| format!("{e:#}"))?;
                 let partition_opts = super::PartitionRunOptions {
                     partition: None,
                     from: None,
@@ -123,16 +129,32 @@ fn default_sub_runner() -> SubRunner {
     )
 }
 
-/// Returns the conventional models directory only for a model-scoped
+/// Returns the owning pipeline's models directory, only for a model-scoped
 /// transformation sub-run. Supplying it to replication/load sub-runs makes
 /// `run()` execute every transformation model after the pipeline itself.
-fn models_dir_for_model_scope(config_path: &Path, model_name: Option<&str>) -> Option<PathBuf> {
-    model_name.map(|_| {
-        config_path
-            .parent()
-            .unwrap_or(Path::new("."))
-            .join("models")
-    })
+///
+/// This must resolve the pipeline's CONFIGURED directory, not a hardcoded
+/// `models`. The DAG builds a node for every model it discovered, then hands
+/// each node's model name to a sub-run along with the directory to find it in —
+/// so a pipeline declaring `models = "transforms/**"` built a correct node whose
+/// sub-run then died with `models directory '<config>/models' not found
+/// (required for --model)`. Discovery and execution have to agree, which is why
+/// both go through [`crate::models_loader::resolve_models_dir`].
+fn models_dir_for_model_scope(
+    config_path: &Path,
+    cfg: &rocky_core::config::RockyConfig,
+    pipeline_name: &str,
+    model_name: Option<&str>,
+) -> Result<Option<PathBuf>> {
+    use rocky_core::config::PipelineConfig;
+
+    if model_name.is_none() {
+        return Ok(None);
+    }
+    let Some(PipelineConfig::Transformation(t)) = cfg.pipelines.get(pipeline_name) else {
+        return Ok(None);
+    };
+    crate::models_loader::resolve_models_dir(&t.models, config_path)
 }
 
 /// Execute `rocky run --dag`: run every pipeline in dependency order.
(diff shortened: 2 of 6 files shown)


hugocorreia90:
## Red team: OpenAI Codex (`gpt-5.6-sol`, `xhigh`) — verdict on this branch was **BLOCK**. Addressed.

Reviewed `66e1600e` against `4f37c1dd`, alongside #1263 and #1266.

### BLOCKER — silent success for valid general glob shapes — **accepted, fixed**

`models` is documented as a *glob* (`config.rs` `default_models_glob`), but my `models_base_dir` split on `**` alone. So `models = "models/*.sql"` stayed intact, got probed as a literal directory that never exists, and produced an empty DAG exiting 0 — with a conventional `models/orders.sql` sitting right there that `main` would have built. A regression, and my unit test covered only `/**` suffixes and no-wildcard input, so it missed exactly this.

Codex was right that the derivation was the problem, and the codebase already had the answer: `tick` carried a hardened `resolve_models_dir` that splits on `['*', '?', '[']` **and** confines the result to the project root. I lifted it to `models_loader` and pointed both DAG discovery and the sub-run resolver at it, so there is now one derivation instead of a fourth hand-rolled one. That also closes the containment gap Codex noted in passing (absolute paths and `..` escapes were accepted with no project-root check).

New tests: `resolve_models_dir_handles_every_glob_shape` covers `models/**`, `models/*.sql`, `models/staging/**`, `models`, `transforms/**` and an absent directory; `resolve_models_dir_refuses_an_escape` pins the containment; and `a_wildcard_glob_still_resolves_its_directory` reproduces Codex's exact scenario end-to-end through the binary.

Two of its probes stay open by choice and are named in the body: `**/*.sql` derives an empty base and resolves to the config directory (unchanged from today), and `run.rs` / `validate.rs` still use the naive `split("**")`. I did not change `run`'s derivation here — that would flip plain `rocky run` from silently no-opping to materializing tables for those configs, which deserves its own PR and its own risk note. Filed as #1267.

### SHOULD-FIX — seed closure partial — **accepted, partly fixed**

Codex found `rocky dag` still using `discover_seeds(...).unwrap_or_default()` (`dag.rs:101-115`) — the same bug one command over, which #1259 did not name. Fixed in this PR: `rocky dag` now propagates too, so it can no longer print a DAG missing its seed nodes and edges while looking complete.

The `seeds_dir.is_dir()` guard still converts a regular file or dangling symlink named `seeds` into "no seeds." That is pre-existing and now stated in the body rather than left implied.

### Confirmed rather than assumed

Useful negative results I have folded into the body instead of leaving as my own assertions:

- **Duplicate model names**: Codex traced `unified_dag.rs:556-570` and `728-813` and confirmed duplicates already fail today via ID coalescing producing `CyclicDependency { nodes: [] }`. So the guard I added replaces an opaque failure with a useful one rather than introducing a new hard failure — which is what I claimed but had only verified behaviourally.
- **Symlinks**: `Path::is_dir()` follows symlinks, so canonicalize-then-dedupe correctly collapses two aliases of one real directory.
- **Test binding**: the amended custom-glob test cannot pass merely because some arbitrary directory holds one model, and the replication-only test cannot pass through a replication no-op.
- **stdout contract**: every new error path fires after `reserve_stdout_for_json()` and leaves stdout uncontaminated; the `--models`/`--dag` guard fires before config dispatch, so the no-project test assumption holds.

PR #1266, opened 2026-07-27
fix(engine/rocky-cli): propagate subdirectory model-load failures

### LINKED ISSUE #1257: run --dag silently drops models that fail to load in a subdirectory
## Description

#1244 made `run --dag` propagate a **top-level** model-load error. But `load_project_models` still walks immediate subdirectories through `if let Ok(sub)` and discards every child error, so a malformed model one directory down is silently omitted and the run reports success.

## Steps to Reproduce

Take any working project and add a broken model in a subdirectory of `models/`:

```
models/
  ok.sql
  ok.toml
  staging/
    broken.sql
    broken.toml   # contents: name = [
```

```bash
rocky run --dag --output json
```

## Actual Behavior

Exit 0, `failed: 0`. The broken model never appears in the node list and is never mentioned on stderr. Reproduced against the default playground POC: `total_nodes: 3, completed: 3, failed: 0`, with the subdirectory model silently absent.

If **every** model lives in the failing subdirectory, this reproduces #1243's original zero-node success exactly — the class #1244 closed is only closed at the top level.

## Expected Behavior

The command exits non-zero and names the malformed model, matching the top-level behavior.

## Suggested Fix

Propagate subdirectory load errors from `load_project_models` in `engine/crates/rocky-cli/src/models_loader.rs` (or at minimum surface them as warnings). Note the loader is shared by `dag`, `list`, `docs`, `preview`, `compliance`, `estimate`, `test`, `tick`, `validate`, `branch` and `scope`, so changing the tolerance affects all of them.

hugocorreia90:
## Summary

`load_project_models` loads the top-level models directory plus one level of immediate subdirectories, but took each subdirectory through `if let Ok(sub)` and discarded the error. A malformed model one level down simply vanished from the model list, and every command built on this loader carried on as though it did not exist.

Through `run --dag` that meant reporting success having built nothing for that subtree — the same silent-success class the top-level load stopped admitting in #1244. Reproduced before this change: a healthy project plus `models/staging/broken.toml` exits 0 with `failed: 0` and never mentions the broken model.

Closes #1257

## Subproject(s) touched

## Blast radius

This loader is shared by `dag`, `list`, `docs`, `preview`, `compliance`, `estimate`, `test`, `tick`, `validate`, `branch`, `scope` and the maintenance-apply policy gate. **Not all of them should get stricter**, which is the main correction this PR received in review.

Strict (a broken model in a subdirectory now fails the command instead of disappearing): `dag`, `list`, `docs`, `compliance`, `estimate`, `test`, `validate`, `branch`, `scope`. That is the same tolerance change #1244 made at the top level.

Deliberately tolerant, via the new `load_project_models_partial`:

- **`resolve_touched_apply_targets`** — the sharp one. It maps physical target FQNs to logical model names so an attribute-scoped policy rule still fires on a maintenance apply, and it feeds compact's and archive's pre-SQL gates. Emptying that map is a **fail-open**: an unresolved target is evaluated against default attributes, so a `layer = "gold"` rule stops matching and a destructive apply can fall through to a permissive default.
- **`optimize`** — advisory downstream-reference counts must not all collapse to zero.
- **`preview`** — per-model budget projection.
- **`tick`** — member freshness budgets, which would otherwise widen to the project default.

Verified against every project in the repository: **0 of 78** fail, now that #1263 has fixed the one pre-existing breakage.

Verified against every project in the repository: **0 of 78** newly fail. The only failure in the sweep is `engine/examples/docs-demo`, which is broken today for an unrelated reason and is fixed by #1263.

## Changes

- Propagate subdirectory load failures, naming the failing directory.
- Visit subdirectories in sorted order, so a project with several broken ones always reports the same one first.
- Keep an absent models directory a non-error. `load_dir_models` returns no models for a directory that does not exist and several callers rely on that (`rocky docs` on a project with no models, among others), so the descent is skipped rather than failing `read_dir`.

## What this does NOT close

The loader reads the top level plus **one** level down. A model at `models/a/b/` is still not loaded at all — not an error, simply absent (#1262). Verified: `models/a/b/lvl2.sql` never appears in the model list, and corrupting its sidecar changes nothing, because the file is never read.

So this PR narrows the silent-drop surface rather than eliminating it. Making the walk recursive would newly discover `.sql` / `.rocky` files that projects may keep under `models/` deliberately unbuilt — archives, scratch, vendored subtrees — and start materializing tables from them. That is a behavior change deserving its own decision, which is why #1262 is filed separately rather than folded in here.

## Test Plan

Four unit tests in `models_loader.rs`:

- [x] `a_broken_model_in_a_subdirectory_is_an_error` — the regression this exists for; asserts the error names the subdirectory.
- [x] `loads_top_level_and_one_subdirectory_level` — the documented depth still works.
- [x] `a_missing_directory_yields_no_models` — pins the tolerance callers depend on.
- [x] `unrelated_files_do_not_fail_the_load` — a README beside the models is not a model.
- [x] Mutation check: restoring `if let Ok(sub)` fails `a_broken_model_in_a_subdirectory_is_an_error` and only that one, so the assertion binds and the other three are not coupled to it.
- [x] Repository sweep with the built binary: 0 of 78 newly fail.
- [x] `cargo fmt -- --check`

## Review

Red-teamed by OpenAI Codex (`gpt-5.6-sol`, `xhigh` reasoning) together with #1263 and #1264; findings and their disposition are in a comment below.

**What I am least confident about:** whether eleven commands getting stricter is the right default versus warning on subdirectory failures and continuing. The sweep proves no in-repo project breaks, but it cannot speak for a user mid-edit whose `models/staging/` does not currently parse — `rocky list` and `rocky docs` now fail for them where they used to return the models that *do* parse. Fail-closed is right for `run --dag`, where the consequence is unbuilt tables reported as success; it is more arguable for read-only commands. What would settle it is deciding whether this loader should have one tolerance or two.

**The most important thing I am missing:** #1262 means the fail-closed guarantee this PR appears to establish is not actually established. A reader of this diff could reasonably conclude "subdirectory models can no longer silently vanish," and that is only true for the first level. The honest framing is that the loader now fails loudly for the depth it reads and stays silent about the depth it does not — and the second half is the part that still bites.

## Checklist

### DIFF
--- a/engine/crates/rocky-cli/src/commands/apply.rs
+++ b/engine/crates/rocky-cli/src/commands/apply.rs
@@ -2375,16 +2375,29 @@ pub(crate) fn resolve_touched_apply_targets(
     // failure just leaves the maps empty and every target falls through as-is.
     let mut fqn_to_name: BTreeMap<String, String> = BTreeMap::new();
     let mut names: BTreeSet<String> = BTreeSet::new();
-    if let Ok(models) = crate::models_loader::load_project_models(&models_dir) {
-        for m in &models {
-            names.insert(m.config.name.clone());
-            let fqn = format!(
-                "{}.{}.{}",
-                m.config.target.catalog, m.config.target.schema, m.config.target.table
-            )
-            .to_lowercase();
-            fqn_to_name.insert(fqn, m.config.name.clone());
-        }
+    // Partial, NOT `load_project_models`: this map is a policy input, so losing
+    // it wholesale because one unrelated subdirectory is malformed is a
+    // fail-open, not a degraded answer. An unresolved FQN is evaluated against
+    // DEFAULT attributes, so a rule scoped to a positive attribute
+    // (`layer = "gold"`) stops matching and the apply can fall through to a
+    // permissive default. Keep every mapping that loaded and warn about the rest
+    // — those specific targets still fall back to defaults, which is exactly the
+    // pre-existing limitation documented above.
+    let (models, load_errors) = crate::models_loader::load_project_models_partial(&models_dir);
+    for e in &load_errors {
+        tracing::warn!(
+            error = %format!("{e:#}"),
+            "some models could not be loaded; policy attributes for their targets fall back to defaults"
+        );
+    }
+    for m in &models {
+        names.insert(m.config.name.clone());
+        let fqn = format!(
+            "{}.{}.{}",
+            m.config.target.catalog, m.config.target.schema, m.config.target.table
+        )
+        .to_lowercase();
+        fqn_to_name.insert(fqn, m.config.name.clone());
     }
 
     let mut touched = BTreeMap::new();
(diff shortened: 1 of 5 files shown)


hugocorreia90:
## Red team: OpenAI Codex (`gpt-5.6-sol`, `xhigh`) — verdict on this branch was **BLOCK**. Addressed.

Reviewed `e07bca96` against `4f37c1dd`, alongside #1263 and #1264.

### BLOCKER — introduced policy fail-open — **accepted, fixed, and now pinned by a test**

This was a real one and I had missed it. `models_loader.rs` going strict meant `resolve_touched_apply_targets` (`apply.rs`) — which takes the loader result through `if let Ok(models)` — dropped **every** FQN-to-model mapping when one unrelated subdirectory was malformed. That map feeds compact's and archive's pre-SQL gates via `gate_maintenance_apply`. The function's own doc comment spells out the consequence: an unresolved target is evaluated against **default** attributes, so a rule requiring a positive attribute (`layer = "gold"`, `classifications = ["pii"]`) stops matching, and it is explicitly *not* guaranteed to fall back to `require_review`.

Net effect: a valid `payments` model tagged `layer = "gold"`, plus an unrelated broken draft in `models/staging/`, and a destructive maintenance apply that was denied before this change could execute after it.

Codex's own "most important thing I'm missing" was runtime confirmation of this chain. I ran it. With the strict loader the touched-target map comes back as `{"wh.gold.payments": Apply}` — the raw physical FQN, unresolved — instead of `{"payments": Apply}`. The gate then never sees the gold attribute. Confirmed, not merely plausible.

Fix: `load_project_models_partial` returns what loaded plus the errors. `resolve_touched_apply_targets` uses it and warns about the directories that failed, so healthy mappings survive and only the genuinely unloadable targets fall back to defaults — which is the pre-existing documented limitation, not a new one. `load_project_models` stays strict for callers that should fail closed.

`touched_targets_resolve_despite_a_broken_sibling_model` pins it, and it is mutation-checked: swapping the partial loader back for the strict one fails it with the raw FQN in the message.

### SHOULD-FIX — tolerant read-only callers erased healthy information — **accepted, fixed, and the body corrected**

Codex was right that my PR body's claim that all eleven consumers hard-fail was inaccurate. `optimize.rs` swallowed the error into an empty DAG and would have reported every `downstream_references` as 0; `tick.rs` caught it and fell back to the project freshness default; `preview.rs` degraded to empty budgets. All three now use the partial loader, so a broken draft costs only that model rather than the whole answer. Body corrected.

### Confirmed rather than assumed

- **Symlinks**: `Path::is_dir()` follows symlinks, so the early return for an absent directory does not skip a symlink-to-directory.
- **Interaction**: after this lands, #1264's `load_all_models` call propagates subdirectory errors where it previously swallowed them. There is no composed `run --dag` subdirectory-error test — only #1264's top-level integration test and this PR's loader unit tests. Named as a gap rather than papered over.

### Still not closed, by design

The loader reads one level down, so a model at `models/a/b/` is not loaded at all — not an error, simply absent (#1262). Verified: corrupting a depth-2 sidecar changes nothing, because the file is never read. This PR narrows the silent-drop surface; it does not eliminate it.

### Re-verified after the rework

- `cargo test -p rocky-cli --lib`: 1281 passed, 0 failed.
- Repository sweep with the rebuilt binary, now that #1263 has landed: **0 of 78** projects fail under the stricter loader.

Issue #1269, opened 2026-07-27
bug(engine): transformation --shadow/--branch writes to production targets


mattfaltyn:
## Description

Transformation runs invoked with `--shadow` or `--branch` execute against the configured production `ModelIr.target` instead of the resolved shadow target. A successful full-refresh shadow run can therefore replace production data while exiting 0.

## Steps to reproduce

1. Configure a DuckDB transformation model targeting `shadowbug.main.orders`.
2. Set the model SQL to `SELECT 1 AS version` and run:

   ```console
   rocky --config rocky.toml --state-path state.redb run --pipeline silver
   ```

3. Change the model SQL to `SELECT 2 AS version`.
4. Run:

   ```console
   rocky --config rocky.toml --state-path state.redb run --pipeline silver --shadow
   ```

5. Query production and list matching tables.

This reproduces deterministically on current `main` and was repeated twice.

## Expected behavior

- `shadowbug.main.orders` remains at `version = 1`.
- `shadowbug.main.orders_rocky_shadow` is created with `version = 2`.
- Selected in-run upstream reads resolve to the upstream model's shadow target.
- A derived shadow target that overlaps any configured production target or another selected model's shadow target is rejected before execution.

## Actual behavior

- Rocky exits 0 with `status = "Success"`.
- Materialization metadata reports `shadowbug.main.orders`.
- Production is replaced with `version = 2`.
- No `orders_rocky_shadow` table exists.

## Root cause

The CLI resolves a `ShadowConfig`, but transformation and model-only dispatch do not thread it into `run_transformation` / `execute_models`. The existing `rocky_core::shadow::shadow_target` helper is therefore never applied to transformation `ModelIr.target` values.

## Rocky version

`rocky 1.67.0`; confirmed against `main@d61608f634663b846942b32f2067282c46b9eb0d` and revalidated after rebasing the fix onto `main@32f5e71d`.

## OS / architecture

macOS 26.5.2 arm64.

## Suggested regression coverage

Add a DuckDB process-level test proving that transformation `--shadow` preserves production and creates the suffixed target. Cover the model-only and branch schema-override entry points at the shared routing boundary. Also cover empty suffixes, schema overrides that route back to production, production/shadow name collisions, and two selected models collapsing to the same shadow target.

hugocorreia90:
Fixed on `main` by #1270 and #1285. Closing — re-verified live rather than by reading the diff.

[...]

PR #1270, opened 2026-07-27
fix(engine/rocky-cli): isolate transformation shadow runs

mattfaltyn:
## Summary

Prevent transformation `--shadow` and `--branch` runs from writing production targets or reading production versions of selected in-run upstream models.

Refs #1269 — this closes the direct transformation and model-only paths. #1269 stays open
until #1272 lands: `rocky run --dag`, and the snapshot and load pipeline kinds, still accept
`--shadow` / `--branch` and write production targets.

Follow-ups opened from review: #1272 (remaining shadow-writes-production entrypoints),
#1273 (shadow object ownership + lifecycle), #1274 (`rocky compare` for transformation
pipelines), #1275 (physical reads produce no DAG edge, so producer and consumer are
co-scheduled).

## Subproject(s) touched

## Changes

- Thread the resolved `ShadowConfig` through transformation and model-only execution.
- Rewrite selected model targets and declared in-run dependency reads to their physical shadow targets before schema creation, execution, state/reuse identity, metadata, and governance capture.
- Preserve production reads for unselected/deferred upstreams.
- Reject derived targets that overlap any configured production target or another selected model's shadow target.
- Fail closed for `content_addressed` and `time_interval` shadow runs until their external storage and partition-state identities can be isolated safely.
- Add DuckDB regression coverage for model-only, two-model suffix-shadow, branch-schema routing, unchanged production tables, unsupported strategies, and unsafe target collisions.
- Document shadow routing, collision rejection, and the current fail-closed strategy limitation.

## Test Plan

- [x] `cargo test -p rocky-cli --features duckdb shadow_` (5 passed)
- [x] `cargo nextest run --all-features` (5,741 passed; 108 skipped)
- [x] `cargo fmt -- --check`
- [x] `npm ci && npm run build` in `docs/` (98 pages; 0 dependency vulnerabilities)
- [ ] `cargo test --all-targets` (all test targets passed in the earlier full run, then the Criterion `binary_startup` benchmark self-deadlocked because it launches a nested `cargo run` against the outer Cargo target lock; the repository's canonical all-feature nextest command is green above)
- [ ] `bash scripts/lint-adapter-boundary.sh` from the repository root (fails on existing imports in unrelated `output.rs`, `compact.rs`, `discover.rs`, `export_schemas.rs`, `restore.rs`, `archive.rs`, `replay.rs`, and `run_content_addressed.rs`; this PR adds no violations, and the workflow's adapter-boundary check is green)

## Checklist

- [x] If this is a CLI JSON schema or DSL syntax change, all consumers and golden fixtures are updated in the same PR (not applicable: no schema or DSL syntax change).

### Engine (`engine/`)

- [x] The canonical all-feature nextest suite passes

### DIFF
--- a/docs/src/content/docs/concepts/shadow-mode.md
+++ b/docs/src/content/docs/concepts/shadow-mode.md
@@ -14,11 +14,25 @@ Shadow mode writes pipeline output to shadow tables instead of (or alongside) pr
 3. A comparison engine checks row counts, schemas, and optionally sample data between shadow and production
 4. Results show pass/warn/fail with detailed diffs
 
-Shadow and branch runs currently reject `content_addressed` and `time_interval`
-models. Those strategies also persist object-storage or partition-state
+:::caution[Shadow isolation covers `rocky run` for transformation pipelines only]
+`rocky run --dag`, and the snapshot and load pipeline kinds, still accept
+`--shadow` and `--branch` but write **production** targets. Do not rely on those
+entrypoints for isolation.
+:::
+
+Shadow and branch runs currently reject `content_addressed`, `time_interval` and
+`ephemeral` models. The first two persist object-storage or partition-state
 identities that cannot yet be isolated by rewriting only the warehouse target.
-Rocky also rejects a derived shadow target that matches any configured
-production target or another selected shadow target.
+Ephemeral models are neither materialized nor inlined into their consumers, so a
+consumer would read the production table and no rewrite could redirect it — give
+the model a materialized strategy to shadow it. Rocky also rejects a derived
+shadow target that matches any configured production target or another selected
+shadow target.
+
+Rocky does **not** yet verify that a derived shadow target is unoccupied by an
+object it does not know about. If a table matching the derived name already
+exists and is not a Rocky model target — a source, a seed, or an ad-hoc table —
+a full-refresh model will replace it. Prefer a dedicated shadow schema you own.
 
 A model that reads another selected model's table is routed to that upstream's
 shadow target whether or not it declares the dependency in `depends_on` —
--- a/docs/src/content/docs/concepts/shadow-mode.md
+++ b/docs/src/content/docs/concepts/shadow-mode.md
@@ -20,6 +20,14 @@ identities that cannot yet be isolated by rewriting only the warehouse target.
 Rocky also rejects a derived shadow target that matches any configured
 production target or another selected shadow target.
 
+A model that reads another selected model's table is routed to that upstream's
+shadow target whether or not it declares the dependency in `depends_on` —
+matching is on the upstream's configured `catalog.schema.table`, so a physical
+read is redirected too. When a reference could resolve to more than one
+selected upstream (two models whose targets share a name in different
+catalogs, read as a bare or partially qualified name), Rocky refuses the run
+rather than guess which one to read.
+
 ## Shadow target rewriting
 
 ### Suffix mode (default)
--- a/docs/src/content/docs/concepts/shadow-mode.md
+++ b/docs/src/content/docs/concepts/shadow-mode.md
@@ -14,6 +14,12 @@ Shadow mode writes pipeline output to shadow tables instead of (or alongside) pr
 3. A comparison engine checks row counts, schemas, and optionally sample data between shadow and production
 4. Results show pass/warn/fail with detailed diffs
 
+Shadow and branch runs currently reject `content_addressed` and `time_interval`
+models. Those strategies also persist object-storage or partition-state
+identities that cannot yet be isolated by rewriting only the warehouse target.
+Rocky also rejects a derived shadow target that matches any configured
+production target or another selected shadow target.
+
 ## Shadow target rewriting
 
 ### Suffix mode (default)
--- a/docs/src/content/docs/reference/commands/core-pipeline.md
+++ b/docs/src/content/docs/reference/commands/core-pipeline.md
@@ -424,8 +424,15 @@ contains a `content_addressed`, `time_interval` or `ephemeral` model — the fir
 two require additional storage or partition-state isolation, and an ephemeral
 model is neither materialized nor inlined, so its consumer would read
 production. They also fail closed when the chosen suffix or schema would collide
-with a production target or another selected shadow target, and when two targets
-differ only by case on a dialect that quotes identifiers.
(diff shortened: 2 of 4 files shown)


hugocorreia90 (commented):
Reviewed with an independent model (Codex/GPT-5) as red team, plus local verification. **CI is fully green on the rebased head and I updated the branch onto current `main` first** (its previous green ran against a four-commit-stale base). Not approving yet — two confirmed defects below, one of which is a new data-loss path.

## What's good, verified rather than assumed

- **The tests bind on the real defect.** I removed only the `apply_shadow_rewrite` call from `execute_models` and re-ran: the two `run_local` end-to-end tests fail with `suffix shadow must not overwrite the production target` (left: 2, right: 1) and the log shows `.main.orders` being written. The six `run.rs` tests still pass because they call the function directly — so the suite cleanly separates "the function works" from "the function is wired", and covers both. 8/8 pass on the head.
- **The `contain_failures` carve-out is correct.** Confirmed `run.rs:7193` selects `plan.layers` in that mode and `augmented_layers` builds nodes from the ledger's own upstream set including physical reads (`containment.rs:443`, test at `:910`). Stale `project.layers` is genuinely unused.
- Single production call site; governance captured after the rewrite; derived DAG edges for rewritten physical reads is a good catch.

## Confirmed blockers

**1. Snowflake rewritten reads cannot resolve the producer's table.** `run.rs:5739` sets `quote_style` only for BigQuery, so `DeferTarget` serialises Snowflake replacements unquoted. But `rocky-snowflake/src/dialect.rs:26-43` deliberately double-quotes every component, with its own comment explaining that unquoted "silently breaks against any schema / table created with quoted lowercase names". So the producer writes `"db"."schema"."orders_rocky_shadow"` and the consumer reads `db.schema.orders_rocky_shadow`, which Snowflake folds to uppercase. Multi-model shadow runs break on Snowflake. (The same pattern exists in `apply_defer_rewrite` at `run.rs:5621`, so defer likely has it too — worth a separate issue.)

Related: `target_identity` lowercases (`run.rs:5735-5737`), so the distinct Snowflake targets `"Orders"` and `"orders"` collapse to one key and an unrelated read can be redirected.

**2. The ephemeral exclusion rests on inlining that does not exist.** The comment at `sql_gen.rs:443` says "the compiler inlines them as CTEs in downstream queries", but I can find no CTE-inlining implementation in `rocky-compiler/src` or `rocky-sql/src` — the only ephemeral handling in the compiler is the dbt importer, which states `materialized='ephemeral'` has no Rocky equivalent. So a consumer's `FROM eph` reads a physical production table, and `shadow_routes_ephemeral_target_but_never_redirects_readers_to_it` asserts exactly that as correct. Suggest failing closed on ephemeral dependencies until inlining is real.

**3. A non-model table can be destroyed** (Codex, confirmed). Collision detection indexes only configured model targets, so a pre-existing source/seed/audit table named `<model>_rocky_shadow` is replaced by `CREATE OR REPLACE TABLE`. Note this path is *new*: before this PR no shadow target was created at all. #1273 tracks the ownership contract; a pre-materialisation "absent or Rocky-owned" check would close it.

## Worth deciding, not necessarily blocking

- **Incremental self-reads** keep the production watermark while writing an empty shadow target, so the shadow table gets only the post-production delta — and a re-run appends it again. Not a whole-table preview. Not reachable via `rocky compare` today (#1274 open), so this may be fine to defer with a doc note.
- **Schema pre-creation ignores the model filter** (`run.rs:7017-7037`), so `rocky run --model a --shadow` can create an unselected model's *production* schema. No data loss, but it breaks isolation.
- **Docs.** `shadow-mode.md` now describes the safety guarantees without mentioning that `--dag`, snapshot and load still write production (#1272). A reader would reasonably conclude `--shadow` is safe everywhere. One sentence naming the unsafe entrypoints would fix it.
- **Coverage** is DuckDB/full-refresh only — no incremental, no Snowflake quoting/case test.

None of this diminishes the core fix, which addresses a genuine data-safety bug and is well tested for the path it covers. Items 1 and 2 look small; item 3 is the one I'd most want closed before this ships.

hugocorreia90:
Thanks @mattfaltyn — the core routing work here is solid, and its tests genuinely bind: removing the `apply_shadow_rewrite` call makes the end-to-end tests fail with production being written, which is exactly the bug being fixed.

I reviewed it, found some dialect and strategy edges, and pushed fixes rather than sending it back — apologies for the volume of commits on your branch, and I've cleaned up the running commentary I left while iterating.

**What I changed on top of your work:**

- **Snowflake quoting.** Rewritten upstream reads were quoted only for BigQuery, but Snowflake stores and resolves a double-quoted identifier exactly as entered while folding an unquoted one to upper case — so a producer wrote `"db"."schema"."orders_rocky_shadow"` and its consumer read a name that resolved elsewhere. Quoting is now derived per dialect and fails closed on an unknown one. `apply_defer_rewrite` had the same defect and shares the helper.
- **Ephemeral models are rejected**, alongside `content_addressed` and `time_interval`. The exclusion rested on CTE inlining that does not exist anywhere in the engine, so a consumer's `FROM eph` was reading production.
- **A corrupted pin.** `engine-wasm-release.yml` pinned `jetli/wasm-pack-action` to a commit sharing a 25-character prefix with the real v0.4.0 and then diverging; `git ls-remote` matches no ref, so the action could never resolve.
- **Case-collision guard** for dialects where identifier case is part of object identity, plus schema pre-creation honouring the model filter, and docs recording the `--dag`/snapshot/load carve-out.

Verified: `cargo nextest run --all-features` 5751 passed, clippy and fmt clean, 15/15 shadow tests, each new guard mutation-checked.

Follow-ups (#1272, #1273, and making `rewrite_upstream_refs` case-aware so the guard can be deleted) will come as a separate PR from me rather than more churn here.

hugocorreia90 (approved):
Approving. The core routing fix is sound and its tests bind on the real defect; the dialect and strategy edges I found are fixed on the branch and every new guard is mutation-checked. All seven required checks are green.

Remaining known gaps are tracked and not regressions from this change: #1272 (`--dag`, snapshot and load still write production under `--shadow`) and #1273 (a derived shadow target is not verified to be unoccupied). Both, plus making `rewrite_upstream_refs` case-aware so the case guard can be removed, will follow in a separate PR.

PR #1265, opened 2026-07-27
chore(deps): bump pinned GitHub Actions trust roots

hugocorreia90:
Supersedes #1249 and #1250, and consolidates the earlier replacement #1267 into one frozen-trust-root bootstrap PR.

This carries the exact Dependabot updates for:
- actions/checkout v7.0.0 -> v7.0.1
- taiki-e/install-action v2.83.4 -> v2.85.2
- astral-sh/setup-uv v8.3.2 -> v9.0.0

It also updates the checker constants that intentionally pin those exact action SHAs.

Validation on exact head:
- python3 -m unittest discover -s .github/tests -p test_*.py -v (100 passed)
- actionlint -shellcheck shellcheck (passed)
- git diff --check (passed)

The enforcing containment job is expected to reject this PR because the checker is a frozen trust root. .github/SECURITY_ENVIRONMENTS.md requires independent security review and a narrow maintainer override for this bootstrap, followed by a fresh containment run from main.

PR #1277, opened 2026-07-28
fix(ci): gate release workflow action sources and repair the wasm-pack pin

hugocorreia90:
Follow-up to #1265 / #1276. **This is a frozen-trust-root change**, so `Credential containment policy` is expected to be red — see the bootstrap procedure in `.github/SECURITY_ENVIRONMENTS.md`.

## The gap

Every substantive action-source rule in `check_credential_containment.py` sits behind a `pr_triggered` guard. Release, publish and deployment workflows are triggered by tag or branch pushes, so they received YAML hygiene checks and nothing else — no action-source validation at all. Those jobs hold `contents: write` and the publishing secrets, which made an unreviewed action source in one of their steps a direct supply-chain path into the binaries `engine/install.sh` serves.

14 jobs across 8 workflows were in that state: `engine-release`, `engine-wasm-release`, `sdk-release`, `dagster-release`, `vscode-release`, `engine-docs`, `engine-evals-live`, `preview-comment`.

Neither did any required status check cover them: the five required contexts are all path-filtered on `engine/**`, so a PR editing only a release workflow triggers none of them.

## The fix

Every job **not reachable from a pull request** may now only run actions in `RELEASE_ACTION_SOURCES`, a pinned allowlist. Parsing fails closed — a step whose `uses:` cannot be read as exactly one pinned remote source is a violation rather than something to skip, and a repository-local action in such a job is rejected outright.

Scoping to *all* non-PR jobs rather than only demonstrably credentialed ones costs nothing: both sets resolve to the same 14 sources, so the broader and simpler invariant is the one enforced.

## A corrupted pin, found while verifying provenance

Every SHA was checked against its upstream tag before being added to the allowlist. One did not resolve:

```
engine-wasm-release.yml pinned:  0d096b08b4e5a7de8c28de67e0cafa345d365d15  # v0.4.0
jetli/wasm-pack-action v0.4.0:   0d096b08b4e5a7de8c28de67e11e945404e9eefa
common prefix:                   0d096b08b4e5a7de8c28de67e  (25 chars, then diverges)
```

`git ls-remote` matches no ref against the pinned value, and the REST API returns `422 No commit found`. The action could never resolve; the pin has been wrong since it was introduced. Corrected, and now asserted from the allowlist so it cannot regress silently.

## Evidence

- `check_credential_containment.py .` → `passed` on the real tree (no false positives)
- `python3 -m unittest discover -s .github/tests` → **104 passed**
- `actionlint` → exit 0
- **Mutation-checked, three cells.** Removing only the two-line enforcement wiring makes 3 of the 4 new tests fail; additionally reverting the wasm pin makes the 4th fail too. The tests bind on what they name rather than passing incidentally.
- `test_every_release_workflow_rejects_an_unreviewed_action` iterates every non-PR workflow rather than hardcoding one, so a sibling release workflow cannot regress silently — which is how this gap arose.

### DIFF
--- a/.github/SECURITY_ENVIRONMENTS.md
+++ b/.github/SECURITY_ENVIRONMENTS.md
@@ -121,6 +121,28 @@ closed only after these pass on the merged revision:
 Record the run URLs in the remediation ledger. Until these pass on `main`, the
 code is merge-ready but RD-026 stays In review.
 
+## Release and deployment action sources
+
+The containment rules above govern jobs reachable from a pull request. Release,
+publish and deployment workflows are a different trust domain: they run from a
+tag or a push, never execute candidate code, and so none of the pull-request
+rules applied to them. They do hold a write token and the publishing secrets,
+which made an unreviewed action source in one of their steps a direct path into
+the artifacts this project ships — `engine/install.sh` serves exactly those
+binaries.
+
+Every job that is not reachable from a pull request may therefore only run
+actions in `RELEASE_ACTION_SOURCES`, the pinned allowlist in the checker. Parsing
+fails closed: a step whose `uses:` cannot be read as exactly one pinned remote
+source is a violation rather than something to skip, and a repository-local
+action in such a job is rejected outright. Adding a new action to a release
+workflow is a trust-root change and follows the bootstrap procedure below.
+
+Pin provenance is not self-verifying. A pinned commit that shares a long prefix
+with the intended tag but diverges is indistinguishable by eye and simply fails
+to resolve at run time. Confirm any new pin against its upstream tag before
+adding it to the allowlist.
+
 ## Updating frozen trust roots
 
 The credential-containment workflow compares the candidate copies of its
--- a/.github/SECURITY_ENVIRONMENTS.md
+++ b/.github/SECURITY_ENVIRONMENTS.md
@@ -143,6 +143,16 @@ with the intended tag but diverges is indistinguishable by eye and simply fails
 to resolve at run time. Confirm any new pin against its upstream tag before
 adding it to the allowlist.
 
+These jobs must also use the canonical block form the checker understands: block
+mapping jobs, a `steps:` list at six-space indentation, no flow-style `jobs: {...}`,
+no job image, and no reusable-workflow call. A step whose scalar spans several
+lines with a continuation beginning `uses:` is rejected as well. Recognising that
+shape needs a real YAML parser, which this checker forgoes so the same trusted
+file can run in the minimal policy job, so it fails closed and the author
+reformats. The restriction is safe in the other direction: a scalar can never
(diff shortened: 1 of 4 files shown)


hugocorreia90:
## Review record

### Bootstrap evidence

`Credential containment policy` is red by construction. The failure is exactly two violations — the two frozen trust roots this PR intentionally changes — reproduced identically by CI and locally:

```
.github/scripts/check_credential_containment.py: frozen trust root differs from trusted base
.github/tests/test_credential_containment.py:    frozen trust root differs from trusted base
credential-containment policy: 2 violation(s)
```

Running the candidate checker against the candidate tree — the post-merge steady state — returns `passed`, exit 0.

### Independent review

Reviewed by Codex/GPT-5 as the independent red team. **It did not produce a formal written verdict**: two separate runs terminated on OpenAI's content filter, the second immediately before writing its report. Its sandbox also had no outbound network, so it could not verify pin provenance.

It did produce three concrete findings before terminating, each of which I reproduced independently and dispositioned:

| Finding | Status |
|---|---|
| Flow-style `jobs: {...}` bypassed every per-job rule (job parser returns nothing or `<unparsed:>`) | **Fixed** — non-PR workflows now require canonical block form, as the PR side already did |
| Completeness check counted `uses:` keys outside `steps:`, rejecting a matrix dimension named `uses` | **Fixed** — counting scoped to the steps block; job-level `uses:` rejected separately by name |
| A multi-line scalar whose continuation begins `uses:` is rejected | **Accepted as a deliberate boundary** — documented and pinned by a test |

On the third: recognising plain/quoted multi-line scalars needs a real YAML parser, which this checker forgoes so the same trusted file can run in the minimal policy job. It fails **closed**, and I verified there is no fail-open dual — four attempts to conceal an executed step behind block, folded and quoted scalars were all caught, because step splitting keys off the six-space prefix unconditionally.

Codex also independently reached the same conclusion on scope: the seven dual-trigger workflows are not source-unchecked, they pass `_check_unprivileged_job` with canonical parsing and full 40-hex pins, are all `contents: read` with zero secrets, and only `extractions/setup-just` sits outside the release allowlist.

### Verification

- checker returns `passed` on the real tree — no false positives
- 110 unit tests; `actionlint` exit 0
- **7 guards, each mutation-checked individually**: removing any one makes the test named after it fail
- probe suites: 12 attack shapes rejected, 10 legitimate shapes clean, control 0 violations
- all 15 allow-listed SHAs verified against upstream by two methods (`gh api` + `git ls-remote`), which is what surfaced the corrupted `wasm-pack` pin

### Note on merge mechanics

`Credential containment policy` is now a required status check on `main` (added today), so this PR is correctly `BLOCKED` and needs the documented narrow maintainer override. That is the intended ceremony for a trust-root change, and this PR is its first live exercise.

PR #1284, opened 2026-07-29
fix(engine/rocky-cli): preserve DAG partition selection

### LINKED ISSUE #1283: bug(engine): run --dag silently discards partition selection
## Description

`rocky run --dag` accepts the time-interval partition flags, but the DAG execution path silently discards them. A historical partition request therefore succeeds while materializing the current/latest partition instead.

The same root cause affects persisted DAG runs created with `rocky plan --dag`: apply reconstructs `PartitionRunOptions`, then calls `run_with_dag` without passing them.

## Steps to reproduce

1. Configure a daily `time_interval` transformation model whose `first_partition` includes `2020-01-01`.
2. Run:

```console
rocky run --dag --partition 2020-01-01
```

3. Inspect the materialized partition or partition state.

The CLI accepts the combination and exits successfully, but executes the partition containing the current UTC time.

## Expected behavior

Every transformation node executed by the unified DAG receives the requested partition selection. `--partition 2020-01-01` must execute only that partition, and the sibling partition options (`--from`/`--to`, `--missing`, `--lookback`, and `--parallel`) must retain their normal `rocky run` semantics.

If DAG partition selection is unsupported, the CLI must reject the combination before execution rather than silently changing it.

## Actual behavior

`engine/rocky/src/main.rs` builds the requested `PartitionRunOptions`, but the `dag` branch calls `run_with_dag` without them. `engine/crates/rocky-cli/src/commands/run_dag_exec.rs` then creates a fresh options value with `partition`, `from`, and `to` set to `None`, `latest` and `missing` set to `false`, no lookback, and `parallel = 1`.

The time-interval runner interprets that empty selection as `PartitionSelection::Latest`, so the command reports success after executing the wrong partition.

## Impact

Backfills and historical repairs invoked through the unified DAG can update the wrong time window without an error. Operators may believe a requested historical partition was rebuilt when it was not, while the latest partition is rewritten unexpectedly.

## Suggested fix

Thread the caller-provided `PartitionRunOptions` through `run_with_dag`, `CliDispatcher`, and the per-node sub-run. Add a dispatch-boundary regression test proving the exact options survive, plus a CLI-level guard against future argument loss for both direct and plan/apply DAG paths.

## Environment

- Rocky: 1.67.0 / `main@d65f1ac0e55e87b466ebf6bca02362d36fd4d255`
- OS: macOS Darwin 25.5.0 arm64

mattfaltyn:
## Summary

Preserve time-interval partition selection when `rocky run --dag` dispatches transformation nodes. The same options now survive persisted `rocky plan --dag` / `rocky apply` replay, so a requested historical partition cannot silently degrade to the latest partition.

Closes #1283.

## Subproject(s) touched

## Changes

- Thread the caller's `PartitionRunOptions` through `run_with_dag`, `CliDispatcher`, and every pipeline sub-run.
- Replay persisted DAG partition options through the same path during `rocky apply`.
- Add an exact dispatcher-boundary test for all partition and skip options.
- Add a DuckDB regression test proving `--partition 2020-01-01` materializes that historical partition through the real DAG execution path.
- Document the corrected behavior in the engine changelog.

## Test Plan

- `cargo test -p rocky-cli --all-features --no-run`
- `cargo test -p rocky --all-features --no-run`
- `commands::run_dag_exec::run_opts_threading_tests::dispatch_passes_the_threaded_run_opts_to_each_sub_run`
- `commands::run_dag_exec::tests::historical_partition_selection_reaches_dag_model_sub_run`
- `cargo fmt -- --check`
- Full Linux test, clippy, format, adapter-boundary, codegen, credential-containment, and release-build jobs in GitHub Actions.

## Checklist

- [x] If this is a CLI JSON schema or DSL syntax change, all consumers (`integrations/dagster/`, `editors/vscode/`) are updated in this same PR (not applicable: no schema or syntax change)

### Engine (`engine/`)

- [ ] `uv run pytest` passes

- [ ] `npm run lint` is clean


hugocorreia90 (changes requested):
Thanks — the diagnosis in #1283 is right and the threading is done in the right place (one `SubRunner` seam, one cloned value per node, seeds/source/test nodes correctly bypassed). I verified the fix rather than just reading it, and I have one substantive change request plus a rebase.

## What I verified

**Both new tests bind.** I mutated the production code and re-ran them:

- Restoring the pre-fix empty `PartitionRunOptions` inside `default_sub_runner` → `historical_partition_selection_reaches_dag_model_sub_run` fails with `left: 0, right: 1`. Confirms the pre-fix path materialized the *now()* partition (empty), not `2020-01-01`. `PartitionSelection::Latest` resolves to exactly one key (`plan_partition.rs::resolve_selection`), and `first_partition` is only read by `Missing`, so the test is genuinely non-vacuous.
- Reverting the dispatcher call argument to `PartitionRunOptions::default()` → both new tests fail (`left: None, right: Some("2020-01-02")` and the row assertion).

**It works on current `main`.** I rebased locally. Six conflict hunks, plus several further `run_with_dag` / `SubRunner` call sites that `main` added and that don't conflict but still need the new argument — so the merge is a bit wider than the conflict markers suggest. Note `apply.rs` auto-merges silently into a six-argument call; I checked it and the resulting order (partition, skip, shadow) is correct. After the rebase all 11 `run_dag_exec` tests pass, including the new one alongside the `--shadow` tests from #1285. On the PR head as-is: `cargo test -p rocky-cli --all-features` → 1351 passed, 0 failed, and `cargo clippy -p rocky-cli --all-targets --all-features -- -D warnings` exits clean.

**The apply-side change is live, not dead code.** `build_and_persist_run_plan` (`plan.rs`) persists all seven fields unconditionally, and `execute_run_plan` reconstructs them before the `--dag` branch, so `&partition_opts` is in scope and correct.

---

## 1. `--parallel` should not be threaded as-is (the one blocker)

This is the part that goes beyond #1283 and I think is wrong as written.

`rocky run --parallel` defaults to **4**; the code this PR deletes hardcoded **1**. So threading the whole struct changes `--dag` concurrency, not just partition selection.

The reason that matters is a deliberate design decision in the non-DAG path. `run.rs::execute_models`:

```rust
// The two special paths (`content_addressed`, `time_interval`) stay serial —
// they push to `output` inline and `time_interval` already self-parallelizes
// per-partition. A layer mixing strategies runs entirely serially.
let can_parallelize = requested_parallel > 1
    && adapter_supports_concurrency
    && !matched.is_empty()
    && matched.iter().all(|(_, _, model)| is_plain_strategy(model));
```

Non-DAG, `--parallel N` therefore gives you *either* N concurrent plain models *or* N concurrent partitions inside a serially-executed `time_interval` model — never both. The ceiling is N.

Under `--dag` there is no such guard: `run_with_dag` builds `DagExecutor::new(dispatcher)` and never calls `with_max_concurrency`, so `max_concurrency` is `None` → `Semaphore::MAX_PERMITS` and every node in a layer runs at once. (`with_max_concurrency` is currently referenced only from `benches/dag_execute.rs`.) Post-PR that becomes *unbounded nodes × up to 4 partitions each* — exactly the multiplication `execute_models` refuses to do. It bites whenever a selection plans more than one partition: `--missing`, `--from/--to`, or even a flagless `rocky run --dag` against a model with a TOML `lookback > 0`.

It also means `--parallel` still doesn't have the semantics #1283 asked it to retain — `rocky run --dag --parallel 1` does *not* force serial execution, because node fan-out ignores the flag entirely.

Two ways out; either is fine by me:

- **Minimal (my preference for this PR):** thread only the six selection fields and keep `parallel: 1` in the sub-run. That fixes the silent-wrong-partition bug — the actual subject of #1283 — with no concurrency change at all, and leaves `--parallel`-under-`--dag` as a separate, honest follow-up issue.
- **Complete:** also wire the flag into `DagExecutor::with_max_concurrency` so `--parallel` means one thing on both paths. Bigger blast radius (it caps today's unbounded DAG fan-out), so probably its own PR.

Either way the CHANGELOG line needs to stop listing `--parallel` / "concurrency" as preserved unless node-level concurrency is actually bounded.

## 2. The `plan --dag` → `apply` half has no test

Both new tests call `run_with_dag` directly; nothing exercises `execute_run_plan`'s `--dag` branch. #1283 explicitly asked for a guard on *both* paths, and this is the path where the options round-trip through JSON, so it has a failure mode the direct path doesn't.

The harness already exists in `apply.rs`'s test module — `minimal_run_plan()` + `write_plan(..., PlanKind::Run, ...)` + `run_apply_run_plan(...)`, as used by `governed_dag_apply_is_refused`. A plan with `dag = true, partition = Some("2020-01-01")` applied against the same DuckDB fixture as the new e2e test would close it in ~30 lines.

## 3. Mixed-grain projects change behavior — worth a CHANGELOG sentence

One `PartitionRunOptions` is cloned to every node, but grain validation happens per model inside the sub-run (`plan_partitions` → `validate_partition_key`). Day and hour keys have incompatible canonical formats (`2020-01-01` vs `2020-01-01T00`), so on a project mixing grains, `rocky run --dag --partition 2020-01-01` now hard-fails the hour-grain node where it previously exited 0.

That's the correct direction and it matches non-DAG `rocky run`, so I'm not asking you to change the behavior — but it turns a green command red for those projects, and the failure is *partial*: because layers fan out unbounded, other nodes in the same layer can commit before the mismatched one fails. Worth one sentence in the CHANGELOG entry. (A pre-dispatch validation pass over every affected model would be the real fix; happy for that to be a follow-up.)

## 4. Nit: the removed `apply.rs` caveat was still true

The replaced comment said the DAG runner "reads config + tooling defaults rather than walking the same flag matrix as `commands::run::run`". That's still accurate — `--filter`, `--env`, `--idempotency-key`, `--cache-ttl` and `--resume`/`--resume-latest` are all still passed as `None` by `default_sub_runner`. (`--defer`, `--var` and `--assume-fresh-state` are rejected outright with `--dag`, so those are genuinely handled.) I'd keep the caveat and add the partition sentence to it rather than replace it.

---

## Adjacent, pre-existing — not this PR

Worth filing separately: `execute_run_plan` takes the apply gate's `Arc<LoadedConfig>` and documents that it "performs NO config load of its own (its former `.ok()` re-load was a swap window between the gate and execution)" — but the `--dag` branch passes only `config_path`, and `run_with_dag` calls `load_rocky_config_fingerprinted` again. That reopens exactly the gate→execution swap window the #1120 single-snapshot invariant closed, for DAG applies only. Threading the existing `Arc` through `run_with_dag` would close it.

---

To summarize: rebase onto `main`, drop `parallel` from the threading (or bound the DAG), add the apply-path test, and touch up the CHANGELOG. The rest is good work — the seam you picked is the right one and the recorder test is a real regression guard.

hugocorreia90:
Superseded by #1290 — thank you, this was a good catch and a correct diagnosis.

#1290 keeps your work: the root cause, the `SubRunner` seam you chose to thread through, and your `historical_partition_selection_reaches_dag_model_sub_run` end-to-end test, which I mutation-checked (reverting the fix makes it fail `left: 0, right: 1`, so it genuinely binds). The commit message and PR description both credit you.

What changed relative to this PR, all of it from the review above:

- `--parallel` is no longer threaded. It rides in the same struct as the selection but answers a different question, and its `rocky run` default is 4 against the 1 this boundary hardcoded — so threading it raised concurrency on the flagless `rocky run --dag` path too. `sub_run_partition_opts` narrows at the dispatch seam and pins per-sub-run concurrency to 1. Giving the flag one meaning on both paths is #1288.
- Added the missing apply-leg test through `execute_run_plan`.
- Restored the `apply.rs` flag-light caveat, since it was still true.
- Reworded the CHANGELOG, and corrected the `--parallel` help text, which promised something `--dag` never delivered.

The rebase was the other blocker — `main` had moved twice under this branch, and `apply.rs` auto-merges into a six-argument call that nobody reviewed.

Closing this one in favour of #1290. Sorry for taking it over rather than iterating here; that's my convention once a change needs more than a couple of rounds, not a reflection on the work.

Issue #1268, opened 2026-07-27
run and validate still derive a models glob's base by splitting on ** alone


hugocorreia90:
## Description

A transformation pipeline's `models` field is a glob. Three places derive its base directory, and they disagree:

- `models_loader::resolve_models_dir` (used by `tick`, and by `run --dag` after #1264) splits on `['*', '?', '[']` and confines the result to the project root. **Correct.**
- `run.rs:1886` splits on `"**"` alone.
- `validate.rs:924` splits on `"**"` alone.

## Impact

With `models = "models/*.sql"` and a conventional `models/orders.sql`:

- `run.rs` derives the literal path `models/*.sql`, finds no such directory, takes the `ModelsDirDecision::Absent` branch, and the transformation pipeline silently no-ops.
- `validate.rs` similarly resolves nothing, so its model-count and DAG checks see an empty project.

Same shape as #1243: a supported configuration producing no work and reporting success.

Neither is a wildcard the docs discourage — `models` is documented as a glob pattern with no restriction to `/**` forms.

## Why this was not fixed in #1264

#1264 fixed the `--dag` path and left `run` alone deliberately. Changing `run`'s derivation flips those projects from "silently builds nothing" to "builds and materializes tables", which is the correct behavior but is a materialization change that deserves its own PR and its own note, not a side effect of a DAG fix.

The interim state is that `rocky run --dag` handles `models/*.sql` correctly while plain `rocky run` still no-ops on it. That is the safe direction of divergence — the DAG builds what the config declares — but it is a divergence, and it should not stay.

## Suggested fix

Point `run.rs` and `validate.rs` at `models_loader::resolve_models_dir` so all four sites share one derivation, and take the project-root containment that comes with it. Note the behavior change for affected projects in the release notes.

Found by Codex (`gpt-5.6-sol`) while red-teaming #1264.

Issue #1286, opened 2026-07-30
branch promote loads models without project-root containment, and derives production targets from them


hugocorreia90:
## Description

`rocky run` and `rocky validate` now refuse a `models` glob whose base resolves outside the project root (#1268). `rocky branch promote` does not: it joins and loads the same path with no containment check, then derives **production** promotion targets from whatever sidecars it finds there.

## Evidence

`commands/branch.rs:1115-1134` derives the base and loads models directly:

```rust
let project_root = config_path.parent().unwrap_or(Path::new("."));
let models_base = pipeline.models.split(&['*', '?', '['][..]).next()
    .filter(|base| !base.is_empty())
    .unwrap_or("models");
let models_dir = project_root.join(models_base.trim_end_matches('/'));
```

No canonicalization, no project-root prefix check — compare `models_loader::locate_models_dir` and `scope.rs`, which both confine the result. `branch.rs:1167-1176` then turns those models into the promotion target list, so an out-of-tree sidecar's `[target]` decides what gets replaced in production via `CREATE OR REPLACE TABLE prod.<x> AS SELECT * FROM branch__<name>.<x>`.

## Impact

A configuration `run` refuses to execute is still consumable by `branch promote`, which is the more consequential verb of the two. The inconsistency is the defect: an operator who fixes their glob after a `run` refusal has no signal that `promote` was reading it all along.

Not a privilege boundary — someone who can edit `rocky.toml` can already point `models` anywhere. It is a misconfiguration guard that one command enforces and another does not.

## Suggested fix

Point `branch.rs` at `models_loader::locate_models_dir` and propagate its error, the way `run` does. The call site is already fallible (`anyhow::bail!` on a missing directory at `branch.rs:1123`), so no new error plumbing is needed.

## Not in scope of #1268

That PR deliberately limited its new refusal to `run` and `validate`. Adding one to `promote` is a separate behaviour change to a separate command and deserves its own note.

Found by Codex GPT-5 red-teaming #1268.

PR #1295, opened 2026-07-31
fix(engine): honor model-file globs

### LINKED ISSUE #1293: bug: literal-prefix model globs silently run no models
## Description

`TransformationPipelineConfig.models` is documented as a glob pattern, but a glob whose filename component has a literal prefix (for example `models/allowed*.sql`) is derived as the literal directory `models/allowed`. Because that path does not exist, `rocky run` takes the missing-models-directory branch, exits successfully, and materializes nothing.

This is an incomplete edge case in the fix from #1268 / #1287. That change correctly made `models/*.sql` resolve to `models`, but deriving the base by splitting at the first wildcard leaves the literal filename prefix attached.

## Reproduction

On `main` at `f85821dea58a5b0839530de5bd0f2504817529c5`, create:

```toml
# rocky.toml
[adapter]
type = "duckdb"
path = "warehouse.duckdb"

[pipeline.selected]
type = "transformation"
models = "models/allowed*.sql"

[pipeline.selected.target]
adapter = "default"
```

Add a valid `models/allowed.sql` plus its sidecar (and, optionally, a
`models/blocked.sql` control), then run:

```console
rocky --config rocky.toml --state-path state.redb --output json run --pipeline selected --parallel 1
```

Observed:

- exit status is 0;
- stderr reports `models/allowed` as the missing models directory;
- JSON reports `"status":"Success"` and `"materializations":[]`;
- the target table is not created.

The same project with `models = "models/*.sql"` materializes the model.

## Expected behavior

Consistent with the field's documented glob contract, the directory portion of
`models/allowed*.sql` should resolve to `models` and only matching model sources
should be loaded. A successful run must materialize `allowed` rather than
silently completing with no work, and it must not materialize a nonmatching
`blocked` model.

## Existing issue / PR check

I searched the repository's complete issue and pull-request history (including closed and merged items) for the literal-prefix examples, `locate_models_dir`, and related glob wording. No existing issue or PR covers this case. #1268 / #1287 covers `models/*.sql`, not a filename component containing both literal text and a wildcard.

mattfaltyn:
## Summary

Closes #1293.

Rocky currently derives a model root from `TransformationPipelineConfig.models` but then loads every model below that root. A valid file glob with a literal filename prefix, such as `models/orders*.sql`, therefore either selects nothing or admits models that the configuration excludes, depending on the command path. A wildcard-free model file such as `models/orders.sql` is also misclassified as a directory and expanded to `models/orders.sql/**`, so it selects nothing.

This change makes the configured glob the shared source of truth for model discovery while retaining a confined directory base for defaults and sidecar lookup. Adjacent selector defects found during implementation are included because they share that resolution path.

## Subproject(s) touched

## Changes

- Resolve pipeline model globs relative to the config file, derive their non-wildcard directory base (or an exact source file's parent), and reject matches that escape that base.
- Preserve wildcard-free `.sql`/`.rocky` files as exact selectors while retaining directory semantics for an existing directory whose name ends in `.sql` or `.rocky`.
- Normalize an explicitly empty model selector to the single default `models/**` pattern.
- Filter SQL and Rocky DSL inputs by the complete glob before parsing or compilation, and skip defaults, groups, and test metadata in directories that contain no matching source.
- Use the same selected model set in run, DAG, apply/policy, branch, validate, GC, restore, maintenance, scope, and column-lineage paths.
- Preserve explicit `run --models` precedence and deduplicate physical files selected by overlapping validation globs.
- Add source-compatible compiler entry points for matching and already-loaded models; normal variable substitution, lowering, contract validation, and cached source-schema behavior remain intact.
- Add regression coverage for literal-prefix, exact-file, relative-config, extension-suffixed directory, wildcard-directory, recursive, traversal, override, DAG, policy, validation, excluded metadata, and run-variable cases.

## Test Plan

- `cargo nextest run --all-features` — 5,813/5,813 passed across 112 binaries; 108 configured tests skipped.
- `cargo check --locked --workspace --all-targets --all-features` — passed.
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — passed.
- `cargo fmt --all -- --check` — passed.
- Focused selector suite — 21/21 passed, including all 8 `models_glob_base` E2E tests.
- Independent red-team review by Averroes — no P1/P2 findings; its `.sql` directory-compatibility finding was fixed and regression-tested before final approval.
- Final GitHub duplicate searches found no existing issue or PR for literal-prefix selection, exact model-file selection, or excluded-directory metadata parsing. #1268/#1287 cover directory wildcard base derivation (`models/*.sql`), not filtering `models/orders*.sql` or `models/orders.sql`.

## Checklist

- [x] If this is a CLI JSON schema or DSL syntax change, all consumers (`integrations/dagster/`, `editors/vscode/`) are updated in this same PR (not applicable; no schema or DSL syntax change)

### Engine (`engine/`)
- [x] `cargo test --all-targets` passes (all-features Nextest run)

- [ ] `uv run pytest` passes (not touched)
- [ ] `uv run ruff check` is clean (not touched)
- [ ] `uv run ruff format --check` passes (not touched)

- [ ] `npm run compile` succeeds (not touched)
- [ ] `npm run test:unit` passes (electron tests run in CI; not touched)
- [ ] `npm run lint` is clean (not touched)

## Review handoff

**What are you least confident about?** Windows-specific separator/glob behavior was not exercised on this macOS host. The existing separator-aware unit cases and shared path-resolution code remain green.

**What’s the most important thing I’m missing?** There is no full E2E dedicated to an exact `.rocky` file selector. Exact `.rocky` resolution is covered at the resolver level and follows the same compiler matching path as the existing Rocky DSL coverage.

**ModelIr / contract semantics preserved: yes** — DAG execution, policy evaluation, and lineage compile the same glob-selected model objects through Rocky's existing lowering, run-variable substitution, and contract validation paths.


hugocorreia90 (changes requested):
Thanks — the core of this is right and the selector work is a real improvement. `models/orders*.sql` and `models/orders.sql` both selecting correctly is the fix #1293 asked for, and routing every command through one resolved glob is the right shape.

Three comment/test corrections before merge, and two things I want to reframe rather than change.

Also: this fixes **#1286** as well (the `locate_models_dir` swap in `discover_transformation_branch_targets`, plus `transformation_branch_targets_refuse_an_escaping_literal_prefix_glob`). Worth adding to the closes list so it doesn't stay open after merge.

---

### 1. The new `run_dag_exec.rs` doc claim is not true, and it replaces the warning that was

The old comment said the loader reads a base directory plus one level of subdirectory, "NOT true glob matching — see #1262". The replacement says `--dag` and a plain run now agree on which files are in scope. They still don't:

- a plain `run` compiles through `Project::load_models_with_db`, which reads **one** directory (`rocky-compiler/src/project.rs:117`)
- DAG discovery and `list models` go through `load_project_models_partial_with`, which reads the root **plus immediate subdirectories** (`rocky-cli/src/models_loader.rs:176`)

So for a `models/staging/` layout they disagree exactly as before. Please restore a scope note rather than assert agreement.

### 2. This PR does not close #1262 — worth being explicit

Related to the above, and I mention it because the PR description reads as though discovery became glob-driven end to end. What landed is a glob **filter** applied over the two unchanged fixed-depth walks; `load_project_models_partial_with` still does one `read_dir` plus one subdir loop with no second descent. `models/a/b/lvl2.sql` is still never read, so `models/**` still cannot reach it.

That's fine as scope — a filter can only narrow, so nothing widens and no behaviour decision is being made here. I just don't want the tree to claim otherwise. `models_loader.rs:196-211` still honestly documents the fixed-depth limit, which is the note that contradicts item 1.

### 3. `api.rs`'s `dag_column_lineage_tolerates_nested_model_layout` now documents the opposite of what it does

The doc still says the lineage compile reads only the root, so the nested `stg` is invisible and only `fct` is compiled. This PR falsifies that — `build_column_lineage_from_models` now takes the already-loaded model set and uses `compile_preloaded_models`, so the nested model does participate. The test stays green because it only asserts node labels.

This is genuinely the nicest change in the PR and it's under-advertised. Please update the comment and add an assertion on the lineage edge so the new behaviour is pinned rather than incidental.

---

### Reframing, not asking for changes

**`models/*.sql` now excludes `.rocky` models and subdirectory models.** I initially read this as an unflagged narrowing and I was wrong — `models` is documented as a glob for model *files* (`rocky-core/src/config.rs:4989`), the changelog entry says only matching `.sql`/`.rocky` sources are parsed, and `project.rs` covers matching `.rocky` loading. It's intended semantics. The one thing I'd ask: state the consequence outright in the changelog, because "only matching sources are parsed" implies rather than says that a `.sql`-suffixed glob drops every `.rocky` model in that directory. A project on `models/*.sql` with DSL models would lose them silently on upgrade.

**`gc` and `restore` became fallible.** These previously could not fail and now propagate a containment error, and the doc explaining the infallibility went with it. I checked whether this was the #1266 shape — a shared helper made stricter, feeding a policy gate an empty result — and it isn't: `gc.rs:1579` resolves the models dir *before* the policy gate at `gc.rs:1636`, so the `?` aborts rather than un-gating anything. It fails closed. Please just call it out in the changelog as a deliberate compatibility change, since a config `gc` tolerated yesterday now errors.

---

Everything above is comment/test/changelog work. I haven't reviewed the whole 20-file surface line by line, so this isn't a blanket approval of the rest — but nothing I did read is a correctness blocker.

hugocorreia90:
Superseded by #1305 — thank you, this was a correct diagnosis and a better fix than the one I shipped alongside it.

**#1305 keeps your work as your commit**, with original authorship: the base-derivation redesign, carrying the resolved glob through every consumer rather than just the base directory, the `locate_models_dir` `Present | Absent` shape, and all 17 of your tests. My changes sit in a second commit on top.

**Part of why this moved rather than iterating is my error.** While this PR was open I merged #1300, which fixed the model-only models-directory derivation — a defect yours already fixed, and fixed *better*: you also carry the glob forward so non-matching siblings are not recompiled. That created the `run.rs` conflict. #1305 resolves it by taking your version wholesale and deleting mine. I confirmed yours is a genuine superset rather than assuming it: both regression tests #1300 added pass unmodified against your logic.

What changed relative to this PR, all of it from the review above plus one new finding:

- **`resolve_touched_apply_targets` fail-open (the substantive one).** An unconfinable models glob degraded to an empty model set, so every physical target stayed unresolved — on the policy input for `rocky compact` and `rocky archive`, both destructive. That function's own doc is explicit that an unresolved target reaches `default_agent_effect` *only* when no configured rule's scope is satisfied by the default attribute set, "it is **NOT** guaranteed to". A permissive rule could have authorized an apply the real model attributes denied. Now refuses with an explicit reason, with a test.
- **The `run_with_dag` doc claim.** It said `--dag` and a plain run "agree on which files are in scope". They apply the same glob but read different file sets — 2-level walk vs 1-level compile. Corrected and pointed at #1262, which #1305 does not close.
- **`dag_column_lineage_tolerates_nested_model_layout`.** Your switch to `compile_preloaded_models` is the nicest change in the diff and was under-advertised — nested models now participate in lineage. But the test still documented the opposite and asserted only node labels. It now pins the `stg -> fct` edge, so a regression to the old root-only read cannot stay green.
- **Changelog placement.** The entry had landed inside the already-released `[1.68.0]` section; moved to `[Unreleased]`, and the `.sql`-glob-excludes-`.rocky` and `gc`/`restore`-fallibility consequences are now stated outright rather than implied.

Two things from my earlier review I want to withdraw on the record, because I was wrong and you were right:

- I called the `models/*.sql` narrowing "silent and untested". It is neither — `models` is documented as a glob for model *files*, your changelog says only matching sources are parsed, and `project.rs` covers matching `.rocky` loading. I only asked that the consequence be spelled out, and did that myself.
- I flagged `gc`/`restore` becoming fallible as the #1266 fail-open shape. It is not: the `?` fires before the policy gate and before any mutation, so it fails *closed*. I verified that rather than leaving the accusation standing.

I also verified your glob semantics empirically rather than by reading: `models/**` matches at one, two and three components, `models/*.sql` correctly excludes subdirectories, `models/ord*.sql` matches `orders.sql`. Worth noting the filter is already correct for the recursive walk #1262 will add — the walk is what limits depth today, not the glob.

Closing in favour of #1305. Sorry for taking it over rather than iterating here; that is my convention once a change needs more than a couple of rounds, and in this case the conflict I created made it the cleaner path. Not a reflection on the work.

mattfaltyn:
All good thank you @hugocorreia90 🙏

PR #1298, opened 2026-07-31
fix(engine/rocky-cli): execute the apply gate's config snapshot on the --dag path

### LINKED ISSUE #1289: bug(engine): `rocky apply` of a `--dag` plan reloads config, reopening the #1120 gate→execution swap window
## Description

`rocky apply` of a `--dag` run plan reloads `rocky.toml` from disk after the apply gate has already read and fingerprinted it, reopening the gate→execution config-swap window that the #1120 single-snapshot invariant closed for every other apply path.

`commands/apply.rs::execute_run_plan` takes the caller's snapshot and documents the invariant on its own parameter:

```rust
// The caller's ONE fingerprinted config snapshot (#1120): the same
// instance the apply gate read. The preflight/seam reads below and the
// `run()` execution all use it — this function performs NO config load
// of its own (its former `.ok()` re-load was a swap window between the
// gate and execution).
loaded: std::sync::Arc<rocky_core::config::LoadedConfig>,
```

The `--dag` branch then passes only `config_path`:

```rust
return crate::commands::run_with_dag(
    config_path,
    state_path,
    output_json,
    &partition_opts,
    &crate::commands::run::SkipRunOptions::default(),
    shadow_config.as_ref(),
)
```

and `commands/run_dag_exec.rs::run_with_dag` loads the file again:

```rust
let loaded = std::sync::Arc::new(
    rocky_core::config::load_rocky_config_fingerprinted(config_path)
        .with_context(|| format!("failed to load config from {}", config_path.display()))?,
);
```

`loaded` is dropped on the floor for this branch. The non-DAG path below it threads the snapshot through correctly, so the DAG apply is the only apply that re-reads.

## Impact

A `rocky.toml` swap between the apply gate and DAG execution makes the DAG execute a different config than the one the gate approved — different adapter, different targets, different governance — while reporting success against the reviewed plan id. This is the same class of defect #1120 fixed inside the DAG (one snapshot for every node) and that the `.ok()` re-load removal fixed for non-DAG apply; the DAG apply seam was missed.

Narrow in practice — it needs a write to `rocky.toml` inside the gate→execute window — but it is a documented invariant that this path silently does not hold, and the plan-identity checks (`config_policy_identity`) run against the gate's snapshot, not the reloaded one.

## Suggested fix

Give `run_with_dag` an optional pre-loaded `Arc<LoadedConfig>` (or a variant that takes one) and thread `execute_run_plan`'s `loaded` into it, so the DAG apply executes the same instance the gate read. `run_with_dag`'s own `rocky run --dag` caller keeps loading it itself. The dispatcher already `Arc::clone`s one snapshot per node, so nothing downstream changes.

Add a regression test in the shape of the existing `a_stored_dag_shadow_plan_carries_its_shadow_routing_into_execution`: mutate `rocky.toml` between constructing `loaded` and calling `execute_run_plan`, and assert the DAG executed the original.

## Environment

- Rocky: 1.67.0 / `main@7a3063fd`
- Pre-existing; not introduced by #1284 / #1288. Surfaced while reviewing #1284.

hugocorreia90:
## Summary

Closes #1289.

`rocky apply` of a `--dag` run plan reloaded `rocky.toml` after the apply gate had already read and fingerprinted it, reopening the gate→execution config-swap window that #1120's single-snapshot invariant closed for every other apply path.

`execute_run_plan` documents the invariant on its own `loaded` parameter — "this function performs NO config load of its own (its former `.ok()` re-load was a swap window between the gate and execution)". The `--dag` branch passed only `config_path` onward, and `run_with_dag` then called `load_rocky_config_fingerprinted` again, dropping the caller's snapshot on the floor. The non-DAG path below it threaded the snapshot correctly, so the DAG apply was the only apply that re-read.

A write to `rocky.toml` inside that window made the DAG execute a different adapter, different targets and different governance than the gate approved, while reporting success against the reviewed plan id — and `config_policy_identity` had already run against the gate's snapshot, not the reloaded one.

## Subproject(s) touched

## Changes

- `run_with_dag` takes the caller's `Arc<LoadedConfig>` and performs no config load of its own. `run_dag_exec.rs` now contains **zero** non-test `load_rocky_config*` calls.
- `execute_run_plan`'s `--dag` branch threads its own `loaded` through, so both apply entrypoints (human and AI-authored — they share `execute_run_plan`) are covered by the one change.
- `rocky run --dag`, which has no gate above it, loads the one snapshot in `main.rs` and hands it down.

### Why the parameter is required rather than `Option`

The issue suggests an optional pre-loaded config. This uses a non-optional one instead, matching the shape the repo's own #1120 remediations chose for `run_seed` / `run_load`: an `Option` leaves the fallback load in place, so a future caller can pass `None` and silently reopen the window, and a regression test could only ever prove that *today's* apply passes `Some`. Requiring the snapshot enforces "no config read" at compile time. `config_path` stays alongside it — `run_with_dag` still needs the path for relative resolution in four other places (`load_transformation_models`, the `seeds` dir, the `CliDispatcher`, and `default_sub_runner`).

## Test Plan

New regression test `a_stored_dag_plan_executes_the_gates_snapshot_not_a_reload`.

**The test named in the issue as a template is vacuous for this purpose.** `a_stored_dag_shadow_plan_carries_its_shadow_routing_into_execution` sets `shadow: true` and asserts an `Err` from the DAG's shadow refusal — which fires *after* the config load, so both the old and new config produce the identical error and nothing is ever materialized. Pre-fix and post-fix outcomes are byte-identical. It is a usable harness but an unusable assertion.

The new test instead composes the harness from `a_stored_dag_partition_plan_rebuilds_that_partition_not_the_latest` (same `execute_run_plan` `--dag` entry, `shadow: false`, expects `Ok` and real rows — so an earlier gate turning it into an error fails it too) with the A/B swap drill from `seed_dispatch_executes_the_captured_snapshot_not_a_reload`: two DuckDB files sharing the stem `proj`, so `catalog = "proj"` is valid under either config and only the physical path differs. Capture the snapshot pointing at warehouse A, overwrite `rocky.toml` to point at B, apply, then assert the rows landed in A **and** that B's file was never created (DuckDB creates the file on open, so opening B at all is observable).

**Mutation-checked, not just green.** Restoring the pre-fix reload inside `run_with_dag` makes the test fail on exactly the intended assertion:

```
panicked at crates/rocky-cli/src/commands/apply.rs:4157:
the swapped-in config must never be read: opening warehouse B would create its file
```

Gates:

- `cargo nextest run --all-features` — **5,799 passed**, 0 failed, 108 skipped.
- Verified `run_dag_exec.rs` has no remaining non-test config load.

## Scope

This closes the **config** swap window only. Model files are still read from disk at DAG execute time, and no models-fingerprint gate is in play because a governed `--dag` apply is refused outright (`apply.rs`: "a `--dag` apply is not yet policy-gated"). That refusal is unchanged.

## Checklist

### Engine (`engine/`)

### DIFF
--- a/engine/CHANGELOG.md
+++ b/engine/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [1.68.0] - 2026-07-31
+### Fixed
+
+- **`rocky apply` of a `--dag` run plan now executes the config its apply gate approved, instead of re-reading `rocky.toml`.** `execute_run_plan` documents on its own `loaded` parameter that it performs no config load of its own — the single fingerprinted snapshot (#1120) is threaded from gate through execution so a `rocky.toml` swap in that window cannot make execution diverge from what was reviewed. The `--dag` branch broke that invariant: it passed only `config_path` onward, and `run_with_dag` loaded and fingerprinted the file again. A write to `rocky.toml` between the gate and DAG execution therefore ran a different adapter, different targets and different governance than the gate approved, while still reporting success against the reviewed plan id — and the plan-identity checks (`config_policy_identity`) had already run against the gate's snapshot, not the reloaded one. The non-DAG apply path threaded the snapshot correctly, so this was the only apply that re-read. `run_with_dag` now takes the snapshot as a **required** parameter rather than an `Option`, so the fallback load is gone entirely and a future caller cannot silently reopen the window — the same shape `run_seed` / `run_load` adopted for #1120. `rocky run --dag`, which has no gate above it, loads the one snapshot in `main.rs` and hands it down. Narrow in practice (it needs a write inside the gate→execute window) but it was a documented invariant the path silently did not hold. (#1289)
 
 ### Added
 
--- a/engine/crates/rocky-cli/src/commands/apply.rs
+++ b/engine/crates/rocky-cli/src/commands/apply.rs
@@ -656,6 +656,12 @@ async fn execute_run_plan(
         // contract.
         return crate::commands::run_with_dag(
             config_path,
+            // The gate's own snapshot — NOT a re-read (#1289). `execute_run_plan`
+            // documents on its `loaded` parameter that it performs no config load
+            // of its own; this branch used to violate that by handing `run_with_dag`
+            // only a path, which then loaded the file again after the gate had
+            // fingerprinted it.
+            std::sync::Arc::clone(&loaded),
             state_path,
             output_json,
             &partition_opts,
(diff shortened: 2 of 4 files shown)


hugocorreia90:

Per `AGENTS.md`, reviewed by **Codex (GPT-5)** via `codex:codex-rescue` — a genuinely different model from the author. Read-only, against the immutable branch ref.

**Verdict: safe to merge. No confirmed code issues.** All six attack lines returned CONFIRMED:

1. **Window closed end-to-end.** No production config loader remains below `run_with_dag` — the dispatcher clones the one Arc, seed dispatch passes it to `run_seed` (which has no config path at all), and `default_sub_runner` passes it to `run::run`. Every remaining `load_rocky_config*` hit in `run_dag_exec.rs`, `run.rs`, `seed.rs`, `run_local.rs`, `load.rs` and `models_loader.rs` is inside `#[cfg(test)]`.
2. **Test is non-vacuous and sound.** `db_b`'s absence is a valid sentinel — `DuckDbConnector::open` creates its file, and adapter-registry construction opens the configured DuckDB immediately. An early failure cannot pass the `.expect`, a zero-node false success cannot pass the A-table query, and the fresh tempdir excludes a stale table.
3. **No caller weakened.** Both production callers pass the gate's own Arc (apply clones the instance it fingerprinted; the AI-authored route does the same). None of the eight converted test calls previously mutated `rocky.toml` mid-flight, so the `dag_snapshot` helper masks nothing.
4. **Direct `rocky run --dag` preserved** — same loader, same error context.
5. **No enum-exhaustiveness / nullability / 3VL surface touched.**
6. **The governed `--dag` refusal is real and unchanged**, including for AI-authored plans.

Two notes carried forward rather than dismissed:

- **One ordering change it identified:** the config load now happens before the outer `tokio::select!`, so an already-pending shutdown is observed only after the synchronous parse. The loader is a file read + parse + validation with no warehouse or state side effect, so this is a cancellation-ordering nuance, not a semantic change. Accepted.
- **Scope boundary worth keeping explicit:** "single snapshot" covers `rocky.toml` only — model SQL, sidecars and seed files remain live execution-time inputs. Acceptable here *because* governed DAG applies are categorically refused; it must be revisited if governed `--dag` is ever enabled. This is stated in the PR description and now in-code.

Its one limitation was environmental: its sandbox rejected `tempfile::tempdir()`, so it could not execute the new test and confirmed the mutation statically instead. That gap is covered — I ran the test and the mutation check locally, and restoring the reload fails it on exactly the intended assertion (`the swapped-in config must never be read`). CI's `Test` job is green on this branch.

Issue #1275, opened 2026-07-27
bug(engine): physical schema.table upstream reads produce no DAG edge, so producer and consumer are co-scheduled


hugocorreia90:
## Description

A model that reads another model's target by its physical `schema.table` name produces no DAG edge, so the two land in the same execution layer and can run concurrently — the consumer may read a table its producer has not written yet.

This is not shadow-specific; it affects any run whose models reference upstreams by physical name.

## Root cause

`rocky_compiler::resolve::classify_table_ref` auto-derives a dependency only from a **bare single-part** name that matches a model name:

```rust
match parts.len() {
    1 => { /* ModelRef if it matches a model name */ }
    2 => TableRefKind::SourceRef { .. },   // never a model
    _ => TableRefKind::RawRef(..),          // never a model
}
```

So `SELECT id FROM main.orders` never resolves to the model that produces `main.orders`, and `resolve_dependencies` records no edge unless the sidecar spells out `depends_on`.

## Steps to reproduce

Two models, neither declaring `depends_on`:

- `orders` → target `main.orders`, SQL `SELECT 1 AS id`
- `mart_qualified` → target `main.mart_qualified`, SQL `SELECT id FROM main.orders`

`rocky_ir::dag::execution_layers` on the compiled project returns a single layer containing both:

```
[["mart_qualified", "orders"]]
```

With `--parallel N` on an adapter reporting `supports_concurrent_execution() == true`, they execute simultaneously.

## Expected behavior

Either the edge is derived from the physical read, or a model whose read set cannot be proven disjoint from the rest of the layer is not co-scheduled with it.

## Notes

The engine already has the machinery. `commands/containment.rs` deliberately does not trust `dag_nodes` as a complete read set — it folds physical-target reads into the failure closure via `rocky_sql::lineage::extract_lineage`, and fails closed on any model whose reads are not provably complete (`rocky_sql::lineage_complete::lineage_is_provably_complete`). The same resolution would supply the missing edges here.

The `D011` diagnostic currently tells users to "remove the depends_on field entirely to let auto-derivation handle everything", which is only safe advice for bare-name references. Either the derivation should cover physical reads, or that suggestion needs qualifying.

Related: #1270 fixed the shadow-routing half of this (a physical read of an in-run upstream was being redirected only when declared). The ordering half is untouched.

hugocorreia90:
[...]

Issue #1236, opened 2026-07-26
ci-diff: target reported removed while another model still occupies it


hugocorreia90:
## Description

When two models resolve to the **same** target and one of them moves away, `ci-diff` reports the
old target as `removed` even though the other model still occupies it. Under the row-identity
contract (a `DiffResult` row is a warehouse target) that is wrong: nothing is abandoned.

## Reproduce

```toml
# models/a.toml and models/b.toml — both target c.s.orders
name = "a"   # and "b"
[target]
catalog = "c"
schema  = "s"
table   = "orders"
```

Retarget only `a` (`schema = "s2"`), commit, then `rocky ci-diff HEAD~1 --models models`.

## Actual

```
a  removed  c.s.orders     <- but b is still there
a  added    c.s2.orders
```

## Expected

`c.s.orders` is still occupied by `b`, so it should not be reported as removed. Only
`c.s2.orders` is genuinely new.

## Pre-existing

Confirmed on `4bf43c8c` (before the target field existed), which produces the same
`a removed` + `a added` pair — the resolved target was simply not visible then, so the wrongness
was not observable. Adding `resolved_target` exposed it rather than causing it.

## Why it happens

`Project::from_models` rejects duplicate `config.name` but says nothing about duplicate targets, so
colliding targets are name-qualified into `{target}#{name}` keys to keep both models in the report
(#1226). That qualification makes Added/Removed a per-*model* judgement, while the contract wants a
per-*target* one.

## Direction

Decide Added/Removed from target **occupancy** across the whole base and head projects — a target is
Removed only when no head model resolves to it, Added only when no base model did — rather than from
one model's assignment. Alternatively, reject duplicate targets at project construction, which would
make the question moot and is arguably right on its own (two models writing one table is a config
error Rocky currently accepts silently).

hugocorreia90:
Triage note — this folds into #1291, but **should not be closed into it yet.**

This issue's own "Direction" offers rejecting duplicate targets at project construction as the option that makes the question moot, and #1291 independently reaches the same conclusion from the scheduling side. That is one decision, not two, and #1291 is where it should be made.

[...]

Issue #1281, opened 2026-07-29
Read the live identifier case-sensitivity setting instead of assuming it


hugocorreia90:
Shadow execution currently **assumes** how a warehouse treats identifier case, because it has no way to ask. Several fail-closed refusals exist only to cover that assumption. Obtaining the real setting would replace all of them with an exact answer.

## Why the assumption is unsafe in both directions

Case-sensitivity is **connection/object state, not dialect state**:

- Snowflake: `QUOTED_IDENTIFIERS_IGNORE_CASE` is a session/account parameter. Rocky renders Snowflake targets double-quoted and does not pin this parameter, so whether `"orders"` and `"Orders"` are one object or two is decided by the account.
- BigQuery: a dataset can be created with `is_case_insensitive`, so the answer varies *per dataset* within one project.

Assuming one answer is fail-open for one question and fail-closed for the other, which is why shadow mode currently keeps two separate rules:

| question | wrong answer causes | current handling |
|---|---|---|
| could two targets be the same object? | two models silently write one shadow table | always fold — refuse the pair |
| does this reference name that routed upstream? | rewrite → read the wrong table; leave → read **production** while writing shadow | neither is safe, so refuse the reference |

Both refusals are conservative placeholders. With the real setting, each becomes a decision rather than a refusal.

## What to implement

A capability on the warehouse adapter that answers "is identifier case part of object identity for this connection/dataset", e.g.:

- Snowflake: `SHOW PARAMETERS LIKE 'QUOTED_IDENTIFIERS_IGNORE_CASE'` once per session.
- BigQuery: the dataset's `isCaseInsensitive` field from the Datasets REST resource, per dataset touched.
- DuckDB / Databricks / Trino: constant `false` (case is not part of identity) — no round trip.

Then feed it into `dialect_case_rules` in `engine/crates/rocky-cli/src/commands/run.rs` so the per-component `IdentifierCaseRules` reflect reality.

## Acceptance

- Two models whose targets differ only by case are refused **only** when the connection actually folds case, and route independently when it does not.
- A reference matching a routed upstream only under folding is resolved rather than refused, in whichever direction the live setting dictates.
- The probe failing is itself fail-closed — an unreadable setting must keep today's conservative behaviour, not fall through to a guess.
- Tests cover both states of the setting, not just the default.

## Context

[...]

PR #1299, opened 2026-07-31
fix(engine/rocky-compiler): report two models targeting one physical table

### LINKED ISSUE #1291: bug(engine): two models with the same physical target run concurrently in the same DAG layer
## Description

Nothing prevents two transformation models from declaring the same `catalog.schema.table`. `rocky-compiler`'s project construction rejects duplicate model *names* (`crates/rocky-compiler/src/project.rs`), not duplicate *targets*, and each model freely declares its own `[target]`.

Under `rocky run --dag` those two models are independent nodes with no edge between them, so they land in the same execution layer — and `DagExecutor` dispatches every node in a layer concurrently with no cap (`DagExecutor::new` leaves `max_concurrency` unset; see also #1288). Two concurrent writers to one physical table.

## Reproduced on `main@f85821de`

One transformation pipeline, two models `a` and `b`, both with `catalog="p" schema="m" table="shared"`:

```
$ rocky dag --models models --output json
nodes:            ['transformation:a', 'transformation:b']
execution_layers: [['transformation:a', 'transformation:b']]
```

Same layer. Pre-existing — it needs neither multiple pipelines nor any recent change.

## Expected

Either refuse at DAG-build time with both model names and the shared target, or serialize same-target writers into different layers. Refusing is probably right: two models writing one table is almost always a config mistake, and the survivor depends on scheduling.

## Why this is filed separately

Surfaced while designing #1261. The obvious collision key — "two pipelines claim the same model" — does not catch this, because the models are *different*. A correct check has to key on the resolved **physical** target: canonical adapter routing identity (Rocky already models this for plan/apply identity in `commands/apply.rs`, where e.g. a DuckDB `path` change is an identity change) plus `catalog.schema.table`, and it must account for per-model adapter overrides (`rocky_core::cross_engine::effective_model_adapter`). Adapter *name* or adapter *type* alone is insufficient.

## Environment

hugocorreia90:
## Summary

Closes #1291. Re-scopes #1236 (see below).

Nothing stopped two transformation models from declaring the same `catalog.schema.table`. `Project::from_models` rejects duplicate model *names* but says nothing about duplicate *targets*, so the two are independent DAG nodes with no edge between them: same execution layer, and `DagExecutor` dispatches a layer with no concurrency cap. Two concurrent writers to one table. Even serially the surviving rows depend on execution order, which is alphabetical.

## Subproject(s) touched

## What it does

Project construction groups models by the physical object their target names and records each colliding group on `Project::target_collisions`. `compile` renders each group as an **E036** error diagnostic naming every participating model.

That error severity is the enforcement: `run.rs` already builds `exec_compile_failed` from error-severity diagnostics keyed by model, excludes those models from execution, and counts them toward `tables_failed`. So the colliding models never execute — no concurrent write — while every model with its own target still builds.

Identity is `rocky_sql::defer::CollisionIdentity`, the same case-folding key shadow-target collision detection already uses. Answering "different" for two spellings that name one warehouse object is the fail-open this exists to prevent.

`ephemeral` models are excluded — they carry a fully populated but phantom target and materialize nothing, so they are never the second writer. (`config.skip` is *not* excluded; it is the skip-if-unchanged gate, not a materialization opt-out.)

## Three design points worth reviewing

**1. A diagnostic, not a `ProjectError`.** The issue asks for a refusal at DAG-build time. A hard construction failure is the wrong instrument here: it is swallowed into silence by every tolerant caller — `ci-diff` logs the head compile at `debug!` and `.ok()`s the base, and the LSP's `Err` arm publishes no diagnostic at all, so the editor would show *nothing*. It would also make any historical ref containing a collision unloadable, which forecloses #1236's own preferred fix (deciding added/removed from target occupancy across base ∪ head needs both projects to load).

Direct evidence this was the right call: `colliding_targets_keep_both_models` (`ci_diff.rs`) — the one test a hard error would have broken — still passes untouched.

**2. No separate execution refusal.** An earlier revision added an explicit `anyhow::ensure!` in `execute_models`. It is deleted, because it was redundant *and worse*: the diagnostic path already prevents the colliding models from executing, and it does so **per model**, whereas aborting the invocation also stranded every healthy model in the project. `a_target_collision_does_not_stop_the_healthy_models` pins that containment.

**3. Not routed through `resolve_diagnostics`.** That field looks like the natural channel, but it currently holds only D011 warnings and is **read nowhere** — so collisions would have been invisible. Merging it wholesale into `compile` would newly surface D011's text, which #1275 shows is misleading advice (it tells users to drop `depends_on` and rely on auto-derivation, which does not cover physical reads). A dedicated `target_collisions` channel avoids entangling the two.

## Scope

The key is `catalog.schema.table`, **not** the adapter routing identity #1291 also asks for. `from_models` takes `Vec<Model>` and no `RockyConfig`, so it cannot resolve an adapter name to a physical destination. That is sound *today* only because the per-model adapter override is coded but not wired — `rocky_core::cross_engine::effective_model_adapter` has zero production call sites, and execution resolves exactly one warehouse adapter per *pipeline*. The in-code doc says so and names what must change when that is wired.

The cross-*pipeline* case (two pipelines claiming one target) is not covered; it belongs next to `reject_models_claimed_twice` in `unified_dag`, which does have the config. #1291's filed repro is single-pipeline.

Not addressed here, and worth its own decision: `rocky import-dbt` derives `target.table` from the dbt `alias`, so two dbt models sharing an alias in one resolved schema now produce a project that reports E036 after the import completes.

## Test Plan

Blast radius was measured before writing the check, not assumed: **zero** duplicate-target groups across 176 models in 84 model roots under `examples/`, `engine/examples/`, `engine/crates/*/tests/` and the playground — so no fixture churn and no `just regen-fixtures`.

New tests:

- `duplicate_target_is_recorded_not_rejected` — construction succeeds and records the group.
- `duplicate_target_identity_folds_case` — `c.s.Shared` and `C.S.shared` are one object.
- `ephemeral_models_do_not_collide`.
- `distinct_targets_do_not_collide` — guards against the check firing on every project.
- `two_models_targeting_one_table_never_both_execute` — end-to-end on DuckDB; the run fails **and** `main.shared` is never created, so an enforcement firing after the first model materialized would still fail this.
- `a_target_collision_does_not_stop_the_healthy_models` — the containment property.

**Mutation-checked, not just green.** Replacing `collide_on_target(&models)` with an empty vec makes both end-to-end tests fail.

Gates:

- `cargo nextest run --all-features` — **5,805 passed**, 0 failed, 108 skipped.

## Checklist

### Engine (`engine/`)


hugocorreia90:

Per `AGENTS.md`, reviewed by **Codex (GPT-5)** via `codex:codex-rescue`. **Verdict: BLOCK**, with three standing findings. All three were correct. Two are fixed in this PR; one is filed.

### S1 — cross-pipeline `--dag` still races. VALID. → filed as #1301

The enforcement is per-*compile*, and `rocky run --dag` compiles one model set per pipeline — so two pipelines whose models claim one target never meet in a single compile, emit no E036, and both nodes fan out.

I had scoped this out, but the reviewer's sharper point stands: **the changelog described the `--dag` fan-out as the motivating harm while not closing the cross-pipeline case.** Within one pipeline `--dag` *is* covered (each sub-run compiles the whole pipeline root, so both writers are present). Across pipelines it is not. The changelog now says exactly that, and #1301 tracks the adapter-routing-aware preflight at unified-DAG construction — which, unlike this check, genuinely needs routing identity, since two pipelines may legitimately share a `catalog.schema.table` on *different* warehouses.

### S2 — `--dag` poisons healthy nodes. VALID, and it refuted my blast-radius claim. → FIXED HERE

Compile-error accounting counted **every** error diagnostic, in scope or not. Every `--dag` node is its own `--model X` sub-run compiling the whole pipeline root, so the colliding pair's E036 errors made *every* node report failure — even nodes that had materialized correctly — and `DagExecutor` then skipped their healthy descendants.

That directly refutes "every other model still builds", which was true only for the non-DAG path my test covered.

Fixed by separating two things the code conflated:

- **exclusion set** — every model with an error diagnostic, so nothing broken executes (unchanged)
- **reported failure** — only models the invocation would actually have run

The broken model's own node still reports it, which is where it belongs. This also fixes the pre-existing general case: `rocky run --model X` no longer fails because some unrelated model is broken.

Pinned by `a_collision_does_not_fail_a_model_scoped_run_of_another_model`, mutation-checked — removing the scope filter fails it.

**Credit where due:** I hit this myself earlier, saw `--model a` return `2 model(s) failed`, and talked myself out of it as "consistent with existing behavior for compile errors". It was the bug.

### S3 — `rocky ai` bypasses the compile pipeline. VALID. → filed as #1302

`rocky-ai` builds a `Project` and typechecks directly rather than going through `compile`/`compile_project`, so it sees only `TypeCheckResult` diagnostics and misses E036 — and it applies `--target` *after* validation, so what is verified is not what is written. Not a warehouse race (a later compile catches it), but a real divergence between compile entry points.

### S4 — case-folding false positives. CONFIRMED as a deliberate tradeoff

On Snowflake (targets always double-quoted) and BigQuery, two targets differing only in case are distinct objects, so folding can refuse a legitimate project. It fails **closed**, matching the existing `CollisionIdentity` contract that shadow collision detection already relies on, and #1281 is the issue that would make this exact rather than assumed. Now stated in the changelog rather than left implicit.

### S5 — test gaps. VALID → two added

No test asserted the E036 code itself. Added `a_target_collision_becomes_one_e036_error_per_model` (asserts the code, the *severity* — a warning would exclude nothing from execution — one diagnostic per participant, and that the message names the target and every model) plus `no_collision_emits_no_diagnostic`.

The remaining requested tests are cross-pipeline DAG coverage, which belongs with #1301 since the behaviour it would pin does not exist yet.

### S6 / other dispositions — CONFIRMED

`Ephemeral` is the complete non-materializing set (`View` still creates a named object; content-addressed writes through its runtime path). No missed in-repo `Project` literal. No codegen cascade — `Project` is not a CLI JSON schema, and E036 reaches consumers through `CompileOutput.diagnostics`, which already exists.

One point I accept without a fix: `Project` is public, so an external `Project { .. }` literal would need the new field. No external consumer is provable from this repo.

hugocorreia90:
Superseded by #1303 — identical branch content. Recreated because GitHub stopped dispatching `engine-ci` for this PR after its first commit; three re-trigger attempts (push, close/reopen, force-push a fresh SHA) all failed to register the engine jobs. See #1303 for the full description and the independent red-team disposition.

PR #1300, opened 2026-07-31
fix(engine/rocky-cli): resolve --model against the named pipeline's models dir

### LINKED ISSUE #1292: bug(dagster): transformation assets execute by bare model name, so a multi-transformation-pipeline project runs models against the wrong adapter
## Description

`dagster_rocky` materializes a transformation asset by passing the node's bare label (the model name) to the CLI, without `--pipeline` (`integrations/dagster/src/dagster_rocky/dag_assets.py`).

On the engine side, a model-only invocation with no `--pipeline` falls back to the **first transformation adapter** it finds (`crates/rocky-cli/src/commands/run.rs`, the model-only fallback: "Without an explicit pipeline, preserve the model-only fallback: first transformation adapter, then first replication adapter").

For a project with one transformation pipeline that is always correct. For a project with two — say `silver` and `gold` with distinct `models` directories and different target adapters — every asset executes against whichever transformation pipeline happens to come first, so `gold`'s models are built into `silver`'s warehouse.

## Why it is not observable today

`rocky run --dag` / `rocky dag` currently fail outright for any project with 2+ transformation pipelines (#1261), so no such project can produce the asset graph that would expose this. Once #1261 lands, these projects build a valid DAG and the mis-routing becomes reachable.

## Related

The asset key itself is also pipeline-blind: transformation asset keys are built from `catalog`/`schema`/`table` only (`integrations/dagster/src/dagster_rocky/translator.py`), and column-lineage edges synthesize `transformation:<bare-model>` ids rather than using the node ids from the payload (`dag_assets.py`). Both assume a model name is globally unique across pipelines.

## Suggested fix

Carry the node's `pipeline` field (already present in `DagNodeOutput`) through the translator and pass `--pipeline <name>` when materializing a transformation asset. Then decide whether the asset key needs pipeline qualification too, or whether uniqueness of `catalog.schema.table` is a guarantee Rocky should enforce.

## Environment

- Filed from a review of #1261; not introduced by it.

hugocorreia90:
## Summary

The **engine half** of #1292. That issue is filed against `dagster_rocky`, which materializes a transformation asset by passing the bare model name with no `--pipeline`, so the engine falls back to the first transformation adapter. Fixing the Python side alone would not close it — this is why.

`rocky run --model X --pipeline P` already resolved the target **adapter** from `P` (`run.rs`, the `--model` arm). It did not resolve the models **directory** from `P`: that came from `--models`, or absent that a hardcoded literal `models`.

Two consequences, one of which I did not expect:

1. **Wrong directory for a multi-root project.** #1292's own scenario is two transformation pipelines with distinct `models` directories. With the directory hardcoded, `--pipeline gold` picked gold's warehouse and then looked for the model under `models/` — either not found, or a *same-named model belonging to the other pipeline*.
2. **The fallback is CWD-relative, not project-relative.** `Path::new("models")` resolves against the process working directory. So `rocky run --model X` with no `--models` only ever worked when invoked from the project root. The mutation check below shows this directly: with the fix reverted, the new test fails with `models directory 'models' not found` even though the project has one.

## Subproject(s) touched

## Changes

When `--model` is given with `--pipeline` and **no** explicit `--models`, derive the models directory from that pipeline's own `models` glob via `models_loader::resolve_models_dir` — mirroring `models_dir_for_model_scope`, which the unified-DAG sub-runner has always used for exactly this reason.

Deliberately narrow:

- **An explicit `--models` still wins.** It is an operator override, and the DAG sub-runner supplies one for every model-scoped sub-run — so that path is untouched.
- **`--model X` with no `--pipeline` is byte-identical to before**, including the CWD-relative `models` fallback. Changing that is a separate behaviour decision.
- A non-transformation `--pipeline` still falls through to the existing, better-worded rejection a few lines below rather than erroring on a missing directory.

## Test Plan

New test `model_only_run_loads_from_the_named_pipelines_models_dir`: two transformation pipelines, two DuckDB warehouses, two model roots.

It uses a **decoy** — a same-named `mart` model in *each* root selecting a different value (`99` in `models/`, `7` in `gold_models/`) — so "found a model" and "found the *right* model" are distinguishable. Querying gold's warehouse directly for `7` proves both halves at once: the right directory (7, not 99) and the right adapter (the table exists in gold's file at all).

**Mutation-checked.** Reverting to `models_dir.unwrap_or_else(|| Path::new("models"))` fails it:

```
--model with --pipeline must resolve that pipeline's models dir:
models directory 'models' not found (required for --model)
```

One assertion I wrote and then removed, recorded because a reviewer might expect it: I initially also asserted silver's DuckDB file is never created. That fails, but not because of this bug — `AdapterRegistry::from_config` eagerly constructs every configured adapter, so its file exists regardless of which pipeline runs. Pre-existing and unrelated; the comment in the test says so rather than leaving a subtly wrong assertion behind.

Gates:

- `cargo nextest run --all-features` — **5,800 passed**, 0 failed, 108 skipped.
- `cargo clippy -p rocky-cli --all-targets --all-features -- -D warnings` — clean.

## What this does NOT close

#1292 stays open. The `dagster_rocky` half is still required: `dag_assets.py` must pass the node's `pipeline` (already present on `DagNodeOutput`) through `resource.py` and `client.py`, and the SDK must stop unconditionally appending `--models self.models_dir` so this derivation is reachable. That half spans two independently released wheels and must follow **after** this engine change is released, since the published `dagster-rocky` resolves `rocky-sdk` from PyPI.

Two findings from tracing that half, recorded so they are not re-derived:

- The transformation **asset key is target-derived** (`translator.py`), so it stays byte-identical — passing `--pipeline` is **not** a breaking change for existing Dagster deployments, and `translator.py` needs no edit.
- **No `just regen-fixtures`** is implicated: `pipeline` already exists in `output.rs`, `schemas/dag.schema.json`, the generated Pydantic types, and `fixtures_generated/dag.json`.

## Checklist

### Engine (`engine/`)


hugocorreia90:
## Self-check: does this newly expose `locate_models_dir`'s strictness?

Flagging this before review because it is the failure shape this repo has been bitten by (#1266: making a shared helper's strictness reachable from a new caller).

`resolve_models_dir` delegates to `locate_models_dir`, which has **two** error paths this arm could not previously reach — a fail-closed canonicalize error (`models_loader.rs:105`) and a project-root containment refusal (`:112`). The new `?` propagates both. Walking every case:

| Config / invocation | Before | After |
|---|---|---|
| `models = "models/**"`, run from the project root | works (`models` resolves via CWD) | works (`<project>/models`) — identical |
| `models = "models/**"`, run from **elsewhere** with `--config /path/rocky.toml` | **broken** — `models` is CWD-relative, so "models directory 'models' not found" | **works** — resolves against the config's parent |
| `models = "../shared/**"` (escapes the root) | errors, "models directory 'models' not found" | errors, "resolves outside the project root" |
| `<project>/models` exists but cannot be canonicalized | may have worked via the CWD-relative path | fails closed |
| no `--pipeline` | unchanged | unchanged (derivation is skipped entirely) |
| explicit `--models` | unchanged | unchanged (the override still wins) |

So: **no config that works today starts failing.** The out-of-root case errors before and after — only the message improves. The one genuinely new refusal is the un-canonicalizable directory, which is fail-*closed* and the direction that helper deliberately chose.

It also fixes a case that was quietly broken: `--model` with `--config` pointing outside the CWD never resolved the models directory correctly, because the fallback is CWD-relative rather than project-relative.

Unlike #1266, the `?` here aborts the command — it does not hand a degraded or empty value to a policy gate.

hugocorreia90:

Per `AGENTS.md`, reviewed by **Codex (GPT-5)** via `codex:codex-rescue`, which additionally cross-checked with a second independent model. **Verdict: BLOCK.** Both findings were correct. Both are fixed in this PR.

*(Process note: this review took four attempts — two agent wrappers died on API errors and one Codex process died with a stale `running` status. The fourth completed. Recorded because "the review passed" and "the review never ran" look identical from the outside.)*

### Finding 1 — the fix still mis-routed when the pipeline's root is absent. FIXED

`resolve_models_dir` collapses `ModelsDir::Absent(path)` to `None`. My derivation then fell straight through `unwrap_or_else(|| Path::new("models"))` back to the **CWD-relative fallback** — re-introducing exactly the mis-routing this change exists to remove.

Trigger: `gold.models = "gold_models/**"`, `gold_models/` absent, and a `models/mart` in the process working directory → `mart` compiles from the wrong directory and executes against **gold's adapter**.

Fixed by using `locate_models_dir` and passing `Absent`'s path through to the existence check, so a missing root fails naming **the pipeline's own path**. Pinned by `an_absent_pipeline_models_root_fails_by_name_not_by_fallback` and mutation-checked: reverting to `resolve_models_dir` makes the error name `models` instead of `gold_models`, and the test catches it.

### Finding 2 — test gap. FIXED

My original test creates the decoy inside the temp project, but the old fallback was **process-CWD**-relative and the test never changes CWD — so it proved "old code cannot find `models`", not "old code builds the decoy". The new absent-root test closes that by asserting on the *name* in the error, which is the discriminating signal and needs no CWD manipulation.

### Correcting my own claim above — the strictness table was wrong

I posted a table asserting **"no config that works today starts failing."** That is **false**, and the reviewer refuted it correctly.

The row I missed: previously, a usable CWD-relative `models` let a model-only run succeed **without ever inspecting the pipeline's glob**. So a project with an out-of-root or un-canonicalizable `P.models`, run from a directory that happens to contain `models/`, succeeded before and now fails.

The correct statement: this **is** a compatibility tightening, not a no-op. It is the fail-closed direction — the alternative is executing a model from a directory the project never pointed at — but it should be described as a deliberate break rather than as costless. The `--config`-outside-CWD case genuinely is a fix, since derivation resolves against `config_path.parent()`.

- **No caller regression.** Explicit `--models` bypasses derivation; without `--pipeline` the fallback is unchanged. The reviewer enumerated `main.rs`, the HTTP API, the DAG sub-runner (already passes an explicit directory), apply, backfill (`pipeline: None`) and MCP (`pipeline: None`), and found no valid caller relying on the old behaviour.
- **Removing the `!silver_db.exists()` assertion was right** — the adapter registry opens every configured DuckDB regardless of selection, so both files exist either way. Asserting `7` in gold's warehouse does prove both dimensions: a wrong directory yields `99`, a wrong adapter yields no table at all.

Its stated open question — whether the newly-enforced containment failure is an acceptable announced break — is a product call, flagged here rather than decided in a PR.

PR #1306, opened 2026-07-31
fix(engine/rocky-compiler): derive a DAG edge from a physical target read

### LINKED ISSUE #1275: bug(engine): physical schema.table upstream reads produce no DAG edge, so producer and consumer are co-scheduled

hugocorreia90:
fix(engine/rocky-compiler): derive a DAG edge from a physical target read

`resolve::classify_table_ref` auto-derived a dependency only from a bare
single-part name matching a model name: a 2-part reference was
unconditionally a `SourceRef` and a 3-part one a `RawRef`, neither of which
is ever a model. So `SELECT id FROM main.orders` produced no edge to the
model writing `main.orders`, `execution_layers` returned both models in one
layer, and under `--parallel N` on an adapter reporting
`supports_concurrent_execution()` the consumer could read a table its
producer had not written yet — silent wrong data on a green run, and nothing
shadow-specific about it.

Qualified references now resolve against a producer index built from every
model's declared `[target]`, covering both the 2-part and the 3-part
spelling. Quoting and case are handled by the same canonicalizer failure
containment already uses for exactly these reads, lifted into
`rocky_sql::identifier` so the two cannot drift on what `"main"."orders"`
means; containment's own tests pass unchanged, which is the evidence the move
was behavior-neutral.

Four decisions, stated because a partial fix here reads as complete:

  * A reference that cannot be narrowed to one producer edges to ALL of
    them, not none. Refusing on ambiguity is the same silent co-scheduling
    being fixed. The motivating case is legitimate rather than an error —
    two models in different catalogs sharing a `schema.table`, which a
    catalog-less read cannot choose between — so this does not lean on the
    E036 duplicate-target error for its soundness.
  * A reference matching this model's own target is a self-read (the
    incremental `MAX(ts)` shape) and contributes no edge at all. The whole
    producer set is dropped, not just this model, so two models declaring
    one target that each self-read do not become a mutual pair.
  * A 2-part name matching no model target still classifies as a source. A
    model target wins over a same-named declared source, because the
    write-then-read hazard is real whatever else the table is called.
  * Bare names are still matched against model names only, never target
    tables — bare names share a namespace with CTE aliases that lineage
    extraction does not filter out. Containment takes the bare lookup on
    purpose; over-containing costs a withheld model, whereas a spurious
    ordering edge perturbs every healthy run and can close a cycle.

Derived edges are applied only while the graph stays acyclic. The resolution
is deliberately generous, so it can pair two models into a cycle the declared
graph does not have, and a cycle is a hard `ProjectError` that `ci-diff` and
the LSP swallow into silence. An edge that would close a cycle is dropped and
reported as a D012 warning instead, mirroring
`ContainmentLedger::augmented_layers`; a project that loads today still
loads. A pre-existing declared cycle is left reporting exactly what it
reported before.

Scope, for the same reason the duplicate-target check stated its own:
resolution happens per compiled model set, so it covers a plain `rocky run`,
`rocky compile`, `rocky plan`, the LSP, and each `rocky run --dag` sub-run
within one pipeline. A physical read of a different pipeline's target still
gets no edge, since `--dag` loads a separate model set per pipeline.

The D011 suggestion, which advised removing `depends_on` "to let
auto-derivation handle everything", now says where derivation stops.

Closes #1275


hugocorreia90:
Holding as draft: the branch this PR was opened from is superseded. A completeness pass found the same defect at a second site — `unified_dag::infer_runtime_dependencies` also matched a reference against a model's last path segment rather than its declared target — so the diff here fixes `resolve` only and would read as complete while `run --dag` kept the bug. Will update to the finished commit and re-gate before marking ready.

hugocorreia90:

**Reviewer: Codex / GPT-5.6-Terra**, run read-only against the first commit. **Verdict: BLOCK, five findings.** Four are fixed with tests; two are accepted as residual and documented in-code and in the CHANGELOG rather than papered over.

| # | Finding | Disposition |
|---|---|---|
| 2 | Whole-set self-read drop was a **fail-open** | **Fixed** — narrowed to exact-target twins. Red D proves it |
| 3a | Greedy guard could sacrifice a proven edge for a speculative one | **Fixed** — exact-catalog edges ranked first |
| 3b | D011/D012 were recorded on `Project` but **never surfaced to any user** | **Fixed** — `resolve_diagnostics` merged into `CompileResult`. Red E |
| 4 | 🔴 **`run --dag` bypassed the fix entirely** — `UnifiedDag::infer_runtime_dependencies` matched a reference's last segment against a node's *logical label*, so `stage_orders` writing `main.orders_current` was invisible to `FROM main.orders_current` even inside one pipeline | **Fixed** — that graph now resolves declared targets through the same canonicalizer and catalog rule. Red F |
| 1 | Physical reads inside a CTE body / derived table / EXISTS yield no edge | **Residual, documented.** Pre-existing limit of `extract_lineage`'s top-level FROM/JOIN walk, shared identically by the bare-name derivation that has always existed — not introduced or widened here. Remedy is an explicit `depends_on` |
| 5 | Cyclic fallback is Θ(N⁴) worst case | **Residual, documented.** Reachable only once a cycle exists (a misconfiguration) and only at that scale; the common path costs two sorts |

Finding 4 is why this PR is worth its size: the first commit fixed `resolve` and would have read as complete while `run --dag` — the concurrent path where the race actually bites — kept the bug.

A follow-up Codex pass on the fix delta was dispatched but had not returned after an hour, so it is **not** part of this evidence. The delta was reviewed by a stronger model instead, which found three more real gaps (missing quote canonicalisation and missing cycle guard in `unified_dag`, missing test for the diagnostics merge). All three are fixed.

## Non-vacuity

Six independent reds, each disabling exactly one mechanism and re-running. Red A (neuter the producer index) reproduces the issue's symptom verbatim:

```
a_physical_target_read_separates_the_execution_layers
  left:  [["mart_qualified", "orders"]]
  right: [["orders"], ["mart_qualified"]]
```

## ⚠️ Behaviour changes on upgrade — reviewers should weigh these

1. **Execution serialises where it previously ran concurrently.** That is the fix, but wall-clock time can increase and `rocky dag` / `rocky plan --output json` `execution_layers` change shape — visible to anything asserting on layer counts.
2. **D011 warnings become visible for the first time.** They were recorded but never surfaced, so a project with a `depends_on`/SQL mismatch will suddenly show warnings. All warning-severity, `has_errors` unchanged, nothing newly excluded from execution — but the diagnostic count in `--output json` grows.
3. A new **D012** appears where physical reads form a cycle.

One modified test, called out so it is not mistaken for a regression: `shadow_routes_undeclared_physical_upstream_reads` asserted `depends_on` was **empty** for a 2-part read. That premise *encoded the defect*. It now asserts the edge is derived; the test's actual shadow-routing assertions are unchanged and still pass.

## Least-confident point

The empty-catalog wildcard — a 3-part read matching a producer that declares no catalog — is a deliberate over-approximation. It can manufacture a false edge and therefore a false cycle, which is dropped with a D012 rather than failing the project, and exact-catalog edges are ranked ahead of it. **Not empirically validated against a live warehouse**, and settling it needs a session where the default catalog differs from the named one.

hugocorreia90:
## ⛔ Independent completeness review: **FIX_FIRST** — not mergeable as-is

A second reviewer, run adversarially for *completeness* rather than correctness, returned five findings. Four are CONFIRMED. Two are HIGH, and one **falsifies a claim made in this PR's own residual-risk list** (and in my comment above — correcting it here).

### F1 (HIGH, CONFIRMED) — the over-approximation leaks out of scheduling into **type checking**, and it errors

Physical edges land in `dag_nodes.depends_on`, which `semantic.rs:284` copies verbatim into `ModelSchema.upstream` (`:463`), which is the trigger for `typecheck::check_join_keys` (`typecheck.rs:2099`, called unconditionally at `:540`) via `upstream.len() >= 2`. It then compares same-named columns across upstreams **the model never reads**, and a pair with no `common_supertype` emits `Diagnostic::error(E001)` at `typecheck.rs:2230` — an error, not a warning.

Repro: `raw_orders` targets `raw.main.orders` with `id` Utf8; `curated_orders` targets `analytics.main.orders` with `id` Int64. `mart` declares `depends_on = ["curated_orders"]` and its SQL reads two-part `main.orders`. Pre-fix `mart.upstream == ["curated_orders"]`, `check_join_keys` returns at the len<2 guard, project compiles clean. Post-fix the catalog-less read matches **both** producers — documented behaviour — so `upstream == ["curated_orders","raw_orders"]`, Utf8 vs Int64 has no supertype, **E001 → `has_errors = true` → `mart` excluded from execution → `rocky compile` and `rocky run` fail on a project that was green before the upgrade.** The phantom upstream is one the model provably never reads (wrong catalog).

**This directly contradicts "All are warning-severity so `has_errors` is unchanged and nothing new is excluded from execution."** The graph change, not the diagnostics merge, is what flips it. No test asserts that physical edges cannot introduce new *errors*. Same root cause also widens the P002 blast-radius lint and `compile_incremental`'s affected set (`compile.rs:460,471`).

### F2 (HIGH, CONFIRMED) — the fix reaches two of **three** ordering consumers

`rocky dag` builds the same `UnifiedDag` at `dag.rs:219` and computes `execution_phases` at `:224` **without ever calling `infer_runtime_dependencies`** — its only production call site is `run_dag_exec.rs:276`. So the orchestrator-facing export carries the declared graph only. `dagster_rocky/dag_assets.py:180` builds `deps_list = node.depends_on` and passes it as `deps=`, so the missing edge becomes a **missing Dagster asset dependency** — the exact #1275 race, in the plane the project documents as its orchestrator contract, while `rocky run --dag` now serialises. The two answers disagree.

It also makes residual-risk claim (a) factually wrong: **`rocky dag`'s `execution_layers` are unchanged by this commit.**

### F3 (MEDIUM, CONFIRMED) — the cycle-drop was ported without either safeguard

`extend_while_schedulable` (`unified_dag.rs:936-957`) drops a cycle-closing edge with **no diagnostic and no `tracing::warn!`**, where the compiler emits D012 for the identical decision. It also re-adds candidates in insertion order with no exact-first ranking — `exact` isn't tracked on that side at all, though Codex finding 3a was fixed compiler-side with `edges.sort_by_key(|e| !e.exact)`. Since that graph also matches label tails, a **speculative** edge can be kept while a **proven** physical-target edge is dropped, silently.

### F4 (MEDIUM, CONFIRMED) — the issue's second repro is untested

The issue body's repro is covered; the repro in the issue *comment* — an executor-level from-scratch failure, "0 executed / 1 failed" — is not. Every new test asserts layer structure, not a run. `e2e.rs` is DuckDB-backed and credential-free, so it could host it.

### F5 (LOW, PLAUSIBLE) — ephemeral parity

`resolve::ProducerIndex::build` skips `Ephemeral` (with a test); `target_by_model` at `run_dag_exec.rs:274-284` does not. The two graphs disagree on a case the commit message claims is aligned.

---

**F1 alone blocks.** It converts a previously-compiling project into a failing one, which is a worse outcome than the race being fixed. Holding as draft.

hugocorreia90:
## Second independent review: **BLOCK** — 7 findings, all CONFIRMED

This is a different reviewer and a different lens (correctness/adversarial) from the completeness pass above. It reached the same verdict independently. **Two reviewers, two blocks, 13 findings between them**, with meaningful overlap — which is itself signal.

| # | Sev | Finding |
|---|---|---|
| 1 | HIGH | **Exact-catalog priority is lost during first-seen dedup**, so a speculative edge can displace a proven one — `resolve.rs` ~518. The `sort_by_key(!exact)` fix does not survive dedup |
| 2 | HIGH | The unconditional **label-matching fallback in `unified_dag`** has no confidence ranking and can silently displace a proven physical-target edge |
| 3 | MED | **`rocky run` (default, non-`--dag`) does not surface D011/D012 to the operator** — so the diagnostics merged in this PR reach `compile` and the LSP but not the most common command |
| 4 | MED | **D011's advice is unsafe**: it suggests removing an explicit `depends_on`, which is wrong given the acknowledged CTE / derived-table / EXISTS blind spot. Following the hint can delete a real edge the resolver cannot re-derive |
| 5 | MED | **Ephemeral parity** — `UnifiedDag` indexes Ephemeral targets, the compiler excludes them; phantom producer edge |
| 6 | LOW | Canonicalizer refusal silently drops the edge with no compensation on the compiler path (narrower in practice than disclosed — many malformed forms fail earlier) |
| 7 | LOW | The empty-catalog wildcard's risk is **understated**: it is a direct contributor to finding 1's exact-edge displacement, not merely a live-warehouse validation gap |

### Where the two reviews agree

Findings 1/2 here are the same defect the completeness pass raised as its F3, reached independently: **the exact-first ranking does not actually protect a proven edge.** Finding 5 here is its F5. Convergence from two different framings is the strongest evidence in this review round.

### What only one reviewer caught

- The completeness pass alone found the **E001 typecheck regression** — the one that breaks previously-green projects — and the **`rocky dag` gap** that leaves dagster-rocky with a missing asset dependency.
- This pass alone found **D011's unsafe advice** and that `rocky run` never shows these diagnostics.

Neither review subsumes the other, which is the argument for running both.

### Disposition

**Not merging.** The cumulative picture is that the fix is directionally right and the producer-index design is sound, but the *edge-selection* layer (dedup, ranking, fallback confidence) does not preserve the guarantees the commit message claims for it, and the graph change has a reach into type checking that was not accounted for.

Concretely, before this can land:
1. **F1 (typecheck E001)** — physical edges must not widen `ModelSchema.upstream` in a way that turns a green project red. This needs a decision, not a patch: either scheduling edges are kept distinct from type-checking upstreams, or `check_join_keys` must not trigger on derived edges.
2. **Exact-edge preservation through dedup** — with a test that a proven edge survives a competing speculative one.
3. **`rocky dag`** — decide whether the orchestrator export joins the fix or is explicitly documented as declared-only. Today the two answers disagree, and dagster-rocky consumes the one that is still wrong.
4. Ephemeral parity, D011 advice, and diagnostic surfacing on the default `run` path.

hugocorreia90:
Blocking finding, found while re-assessing this draft. **A derived edge can turn a project that compiles today into one that does not** — which is the opposite of what this PR's own comment claims.

## The chain, every link read

1. This PR derives edges into `DagNode.depends_on` (`resolve.rs`, `resolve_dependencies`).
2. `semantic.rs:285` copies `node.depends_on` **verbatim** into `ModelSchema.upstream` (`:463`).
3. `typecheck.rs:2099` gates the join-key check on `upstream.len() >= 2`. A model that declared 0 or 1 upstreams and now has 2+ derived ones **starts being checked** where it previously returned early.
4. `check_join_keys` compares same-named columns across upstreams; a pair with no `common_supertype` emits `Diagnostic::error(E001)` — an **error**, not a warning (`typecheck.rs:2229-2237`).
5. `compile.rs:352` seeds the diagnostic list from type-check: `let mut diagnostics = type_check.diagnostics.clone();`
6. `compile.rs:368` sets `has_errors = diagnostics.iter().any(is_error)` → **true**, and the model is excluded from execution.

## The comment this PR adds is refuted by the same PR

`compile.rs:364-365`:

> All are warning-severity, so `has_errors` below is unchanged and nothing new is excluded from execution.

That is true of D011/D012 — the diagnostics it is attached to. It is false as a statement about this change, because **the graph edit flips `has_errors`, not the diagnostic edit**. The comment is doing justifying work it cannot support, and it sits directly above the line it is wrong about.

## Concrete repro shape

Two producers, `a` → `main.a` and `b` → `main.b`, each with a column `id` — one `VARCHAR`, one `BIGINT`. A consumer `c` with SQL `SELECT ... FROM main.a JOIN main.b USING (id)` and **no** `depends_on`.

- Today: `c` has 0 upstreams, `check_join_keys` returns at the `< 2` guard, project compiles.
- With this PR: `c` has 2 derived upstreams, the types have no common supertype, E001, `has_errors = true`, `c` excluded.

The same shape widens the P002 blast-radius lint and `compile_incremental`'s affected set (`compile.rs:460,471`), for the same reason.

## Why this is a design question, not a patch

`depends_on` is doing two jobs: it is the **scheduling** graph and it is the **type-checking** upstream set. They are the same field today. This PR needs one of:

- carry derived edges separately (e.g. `DagNode.declared_depends_on` alongside the full set, with `semantic.rs` reading the declared one), so type-checking keeps its current input while scheduling gets the new edges; or
- gate `check_join_keys` on declared upstreams explicitly.

The first is cleaner — the two consumers genuinely want different sets — but it touches `rocky-ir`, `rocky-compiler` and every `DagNode` construction site, so it wants deciding before implementing.

## State of this branch

- **26 commits behind `main`**, and `git rebase origin/main` conflicts in `resolve.rs`. Any CI green recorded on this PR is against a base that no longer exists, so it is not evidence.
- Prior history worth remembering: this branch was once 13/13 green with a compile-breaking regression that author gates, six non-vacuity checks and a review round all passed. Whatever lands here needs a rebase plus a full re-gate, not a re-read of old results.

Leaving as a draft. The underlying bug (#1275) is real and severe — silent wrong data from co-scheduled producer and consumer — but shipping it must not trade that for refusing projects that work today.

hugocorreia90:
Closing as superseded by #1352, which fixed #1275 with a better design than this draft had.

## Why this one should not be revived

I left a blocking finding here: deriving edges into `DagNode.depends_on` feeds `ModelSchema.upstream` (`semantic.rs:285`), which triggers `check_join_keys` at `upstream.len() >= 2`, which can emit **E001** — an error, not a warning — turning a project that compiles today into one that does not.

**#1352 avoids that, by ordering rather than by a separate field.** `augment_physical_read_edges` (`run.rs:7483`) mutates `compile_result` *after* compilation, alongside the defer and shadow rewrites. So the semantic graph and type check have already been built from the **declared** `depends_on`, and the derived edges reach the scheduler without ever reaching `check_join_keys`. Same outcome as the declared-vs-derived split I suggested, at a fraction of the blast radius — no `rocky-ir` change, no touching every `DagNode` construction site.

It is also the more complete fix on the merits: a shared `physical_edges` module feeds both the plain-run layer computation and the `UnifiedDag` executor, so the two schedulers cannot drift — which was the second half of what I flagged here, and which this branch only ever addressed by changing two call sites in parallel.

## What this branch had that is worth not losing

Nothing that #1352 lacks, as far as I can tell. The one thing I would keep an eye on is the same-pipeline/cross-pipeline asymmetry: this branch went through `resolve::classify_table_ref`, which is per-`Project`, and needed a second mechanism for `--dag`. #1352 starting from one shared derivation is precisely why it does not have that problem.

This branch was also 26 commits behind and conflicting on rebase, with a history of being 13/13 green while carrying a compile-breaking regression. There is no version of reviving it that is cheaper than what already landed.

PR #1307, opened 2026-07-31
fix(engine/rocky-cli): load models below the first subdirectory level

### LINKED ISSUE #1262: Models below the first subdirectory level are never loaded, despite the default models/** glob

hugocorreia90:
fix(engine/rocky-cli): load models below the first subdirectory level

`load_project_models` read the models root plus exactly one level of
immediate subdirectories. A model at `models/a/b/` was never loaded — not
an error, simply absent — even though the default transformation glob is
`models/**`, which reads as fully recursive. Replacing that model's `.toml`
with malformed TOML changed nothing, because the file was never opened and
so could not even fail. Every command on this loader inherited it: `dag`,
`list`, `docs`, `preview`, `compliance`, `estimate`, `tick`, `validate`,
`branch`, `scope`, and `run --dag`, which built nothing for the deeper
subtree and reported success.

The walk is now a pre-order depth-first traversal of the whole tree, over an
explicit stack rather than recursion so a deep tree cannot overflow. It keeps
the per-directory decomposition, so `_defaults.toml`, `groups/` and named
test definitions still resolve beside the models they apply to. Children are
sorted at every level, so discovery order no longer depends on `read_dir`.
Symlinked subdirectories are still followed, as they always were, and a cycle
is broken by entering each directory at most once keyed on its resolved
identity. Nesting past a 64-level ceiling is reported rather than silently
truncated.

Nothing new is excluded: hidden directories were never filtered out and are
not filtered out now — adding an exclusion here would be a fresh silent drop
of exactly the kind this closes.

The compile path stays flat and is left alone deliberately. `rocky run`
without `--dag` loads through `Project::load_models_with_db`, which reads a
single directory, so it still sees only the models sitting directly in the
root while `rocky list models` and `rocky dag` now see the whole tree.
Widening it materializes tables that projects have never had built, which is
a change that deserves its own PR and its own release note. `rocky test
--declarative` keeps its own shallower loader for the same reason.

Canary: all 112 in-repo projects produce identical `list models`, `dag` and
`validate` output before and after, and the credential-free POCs run
end-to-end with identical logs and exit codes. No in-repo project nests
models more than one level.

Closes #1262

---

**Review question worth settling before merge:** this leaves the compile path
flat on purpose, so after it `rocky list models` and `rocky dag` see the whole
tree while `rocky run` without `--dag` still sees only the root. That is a
smaller blast radius, but it means `list models` can show a model `rocky run`
will not build. Either that asymmetry is acceptable and documented, or the
compile path should follow — worth a deliberate call rather than inheriting it.

Closes #1262

### DIFF
--- a/docs/src/content/docs/advanced/troubleshooting.md
+++ b/docs/src/content/docs/advanced/troubleshooting.md
@@ -16,7 +16,7 @@ A model references another model that doesn't exist in the project.
 **Possible causes:**
 - The referenced model file is missing or misnamed
 - The `name` field in the `.toml` sidecar doesn't match the file name
-- The model is in a subdirectory that Rocky isn't scanning
+- The model is in a subdirectory. `rocky list models` and `rocky dag` walk the whole tree under `models/`, but the compiler reads only the models directory itself, so `rocky compile` treats `models/staging/stg_orders.sql` as absent
 
 **Fix:** Check that the referenced model exists in `models/` and its `name` field matches. Rocky auto-discovers dependencies from SQL table references; bare names matching model file names become DAG edges.
 
--- a/docs/src/content/docs/reference/commands/development.md
+++ b/docs/src/content/docs/reference/commands/development.md
@@ -543,7 +543,7 @@ rocky list deps <model>      # What this model depends on
 rocky list consumers <model> # What depends on this model
 ```
 
-All subcommands support `--output json` via `-o json`. Models are discovered from the `models/` directory (and immediate subdirectories for the common `models/{layer}/` layout).
+All subcommands support `--output json` via `-o json`. Models are discovered from the `models/` directory and every directory beneath it, so both `models/{layer}/` and `models/marts/core/` are picked up. Each directory is read on its own, so a per-directory `_defaults.toml` applies to the models beside it. Nesting deeper than 64 levels is reported as an error rather than skipped.
 
 See the [CLI Reference](/reference/cli/#rocky-list) for full examples and JSON output schemas.
 - [`rocky validate`](/reference/commands/core-pipeline/#rocky-validate) -- validate config after registering the adapter
--- a/engine/crates/rocky-cli/src/api.rs
+++ b/engine/crates/rocky-cli/src/api.rs
@@ -1926,7 +1926,7 @@ mod tests {
     /// `--column-lineage` must not fail on the ordinary nested layout: staging
     /// models one level down feeding marts at the root.
     ///
-    /// The model loader reads the root plus one level below; the lineage compile
+    /// The model loader walks the whole tree below the root; the lineage compile
     /// reads only the root. So `fct` is compiled while the `stg` it selects from
     /// is invisible, and the compiler reports `unknown dependency 'stg'`. That
     /// is an artifact of the shallower read, not a defect in the project.
--- a/engine/crates/rocky-cli/src/commands/branch.rs
+++ b/engine/crates/rocky-cli/src/commands/branch.rs
@@ -1068,8 +1068,8 @@ async fn discover_replication_branch_targets(
 
 /// Transformation pipeline path for [`discover_branch_targets`].
 ///
-/// Walks the model files under the pipeline's `models` glob (top-level +
-/// immediate subdirectories — same surface `rocky list models` and the
+/// Walks the model files under the pipeline's `models` glob (the root and
+/// every directory below it — same surface `rocky list models` and the
 /// catalog scope resolver use) and emits one `(prod, branch_source)` pair
 /// per model. Each model's sidecar `[target]` block supplies the production
 /// catalog/schema/table; the branch source rewrites the schema to the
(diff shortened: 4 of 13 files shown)


hugocorreia90:
Holding as draft rather than merging on green, for two reasons.

**1. The authoring agent is still iterating.** Its worktree has moved past this PR's head twice since I opened it — most recently `fa87d740` → `f654a104`, +102/−27 with 113 lines in `models_loader.rs`. A committed worktree is not a finished one.

**2. Its sibling PR (#1306) was 13/13 green with a confirmed compile-breaking regression in it.** Two independent reviewers found 13 findings there that full-suite green, six non-vacuity reds, and an already-addressed Codex round had all passed. This PR has not had an independent review yet, and the same reviewers are queued behind their build agents.

Note also that #1306 and this PR **overlap on `run_dag_exec.rs`** — whichever merges second needs a rebase and a full re-gate, not a stale-base green tick. That is this repo's signature cross-PR break.

The open design question from the PR body still stands: this leaves the compile path flat, so `list models` / `dag` see the whole tree while `rocky run` without `--dag` sees only the root — `list models` can show a model `rocky run` will not build.

hugocorreia90:
## ⛔ Two independent reviews: **BLOCK** and **FIX_FIRST** — not mergeable

Both reviewers reproduced their findings empirically. Four HIGH, all CONFIRMED. The common root is the one thing this PR deliberately left alone: **discovery went recursive while the compile path stayed flat.**

### H1 — `run --dag` goes from deterministic exit 0 to **non-deterministic exit 1**, and takes a previously-green model with it

The widened walk (`models_loader.rs:187`) makes the unified DAG emit a node for the nested model. `run_dag_exec.rs:81-96` dispatches every same-layer node as its **own concurrent sub-run**, each opening the single redb writer at `models/.rocky-state.redb`. The loser of that race dies with *"state store … is locked by another process."*

Measured on 6 freshly-created project dirs with `models/top.*` + `models/a/b/lvl2.*`:

```
lvl2 = failed 6/6   ("failed to open state store … locked by another process")
top  = failed 3/6   ← previously green
exit 1              ("DAG execution had 1 failed node(s)")
```

The residual-risk list predicts the wrong failure mode — it is not `model 'lvl2' not found`.

### H2 — the reported bug is not closed end to end

A deep model becomes a DAG node, but its sub-run **compiles only `models/` flat**, fails at `run.rs:7167-7171` with `model 'deep' not found`, and the DAG exits with a failed node instead of materialising it. The asymmetry flagged in this PR's description is not merely a reporting inconsistency; it is a functional break.

### H3 — `branch promote` gains a fail-open validation/execution mismatch

The breaking-change gate compiles the **flat root** before recursive discovery widens the execution set. With only `models/a/b/orders.sql` present, the gate's root compile returns `NoModels`, `evaluate_breaking_change_gate` records *"gate skipped"* and returns `None` — and then recursive discovery plans and executes the deep model's production replacement **unchecked**, with no `--allow-breaking`.

### H4 — the E036 rebuttal does not hold, and the consequence is data loss

I verified the author's rebuttal as far as it went — `discover_transformation_branch_targets` genuinely never compiles, and the promote-path compile at `branch.rs:1459` is gate-only with skip-on-error. **That was not far enough.** The reviewer establishes that the promote path *does* reach compile and **E036 is returned inside an `Ok(CompileResult)` whose `has_errors` is ignored**.

So the bypass is not merely pre-existing-and-untouched. Before this commit, a deep colliding model was never planned or promoted at all — E036 never had a chance to see it. After it, a root model `current` and `models/a/b/replacement` targeting the same physical table both reach execution as `CREATE OR REPLACE`, and **the later branch source overwrites the first.**

That is the difference between "an unvalidated path exists" and "an unvalidated path now receives inputs that silently destroy data."

---

**Disposition:** the recursive walk itself is well-built — deterministic ordering, cycle-safe, depth-ceilinged, with unusually strong evidence (112-project discovery-parity and 49-POC execution-parity canaries, both dual-binary, zero model-set differences). The defect is not in the walk. It is that **three consumers assume the flat model set** and were not widened with it. Closing #1262 means deciding what the compile path does, not just what the loader does.

hugocorreia90:
Superseded by #1328 (`babb6eea`), which is on `main` and closed #1262.

This PR widened the CLI loader alone. Two independent reviews returned four CONFIRMED HIGHs against it, every one traceable to the same shape: `models/**` was read by **five** different scanners — the CLI loader, the compile path, the three sidecar loaders, the promote gate's compile, and preview's scan — and widening one while the others stayed flat produced inconsistencies worse than the original gap.

#1328 routes all of them through one walk instead. Closing this rather than rebasing it; the branch stays if any of its intermediate reasoning is worth reading.

PR #1314, opened 2026-08-01
fix(engine/rocky-core): retry a transient holder of the state writer lock

### LINKED ISSUE #1234: flaky: two rocky-cli tests fail under CPU load with cargo test (pre-existing)
## Description

Two `rocky-cli` unit tests fail intermittently under `cargo test` when the machine is loaded. They
are **pre-existing** — not introduced by any recent change — and both pass reliably on an idle
machine, which is why they have not been caught.

- `commands::run::tests::auto_sweep_within_interval_skips_without_updating_stamp`
- `commands::discover::tests::new_source_diff_reports_only_unseen`

## Evidence

Measured while reviewing #1226. Two trees: the PR branch, and its own base commit `e88b2acb` with
**no** PR changes applied.

| condition | base `e88b2acb` | PR branch |
|---|---|---|
| idle | 9/9 pass | 8/8 pass |
| 8 CPU spinners | **2/6 fail** | **2/6 fail** |

Identical failure rate and identical pattern on both trees, so the flake is independent of the
change under review. My first (idle-only) baseline made it look like a regression; it isn't.

## Reproduce

```bash
cd engine
for c in $(seq 1 8); do (while :; do :; done) & done
for i in $(seq 1 6); do cargo test -p rocky-cli --lib 2>&1 | grep "^test result"; done
kill %1 %2 %3 %4 %5 %6 %7 %8
```

## Notes

- `cargo nextest` (what CI runs) forks a process per test and does not reproduce this, so CI stays
  green — the exposure is local `cargo test`, which `engine/AGENTS.md` lists as the everyday loop.
- `auto_sweep_within_interval_skips_without_updating_stamp` uses only a `TempDir` + redb and no
  sleeps, so plain wall-clock sensitivity does not obviously explain it. Root cause not diagnosed.

## Suggested direction

Diagnose under load rather than idle. If the cause is scheduling sensitivity, pin the tested
interval/clock explicitly rather than relying on elapsed real time.

hugocorreia90:
## Summary

Closes #1234.

That issue is filed as two flaky `rocky-cli` tests. It is not a test defect — **the same fault silently skips end-of-run retention in production**, and the issue's suggested direction (pin the interval/clock) does not address it, because neither test is time-sensitive and neither fails on its assertion.

## Root cause

`StateStore::open` acquires its advisory writer lock with a **single, zero-retry** `flock`, so any momentary holder becomes a hard `LockHeldByOther`. The redb open *immediately below it* already retries transient contention on a 5×50 ms budget — the advisory acquire was the only unprotected step in `open_inner`.

The holder is momentary far more often than it looks. A lock on a file this process itself created can be briefly contended when another **thread** of the same binary spawns a child process, because the descriptor is inherited across the fork window. That is exactly why `cargo test` (threads in one process) exposes it and `cargo nextest` (a process per test) does not — which is why CI has stayed green.

**Not test-only.** `run`'s end-of-run auto-sweep opens the store the same way and swallows the error as a `tracing::warn`, so retention silently does not happen.

## Subproject(s) touched

## Changes

A bounded retry on the advisory acquire, on the same budget as the redb open beside it, scoped to `StateStore::open_inner` only.

- The retry re-attempts the lock on the **already-open file** rather than reopening it, so a genuine external holder is not raced against.
- Only a *contended* lock retries; an I/O error is terminal.
- A holder that never releases still fails after the budget, so **cross-process exclusion is unchanged** — `second_writer_fails_when_lock_held_externally` passes untouched.
- The public `try_acquire_writer_lock` keeps its single-attempt, non-blocking contract. `state_sync`'s download path and the apply seam deliberately rely on failing fast rather than blocking, so widening it there would change behaviour they depend on.

`try_acquire_writer_lock` is split into `open_lock_file` + `lock_once` so both the single-attempt and retrying paths share one implementation rather than duplicating the open.

## Test Plan

**The issue's own reproduction, with a load-matched baseline** — the measurement that actually settles a flake:

| | 12 runs, 8 CPU spinners, `--test-threads=16` |
|---|---|
| pre-fix | **4 failed** |
| post-fix | **0 failed** |

Both binaries built from this tree; only the acquire differs. Every pre-fix failure is one of the two tests #1234 names (`new_source_diff_reports_only_unseen` in all four, `auto_sweep_within_interval_skips_without_updating_stamp` in one), and the panic is `reopen: LockHeldByOther` — the predicted mechanism, not a coincidental failure. The spawn-heavy `commands::ci_diff::` module is included in the run because it is what amplifies the contention.

New unit test `open_retries_the_advisory_lock_and_succeeds_after_brief_hold`, and it is **deliberately deterministic**. The first version slept a fixed 100 ms against the 5×50 ms budget and hoped the release landed inside it — under load both stretch, so the test would itself have been load-sensitive, which is the defect class this PR fixes. It now follows the pattern the crate already established for the redb-open retry: a thread-local observer the holder polls, releasing the moment the opener provably enters the retry loop. No wall-clock racing. (That hook's own docstring says to do exactly this "instead of sleeping a fixed duration and racing the retry budget".)

**Mutation-checked** — reverting to the single attempt fails it with `opener never hit lock contention — it did not retry`.

Gates:

- `cargo nextest run --all-features` — full suite green.

## Deliberately not changed

The auto-sweep's `warn`-and-skip on a lock error stays. With the retry, the transient case no longer reaches it; what remains is *genuine* contention — another writer holds the lock — where skipping the sweep is reasonable, since that writer will run it. Turning it into a hard failure is a separate behaviour decision and not needed to close this issue.

## Checklist

### Engine (`engine/`)


hugocorreia90:

Per `AGENTS.md`, reviewed by **Codex (GPT-5)** via `codex:codex-rescue`, against `f42be11c`.

**Verdict: BLOCK on test reliability only; implementation semantics sound.** Its single blocking finding was already fixed in `6b0a5588`, pushed before the review returned — and it independently prescribed the same remedy I had applied, which is a useful confirmation rather than a coincidence.

### The blocking finding — already resolved

> load-sensitive fixed-sleep test … Synchronize release to an observed first `WouldBlock`; **do not merely widen the sleep margin**.

Correct, and it named the sharper version of the risk than I had: the release thread's 100 ms sleep only *starts* when that thread is scheduled, while the opener has just four 50 ms sleeps before exhaustion — so the two are not merely both stretching, they can stretch independently.

`6b0a5588` replaces it with the observer pattern the crate already uses for the redb-open retry: the holder polls a shared counter and releases the moment the opener provably enters the retry loop. No wall-clock racing, and no widened margin. Still mutation-checked — reverting the fix fails it with `opener never hit lock contention — it did not retry`.

Writing a load-sensitive test inside a PR that fixes load-sensitivity was an own-goal; the crate's own observer docstring already said not to do it.

1. **Exclusion is not weakened.** The lock is still taken before redb; five failed attempts still return `LockHeldByOther`. Reusing the fd cannot create two owners — it is retried only after `WouldBlock`, and independent same-process opens stay exclusive.
2. **No caller now blocks.** Public `try_acquire_writer_lock` remains single-attempt with no sleep. `state_sync` retains its own async retry wrapper; the apply path's only direct occurrence is a test, and its production ledger path already deliberately retries `StateStore::open`. It also checked the deadline question I raised specifically: `run_watch` has no deadline, the scheduler explicitly handles bounded contention, and `serve` sweeps on the blocking pool without a request deadline.
3. **Error classification is right.** Only `WouldBlock` → retryable `LockHeldByOther`; every other `fs4` failure stays terminal `LockIo`, so a permanent fault is not blindly retried. With the attempt count fixed at five, the terminal `unreachable!` is genuinely unreachable.
5. **The auto-sweep's warn-and-skip stays defensible** — and it added a detail I had not: the sweep leaves its timestamp **unstamped** on failure, so a later run retries it, and the manual sweep escape hatch remains. It defers cleanup rather than hiding the failure.

Its stated open question — the practical frequency of the old test's flake — is moot now that the timing dependency is gone entirely.

Issue #1310, opened 2026-08-01
bug(engine): branch promote executes colliding duplicate-target models — last CREATE OR REPLACE wins, E036 never consulted


hugocorreia90:
## Description

`rocky branch promote` executes colliding duplicate-target models back to back — the later `CREATE OR REPLACE` silently overwrites the earlier one, and E036 (the duplicate-target diagnostic #1303 added) is never consulted anywhere on the path.

**This is live on main today at the root level.** Two top-level models targeting the same physical table both get planned and both get executed; whichever the loader ordered last wins. No error, no warning, exit 0.

## Root cause

The promote path never compiles for discovery, and never checks compile diagnostics at all:

- Discovery loads raw models: `branch.rs:~1140` — `let all_models = crate::models_loader::load_project_models(&models_dir)?;` — then plans one promote step per non-ephemeral model. No compile, no collision check.
- The **only** compile on the promote path is the breaking-change gate (`branch.rs:1459`), which consumes only its own comparison fields and **gate-skips on error** (`push_skip(...); return None`).
- `grep -n has_errors branch.rs` → **zero hits**. #1303 shipped E036 as an error-severity diagnostic inside `Ok(CompileResult)`; nothing destructive ever reads it.
- Execution (`branch.rs:~800`) iterates the planned steps and runs each statement: `for step in targets { adapter.execute_statement(&step.statement) }`.

## Reproduce (depth 0, main)

```toml
# models/current.toml and models/replacement.toml — both target wh.main.orders
```

`rocky branch promote …` → two promote steps, two `CREATE OR REPLACE` against `wh.main.orders`, last one wins. `rocky compile` on the same tree reports E036 — but promote neither runs it nor reads it.

## Why now

Found during independent review of PR #1307 (recursive model loading, #1262). That review framed it as "recursion widens the bypass" — deep collisions previously invisible to discovery would start reaching promotion. Verifying the claim showed the bypass is **already live at depth 0**; recursion only enlarges the input population. Both of that PR's independent red-team rounds converged on the same remedy without seeing each other:

> a shared collision-validation contract for destructive consumers of a raw model set — `gc`, `apply`, `branch promote`.

## Suggested direction

Destructive consumers must not act on a raw `load_project_models` result. Either:
- compile and **fail closed on `has_errors`** before planning (promote is not latency-sensitive), or
- run the duplicate-target check directly on the loaded set (the #1303 check is already factored) and refuse with the E036 text.

[...]

Issue #1288, opened 2026-07-30
bug(engine): `--parallel` does not bound DAG node fan-out and means something different under `--dag`


hugocorreia90:
## Description

`--parallel` means two different things depending on whether `--dag` is set, and under `--dag` it does not bound the thing that actually fans out.

`rocky run --parallel N` documents itself as "concurrency for warehouse-bound work (default 4; pass `--parallel 1` to force fully serial)". On the non-DAG path that holds, and it is deliberately never multiplicative — `commands/run.rs::execute_models` excludes the two special strategies from concurrent model execution:

```rust
// The two special paths (`content_addressed`, `time_interval`) stay serial —
// they push to `output` inline and `time_interval` already self-parallelizes
// per-partition. A layer mixing strategies runs entirely serially.
let can_parallelize = requested_parallel > 1
    && adapter_supports_concurrency
    && !matched.is_empty()
    && matched.iter().all(|(_, _, model)| is_plain_strategy(model));
```

So a non-DAG run gives you *either* N concurrent plain models *or* N concurrent partitions inside a serially-executed `time_interval` model. The ceiling on concurrent warehouse queries is N.

Under `--dag` there is no such guard. `commands/run_dag_exec.rs::run_with_dag` builds the executor with

```rust
let executor = DagExecutor::new(dispatcher);
```

and never calls `DagExecutor::with_max_concurrency`, so `max_concurrency` stays `None` and `dag_executor.rs` falls back to `Semaphore::MAX_PERMITS` — every node in a layer runs at once. `with_max_concurrency` is currently referenced only from `benches/dag_execute.rs`.

Two consequences:

1. `rocky run --dag --parallel 1` does **not** force serial execution. Node fan-out ignores the flag entirely.
2. If the flag were also honored per sub-run, the result would be *unbounded nodes × N partitions* — exactly the multiplication `execute_models` refuses to do.

#1283 asked for `--parallel` to "retain [its] normal `rocky run` semantics" under `--dag`. The partition-selection half of #1283 is fixed in #1290, which deliberately pins per-sub-run concurrency to 1 rather than threading the flag, precisely to avoid (2). That leaves (1) open, and it is what this issue tracks.

## Expected behavior

`--parallel N` should mean one thing on both paths. The natural reading for `--dag` is that it bounds node fan-out — `DagExecutor::with_max_concurrency(n)` — since the DAG already parallelizes at the node level and each sub-run is one pipeline (or one model).

## Why it wasn't done in #1288

[...]

PR #1318, opened 2026-08-01
fix(engine/rocky-cli): take turns on the state file instead of racing it under run --dag

### LINKED ISSUE #1312: bug(engine): run --dag fails deterministically for any pipeline with two independent models — same-layer sub-runs race the zero-retry writer flock
## Description

`rocky run --dag` fails **deterministically** for any transformation pipeline containing two or more independent (same-layer) models. This is the most basic multi-model shape there is, and it fails on the stock released binary with two `SELECT 1` models.

## Reproduce (stock `rocky 1.68.0`)

```toml
# rocky.toml
[adapter]
type = "duckdb"
path = "probe.duckdb"

[pipeline.probe]
type = "transformation"
models = "models"

[pipeline.probe.target.governance]
auto_create_schemas = true
```

`models/alpha.sql` and `models/beta.sql` each `SELECT 1`, sidecars with `full_refresh` + a `[target]`, **no `depends_on`**. Then `rocky run --dag`.

## Measured matrix (pre-registered controls, 12 hot runs + 4 controls)

| Shape | Result |
|---|---|
| 2 independent models, `--dag`, `execution.concurrency = 4` | **6/6 exit 1** |
| 2 independent models, `--dag`, no execution block (defaults) | **3/3 exit 1** |
| 2 independent models, `--dag`, `execution.concurrency = 1` | **3/3 exit 1** |
| 1 model, `--dag` | exit 0 |
| 2 **chained** models (`depends_on` → two layers), `--dag` | 3/3 exit 0 |
| 2 independent models, plain `rocky run` | exit 0 |

Three things the matrix pins:

1. **`execution.concurrency` does not govern DAG node dispatch** — `concurrency = 1` still fails, so there is no config workaround short of adding fake `depends_on` edges.
2. Exactly **one** node fails per run (the other succeeds), so this is concurrent same-layer sub-runs racing, not the outer process holding the lock.
3. Layering serializes and passes — which is why chained-model projects never see it.

## Mechanism

Each same-layer node is dispatched as its own concurrent sub-run; every sub-run opens the state store as a **writer**. The loser hits the exclusive advisory flock, which has **zero retry** (`try_acquire_writer_lock`, `rocky-core/src/state.rs:564-576` — single `FileExt::try_lock`, contrast the redb open right below it which retries 5×50ms):

```
"state store at models/.rocky-state.redb is locked by another process;
 only one rocky writer can run at a time against the same state file"
→ node failed → "DAG execution had 1 failed node(s)" → exit 1
```

The same log line also shows `auto-sweep: failed to open state store; skipping` — the silent retention-skip production reach documented on #1234. Same root defect, louder symptom.

## Why it stayed latent

- Every in-repo project that exercises `--dag` has chained models (layered → serialized) or a single model per pipeline — the 112-project / 49-POC parity canaries run clean because **no fixture anywhere is two independent models under `--dag`**.
- `dagster-rocky` materializes per-asset via `rocky run --model X`, never `--dag`.
- Plain `rocky run` is monolithic (one writer) and unaffected.

The per-model sub-run dispatch dates to the #1140 fix era; this failure shape appears to have shipped with it.

## Fix direction

A short flock retry (the natural #1234 fix, ~250ms) is **not sufficient here**: real models run seconds to minutes, and the loser's retry budget expires while the winner's sub-run legitimately holds the lock for its whole duration. The DAG executor needs one of:

- serialize sub-run dispatch **per state path** (nodes sharing a state file queue; genuinely independent state files still parallelize),
- a shared writer handle threaded through same-path sub-runs, or
- a wait-until-free lock acquisition for DAG sub-runs, bounded by heartbeat rather than a fixed budget.

`#1234`'s zero-retry flock is the shared root; its short-retry fix helps the transient class (auto-sweep, read-mostly collisions) but not this one.

## Relations

- Found while running the discriminating experiment from PR #1307's review: #1262's recursion makes nested layouts produce this ≥2-independent-node shape (measured 6/6 there too) — but the flat case **already fails today**, so this is not a #1262 regression risk; it is a standing defect that gates #1262's v2.
- #1234 — same zero-retry flock, transient-symptom side.
- #1140 — the dispatch design this dates to.

hugocorreia90:
Closes #1312.

## The bug

Same-layer DAG nodes are dispatched concurrently with no bound, and every pipeline-bound sub-run drives `run()`, which holds the state store's exclusive writer flock for its whole duration. With two or more independent models in one pipeline, the first sub-run to open the store won and every sibling sharing the state file died on `LockHeldByOther` — deterministically, on a project as small as two `SELECT 1` models. Measured **12/12 exit 1** on stock 1.68.0 across `execution.concurrency` 4 / default / **1** (the knob never governed DAG node dispatch). Chained models dodged it only because layering serialized them — which is why every pre-existing `--dag` fixture passed.

## The fix

A flock retry cannot absorb this: its budget is milliseconds and the winner legitimately holds the lock for as long as its models take (which is why this is complementary to #1314, not competing). Instead the dispatcher owns a **turnstile keyed by state path**: each pipeline-bound node takes the exclusive turn for its state file before its sub-run starts and holds it for the sub-run's duration — exactly as long as `run()` holds the flock.

Keyed by path because *that* is the invariant — one writer per state file — not "one sub-run at a time": nodes writing genuinely distinct state files still overlap, and a rendezvous test pins that so the turnstile can never quietly become a global serializer. Seeds bypass it (`run_seed` never opens the store). The outer `run_with_dag` process opens no store around execution — verified — so every open on the DAG path happens inside a sub-run, and the turnstile covers all of them.

Today the dispatcher threads one canonical path to every pipeline-bound node, so those nodes execute one at a time — strictly better than what it replaces: before this gate, "concurrency" between such nodes meant one ran and the rest crashed. Bounding node fan-out properly is #1288's territory; this PR is the correctness floor under it.

## Evidence

- **Dispatch-seam invariant**, mutation-checked: an in-flight high-water probe over two same-path nodes asserts the mark never exceeds 1. Neutering the `turn()` acquisition fails it with the mark at 2, while the distinct-paths rendezvous still passes — red in exactly one direction.
- **Over-serialization guard**: two dispatchers sharing one turnstile with distinct paths must overlap; a 2-party barrier deadlocks (→ timeout failure) if cross-path serialization ever appears.
- **End-to-end**: a new real-binary test drives the #1312 fixture twice — create-state and reopen-state rounds — asserting exit 0, `failed == 0`, `completed == 2`.
- **The exact fixture that failed 12/12 on stock 1.68.0 passes 8/8** against this branch across all three concurrency variants.
- Gates: fmt clean · clippy `--all-targets --all-features -D warnings` clean · nextest **5829/5829** (108 skipped, live-infra).

## Residual, stated outright

A project whose pipelines declare **distinct models targeting the same physical table** previously had one `--dag` node crash on the lock — an accident of this very bug, since the crash also hit healthy projects. Both now build, last write wins. That is **exactly what the primary path already does** (`rocky run --pipeline` per pipeline: both exit 0, last write wins — verified empirically on stock), so the DAG converges on standing behavior instead of crashing by accident. Duplicate-target *enforcement* is #1310 (evidence posted there); the DAG's duplicate-*name* refusal is unaffected.

Independent red-team review (Codex) in flight; merge held for it plus green CI.


hugocorreia90:

**Reviewer: Codex / GPT-5.6-Terra**, adversarial pass over the diff + worktree. **Verdict: FIX_FIRST.** The mechanism cleared every hard attack — its words: *"no deadlock, cancellation leak, bypass, test race, or #1314 conflict was found"* — with zero findings on the turnstile itself, coverage of the dispatch arms, and the #1314 interaction. Two findings survived, one of which I **refuted empirically** rather than fixed; the other is fixed in this PR.

### Finding 1 — "the wildcard scope removes valid quality/snapshot concurrency" → **REFUTED by experiment**

The claim: gating *all* pipeline-bound kinds (not just transformations) forfeits legitimate same-layer overlap for quality/snapshot nodes.

The discriminating experiment (two independent quality pipelines, shared state path):

| binary | result |
|---|---|
| stock 1.68.0, tables present | **4/4 exit 1** — both checks finish their warehouse work ("quality pipeline complete … in 8ms/9ms"), then one node dies at the end-of-run state-store open |
| stock 1.68.0, empty database | **3/3 exit 1** — same crash; the store open happens regardless of check outcomes |
| this branch | **2 completed / 0 failed, exit 0** |

So quality nodes *do* overlap on warehouse work — and the overlap always ended with a node failing at the store. There was no valid concurrency to remove; there was concurrency that always crashed. **Narrowing the turnstile to transformation nodes would have left every other pipeline-bound kind broken.** The wildcard scope stands, and a new e2e (`run_dag_two_independent_quality_pipelines_both_complete`) pins the non-transformation half so the scope is load-bearing, not accidental.

### Finding 2 — the published `layer` contract now makes a false claim → **CONFIRMED, fixed**

`DagRunNodeOutput.layer` was documented as *"nodes in the same layer ran concurrently"* — false after this PR for nodes sharing a state file (and false-in-outcome before it, when those nodes crashed). Reworded to what is true: a layer bounds **ordering, not wall-clock overlap**; seed nodes and distinct-state-file nodes still run concurrently. Doc-comment change on an `*Output` struct ⇒ the codegen cascade (schemas / SDK / TS descriptions) rides in this PR.

### The reviewer's residual, answered

Its stated least-confident point was the operational severity of serializing local quality/snapshot work. The experiment answers it: the alternative to serialization was not overlap, it was failure — throughput regression relative to a crash is not a regression. Proper node-fan-out control remains #1288's territory.

Also from the stop-gate review round: a cross-pipeline duplicate-target project under `--dag` now converges on the primary path's standing last-write-wins behavior instead of crashing by accident — measured three-leg on stock, documented in the CHANGELOG residual, evidence posted to #1310.

Issue #1255, opened 2026-07-27
soak-verdict G3 reports PASS on a demonstrably leaking build — the RSS gate cannot see a ~12 MB/24h leak


hugocorreia90:
`just soak-verdict` returned `PASS` for a 24h run whose memory curve is statistically indistinguishable from the run that originally *found* the JobRegistry leak. The gate is not wrong about anything it measures — it simply cannot resolve a leak of this size.

## The gate

`scripts/soak_verdict.py:356-379`, gate **G3**:

```python
if growth_pct > 10 and delta_mb > 20:   # :373
```

The conjunction is deliberate — the comment says it exists "so drift on a small baseline cannot trip it" — but the 20 MB absolute floor is above the magnitude of the leaks this harness is actually being used to hunt.

## Measured

| run | growth | delta | G3 | reality |
|---|---|---|---|---|
| `.soak/soak-jobcache-20260726-003634` | 40.45% | 11.16 MB | **PASS** | grows ~482 KB/h through the final quarter |
| `.soak/soak-20260724-171021` (the run that *found* the leak) | 45.0% | 11.77 MB | **would PASS** | the leak that motivated #1232 |

Both clear the percentage test comfortably and neither reaches the absolute one, so the conjunction never fires. A gate that passes the very run used as the leak's existence proof is not providing evidence.

## Why the percentage alone is not the fix either

Raising or splitting the thresholds is not sufficient, because a *correctly fixed* build is also expected to show large early growth: the bounded cache fills once (job 512, ≈8h32m at 1 job/min), and that one-time fill is not a leak. A naive percentage gate would fail a healthy run.

The discriminating measurement — pre-registered for this soak before it started — is the **final quarter (18–24h)**, where a bounded structure must be flat. Flat, not declining: freed entries return to allocator free lists rather than the OS, so RSS holding at high-water is a pass.

Equivalently, and more robustly, the **inflection at the fill point** is the signal. For this run:

```
pre-fill  (2.0–8.53h):  +506.9 KB/h
post-fill (8.53–24h):   +481.9 KB/h     → no inflection
```

## Suggested change

Fold the final-quarter rule into G3 rather than replacing the existing conjunction:

- keep the current percentage-and-absolute check as a coarse catastrophic-growth trip;
- add a final-quarter gate: least-squares RSS slope over the last 25% of the run, failing above a threshold on the order of ~100 KB/h, with a documented noise band;
[...]

PR #1319, opened 2026-08-01
fix(scripts): fail the soak verdict closed when the RSS growth gate has no input

hugocorreia90:
## Summary

Closes the last residual of #1255.

G3b (the final-quarter RSS slope gate) and G4 (fd) both guard their inputs with `is not None` and emit an explicit `INVALID` when a window is empty — #1313 added exactly that for fd, on the reasoning that *"a verdict that never weighed the fd gate is not a soak PASS."*

**G3a still guarded on truthiness and had no `else`.** `median` returns `None` for an empty window, so a run missing baseline or final-hour `rss_kb` readings skipped the RSS growth gate in silence and could still report `PASS`. G3a was the last gate that could quietly not run.

This is the same defect class #1313 closed, on the gate immediately above it.

## Subproject(s) touched

## Changes

`is not None` instead of truthiness, plus an `INVALID` finding naming which window was empty — mirroring G4's wording and structure so the two read the same way.

Worth noting the truthiness form was not merely stylistic: an RSS median is never legitimately `0`, so the only value it could suppress is `None` — the missing-data case. It silently converted "could not measure" into "nothing to report".

**But it was also load-bearing by accident, which the widening exposed.** `is not None` admits `base_rss == 0`, and the growth computation divides by it. Truthiness had been rejecting `0` incidentally; removing it introduced a `ZeroDivisionError` on exactly the bad-reading class this guard exists for. A zero baseline is therefore treated as INVALID in its own arm, with its own message — no live process has RSS 0, so a 0 is a bad sampler reading, and the gate must report that rather than raise.

## Test Plan

New self-test fixture `rss outage across the tail window is INVALID`, cut from the **sample** timeline rather than `planned_end` — the same correction the fd fixture documents, since nulling from `planned_end - 3600` leaves one live reading at the window edge whose median then feeds the gate normally.

The existing control (`the 24h fixture itself passes with fd intact`) already guards against the new check passing for an unrelated reason.

Second fixture `a zero RSS baseline is INVALID, not a crash`, which would raise without the zero arm.

**Both mutation-checked** — forcing the `None` guard off makes the self-test exit 1 on the outage check; removing the `== 0` arm makes it raise `ZeroDivisionError` on the zero-baseline check.

`python3 scripts/soak_verdict.py --self-test` — PASS.

## Context on #1255

The rest of that issue is already fixed on `main` and I verified each piece rather than assuming:

- **G3b exists** — final-quarter least-squares RSS slope, evaluated independently of G3a, with a documented calibration block. It is one-sided, which matches the issue's "Flat, not declining: freed entries return to allocator free lists rather than the OS".
- **The report states that a PASS means no gate fired**, which was the issue's closing ask.
- **The window is cut from the sample timeline**, so a run whose readings stop early cannot shrink its own final quarter out of the gate.

With this, I believe #1255 is fully addressed and can close.

## Checklist

### DIFF
--- a/scripts/soak_verdict.py
+++ b/scripts/soak_verdict.py
@@ -484,7 +484,32 @@ def gate_resources(run: dict) -> list[dict]:
     rss_slope = ols_slope_per_hour(
         [(s["t"], float(s["rss_kb"])) for s in samples if s.get("rss_kb") is not None and s["t"] >= warm]
     )
-    if base_rss and tail_rss:
+    # `is not None`, not truthiness — and an explicit INVALID when a window is
+    # empty, matching what G4 does. A missing RSS reading is not a small RSS
+    # reading: `median` returns None for an empty window, so the truthiness
+    # form let a run with no baseline or no final-hour readings skip this gate
+    # in silence and still report PASS. G3b guards its own window the same way;
+    # G3a was the last gate that could quietly not run.
+    if base_rss is None or tail_rss is None:
+        missing = [
+            label
+            for label, value in (
+                ("baseline (first hour after warmup)", base_rss),
+                ("final hour", tail_rss),
+            )
+            if value is None
+        ]
+        findings.append(
+            {
+                "gate": "G3a",
+                "status": "INVALID",
+                "detail": "RSS growth gate could not evaluate — no rss_kb readings "
+                f"in the {' or the '.join(missing)} window; sampling stopped or "
+                "the process was not observable. A verdict that never weighed "
+                "the RSS gate is not a soak PASS.",
+            }
+        )
+    else:
         growth_pct = (tail_rss - base_rss) / base_rss * 100.0
         delta_mb = (tail_rss - base_rss) / 1024.0
         run["metrics"]["rss"] = {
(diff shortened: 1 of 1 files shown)


hugocorreia90:

Per `AGENTS.md`, reviewed by **Codex (GPT-5)**, cross-checked with a second independent model.

**One finding, and it was correct: my new test did not bind the gate it claimed to test.** Fixed in `99da6f16`.

### The finding

The outage fixture asserted only the run **verdict** (`INVALID`). But G3b independently reports `INVALID` for that same fixture — its final quarter is nulled too — so the assertion passed on `origin/main`, without the G3a fix, and proved nothing about G3a.

I verified this directly rather than taking it on trust, running the identical fixture through both versions:

```
origin/main   verdict=INVALID  gates=['G3b:INVALID']
this branch   verdict=INVALID  gates=['G3a:INVALID', 'G3b:INVALID']
```

**And my earlier "mutation-checked" claim was wrong.** That run exited 1 on the zero-baseline check I had added in the same batch — not on the outage check. I attributed a failure to the wrong assertion and reported it as evidence.

Now fixed two ways: the check asserts `"G3a:INVALID" in gates` rather than the verdict, and the mutation is the *real* pre-fix shape — guard present, finding silently not emitted — which names this check explicitly:

```
SELF-TEST FAIL: rss outage makes G3a itself INVALID: expected True, got False
```

1. **The INVALID is load-bearing.** Aggregation prioritises `INVALID` over `FAIL`/`PASS` and maps it to exit 2, so this genuinely changes the verdict rather than adding an ignored finding.
2. **The zero-denominator guard is warranted, and the input is reachable** — it independently confirmed the sampler forwards `ps` RSS without positive-value validation, so `0` is admissible at the boundary. That is the arm I added after realising the widening from truthiness to `is not None` had exposed the division; good to have it corroborated rather than assumed.

Its remaining open question — whether a supported host's `ps` can actually report zero RSS for a live process — is not settleable from the repo, and does not change the handling: an unusable reading should report INVALID either way.

PR #1323, opened 2026-08-01
fix(engine/rocky-cli): refuse to promote models that share one physical table

### LINKED ISSUE #1310: bug(engine): branch promote executes colliding duplicate-target models — last CREATE OR REPLACE wins, E036 never consulted

hugocorreia90:
## Summary

Closes #1310.

`rocky branch promote` plans one `CREATE OR REPLACE` per selected model and executes them back to back. Two models resolving to the same physical `catalog.schema.table` therefore let **execution order decide what a production table ends up holding** — silently, with exit 0.

This is a destructive-verb bypass, and it is a gap in my own #1291 work. E036 does not cover it: `promote` never compiles for discovery, so the duplicate-target diagnostic is not merely *unread* on this path — it is never *produced*. #1291's own changelog entry scoped detection to "everything that compiles one model set", and promote is not one of those.

## Subproject(s) touched

## Changes

`collide_on_target` becomes `pub` on `rocky_compiler::project`, with a doc note saying why: a destructive consumer needs the collision answer over a **raw model set**, without paying for full dependency resolution — which would additionally fail the promote on unrelated unparseable SQL elsewhere in the project, a strictly worse outcome for a verb that is otherwise happy to promote a subset.

`discover_transformation_branch_targets` then refuses before planning any statement, naming both models and the shared target.

**Scoped to the selected set, not the project.** The check runs on the models this invocation would actually promote, so `--filter model=<name>` — the workaround the error message itself suggests — keeps working. A project containing a collision it is not promoting is not blocked.

Ephemeral models are excluded, inherited from `collide_on_target`: they carry a populated but phantom target and materialize nothing, so they are never the second writer.

## Test Plan

Two tests, both in `branch.rs`:

- `promote_refuses_two_models_targeting_one_table` — the refusal.
- `promote_allows_a_filter_that_selects_one_side_of_a_collision` — the escape hatch the error advertises actually works. Written because an error message that suggests a workaround it does not honour is worse than no message.

**Mutation-checked.** Reverting the refusal makes the first test fail and correctly leaves the second passing — so the second is pinning the filter path rather than passively agreeing with the first.

- `cargo nextest run --all-features` — 5,837 passed, 0 failed, 108 skipped.
- `cargo clippy --all-targets --all-features -- -D warnings` — clean.

Re-run after rebasing onto `176bdfb2`, since #1291 and #1321 landed under this branch and a stale-base green is not evidence.

**Whether refusing is right for `promote` specifically, versus last-write-wins being the promote contract.** `rocky run --pipeline` has always let the last write win for this shape, and #1312's changelog says so outright. The argument for diverging here is that promote's whole purpose is publishing to a production location — an order-dependent production table is a different severity than an order-dependent dev rebuild — and that #1291 already chose refusal for the compile path, so tolerating it on the *more* destructive verb would be backwards. But it is a divergence, and someone relying on the old behaviour to promote a "last one wins" pair will now get an error.

## What I am most likely missing

**Other destructive verbs with the same shape.** I checked `promote` because #1310 named it. `gc`, `restore`, `compact` and `archive` operate on targets rather than planning per-model writes, so the collision does not arise the same way — but I have not exhaustively swept every path that plans one statement per model from a raw model set. If there is another, this fix does not reach it, and the `pub` helper is the thing that makes fixing it cheap.

## Checklist


hugocorreia90:
Coordination note — this PR raced #1322, which merged first (`1abae06c`, ~75 minutes after this branch was cut): the same #1310 promote-leg refusal, independently built. Neither session saw the other's in time; the branch here now conflicts on `branch.rs`.

**What #1322 shipped:** a discovery-local refusal — case-insensitive target key, post-`--filter` set semantics (pinned by test: a targeted promote out of a colliding project still works), mutation-checked non-vacuity, nextest 5838/5838.

**What it deliberately left on #1310:** the gc/apply legs, a compile-consulting gate (`has_errors` is still ignored where promote *does* compile), and the cross-set ambiguity when `--filter` selects one owner of a shared target.

**Suggested reconciliation:** if the `rocky-compiler/src/project.rs` change here factors the #1303 duplicate-target check into the reusable shared mechanism (the direction both #1307-review red-teams converged on), rebase this PR onto main and **rescope it to that delta** — it would be the better foundation for the remaining legs, and the discovery-local check in `1abae06c` could then delegate to it. If the project.rs change was only in service of the promote check, this is superseded — close in favor of #1322 and carry anything sharper (test cases, error-text improvements) as a small follow-up.

hugocorreia90:
Superseded by #1322, which I opened earlier for the same issue and which merged as `1abae06c` — I lost track of it across a context boundary and re-implemented the fix on a second branch. Closing this one; #1322's version is on `main`.

The two implementations were equivalent in behaviour. This one reused the compiler's `collide_on_target` rather than detecting collisions inline, and #1322 has the better test set (three cases including a distinct-targets control). The one thing worth keeping from here — sharing the compiler's collision key instead of respelling the fold — is going up separately, together with the apply-seam gap the red team found in the merged version.

Deleting the branch.

PR #1328, opened 2026-08-01
fix(engine): load models below the first subdirectory level — everywhere, through one walk

### LINKED ISSUE #1262: Models below the first subdirectory level are never loaded, despite the default models/** glob

hugocorreia90:
Closes #1262.

## The family, not the instance

`models/**` reads as fully recursive, but **five different scanners** each read the tree their own way: the CLI loader (root + one level), the compile path (root only), the three sidecar loaders (single directory), the promote gate's compile (root only), and preview's scan (root only). A model at `models/a/b/` was absent everywhere — not listed, not compiled, not built — and a malformed sidecar there could not even fail, because nothing ever opened it.

The first attempt (#1307) fixed the loader alone; two independent reviews returned four CONFIRMED HIGHs, every one traceable to widening one scanner while the others stayed flat. This PR fixes the family: **one walk in `rocky-core`, consumed by every scanner.**

| Commit | What |
|---|---|
| walker | `rocky_core::model_walk` — pre-order, sorted children (deterministic order as a contract), symlinks followed as always with cycles entered once on resolved identity, per-entry read errors surfaced not `.flatten()`-dropped, 64-level ceiling that **errors** rather than truncates. Its own semantics tests |
| compiler | `Project` loading walks — strict on walk errors via the existing `ModelError` io conversion (no enum ripple); `has_matching_model_source` walks so a deep-only glob project is not `NoModels` |
| sidecars | `*_from_tree` loaders (surrogate keys / unit tests / column docs), first-wins union so flat projects behave byte-identically; **nine call sites** moved over, incl. the compile-paired `ExecutionExtras` build and rocky-engine's test discovery — pre-PR, a deep model would have **materialized without its surrogate keys** |
| tolerant paths | `optimize` and `tick` bound the partial loader's error half as `_load_errors` and dropped it — both now warn per error (matching apply's established pattern), so incomplete discovery is never reported as clean success |
| e2e | nested `run --dag` on the real binary: deep model discovered → its sub-run compiles the same tree (no `model not found`) → same-layer nodes serialize on the #1312 turnstile → **both models materialize, both rounds** |

## How the v1 review findings land

- **H1 (dag race, measured 6/6):** closed by #1312/#1314 on main; the nested e2e exercises exactly that path green.
- **H2 (deep node's sub-run compiles flat → `model not found`):** closed by the compiler commit; pinned by the e2e.
- **H3 (promote gate compiles flat root while discovery recurses → gate skipped, unchecked promote):** closed **by construction** — the gate's HEAD compile goes through the shared walk, and base-ref extraction was verified already recursive (`git ls-tree -r`, structure-preserving). The remaining "compile failed → gate skipped" fail-open is pre-existing and tracked on #1310, deliberately out of scope here.
- **H4 (E036 widening into promote):** the enforcement legs landed independently (#1322, #1324, #1326's shared `CollisionIdentity`) — a deep collision now meets a refusal, not a silent overwrite.
- Review-flagged siblings **deliberately deferred** (unchanged behavior, tracked on #1262): `preview.rs::ModelDag::scan` (root-only — *shallower than the loader ever was*, a pre-existing gap), the LSP diagnostic scan's `.flatten()`.

## Evidence

- **112-project dual-binary canary — zero diffs**: every in-repo project, `list models` + `validate`, main-built binary vs this branch. No flat project changes behavior. (Known one-sidedness: zero in-repo projects nest ≥2 — which is *why* they're clean; the nested behavior is pinned by the e2e and unit tests instead.)
- Full workspace suite **5863/5863** on the rebased tip (main moved three times during the build; rebased + fully re-gated each time). fmt/clippy clean throughout.
- 14 new tests across walker semantics, compiler depth, sidecar trees, tolerant-path warns, and the end-to-end fixture that failed 6/6 on stock 1.68.0.

## Behavior changes, stated outright

1. Projects with deliberately-unbuilt nested trees under `models/`: those models now load, compile, and (under `models/**`) materialize; a duplicate-target collision with a deep model is now reported/refused where the deep model was previously invisible.
2. The promote breaking-change gate now sees deep models on both sides — a breaking change in one is now caught.
3. Deep sidecars now apply — a nested model gains its surrogate keys / unit tests / column docs.

Independent red-team pass dispatched on this diff; merge held for it plus green CI.


hugocorreia90:
Additional evidence — the same-set guarantee's last unpinned corner, checked empirically (the canary couldn't see it: no in-repo project combines nesting with a narrow glob).

**Filter parity at depth, branch binary, nested fixture (`models/shallow.sql` + `models/deep/nested.sql`):**

| glob | `dag` sees | `run` materializes |
|---|---|---|
| `models/*.sql` (cannot cross a separator) | `shallow` only | `shallow` only |
| `models/**` | both | both |

The CLI loader's per-directory glob matching and the compiler's include-fn **agree at depth in both directions** — the deep model is excluded together under the narrow glob and included together under the recursive one. No surface disagreement, which was the defect family this PR exists to close.

hugocorreia90:

**Reviewer: Codex (GPT-5.x) via the codex plugin**, two passes. The first ran ~75 minutes across every attack surface and wedged before its verdict (task retained server-side; its wrapper correctly refused to fabricate conclusions from a command transcript). A fresh pass, hard-scoped to the two decisive questions, returned **VERDICT: BLOCK — two findings, both CONFIRMED, both now fixed with regression tests.**

### F1 — filter parity broke on malformed metadata in non-participating directories → **fixed** (`0d7437f6`)

With `models/orders*.sql`: a deep directory holding only non-matching sources plus a malformed `_defaults.toml` was **skipped by the CLI matching loader** (the #1305 contract) but **parsed by the compiler** before the filter applied — listing succeeded while compiling failed, the exact surface-disagreement family this PR exists to close. My own parity experiment used clean sidecars only, which is why it couldn't see this. The review also exposed that 2a had quietly made `load_dir_models_matching`'s per-directory pre-check tree-wide — the same asymmetry planted inside the CLI path.

Fix: **one participation rule at both walks** — a directory contributing no model under the active filter has its metadata skipped entirely, identically; the per-directory pre-check is per-directory again. Pinned by the reviewer's exact repro asserting both loaders skip together.

### F2 — the policy gate degraded to defaults when the compile failed → **fixed** (`b85c28b9`)

The #1266 class, recurring despite the standing memory — caught by fresh eyes: `model_attributes` turned ANY compile failure into an empty map, so an attribute-scoped deny (`layer = "gold"`) stopped matching and `default_agent_effect` decided a destructive apply. This PR's new failure modes (unreadable subtree, the 64-level ceiling — reachable from an *irrelevant* deep subtree added beside a previously-built promote plan) made that degradation reachable on healthy projects.

Fix, with the failure classes deliberately split — and the split was **forced by five existing tests**, not chosen in the abstract:
- `NoModels` stays a **true empty**: replication-only estates legitimately gate maintenance applies with no model sidecars; the first, blanket version of this fix broke exactly those tests.
- Any other failure over models that exist **refuses the gate** (`PolicyGate::Deny`, compile error as the reason) — a gate that cannot establish what it is gating does not gate on defaults. Same posture as the maintenance-apply refusal already on main.
- `required_verify_after` fails **safe** instead: attributes unavailable ⇒ verify every touched model.

Regression test: `evaluate_apply_policy` over a failing-compile models dir with an attribute-scoped deny in force ⇒ `Deny`, never fall-through allow.

### Evidence refresh

Post-fix: full suite **5865/5865**, clippy clean, fmt clean. Earlier evidence stands: 112-project dual-binary canary zero diffs; empirical filter parity at depth in both directions (clean-sidecar case); nested `run --dag` e2e green both rounds.

### The standing questions

**Least confident:** whether other `.ok()`-style consumers of compiles exist beyond the ones this review swept — the fresh pass grep'd the gate-feeding paths specifically; a fleet-wide sweep of tolerant compile consumers is the natural follow-up and belongs with #1310's contract work.

**Most important thing possibly missed:** the first (wedged) review covered surfaces the scoped pass did not re-examine — deadlock/cancellation, perf of per-keystroke LSP walks, tracing correctness. Those had zero findings in the #1312 round's identical checks and the LSP path is exercised by the suite, but they carry one fewer independent pass than the two decisive questions do.

Issue #1304, opened 2026-07-31
perf(engine): list_runs deserialises the whole RUN_HISTORY table to return 50 rows


hugocorreia90:
## Description

`StateStore::list_runs(limit)` deserializes **the entire** `RUN_HISTORY` table on every call, then sorts it, then throws away all but `limit` rows.

`engine/crates/rocky-core/src/state.rs:2205`:

```rust
pub fn list_runs(&self, limit: usize) -> Result<Vec<RunRecord>, StateError> {
    let txn = self.db.begin_read()?;
    let table = txn.open_table(RUN_HISTORY)?;
    let mut runs = Vec::new();
    for entry in table.iter()? {                     // every row
        let (_, value) = entry?;
        let run: RunRecord = serde_json::from_slice(value.value())?;   // fully deserialised
        runs.push(run);
    }
    runs.sort_by_key(|run| std::cmp::Reverse(run.started_at));
    runs.truncate(limit);                            // ...then 50 are kept
    Ok(runs)
}
```

It cannot early-exit because the table is keyed by `run_id`, so iteration order is unrelated to `started_at` and the full set must be materialised before it can be ordered.

`RunRecord` is not small — it carries a `models_executed` vector — so this is a full deserialisation of the entire run history, not just a key scan.

## Measured cost

From the controlled experiments in #1256: the working set attributable to `GET /api/v1/runs` scales at **~8.59 KB per row in `RUN_HISTORY`** — a 12.1 MB gap between a server whose history holds 1440 runs and one with 0, both serving the same endpoint. At a fixed history size it plateaus (it is a high-water mark, not a leak), but the plateau climbs with every run recorded.

For context, that measurement accounts for ~92–96% of the RSS growth in both 24h scheduler soaks, which is how it was found.

## Impact

1. **Memory and latency scale with total history, not with the requested page.** Every `GET /api/v1/runs`, every `rocky history`, and — via `get_model_history` → `list_runs(100)` (`state.rs:2229`) — every model-metrics and average-duration lookup pays the whole-table cost. A long-lived estate pays more for the same 50 rows indefinitely.

[...]

PR #1330, opened 2026-08-01
feat(engine/rocky-mcp): read-only scheduler surface — schedule_status + a history trigger filter

hugocorreia90:
The first slice of the scheduler agent-operations surface, **deliberately read-only**: the mutating half (`trigger_pipeline`/`resume`) needs a policy-capability vocabulary decision that stays deferred until real demand shows; `pause` rides separately (it carries a cross-crate skip-reason ripple).

## What ships

- **`schedule_status`** — the `GET /api/v1/schedule` projection as an MCP tool, through the **same non-printing core and the same `.rocky` derivation** the API route and the reconciler use, so the snapshot reports against the tick lock a `serve --scheduler` or a cron `rocky tick` actually holds. Stored state only — it does not evaluate demand (side-effect free), and the tool description says so, plus what a past `next_fire_at` means (an overdue pipeline, not a future promise).
- **`history` gains an optional `trigger` argument** (`"Schedule"`, `"Webhook"`, … — the wire vocabulary), applied **below the 50-run recency cap** via a new `StateStore::list_runs_matching`. Filtering the capped result instead lets a busy project's manual runs crowd every scheduler run out of the window — the exact lie-by-omission called out on #1304, whose cheap half this is.

## Evidence

- Below-the-cap contract pinned: a matching run older than **55** non-matching ones is still returned (and a precondition assert proves it sits outside the plain cap).
- MCP tool list + schemas re-pinned (29 tools); roundtrip suite 66/66.
- Full workspace **5873/5873** · clippy `-D warnings` clean · fmt clean.
- **No codegen cascade**: `HistoryOutput`'s shape is untouched (the filter narrows rows, not fields) and `ScheduleStatusOutput` is reused as-is.

## Scope notes

`rocky_dir_for_config` widened `pub(crate)`→`pub` so the MCP layer shares the exact lock-path derivation instead of re-deriving it. The deferred mutating surface and its `PolicyCapability` question, incident bundles, and the brief section remain tracked in the P4 plan; #1304's real half (the `list_runs` full-scan) is unchanged.

Independent red-team pass dispatched; merge held for it plus green CI.

### DIFF
--- a/engine/crates/rocky-cli/src/commands/history.rs
+++ b/engine/crates/rocky-cli/src/commands/history.rs
@@ -141,10 +141,32 @@ pub fn history_runs_output(
     since: Option<&str>,
     audit: bool,
 ) -> Result<HistoryOutput> {
+    history_runs_output_filtered(state_path, since, audit, None)
+}
+
+/// [`history_runs_output`] restricted to runs whose `trigger` matches.
+///
+/// The filter applies BELOW the 50-run recency cap (#1304): filtering the
+/// capped result instead would let a busy project's manual runs crowd every
+/// scheduler run out of the window, reporting "no matching runs" over a
+/// project full of them.
+pub fn history_runs_output_filtered(
+    state_path: &Path,
+    since: Option<&str>,
+    audit: bool,
+    trigger: Option<&str>,
+) -> Result<HistoryOutput> {
     let store = StateStore::open_read_only(state_path)?;
     let since_ts = parse_since(since)?;
 
-    let runs = store.list_runs(50)?;
+    let runs = match trigger {
+        Some(trigger) => {
+            // Match on the same Debug form the output serializes (`"Schedule"`,
+            // `"Webhook"`, ...) so the filter vocabulary IS the wire vocabulary.
+            store.list_runs_matching(50, |run| format!("{:?}", run.trigger) == trigger)?
+        }
+        None => store.list_runs(50)?,
+    };
     let filtered: Vec<_> = if let Some(ts) = since_ts {
         runs.into_iter().filter(|r| r.started_at >= ts).collect()
     } else {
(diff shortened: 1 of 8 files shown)


hugocorreia90:

**Reviewer: Codex (GPT-5.x), scoped pass.** All the structural checks came back clean in its words — `list_runs(limit)` is *"the old loop/sort/truncate body moved unchanged behind an always-true predicate"*; the trigger predicate adds one allocation per scanned row with *"no borrow or unwind corruption"*; no-schedule configs resolve to zero pipelines; `rocky_dir_for_config` is *"a pure, total path derivation with no hidden crate-only precondition."* Four findings survived; **all four fixed** in `964c7d40`:

| Finding | Fix |
|---|---|
| 🔴 **Not side-effect free**: redb's read-only open goes through `Database::create`, so a status read of a fresh project **materialized the state file** — on the one endpoint documented side-effect free (the API route shared the defect) | Existence guard before the open, in the shared core. **Pinned + mutation-checked**: neutering the guard fails the new test with the file created |
| Wire errors carried the alternate anyhow chain → **absolute project paths on the wire** | Top-level message only |
| `result_types` documented the trigger vocabulary as `"Scheduled"` where the wire and the new filter say `"Schedule"` | Aligned, filter cross-referenced |
| A scheduler on an explicit `--state-path` override is invisible to this tool (it reads the canonical path) — silently empty snapshot | Stated outright in the tool description, with the API-route alternative |

Post-fix: **5874/5874** · clippy clean · fmt clean · MCP roundtrip re-pinned.

Housekeeping note: the earlier #1328 review's first Codex task (`task-msa64k8v-ohm6r3`) is still marked running server-side and needs a user-invoked `/codex:cancel` — I can't invoke it.

PR #1334, opened 2026-08-01
feat(engine): runtime schedule pause — agent pause, human pause/resume

hugocorreia90:
P4's pause slice, completing PR1's scope (read-only surface merged as #1330).

## What ships

**Core** — `ScheduleStateRecord.paused` (serde-defaulted, additive): a durable runtime hold the reconciler consults after the cursor fetch, suppressing every demand source with a new `paused` skip recorded each tick — a paused pipeline is never metrically identical to healthy-idle. **State-level on purpose**: a running `serve --scheduler` holds ONE fingerprinted config snapshot, so a `[schedule] enabled = false` edit cannot reach it until restart; this flag can. The wire string threads the full exhaustive-pin chain (SkipReason → TickSkipReason → tick output mapping → metric label fn → `SCHEDULER_SKIP_REASONS` → the cross-crate guard), which is how every site was enumerated — by the compiler and the pins, not by grep.

**Agent surface** — MCP `pause_schedule`: mutating, safe-direction, `confirm: true` required, refuses a pipeline with no `[schedule]` block (the hold must attach to something the reconciler consults, never a stray cursor). **Pause-only on purpose** — resume re-enables autonomous runs, so per the recorded P4 decision it is human-only.

**Human surface** — `rocky state schedule pause|resume <pipeline>`: a `state` subcommand (no new top-level verb), typed `ScheduleHoldOutput`, idempotent (`changed: false` on a repeat), refusal creates nothing (pinned).

## The stop-gate finding, threaded through both surfaces

A hold controls whichever scheduler reads the **same state file** — so a pause against the wrong store must never be silently successful. Both the CLI output (both modes) and the MCP response **name the state file acted on**, and the subcommand help spells out the `--state-path` targeting interplay.

## Evidence

- Demand-level behavior pinned both ways: paused ⇒ no demand + the skip recorded; the identical evaluation with the flag cleared is due again (the skip provably came from the flag).
- Atomic store helper (single write txn read-modify-write — a caller-side get-then-put would race the reconciler's cursor advances).
- CLI round-trip: pause → flag set → idempotent repeat → resume → cleared; no-`[schedule]` refusal creates no store.
- Full workspace **5883/5883** · clippy clean · MCP tool list re-pinned at 30 · codegen cascade complete (schema + SDK + TS + barrels wired — the two silent-failure surfaces — + openapi).

Independent red-team pass dispatched; merge held for it plus green CI.


hugocorreia90:

**Reviewer: Codex (GPT-5.x), scoped pass. Verdict: BLOCK.** Findings and dispositions:

| Finding | Disposition |
|---|---|
| 🔴 **Webhook bypass (CONFIRMED, the verdict's basis):** webhook consumption constructs its demand outside `evaluate_one`'s pause gate — a paused pipeline still executed every spooled delivery | **Fixed** (`12da44a5`): the gate now runs inside webhook consumption with **defer-never-drop** semantics — the pending file stays untouched (an accepted 202'd delivery survives the pause), a `paused` skip is recorded each tick, resume fires exactly the deferred demand. **Pinned at the tick level, exactly the test the review demanded** (paused + spool file through `tick_once` ⇒ 0 runs, file pending, skip recorded; resume ⇒ fires once), and **mutation-checked** (gate neutered ⇒ the test fails with the demand executing) |
| **Old-binary field-drop (CONFIRMED):** a pre-pause resident scheduler's whole-record `touch_schedule_evaluated` rewrite silently erases the unknown `paused` field; schema stays v21 so old binaries aren't fenced | **Accepted as a documented mixed-version limitation, not fenced.** Reasoning on the record: the window exists only while an old scheduler binary keeps running across an upgrade; the feature is experimental; and fencing via a v22 schema bump would hard-stop every old binary over one bool — disproportionate. The CHANGELOG and both surfaces' docs state it: *a pause is only durable once the resident scheduler runs a version that knows the field.* If Hugo prefers the fence, it's a one-line bump + the standard mismatch policy |
| **Serialization (CONFIRMED clean):** serde-default round-trips; no `deny_unknown_fields` | No action |
| **MCP confirm/config gate (CONFIRMED clean):** absent/false confirm rejected before any access; `schedule()` exhaustive over all five pipeline variants | No action |
| **CLI contention (CONFIRMED, loud):** pause during a live child's whole-run store hold fails after the short retry with the acted-on path + "locked by another process", nonzero exit | Accepted; the reviewer noted the message doesn't say "retry" — added to the error text is left as polish, the failure is already loud and actionable |

Also fixed this round: the SDK barrel genuinely carries `ScheduleHoldOutput` now (the barrel-completeness CI test — added in an earlier session precisely for this class — correctly failed and is now green; proven by actual import).

Post-fix: **5884/5884** · clippy clean · barrel test 90/90 · MCP roundtrip 30 tools.

**The reviewer's residual, answered in-code:** whether a pause should cancel retries of an already-claimed demand — current semantics: an in-flight claim completes (pause suppresses *new* demand, never kills running work), which matches the drain philosophy; stated in the tool description.

PR #1339, opened 2026-08-01
feat(engine): scheduler incident bundles + a brief scheduler section

hugocorreia90:
Two halves of the resident scheduler's failure-visibility story: when a scheduled run fails, the facts land on disk as a structured incident bundle, and `rocky brief` now carries a Scheduler section so the digest a failure hook posts already points at them.

## Incident bundles (`rocky-core/src/schedule/incidents.rs`)

The reconciler writes one JSON file under `.rocky/incidents/` at the moment a scheduler-submitted attempt finalizes as `Failure`: pipeline, demand source, occurrence (`logical_ts`), submission id, exit code, attempt count, the consecutive-failure count after this failure, and retrieval **commands** (`rocky history`, `rocky trace`) — citations, not narration.

- **Drain interruptions write nothing** — a graceful shutdown is not an incident. Pinned at tick level in both directions.
- Keep-last-50 sweep; timestamp-prefixed names make lexicographic order chronological. Single-writer by construction (bundles are written inside the tick, ticks serialize on the tick lock).
- `write_incident` failures warn and never fail the tick — the run's outcome is already committed, and diagnostics that break the thing they diagnose are worse than none.
- Hostile pipeline names are sanitized out of the filename (test pins containment).

## Brief scheduler section

`BriefOutput` gains `scheduler`: runtime holds (paused pipelines by name), consecutive-failure streaks worst-first, scheduler-spawned runs in the window, and the incident spool (count + newest path).

Semantics that needed deciding:
- The trigger filter is `Schedule` **or** `Webhook` — both are scheduler-spawned (`RunTriggerKind::for_source` is the ground truth) and both flow through the same claim machine and incident hook.
- `failed_in_window` counts `Failure | PartialFailure` only — the skip statuses did no work and are neither success nor failure, mirroring `RunStatus::terminal_outcome`.
- Fail-closed per the digest's section rules: an unreadable config **or an unreadable schedule cursor** renders `unavailable` with a note (a cursor read error rendering as "not paused" would be exactly the smoothed-over all-clear the rules forbid); nothing scheduled renders `no_data`.

## Evidence

- 5897/5897 engine tests, 204 SDK tests, clippy clean, fmt clean, on a base current with `main`.
- Mutation checks (mechanism neutered → the pinning test goes red, then restored): drain-suppression both directions at tick level; paused-flag ignored; webhook trigger dropped from the filter; skip-statuses counted as failures. Each died at its distinct assertion.
- Schema, Pydantic, TypeScript and OpenAPI bindings regenerated together; a rerun on the rebased tree shows zero drift. No barrel change (`BriefOutput` pre-existed; the nested section types live inside `brief_schema`). No dagster fixture covers `brief` (checked — no recapture needed).
- Docs: incident-bundles subsection in the running-without-an-orchestrator guide; `rocky brief` reference updated.

Deliberately out of scope: webhook delivery of incident bundles (follow-up if bundles prove insufficient) and any stderr capture (child stderr inherits into serve's log by design; the trace pointer + run id deliver the citations).


hugocorreia90:
## Red-team disposition

Independent reviewer: GPT-5.x (Codex) against the initial head — verdict **BLOCK**, 10 findings. All 10 verified against the code; 9 fixed in the hardening commit, 1 sub-item accepted with rationale. Point by point:

1. **Recovery paths emitted no bundle** — fixed. `record_incident` is now the shared seam behind the direct arm, the stuck resolver, the missing-record grace, and the orphan sweep. The orphan test asserts the bundle; the resolver test asserts exactly one bundle across a follow-up tick (no double-write — the resolved claim is `Released`, not `Submitted`, so the resolver cannot re-fire).
2. **Late-drain race** — fixed at the true discriminator. `RunOutcome.drain_interrupted` is set by the spawner only when *it* refused or terminated the child for a drain; the reconciler keys on that record instead of re-sampling the flag. New test: a child that failed on its own while a drain happened to be raised still records its incident; a spawner-terminated child does not.
3. **Capped page → false all-clear** — fixed. The section now queries `list_runs_matching` with the trigger+window predicate below the recency cap. New test: a scheduled failure surfaces past a newer manual flood with a scan cap smaller than the flood (fails under page-then-filter — mutation-checked).
4. **Partial-failure inconsistency** — fixed by inclusion. `Partial` writes bundles (explicit `outcome` field); partial trips partial backoff exactly like failure trips failure backoff, so it is an incident. Docs updated to say so.
5. **Sweep could delete unrelated files** — fixed. A symlinked `incidents` directory is refused outright (write path and brief read path); the sweep deletes only regular files whose names the writer itself produces (`is_bundle_name`); deletion errors propagate instead of being discarded. Tests: foreign files survive bundle churn; a symlinked dir is refused with the victim intact.
6. **Inventory not fail-closed / newest unreliable** — fixed for the substantive parts: only `is_bundle_name` files count, per-entry read errors fail the section closed, symlinked dir refused. The same-second tie is accepted as a documented 1-second resolution: two bundles recorded in the same second are both "newest" for the digest's purpose, the count stays exact, and the path always opens a real bundle — parsing every bundle to disambiguate a 1-second tie adds failure modes for no operational gain.
7. **Fabricated `consecutive_failures: 0`** — fixed. The field is `Option<u32>`: a cursor read error yields `None`, and the orphan sweep — which deliberately never touches the per-pipeline scalars — reports `None` rather than a value unrelated to the orphaned failure.
8. **`attempts` was claim-lifetime** — fixed: `cycle_attempts` (with its serde-default caveat documented on the field).
9. **Dishonest pointers/docs** — fixed. The jobs/schedule endpoints are labeled resident-scheduler-only inside the pointer strings themselves; the guide lists the pointers actually emitted (the `rocky trace` claim is gone); the brief reference scopes the citation rule to event lines, with the scheduler's posture lines carrying names/counts and the incident line carrying the bundle path.
10. **Happy-path-only binding** — addressed by the tests above plus four fresh mutation checks (orphan emission dropped; drain discriminator reverted to flag re-sampling; sound query reverted to page-then-filter; sweep pattern loosened to any-json). Each goes red at its pinning assertion.

On the reviewer's stated residual — a durable, idempotent incident identity across crashes: the claim CAS is the idempotency boundary (a bundle writes only on the CAS winner's turn, once per state transition), but a crash inside the window between CAS-win and the filesystem write can still lose that single bundle. Accepted as a documented limitation: the failure remains visible in `rocky history` and the cursor's streak, and closing the window would mean a WAL for a diagnostic artifact.

Gates on the rebased head: 5906/5906 workspace tests, clippy clean, fmt clean, codegen drift-free. A follow-up independent pass on the fix delta is running before merge.

hugocorreia90:
## Delta-review disposition (round 2)

The independent pass on the hardening commit confirmed findings 1–3 and 5 **closed** (all four terminal seams emit; the drain discriminator is the spawner's own record with timeout kills still counting as incidents; the sound below-cap query; honest Option facts and Partial emission), found no missed terminal seams, no resolver double-write, no orphan-Partial gap, and no panic path in the name check — but returned BLOCK on three residuals, all real. Fixed in the follow-up commit:

1. **`is_bundle_name` too loose** (empty pipeline segment `…Z--id.json` accepted; digit-shaped nonsense timestamps accepted): the timestamp must now be a plausible calendar instant (month 1–12, day 1–31, hour/minute/second in range) and the `<pipeline>-<id>` remainder must have a non-empty segment on both sides of a separator. The strictness test carries the exact probes from the review.
2. **Bundle-named non-files counted**: the brief inventory now counts regular files only — a directory or symlink wearing a bundle-shaped name is not a bundle (pinned by a test with a `…Z-fake-dirbundle.json` directory).
3. **Information regression on the inventory error arm**: an incidents-directory failure no longer zeroes `runs_in_window` / `failed_in_window` that were already computed soundly; the symlink-refusal test now pins the preserved failure count.

Each fix mutation-checked (grammar loosened / file-type check dropped / counts re-zeroed — all red at the pinning assertion). Gates on this head: 5906/5906 workspace tests, clippy clean, fmt clean, base current with `main`.

Accepted-with-rationale, unchanged from round 1 and re-confirmed by the reviewer as the acknowledged limitation: the crash window between terminal CAS and the filesystem write can lose that single bundle (visible still in `rocky history` and the cursor streak); a durable outbox for a diagnostic artifact is deliberately out of scope. Same-second `latest_incident` ties remain a documented 1-second resolution.

hugocorreia90:
## Delta-review disposition (round 3) + one stop-review catch

Round 3 confirmed residuals (b) and (c) closed and BLOCKed on the bundle-name matcher from both directions — my independent month/day bounds accepted impossible dates (Feb 30 class), and the boundary-hyphen rule rejected names the writer legitimately produces (pipeline names are unconstrained config keys, so `-orders` is scheduler-reachable; those bundles would have been invisibly exempt from retention and the brief inventory). Both fixed by making agreement structural rather than re-derived:

- The timestamp is now validated as a real calendar instant with the same chrono grammar the writer formats with (`%Y%m%dT%H%M%S`) — Feb 30 and friends are out, real leap days are in.
- The writer normalizes filename segments (empty → `_`, boundary hyphens → `_`), so the matcher's grammar is exactly the writer's output for **any** pipeline name or submission id. A cross-product agreement test (hostile pipelines × hostile ids, including the review's exact probes) pins the coupling; the flagged test gaps (`runs_in_window` on the error arm, calendar validity) are closed.

A stop-time review then caught what the normalization made easier to hit: sanitize/normalize are lossy, so two **distinct** incidents can land on the same filename in the same second (`a/b` vs `a_b`), and the write was a silent overwrite. Writes now use `create_new` with a numbered variant on collision — both bundles survive, and the variant still matches the bundle grammar. (It also caught a clippy pass-by-value warning in the same edit — fixed.)

All three fixes mutation-checked (calendar parse dropped / normalization dropped / `create_new` reverted to plain write — each red at its pinning assertion). Gates on the rebased head (now on top of the two-pipelines-one-table refusal): 5915/5915 workspace tests, clippy clean, fmt clean.

hugocorreia90:
## Delta-review disposition (round 4) + one more stop-review catch

Round 4 confirmed the round-3 probes fixed (calendar validation, writer↔matcher agreement, collision variants grammar-valid, sweep counting) and BLOCKed on one construction it traced precisely: unpadded numbered variants broke the sweep's lexical-order assumption (`-10` sorts before `-2`), and first-free slot probing re-created exactly the slot the sweep had just deleted — under a heavy same-name collision flood, retention froze while every newest bundle was deleted right after being written, with `write_incident` returning a path that no longer existed.

Fixed structurally: variants are zero-padded (`-002`…`-999`) and the slot is chosen as **max-existing + 1, never first-free** — a swept slot stays dead, retention rolls forward, and the returned path always exists. The regression test runs the review's own trace (a flood past the retention cap on one base name) and asserts every returned path exists and the newest bundle survives its own sweep. The intra-group ordering subtlety (ASCII `-` < `.` makes a base sort after its variants) is documented as within the name format's 1-second resolution. The reviewer's pre-existing note — out-of-range chrono years through the public `write_incident` API produce signed `%Y` output the fixed-width matcher disowns — is accepted: not reachable from the scheduler (`recorded_at` is the tick's `now`), and it predates this PR's scope.

A stop-time review then caught a regression all four rounds missed: the finding-8 fix (bundle `attempts` = `cycle_attempts`) changed a **shared binding** and silently moved the tick's public `ExecutedDemand.attempts` — the frozen tick contract's total-submissions audit counter — to per-cycle. The counters are different on purpose and now bind separately; a standing demand failing across two cycles pins the divergence (tick reports 2 total, bundle reports 1 for the cycle), mutation-checked in both directions.

Gates: 5917/5917 workspace tests, clippy clean, fmt clean, base current.

hugocorreia90:
## Delta-review disposition (round 5)

Round 5 confirmed the round-4 trace dead (max+1 selection, swept slots stay dead) and the attempts decoupling correct with no other consumer wanting the cycle counter — and BLOCKed on three residuals, each fixed:

1. **`-1000` namespace escape**: the pre-loop increment had no bound, so a full variant group minted a 4-digit name the max-scan cannot parse and the sweep sorts wrongly — the round-4 freeze one namespace up. The namespace is now exactly `-002`…`-999` and **fails closed** at the boundary (`io::Error`, warn-never-fail at the callers) — an explicit exhaustion contract instead of stepping outside the parseable namespace. Pinned by a pre-seeded-full-namespace test; the bound is mutation-checked.
2. **Raced empty-group `-001`**: an `AlreadyExists` on the bare base name now jumps to `-002` (the base counts as occupancy 1), so `-001` is never minted by any path.
3. **Published `attempts` description false**: the schema doc claimed "1 unless in-tick retries fired", which the two-cycle divergence test disproves — a standing demand's counter spans released cycles. Reworded to the lifetime audit-counter truth; bindings regenerated (doc-only schema change).

The flagged test gaps are closed: the divergence test now asserts both bundles exist with distinct submission ids and that the newest is cycle 2's; the exhaustion boundary has its own test. The reviewer's speculative note on external `write_incident` callers making races more reachable is accepted as-is — no in-tree caller bypasses `record_incident`, and the collision loop plus the exhaustion contract are exactly the defense an external caller gets.

Gates on the rebased head (now on top of the shadow-compare feature): 5921/5921 workspace tests, clippy clean, fmt clean.

PR #1342, opened 2026-08-01
fix(engine/rocky-ir): validate decimal assignment capacity

### LINKED ISSUE #1338: bug(engine): load contracts accept decimals with insufficient integer capacity
## Description

Load-contract type validation accepts landed numeric types that do not fit in the contract's declared `DECIMAL(p,s)` type.

`validate_contract_typed` delegates known types to `is_assignable`. For Decimal combinations, `is_assignable` asks whether `common_supertype(from, to) == to`, but `common_supertype` independently takes the maximum precision and maximum scale. That does not preserve integer-digit capacity (`precision - scale`).

For example:

- Landed `DECIMAL(10,2)` has 8 integer digits, but a `DECIMAL(10,4)` contract has room for only 6. Rocky accepts it.
- Landed `BIGINT` may need 19 integer digits, but a `DECIMAL(19,10)` contract has room for only 9. Rocky accepts it.

The contract gate can therefore promote a staging table even though legal values of the landed type cannot be represented by the required type.

## Steps to Reproduce

Against `main` at `e4750f6fe798d8b59ac4b67421f046e19a64a8fb`:

1. Construct a load contract requiring `amount DECIMAL(10,4)`.
2. Pass a landed `ColumnInfo` for `amount DECIMAL(10,2)` to `validate_contract_typed`.
3. Inspect the returned `ContractResult`.

Equivalent focused assertion:

```rust
let contract = ContractConfig {
    required_columns: vec![RequiredColumn {
        name: "amount".into(),
        data_type: "DECIMAL(10,4)".into(),
        nullable: true,
    }],
    ..Default::default()
};
let landed = vec![ColumnInfo {
    name: "amount".into(),
    data_type: "DECIMAL(10,2)".into(),
    nullable: true,
}];

assert!(!validate_contract_typed(&contract, &landed).passed);
```

The assertion fails because the result currently passes. Replacing the landed type with `DECIMAL(10,4)` is a passing control. The same false pass occurs for landed `BIGINT` against required `DECIMAL(19,10)`.

## Expected Behavior

A known landed type should satisfy a known Decimal contract only when every value representable by the landed type fits within the contract type. `DECIMAL(10,2)` should fail a `DECIMAL(10,4)` requirement, and `BIGINT` should fail `DECIMAL(19,10)`.

## Actual Behavior

Both incompatible pairs pass `validate_contract_typed`, producing no `required_column_type` violation.

The current calculation is:

```rust
// Decimal + Decimal
precision: max(p1, p2),
scale: max(s1, s2),

// Assignment succeeds when that common type equals the destination.
common_supertype(from, to).is_some_and(|sup| sup == *to)
```

For `DECIMAL(10,2) -> DECIMAL(10,4)`, the computed common type is `DECIMAL(10,4)`, despite the destination having fewer integer digits.

## Rocky Version

`main` at `e4750f6fe798d8b59ac4b67421f046e19a64a8fb`; latest tagged engine release is `engine-v1.68.0`.

## OS / Architecture

macOS 26.6 arm64 (Darwin 25.6.0).

## Configuration

No warehouse credentials or external configuration are required; the failure reproduces through the public `rocky-core` contract API.

mattfaltyn:
## Summary

Closes #1338.

When both landed and expected types normalize, load-contract validation now requires a destination `DECIMAL(p,s)` to preserve both the landed type's fractional scale and its integer-digit capacity. This prevents incompatible normalized Decimal and integer sources from passing the promotion gate.

## Subproject(s) touched

## Changes

- Handle Decimal-to-Decimal and integer-to-Decimal compatibility directionally in `is_assignable`, without changing shared compiler type inference.
- Add IR-level passing and failing capacity cases.
- Add load-contract regression coverage for the same cases and assert the `required_column_type` violation.

## Test Plan

- [x] `cargo test -p rocky-ir test_is_assignable_decimal_capacity`
- [x] `cargo test -p rocky-core test_typed_decimal_contract_checks_full_capacity`
- [x] `cargo nextest run --all-features` (GitHub Actions)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` (local and GitHub Actions)
- [x] `cargo fmt -- --check` (local and GitHub Actions)

## Checklist

- [x] Not a CLI JSON schema or DSL syntax change; no consumer updates are required

### Engine (`engine/`)

- [ ] `cargo test --all-targets` passes

`cargo test --all-targets` executes the default-feature test suite successfully but cannot exit on current `main`: the existing Criterion `binary_startup` benchmark launches nested `cargo run`, which waits on the parent Cargo target-directory lock. The repository's canonical all-feature Nextest CI command is listed above and must pass before this PR is marked ready.


hugocorreia90 (commented):
Thanks for this — the core change is sound, and I want to lead with the part I checked hardest, because this function has a history that makes "stricter" worth proving rather than assuming.

## The change is a pure tightening — verified, not assumed

`is_assignable` sits at a validation gate, so the question that matters is whether any pair that *failed* before now *passes*. I ran the transition against `main` for all three new arms:

| pair | `main` | this PR |
|---|---|---|
| `(Decimal{p1,s1}, Decimal{p2,s2})` | `common_supertype` = `Decimal{max(p1,p2), max(s1,s2)}`, so true iff `p1<=p2 && s1<=s2` | `s1<=s2 && (p1-s1)<=(p2-s2)` |
| `(Int32, Decimal{p,s})` | supertype `Decimal{max(p,10), s}`, so true iff `p>=10` | `p-s >= 10` |
| `(Int64, Decimal{p,s})` | true iff `p>=19` | `p-s >= 19` |

In each case the new condition **implies** the old one — for the integer arms, `p-s >= 10` gives `p >= 10` directly; for decimal-to-decimal, `s1<=s2` plus `(p1-s1)<=(p2-s2)` gives `p1<=p2`. So nothing that was rejected becomes accepted. That is the property I most wanted to hold here, and it does.

Two smaller things that also check out: the `10`/`19` digit constants are right for `i32`/`i64` and match what `common_supertype` already uses (`max(10)` / `max(19)`), so the two functions stay consistent; and `checked_sub` makes a `p < s` type fail *closed* rather than panicking, which is the right direction.

Blast radius is contained: `is_assignable` has exactly one production caller, `landed_type_conforms` (`contracts.rs:328`), so this only reaches load-contract validation.

## Needed before merge

**A changelog entry.** This is a user-visible tightening at a gate — a load contract that passed yesterday can fail today, e.g. landed `INTEGER` into `DECIMAL(10,2)`, or `DECIMAL(10,2)` into `DECIMAL(10,4)`. The first thing an affected user sees is a previously-green pipeline going red, so `engine/CHANGELOG.md` under `## [Unreleased]` → `### Fixed` should name the newly-rejected shapes and why they were unsafe. Happy to suggest wording if useful.

## Worth doing while you're here

**Scope the claim about the gate.** The summary says this "prevents incompatible Decimal and integer sources from passing the promotion gate". That holds *when both types normalize*. `is_assignable` returns `true` whenever either side is `Unknown` (`types.rs:206-208`), so a landed type spelled in a dialect form that does not normalize still passes unconditionally. Not yours to fix and please don't try in this PR — but it is why this area is sensitive, and the PR text reads slightly stronger than what the code can deliver.

**A stale doc comment that this PR makes more wrong.** `contracts.rs:316-320` still says "Otherwise compares by equality", directly above a doc block describing assignability. With this change it is neither equality nor plain assignability but capacity-aware assignability. Cheap to correct while the file is open.

**Two controls in `test_is_assignable_decimal_capacity`.** All six current cases exercise the new arms, so a refactor could break the surrounding behaviour without reddening anything:
- identity still passes — `is_assignable(decimal(10,2), decimal(10,2))` — which exercises the `from == to` early return that the new `match` now sits below
- the reverse direction still fails — `is_assignable(decimal(12,4), decimal(10,2))`
- and optionally a `p < s` case such as `decimal(2,4)`, to pin the `checked_sub` fail-closed behaviour that is currently implicit

**Mutation evidence.** A green CI run on a newly added test does not show the test *binds*. The cheap check: replace the integer-digit comparison inside `decimal_fits` with `true` and confirm both new tests go red, and that nothing else does. If you would rather not, say so and I will run it — but it is worth having on the record for a change at a gate.

None of the above questions the approach; the transition analysis above is the part I would have wanted a reviewer to do, and it came out clean.

mattfaltyn:
@hugocorreia90 Thank you for the careful review and for verifying that this change is a pure tightening. I incorporated the feedback.

- Added an Unreleased changelog entry with the unsafe examples and normalized type scope.
- Corrected the stale contract comment.
- Added identity, reverse direction, and `precision < scale` controls.
- Updated the PR summary to state that the behavior applies when both types normalize.

I also ran the suggested mutation check. Both regression tests failed when the integer capacity comparison was disabled, and the final tests and CI are green. I left `Unknown` handling unchanged as suggested.

Thanks again for the precise and thoughtful review.

hugocorreia90:
One datum to complete the record, complementing the review above: the caller-level blast radius of tightening `is_assignable`. Swept every consumer across the workspace (including through the `rocky-ir` and `rocky-compiler` re-exports): the only call site is `contracts.rs::landed_type_conforms` — the exact gate #1338 is about. Nothing in drift's safe-widening classification, the compiler's assignment checks, or any policy surface consumes it, so the stricter arms cannot change behavior anywhere but the contract validation this PR intends to tighten. (The history that makes this sweep worth stating: a previous strictness change to a shared helper reached a policy gate through a tolerant caller nobody re-checked.)

Issue #1351, opened 2026-08-02
bug(engine): UnifiedDag::infer_runtime_dependencies matches by node label, not physical target — collisions overwrite, lineage failures silently skipped


hugocorreia90:
Found during the #1275 design review (independent red team). `UnifiedDag::infer_runtime_dependencies` (`engine/crates/rocky-core/src/unified_dag.rs` ~935) derives runtime edges for `run --dag` by:

- indexing producers by **lowercased node label** and matching only the **last segment** of a physical read — the configured `[target]` (schema/table, catalog template) is never consulted, so a model whose target table differs from its model name derives no edge (`orders` → `main.orders_v2` is invisible);
- storing producers in a `HashMap` keyed by that label, so two nodes with colliding labels **silently overwrite** each other and one loses all derived edges;
- `continue`-ing on `extract_lineage` failure (~957), so an unparseable model contributes no edges with no signal anywhere.

Consequences: co-scheduling races of the #1275 class survive inside `run --dag` even where the label heuristic looks like it should catch them, and they do so silently. The #1275 fix direction (one shared target-aware, dialect-aware edge derivation consumed by both the Project layer computation and this inference) subsumes the first item; the collision-overwrite and silent-failure behaviors are defects regardless of that design.

PR #1352, opened 2026-08-02
fix(engine): order physical-read consumers after their producers in both schedulers

### LINKED ISSUE #1275: bug(engine): physical schema.table upstream reads produce no DAG edge, so producer and consumer are co-scheduled

hugocorreia90:
Fixes #1275. A model reading another in-run model's target by its physical name (`main.orders`, `db.main.orders`, or a bare `orders_v2` resolved through the connection's search path) had no DAG edge — multi-part references are external to compile-time resolution — so producer and consumer could share an execution layer and race under `--parallel`. And the two run entry points schedule independently: plain `run` iterates `project.layers` while `run --dag` phases a `UnifiedDag` through its own executor, so a one-scheduler fix leaves the same race reachable one flag away (this exact gap failed the first design review).

## Shape

One shared derivation, `rocky_core::physical_edges`, feeds both schedulers:

- Each model's **rendered target components** (catalog/schema/table) are matched componentwise — case- and quote-folded on **both** sides — against every model's SQL read set (`referenced_tables`, which already lowercases; the fold is load-bearing for the config side). 3-part reads match (catalog, schema, table); 2-part match (schema, table); **bare reads match by table component alone**, because search-path resolution is connection state Rocky cannot observe. A spurious match is an extra ordering constraint — the safe error direction; a missed match is the race.
- **Plain run**: `augment_physical_read_edges` pushes edges into the runtime `dag_nodes` copy and recomputes `execution_order`/`layers` (the shipped shadow-rewrite pattern), skipped under `[resilience] contain_failures` whose ledger already orders physical readers. The compile-time graph, `E001` surface, and `rocky dag` export are untouched.
- **`run --dag`**: `unified_dag::infer_physical_dependencies` inserts `DataDependency` edges before phase computation. The existing label heuristic stays for seed/load matching; it was blind to renamed targets (a producer targeting `orders_v2` under model name `orders_model` derived nothing).

## Cycle and incompleteness policy

- A candidate edge that would close a cycle (mutual physical reads) is **skipped deterministically** inside the derivation and surfaced as a warning — the pair serializes in one defined order instead of refusing runs that execute today. `execution_phases`' hard cycle error can therefore never be triggered by a derived edge.
- Models whose SQL cannot be parsed for reads, and colliding targets, surface as structured warnings — `RunOutput.scheduling_warnings` / `DagRunOutput.warnings` — never a silent pass.

## Evidence

- 12 new tests: derivation unit tests (2-part/3-part/bare matching, renamed targets, both-sides folding, mutual-read determinism, declared-edge cycle guard, unparseable reporting, collision multi-match, compile-edge dedup), plain-run layering incl. the issue's exact repro (`[[orders], [mart_qualified]]`) and the mutual-read warning path, UnifiedDag phase-ordering incl. the renamed-target case the label heuristic misses and the no-refusal cycle case.
- 5 mutation checks, each red at its pinning assertion: cycle guard dropped, 2-part matching dropped, bare-table matching dropped, target-side folding dropped, dag edge-insertion dropped. (A read-side folding mutation proved equivalent — `referenced_tables` already normalizes; documented in the module.)
- 5940/5940 workspace tests, 204/204 SDK tests, clippy clean, fmt clean, codegen bindings regenerated together (fixtures unaffected — the new fields serialize only when non-empty).
- Pre-existing containment-path tests pass unchanged (that path's own physical-read ordering is deliberately left in charge under `contain_failures`).

Out of scope, tracked: #1351 (the label heuristic's collision-overwrite and silent-failure behaviors for non-transformation nodes), dagster-side ordering (consumes `rocky dag`, unchanged), and fail-closed escalation on case-unobservable adapters (Snowflake `QUOTED_IDENTIFIERS_IGNORE_CASE`, BigQuery `is_case_insensitive` — #1281 tracks reading the live setting).

### DIFF
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -6037,6 +6037,13 @@
           },
           "version": {
             "type": "string"
+          },
+          "warnings": {
+            "description": "Scheduling warnings from physical-read edge derivation: mutual reads serialized by a deterministic order, models whose SQL could not be parsed for reads, and colliding targets. Empty when none.",
+            "items": {
+              "type": "string"
+            },
+            "type": "array"
           }
         },
         "required": [
@@ -16307,6 +16314,13 @@
               "string",
               "null"
             ]
+          },
+          "scheduling_warnings": {
+            "description": "Scheduling warnings from physical-read edge derivation (#1275): mutual physical reads serialized by a deterministic order, models whose SQL could not be parsed for table references, and colliding targets. Empty (and omitted) when none.",
+            "items": {
+              "type": "string"
+            },
+            "type": "array"
           },
           "shadow": {
             "description": "True when running in shadow mode (targets rewritten).",
--- a/editors/vscode/src/types/generated/dag_run.ts
+++ b/editors/vscode/src/types/generated/dag_run.ts
@@ -39,6 +39,10 @@ export interface DagRunOutput {
    */
   total_nodes: number;
   version: string;
+  /**
+   * Scheduling warnings from physical-read edge derivation: mutual reads serialized by a deterministic order, models whose SQL could not be parsed for reads, and colliding targets. Empty when none.
+   */
+  warnings?: string[];
   [k: string]: unknown;
 }
 /**
--- a/editors/vscode/src/types/generated/run.ts
+++ b/editors/vscode/src/types/generated/run.ts
@@ -167,6 +167,10 @@ export interface RunOutput {
   quarantine?: QuarantineOutput[];
   resumed_from?: string | null;
   /**
+   * Scheduling warnings from physical-read edge derivation (#1275): mutual physical reads serialized by a deterministic order, models whose SQL could not be parsed for table references, and colliding targets. Empty (and omitted) when none.
+   */
+  scheduling_warnings?: string[];
+  /**
    * True when running in shadow mode (targets rewritten).
    */
   shadow?: boolean;
--- a/engine/crates/rocky-cli/src/commands/run.rs
+++ b/engine/crates/rocky-cli/src/commands/run.rs
@@ -5993,6 +5993,82 @@ pub(crate) fn dialect_case_rules(
 /// Route the models built by this invocation and their in-run dependency reads
 /// to shadow targets. Deferred or otherwise unselected upstreams stay on their
 /// production targets.
+/// Physical-read ordering for NON-shadow runs (#1275): a model that reads
+/// another in-run model's rendered target by physical name has no
+/// compile-time edge (multi-part refs are external to `classify_table_ref`),
+/// so the pair can share an execution layer and race under `--parallel`.
+/// Derive the missing edges from target components (shared derivation with
+/// the `run --dag` scheduler), apply them to the runtime DAG copy, and
+/// recompute the plan — the compile-time graph, `E001` surface, and
+/// `rocky dag` export stay untouched. Shadow runs are excluded: shadow's own
+/// rewrite derives edges for every redirected read and refuses the unsafe
+/// remainder.
+///
+/// Cycle-closing candidates never leave the derivation (skipped
+/// deterministically, reported as warnings), so this recompute cannot turn
+/// a compiling project into a refused one.
+fn augment_physical_read_edges(
+    compile_result: &mut rocky_compiler::compile::CompileResult,
+    contain_failures: bool,
+    warnings: &mut Vec<String>,
+) -> Result<()> {
+    let inputs: Vec<rocky_core::physical_edges::PhysicalEdgeModel<'_>> = compile_result
+        .project
+        .models
+        .iter()
+        .map(rocky_core::physical_edges::PhysicalEdgeModel::from_model)
+        .collect();
+    let existing: Vec<(String, String)> = compile_result
+        .project
+        .dag_nodes
+        .iter()
+        .flat_map(|n| {
+            n.depends_on
+                .iter()
+                .map(|p| (n.name.clone(), p.clone()))
+                .collect::<Vec<_>>()
+        })
+        .collect();
+    let derived = rocky_core::physical_edges::derive_physical_edges(&inputs, &existing);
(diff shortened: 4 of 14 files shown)


hugocorreia90:
## Red-team disposition (round 1) + two stop-review catches

The review ran against head `1354501d` and returned BLOCK. Every finding traced; the fixes are on the branch (`5a0e6974`, `f52c2905`):

1. **Both schedulers covered / green-to-red closed** — confirmed by the review; no action.
2. **Bare table-component matching absent** — the review was right and the PR description was wrong about its own tree: the commit that claimed to add bare matching (`314c9f73`) carried only regenerated bindings — a mutation-restore had wiped the uncommitted source edit before `git add -A` ran, so the message described code that never landed. Now actually present, with its tests, in `f52c2905`. (Process note recorded: verify a commit's DIFF, not its message, after any restore.)
3. **False compile-time `ModelRef` edge suppressing a true derived edge** (bare read of an external table that happens to share an unrelated model's NAME) — real, root cause is the compiler's name-only bare matching, which predates this PR; this PR's cycle-skip can promote it from "false edge exists" to "true edge suppressed". Mitigated structurally by specificity ordering (below) where the collision involves derived candidates; the compile-declared case is filed as a follow-up with the shadowing semantics question, since rewriting declared edges at run time is not this PR's business.
4. **Cross-catalog two-part false matches** — same suppression class; mitigated by the same mechanism: candidates now process **most specific first** (3-part, then 2-part, then bare), so a loose candidate can never occupy the graph first and cause the guard to discard an exact true edge. Pinned by an adversarial-order test.
5. **Remove-without-reconsideration** — fixed in `5a0e6974` (a second pass re-checks every name-level skip against the live graph and reinstates safe directions), pinned by a through-intermediate reinstatement test. This was also caught independently by a stop-time review before your report arrived.
6. **Ephemeral phantom producers** — fixed: ephemeral models (no materialization) are excluded from the producer index; a read of their configured name is a read of something external. Pinned.
7. **Warnings lost on pre-result error paths** (`--dag --shadow` refusal, executor-level errors) — accepted as documented: the JSON error envelope has no warnings channel; the tracing::warn lines still fire. **Mislabeled insertion-guard warning** — fixed: full-graph skips no longer claim "mutual physical reads".
8. **No plain-run bypass** — confirmed by the review; no action.
10. **Quadratic guard cost** — the UnifiedDag guard now maintains one incrementally-updated adjacency map instead of rebuilding per candidate. The derivation's per-candidate DFS remains O(V+E) each over hoisted structures; a 2,000-model benchmark is deliberately out of scope for this fix.
11/12. Equivalent-mutant claim and label contract — confirmed by the review; no action.

Gates on the new head: 5945/5945 workspace tests, fmt clean; a further verification round runs before merge.

hugocorreia90:
## Rounds 2–3 disposition — converged

**Round 2 (the specificity/ephemeral/second-pass delta): APPROVE, all six items closed, no new findings** (independent concurrence). Round 3, scoped to the reciprocal-bare commit, confirmed the withdrawal mechanics, key direction, non-resurrection, and warning plumbing — and found exactly one defect: a bare candidate skipped while a soon-withdrawn guess was in the graph stayed stale (the three-way chain). Fixed as the reviewer prescribed: withdrawals now mark the pair ambiguous and **restart the decision pass to a fixpoint**, so every decision derives from a consistent graph (the monotone ambiguous set bounds restarts at one per contradicting pair). Pinned by the reviewer's own three-model construction, mutation-checked (fixpoint restart neutered → red).

**The remaining stale-read surface is deliberate and now has an owner**: contradicting bare guesses and unparseable models co-schedule with structured warnings and exit 0 — an orchestrator that ignores `scheduling_warnings` can still succeed after a stale read. Making those warnings fatal is a semantics change for existing projects, so it is filed as the explicit opt-in follow-up (#1355) rather than smuggled into this fix; the compiler's bare-`ModelRef` false edges are #1354, and live case-sensitivity reading is #1281.

Final gates: 5948/5948 workspace tests, clippy clean, fmt clean, base current. Merging on green checks.

Issue #1355, opened 2026-08-02
run: opt-in strict mode that fails on unresolved physical-read ordering warnings


hugocorreia90:
#1352 derives physical-read ordering edges and surfaces what it cannot safely resolve as structured warnings (`RunOutput.scheduling_warnings` / `DagRunOutput.warnings`): contradicting bare-read pairs (no edge derived), models whose SQL failed reference extraction, and colliding targets. The run still exits 0 — the warnings are advisory, and a consumer that ignores them can read stale physical targets and report success.

Follow-up decision: an opt-in `--strict-scheduling` (or `[execution] strict_scheduling = true`) that turns any non-empty scheduling warning into a run refusal, for estates that want fail-closed ordering. Kept out of #1352 deliberately: making warnings fatal changes run semantics for existing projects and needs a product call on the default.

Related roots: #1354 (compiler bare-ModelRef false edges), #1281 (read the live case-sensitivity setting).

hugocorreia90:
Design groundwork, from starting this and stopping deliberately partway. No code landed — a config key that silently does nothing is worse than no key — but the shape is settled and the cost is now known.

## `[run]`, not `[execution]`

The description offers `[execution] strict_scheduling`. That does not work: `ExecutionConfig` is **per-pipeline** (`pipeline.execution.concurrency`), and `rocky run --dag` derives physical-read ordering *across* pipelines, so a per-pipeline switch has no single answer on the path that most needs it.

`RunConfig` is project-level and already hosts exactly this shape — `skip_unchanged` is an opt-in default-`false` switch with a CLI flag that overrides it for one invocation. `[run] strict_scheduling` should follow it.

## Two enforcement points, and they must not drift

Warnings are produced in two places:

- **plain run** — `augment_physical_read_edges(...)` fills `output.scheduling_warnings` (`run.rs:7519-7523`)
- **`--dag`** — `label_report.warnings()` plus `physical_edges::derivation_warnings(&derived)` collect into `physical_edge_warnings` (`run_dag_exec.rs:307-309`), surfaced at `:395`

Both sit before execution, so refusing at either point is correct. They should call **one** shared function rather than each implementing the check — this repo has already been bitten by a flag meaning different things on the two paths (`--parallel`, #1288), and "strict" quietly differing by how the run was started is the same failure.

## The cost that stopped me

[...]

Issue #1354, opened 2026-08-02
compiler: a bare read matching an unrelated model's NAME creates a false edge that can suppress true physical-read ordering


hugocorreia90:
Found in #1352's review (finding 3). `classify_table_ref` resolves a bare single-part read to a `ModelRef` purely by model-name match. A model that bare-reads an EXTERNAL table (`FROM customers`) while an unrelated model happens to be NAMED `customers` (targeting, say, `prod.customers_v2`) gets a false compile-time dependency on it.

Pre-existing behavior — but #1352's run-time cycle guard trusts declared+compile edges, so the false edge can now cause a TRUE derived physical-read edge to be skipped as a cycle-closer: the reader runs before its real producer and reads stale data (the review's `a`/`customers`/`rollup` construction).

Direction to evaluate: bare `ModelRef` resolution should consider whether the name is shadowed by a physical table the model set can see (or at minimum, run-time derivation should be able to rank compile-derived bare-name edges below exact physical matches when resolving cycles). Needs the shadowing-semantics decision before implementation.

hugocorreia90:
Grounded while implementing this (PR #1628). **Option A is not shippable as specified**, for a reason the issue does not consider.

`rocky_engine::executor::execute_locally` materializes every model as `CREATE OR REPLACE TABLE <model name> AS <sql>`, walking `project.layers` and **ignoring the configured target**. `rocky test` and `rocky ci` run through it. On that path a bare `FROM customers` really does return model `customers`'s output, so the name edge is the only thing ordering the pair — dropping it makes the reader run first and fail. An existing test caught it: `closure_includes_physical_readers_of_seed_targets`.

Second reason: `semantic.rs` walks `project.execution_order`, and `SELECT *` expands an upstream only if its schema was inserted already. Removing an edge reorders that walk, so a `SELECT * FROM customers` reader can expand to nothing and fail its contract with **E010**. A compile that passes today would refuse.

So the graph has to encode one of two execution semantics and both are shipped. **The question that unblocks this issue:** should `execute_locally` materialize at the configured target?

- **Yes** — option A becomes shippable, and the compile graph can encode warehouse semantics for everyone.
- **No** — the name edge is correct on the local path, and the warehouse-only fix has to live below the compile graph, in both schedulers (#1629 is the second one).

[...]

PR #1356, opened 2026-08-02
fix(engine): label-based inference — all claimants ordered, parse failures surfaced, cycle-guarded

hugocorreia90:
Fixes the observable halves of #1351 in `run --dag`'s label-based dependency inference — **and deliberately nothing else**. Five review rounds established that ANY label edge added beyond main's single-slot graph (all-claimant fan-out, non-transformation-only fan-out, either pass order) can suppress an exact physical dependency through the shared cycle guards, because the cross-pass provenance contract does not exist yet. That contract is design issue **#1357**; the ordering improvements move there wholesale.

What ships:

- **The dependency graph is byte-for-byte main's.** The legacy winner's edge (last claimant in build order — identical to the single-slot `HashMap` insert) derives exactly as before; genuine reciprocal label reads keep their loud `execution_phases` refusal. Zero new edges, zero removed edges.
- **Label collisions become visible.** The producer index is a multi-map used as a collision detector: two-plus nodes sharing a lowercased label are reported, with the honest statement that only ONE of them orders the reader and the rest are NOT ordered — declare `depends_on` or rename.
- **Parse failures become visible.** A transformation whose SQL fails table-reference extraction previously contributed no edges with no signal; it now lands in the inference report.
- `infer_runtime_dependencies` returns a `LabelInferenceReport`; the caller merges its warnings into `DagRunOutput.warnings` alongside the physical-derivation warnings.

Evidence: collision-observability, clean-mutual-refusal (status quo pinned), and surfaced-parse-failure tests; the single-slot silent-drop and silent-continue regressions were mutation-checked red in earlier rounds and the final shape keeps those pins. Full workspace gates green; clippy `-D warnings` clean; fmt clean.

Review record: rounds 1–5 (BLOCK×4 on successive ordering schemes, each with a concrete reversal construction) are on this PR with dispositions — they are the design record for #1357. Related: #1352 (merged physical pass), #1354 (compiler bare-`ModelRef`), #1355 (opt-in strict scheduling), #1281 (live case-sensitivity).

### DIFF
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -6039,7 +6039,7 @@
             "type": "string"
           },
           "warnings": {
-            "description": "Scheduling warnings from physical-read edge derivation: mutual reads serialized by a deterministic order, models whose SQL could not be parsed for reads, and colliding targets. Empty when none.",
+            "description": "Scheduling warnings from dependency inference — physical-read derivation (mutual reads serialized deterministically, unparseable models, colliding targets) and label inference (label collisions where only one claimant orders the reader, unparseable models). Empty when none.",
             "items": {
               "type": "string"
             },
--- a/editors/vscode/src/types/generated/dag_run.ts
+++ b/editors/vscode/src/types/generated/dag_run.ts
@@ -40,7 +40,7 @@ export interface DagRunOutput {
   total_nodes: number;
   version: string;
   /**
-   * Scheduling warnings from physical-read edge derivation: mutual reads serialized by a deterministic order, models whose SQL could not be parsed for reads, and colliding targets. Empty when none.
+   * Scheduling warnings from dependency inference — physical-read derivation (mutual reads serialized deterministically, unparseable models, colliding targets) and label inference (label collisions where only one claimant orders the reader, unparseable models). Empty when none.
    */
   warnings?: string[];
   [k: string]: unknown;
--- a/engine/crates/rocky-cli/src/commands/run_dag_exec.rs
+++ b/engine/crates/rocky-cli/src/commands/run_dag_exec.rs
@@ -269,7 +269,7 @@ pub async fn run_with_dag(
         .flatten()
         .map(|m| (m.config.name.clone(), m.sql.clone()))
         .collect();
-    unified_dag::infer_runtime_dependencies(&mut dag, &sql_by_name);
+    let label_report = unified_dag::infer_runtime_dependencies(&mut dag, &sql_by_name);
 
     // Physical-read ordering (#1275): the label heuristic above is blind to
     // configured `[target]`s — a model reading another model's physical
@@ -285,7 +285,8 @@ pub async fn run_with_dag(
             .map(rocky_core::physical_edges::PhysicalEdgeModel::from_model)
             .collect();
     let derived = unified_dag::infer_physical_dependencies(&mut dag, &physical_inputs);
-    let physical_edge_warnings = rocky_core::physical_edges::derivation_warnings(&derived);
+    let mut physical_edge_warnings = label_report.warnings();
+    physical_edge_warnings.extend(rocky_core::physical_edges::derivation_warnings(&derived));
     for w in &physical_edge_warnings {
         tracing::warn!("{w}");
     }
--- a/engine/crates/rocky-cli/src/output.rs
+++ b/engine/crates/rocky-cli/src/output.rs
@@ -5328,9 +5328,11 @@ pub struct DagSummaryOutput {
 pub struct DagRunOutput {
     pub version: String,
     pub command: String,
-    /// Scheduling warnings from physical-read edge derivation: mutual reads
-    /// serialized by a deterministic order, models whose SQL could not be
-    /// parsed for reads, and colliding targets. Empty when none.
+    /// Scheduling warnings from dependency inference — physical-read
+    /// derivation (mutual reads serialized deterministically, unparseable
+    /// models, colliding targets) and label inference (label collisions
+    /// where only one claimant orders the reader, unparseable models).
+    /// Empty when none.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub warnings: Vec<String>,
     /// Total nodes across all layers.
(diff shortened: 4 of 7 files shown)


hugocorreia90:
Stop-time review catch, fixed before the main review round returns: the blanket cycle guard downgraded genuinely reciprocal label reads — which the single-slot heuristic always REFUSED loudly — into a silent stale-read success. The guard is now scoped to exactly the edges the all-claimant fan-out ADDS (claimants the old last-wins insert dropped): those can never introduce a refusal the old behavior did not have, while the legacy winner's edge inserts unguarded so a genuine SQL cycle keeps its loud refusal. Both semantics pinned (`mutual_label_reads_still_refuse_loudly`, `a_fanout_claimant_edge_never_introduces_a_new_refusal`), and the legacy discriminator is mutation-checked in both directions (forced-legacy and forced-fanout each go red).

hugocorreia90:
## Round-1 disposition

Both findings fixed:

1. **Combined-pass reversal** — the priority contract is now structural: `run --dag` runs the target-aware **physical pass first**, and the label pass distinguishes *why* a legacy edge closes. Closing against declared/build edges alone keeps the old loud refusal (genuine SQL cycle, status quo); closing **solely through physical-pass edges** means exact target-component evidence contradicts the label guess — the physical edge wins and the guess is skipped with a warning. Fan-out label additions stay guarded against the complete graph. Two regression tests run both passes in the caller's order: your reader/`shared`/Load construction (the fan-out side) and the sole-claimant variant (the legacy branch — which also fixes the one-claimant reversal you noted as inherited from #1352). The legacy branch's discriminator is mutation-checked red.
2. **"ALL claimants" warning** — reworded to "every claimant the cycle guard admitted".

Full workspace gates green on the new head; clippy/fmt clean; base current.

hugocorreia90:
## Scope decision after round 3 — shrunk to the uncontested core

Round 3 found the mirror of round 1: with physical-first ordering, **case-folding can manufacture a false tier-0 match on a case-sensitive store** (quoted `"Foo"` vs unquoted `FOO` are distinct objects), and the false physical edge then suppresses the true label edge. Together the rounds establish that *every* static cross-pass priority leaks in some corner, because identifier identity is connection state Rocky cannot observe (#1281) — label-first sacrifices the round-1 corner, physical-first sacrifices the round-3 corner.

Rather than trade corners a fourth time, the branch is **reset to the uncontested core** (`8cabab77`), which every round agreed on and which is strictly better than main with **no pass-order change**:

- all-claimant fan-out (no producer silently dropped), with fan-out edges guarded so they can never introduce a refusal the single-slot heuristic did not have — legacy edges land first, byte-for-byte the old graph;
- parse failures surfaced instead of silently skipped;
- collisions reported; honest warning text.

Both reversal corners **predate this PR** (the one-claimant label-vs-physical case is inherited from #1352's pass ordering, as round 1 noted) and are now tracked as the explicit cross-pass priority-contract design issue **#1357**, which needs #1281 (live case-sensitivity) and the #1354 compiler-edge question before any scheme can be sound. #1355's opt-in strict mode remains the fail-closed escape hatch.

Gates on the reset head were green when it was first pushed (5953/5953, clippy/fmt clean); CI reruns now.

hugocorreia90:
## Round-4 disposition

Both findings fixed on top of the shrunk core:

1. **Fan-out edge suppressing an exact physical dependency** — closed structurally, not by ordering policy: fan-out is now restricted to **non-transformation claimants** (seed/load/replication). Their edges never enter the physical pass's transformation-pair existing-projection, so the suppression your `A`/`a`/`r` construction exploited is impossible by construction. A transformation claimant that loses last-wins is deliberately left to the target-aware physical pass, which orders transformation pairs with strictly better evidence — your construction now derives exactly the true tier-0 edge and runs `r` before `A` (pinned as a test through both passes in caller order; the kind-filter is mutation-checked red). This keeps the PR's core promise honest: nothing in the diff is worse than main in any corner, because the only label edges added beyond main's graph can never conflict with physical evidence.
2. **Collision warning overclaim** — the reset had resurrected the "ALL claimants" text; it now states exactly what is ordered (legacy winner + admitted non-transformation claimants).

The cross-pass priority question for transformation pairs remains #1357, unchanged. Full gates rerunning; merging on green + this round's confirmation.

hugocorreia90:
One clarification for the record on the remaining suppression surface: after the round-4 fix, the only label edge that can still enter the physical pass's transformation-pair existing-projection is the **legacy winner's** edge — selected by `claimants.last()` over the same node order main's last-wins `HashMap::insert` produced, i.e. byte-for-byte the edge main derives today. Fan-out edges are non-transformation-only (structurally outside the projection: both endpoints must be transformation nodes, unified_dag.rs:1212-1214), so every label edge this PR ADDS beyond main's graph is incapable of suppressing physical evidence. The legacy transformation edge's conflict with physical evidence is main's inherited corner, owned by #1357 — not modified in either direction here.

hugocorreia90:
Round 5 (dispatched against the now-abandoned non-transformation fan-out variant) is superseded by the scope convergence above — its subject no longer exists on the branch. The final head adds zero edges, so the suppression class it was probing is structurally out of scope.

hugocorreia90:
## Round-6 disposition

The round's two blockers were both text-consistency classes, now closed:

1. **`depends_on` remediation** — already replaced one commit before the round returned: the warning now recommends renaming the colliding producers and states explicitly that `depends_on` cannot order a model against a seed or load.
2. **Contradictory documentation** — the producer-map comment, `LabelInferenceReport` Rustdoc, and `DagRunOutput.warnings` schema description all described the reverted all-claimant ordering or physical-only categories; all four now state the shipped observability-only behavior (bindings regenerated for the description change).

The round also independently **confirmed graph invariance** — the core claim — with terra concurrence, and all 54 unified_dag tests passing. On the "naturally built collision" evidence note: seed/load label collisions against a model are constructible only by hand-appending nodes precisely because `build_unified_dag` refuses duplicate cross-pipeline MODEL claims — the manufactured-node test is the reachable construction for the remaining kinds, and the merged CLI delivery path is the same `warnings` channel #1352's tests already exercise end-to-end.

Merging on green: the graph is byte-for-byte main's, everything main hid is now visible, and the six-round record is #1357's design input.

Issue #1348, opened 2026-08-02
bug(dagster/sdk): DAG discovery forces a whole-project --models, so a multi-transformation project is refused before any asset exists


hugocorreia90:
## Description

`RockyClient.dag()` always sends `--models self.models_dir` (`client.py:1136-1143`). The engine treats that as an **explicit whole-project override** and clones the one loaded list into *every* transformation pipeline:

```rust
// An explicit whole-project override: every transformation pipeline
// genuinely does declare this one directory — and a project with two of
// them is then refused by name ...
Some(dir) => {
    let models = load_all_models(dir)?;
    for (name, pipeline) in &cfg.pipelines {
        if matches!(pipeline, PipelineConfig::Transformation(_)) {
            by_pipeline.insert(name.clone(), models.clone());
        }
    }
```
— `engine/crates/rocky-cli/src/commands/dag.rs:141-158`

`build_unified_dag` then hits `reject_models_claimed_twice` (`unified_dag.rs:326`), because every model is now claimed by every transformation pipeline.

## Consequence

**A project with two or more transformation pipelines cannot produce a Dagster asset graph at all.** `RockyComponent` caches `rocky.dag(...)` (`component.py:874-879`, `972-985`), and that call errors before any definitions are built.

The engine's behaviour here is deliberate and defensible — if you *really* say "all pipelines use this one directory", then two pipelines really do both claim every model, and refusing is the honest answer. The defect is that the SDK sends that override **unconditionally**, so a caller who never asked for a whole-project override gets one.

## Why this is the blocker, not #1292

#1292 fixes adapter routing on the *execution* side. I initially wrote that it became reachable when #1261 landed. That was wrong: the scenario is still unreachable, because discovery refuses first. #1292's fix is a prerequisite, not a completion.

## Suggested direction

Distinguish "the caller left `models_dir` at its default" from "the caller explicitly wants a whole-project override", and have the component's DAG call omit `--models` so each pipeline resolves its own configured root (the `None` branch at `dag.rs:164-186`, which is what `rocky run --dag` already does).

A sentinel default (`models_dir: str | None = None`) is the honest shape, but it is a public field on `RockyClient` and touches every model-aware command, so it wants deciding rather than assuming.

## The coupled constraint — please do not fix this half alone

[...]

PR #1358, opened 2026-08-02
fix(engine): resolve a bare --model run's pipeline once, and refuse when it is ambiguous

### LINKED ISSUE #1350: bug(engine): the model-only pipeline fallback trusts the pipeline it picked for the adapter and models glob, but hardcodes its governance off
## Description

`rocky run --model X` with no `--pipeline` falls back to the first transformation pipeline. It then uses that pipeline for **two** of three fields and ignores it for the third:

```rust
if let Some(t) = rocky_cfg.pipelines.values().find_map(...Transformation...) {
    (t.target.adapter.clone(), false, Some(t.models.as_str()))
    //                          ^^^^^ not `t.target.governance.auto_create_schemas`
}
```
— `engine/crates/rocky-cli/src/commands/run.rs:1471-1475`

The explicit-pipeline branch twenty lines above reads it from the pipeline:

```rust
PipelineConfig::Transformation(t) => (
    t.target.adapter.clone(),
    t.target.governance.auto_create_schemas,
    Some(t.models.as_str()),
),
```
— `run.rs:1458-1462`

## Why it matters

`auto_create_schemas` issues `CREATE SCHEMA` DDL for the models an invocation will build (`run.rs:7460-7478`). So for the *same* project and the *same* single pipeline, `rocky run --model X` and `rocky run --pipeline P --model X` disagree about whether a missing target schema is created — even though the fallback resolved `P` and used its adapter and its `models` glob.

The full `rocky run` path honours the setting (`run.rs:2845`), so the model-only fallback is the odd one out.

## Which way is right?

Unclear from the code, and that is the point — there is no comment either way. Two readings:

- **The hardcoded `false` is a deliberate safety choice**: a model-scoped run should never emit DDL, whatever the config says. If so, it should say so, and arguably the *explicit* branch should be `false` too.
- **The hardcoded `false` is an oversight**: the fallback already committed to pipeline `t`, so honouring `t`'s governance is the consistent behaviour. The full run path agrees.

Either is defensible; being inconsistent is not, because it makes an unrelated flag (`--pipeline`, which reads as "route to the right adapter") silently decide whether DDL is emitted.

## How this surfaced

#1347 threads `--pipeline` through the Dagster asset path so models route to the right adapter. That is correct for routing, but it moves **every** Dagster deployment — including single-pipeline ones — from the fallback branch to the explicit branch, so any project with `auto_create_schemas = true` starts creating schemas where it previously failed.

I have documented that consequence in the dagster changelog rather than suppressing it, because honouring a setting the user wrote is the more defensible half of the inconsistency. But the inconsistency itself should be resolved here, in the engine, rather than depending on which flag a caller happens to pass.

## Acceptance

- The two branches agree about `auto_create_schemas`, in whichever direction is decided.
- The decision is stated in a comment, so the next person does not have to infer it.
- A test pins that `--model X` and `--pipeline P --model X` agree for a single-pipeline project.

## Related

`auto_create_catalogs` sits in the same `governance` block; worth checking whether it has the same split before fixing just the one field.

hugocorreia90:
Closes #1350.

## The bug

`rocky run --model <X>` without `--pipeline` resolves the adapter and the models glob from the first transformation pipeline, but hardcodes that pipeline's `auto_create_schemas` to `false`. So a model whose target schema does not exist yet fails on the bare form and succeeds on `--pipeline <that same pipeline> --model <X>`. One pipeline, two answers.

The literal was a fossil rather than a decision. Before #1305 the fallback captured only `t.target.adapter.clone()` from inside a `find_map`, so `t` was consumed and the governance flag was not in scope to read — the `false` was the only thing that could go there. Verified against `45843a01^`.

## The fix

The whole decision — adapter, governance, models glob — moves into one named function, `resolve_model_run_target`, that both branches go through. The bare form reads the pipeline's own `auto_create_schemas`, so it answers identically to naming that pipeline.

## Behaviour change: multi-transformation projects now refuse

That equivalence only holds when there is a single transformation pipeline to commit to. With several, "the first one" is a guess, and two things make the guess expensive:

- model→pipeline attribution can legitimately diverge from "first" (#1292, #1348), and
- the pipeline picked chooses the **adapter the model materializes on**.

So a wrong guess is not a wrong schema-creation flag, it is DDL against the wrong warehouse. Making the fallback *more* capable — governance on — without fixing attribution would escalate a read-only hole into a writing one.

The bare form therefore refuses when a project has several transformation pipelines:

> this project has multiple transformation pipelines, so a bare --model run cannot know which pipeline (and adapter) the model belongs to; pass --pipeline \<name\> to state it

**This breaks an invocation that works today.** A project with 2+ transformation pipelines running `rocky run --model X` currently succeeds against the first pipeline's adapter, without saying which one it chose; after this it errors. `--pipeline` and `--model` are independent parameters of the same entry point, so the workaround the message names is available. This is scope beyond #1350's literal ask — say the word and I'll split it out and land the governance fix alone.

Single-transformation projects, the overwhelmingly common shape, are unaffected except that governance now works.

## Tests

Three, all driving `resolve_model_run_target` rather than restating the rule:

- `model_only_fallback_matches_the_explicit_pipeline_answer` — asserts the fallback tuple **equals** the explicit-pipeline tuple, so it cannot pass by coincidence on one field.
- `multi_transformation_fallback_refuses_instead_of_guessing` — the bare form errors, and naming a pipeline still honors its governance.
- `replication_only_fallback_keeps_the_adapter_with_governance_off` — the third arm. The extraction rewrote it from an `or_else` chain into an `else`, so "prior answer preserved" is pinned by a test instead of asserted in prose.

## What I'm least confident about

Whether refusing beats warning-and-continuing. I chose refusal because the failure it prevents is silent and destructive rather than noisy, and because the right answer becomes knowable once #1292/#1348 land real attribution — at which point this refusal is replaced by resolution, not by a better guess.

## What I'm most likely missing

A project shape where several transformation pipelines exist but a bare `--model` run is nonetheless unambiguous — for example, several pipelines whose globs are disjoint and where the named model matches exactly one. That is precisely the attribution #1292/#1348 are about; this change treats it as unresolved rather than resolving it here, so such a project pays a refusal it could in principle avoid.


hugocorreia90:
## Round-1 disposition

All findings addressed:

- **Wrong-adapter DDL (the blocker)** — governance is now honored **only when the fallback's commitment is unambiguous**: exactly one transformation pipeline. With several, model→pipeline attribution can diverge from "first" (#1292/#1348), so schema creation stays off there and the operator is warned to pass `--pipeline`. The scope guard is pinned by a multi-transformation test (fallback off, explicit still honored) and mutation-checked red. The escalation the review constructed is no longer reachable: bare `--model` can emit DDL only where the resolved pipeline is provably the model's own.
- **Compile-failed models still trigger schema DDL** — real, pre-existing on both the explicit and fallback paths (the collection filters by name/set, not compile outcome); filed as **#1359** with the fix direction (collect from the post-compile executable set), rather than expanding this PR.
- **Doc misattachment** — `model_phase_ok` has its rustdoc back; the extracted resolver's doc now also states the sole-pipeline scoping.
- **Test anchoring** — the parity test binds all three fields via tuple equality; the sole-guard and fossil regressions are both mutation-checked.

`rocky plan` reachability and dry-run findings confirmed as stated (no change needed). Gates green on the new head.

hugocorreia90:
## Round-2 disposition — the blocker closed by refusal, not by narrowing

The round held the PR to the full wrong-adapter class (materialization DDL, not just schema DDL). Accepting that standard rather than rebutting on main-parity: bare `--model` on a **multi-transformation project now refuses outright** with a message asking for `--pipeline` — no execution on a guessed pipeline at all, so no DDL of any kind can reach a wrong adapter through the bare form. The sole-transformation shape (where the fallback's commitment is provable) honors governance as this PR intended. Behavior change stated plainly: multi-transformation bare `--model` runs, which today execute against a guessed first pipeline (the very defect #1292/#1348 document), become loud refusals — fail-visible over fail-wrong, the same contract the `--dag --shadow` path already applies.

Both semantics pinned (`multi_transformation_fallback_refuses_instead_of_guessing`, sole-pipeline parity), refusal message asserted, clippy `-D warnings` clean locally with CI's exact invocation (the earlier CI red was a stray blank line from the doc-adjacency fix — repaired). The two-adapter end-to-end attribution regression the round asked for belongs with the #1292/#1348 attribution fix, where attribution actually becomes computable.

hugocorreia90:
**Sequencing: merge after #1347.** The stop-time review's last catch is the caller side: today's `dagster_rocky`/SDK `run_model` invokes bare `--model` with no way to name a pipeline, so on multi-transformation projects this PR's refusal would turn their (already-wrong-adapter, #1292) execution into a hard failure with no workaround until the callers can comply. #1347 is exactly that caller fix — the dagster resource passes the node's owning `pipeline` through (its changelog cross-references this PR's #1350 interplay) — so the order is: **#1347 first, then this**. After #1347, dagster/SDK callers always name the pipeline and never hit the bare fallback; direct CLI users on multi-transformation projects get the loud refusal with a stated remedy instead of silent wrong-adapter DDL.

Holding this PR until #1347 lands; no further changes planned on this branch.

PR #1360, opened 2026-08-02
fix(engine): preview selects the base run from the ref the caller named

### LINKED ISSUE #1345: bug(engine): preview picks a base run by "not this branch" rather than by the base it was given
## Description

`rocky preview diff` takes a `--base` ref, then ignores it when choosing which run to compare against.

Both selection sites resolve the base as *"the newest run whose `git_branch` is not the feature branch"*:

```rust
let base_run = store
    .list_runs_matching(1, |r| r.git_branch.as_deref() != Some(branch_name))?
```

That keeps two kinds of run it arguably should not:

1. **A run on some unrelated third branch.** `--base main` will happily pair against the newest run on `experiment/foo` if that is more recent than any `main` run.
2. **A run with no branch at all.** `RunRecord::git_branch` is documented as `None` for "detached HEAD or non-git checkouts", which is not a branch and certainly not *the* base.

## Concrete failure

Runs, newest last: `t1` on `main`, `t2` detached (`git_branch: None`), `t3` on `feature`.

`rocky preview diff --name feature --base main` should compare `t3` against `t1`. It compares `t3` against **`t2`**, and the output labels the base as `main` — so the diff is against something else entirely, under a name that says otherwise.

## This is pre-existing

The predicate is a faithful port of the earlier loop's `else if` arm, which classified every non-matching run — `None` included — as base. #1332 changed *when* the predicate is applied (before the recency cap instead of after), not what it means, and deliberately did not widen its scope. So this is not a regression; it is a defect that the #1332 work preserved rather than fixed, and said so.

## Why it is worth separating

"Which runs are eligible to be the base" is a different question from "does the recency cap hide them", and it has its own answer. `run_preview_diff` already receives `base_ref` and threads it into the output label, so the information needed to select correctly is present and unused — the fix is likely a predicate change plus a decision about what to do when the named base has no recorded run.

## Suggested direction

Select on the base the caller named:

```rust
|r| r.git_branch.as_deref() == Some(base_ref)
```

with an explicit decision for the empty case. Falling back to "newest other branch" would restore today's behaviour as a *documented* fallback rather than an accident; reporting "no run recorded on `main`" is arguably better, since a diff against an unnamed run is not the diff that was asked for.

Worth confirming first that `base_ref` is always a branch name and not sometimes a sha or a tag — if it can be either, selection needs to handle both or the fallback needs to be explicit about which it handled.

## Acceptance

- A base run is selected from the ref the caller named, or the fallback is documented and named in the output.
- A detached-HEAD run (`git_branch: None`) is not silently chosen as the base for a named branch.
- A test with the three-run shape above.

## Provenance

Found by independent red-team review (Codex/GPT-5) of #1344, which agreed the port was faithful but pointed out that faithful is not the same as correct. I had flagged the `None` case there as my least-confident point and kept it deliberately; this is the follow-up.

hugocorreia90:
Fixes #1345. `rocky preview diff --base main` resolved its base run as "the newest run whose branch is not the feature branch" — happily pairing against an unrelated third branch or a detached-HEAD run (`git_branch: None`) while labeling the result `main`. The issue's exact repro (`t1` on main, newer `t2` detached, `t3` on feature) compared `t3` against `t2` under `main`'s name.

Selection now works from the ref the caller named, and only from it:

- **Primary**: the newest run whose recorded **branch equals `--base`, or whose recorded commit sha matches it** (exact, or git-style hex prefix ≥7) — Rocky's own preview workflow passes `github.event.pull_request.base.sha`, and records carry the sha in `git_commit`, so the checked-in workflow works under the named contract.
- **Missing base = refusal, not a stand-in**: when nothing matches, the comparison is not computed against anything else — empty summary, a `base_note` on `PreviewDiffOutput` explaining the absence and remedy, a `> ⚠️` callout in the markdown (reachable on exactly this path), and the VS Code command surfaces the note as a warning instead of a clean zero-diff message. The bisection path gates on **both** runs, so a refusal stays a zero-exit no-op instead of opening adapters.
- **Self-comparison refused**: `--name X --base X` previously diffed one run against itself and reported every model unchanged; it now fails fast with a clear message.
- **The cost preview is untouched**: its unnamed selection remains byte-for-byte main's (newest run not on the branch, detached CI cost baselines deliberately eligible).

Evidence: the issue's t1/t2/t3 repro; sha and hex-prefix base tests; refusal-with-note test; self-comparison refusal test; detached-eligibility pinned for the cost path; primary predicate, sha arm, and self-guard each mutation-checked red. Bindings regenerated for the optional `base_note` (omitted when absent; the SDK models it via the generated types — verified no hand-model parity needed).

### DIFF
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -12732,7 +12732,7 @@
         "description": "JSON output for `rocky preview diff`.\n\nCombines the structural diff (column added/removed/type-changed) from the existing `rocky_core::ci_diff` machinery with a row-level diff produced by either the sampled or the checksum-bisection algorithm. Each per-model entry's `algorithm` field carries a `kind` discriminator (`\"sampled\"` or `\"bisection\"`) plus the matching payload.\n\n**Sampled correctness ceiling.** The sampled algorithm reads `LIMIT N` rows ordered by primary key (or first column). Changes outside that window appear as no-change unless the row count itself differs; `sampling_window.coverage_warning = true` flags that risk. **Bisection** walks the chunk lattice exhaustively over a single-column primary key; `bisection_stats.depth_capped = true` flags the rare case where the recursion bottomed out at the depth cap before reaching leaf size. `summary.any_coverage_warning` rolls both signals up.",
         "properties": {
           "base_note": {
-            "description": "Present when no run was recorded on `base_ref` and the diff fell back to the newest run on another named branch — names the stand-in so the diff never silently wears the base's name (#1345).",
+            "description": "Present when no run was recorded on `base_ref`: the diff is NOT computed against any stand-in (an empty summary is returned) and this note explains the absence and the remedy (#1345).",
             "type": [
               "string",
               "null"
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -12731,6 +12731,13 @@
       "PreviewDiffOutput": {
         "description": "JSON output for `rocky preview diff`.\n\nCombines the structural diff (column added/removed/type-changed) from the existing `rocky_core::ci_diff` machinery with a row-level diff produced by either the sampled or the checksum-bisection algorithm. Each per-model entry's `algorithm` field carries a `kind` discriminator (`\"sampled\"` or `\"bisection\"`) plus the matching payload.\n\n**Sampled correctness ceiling.** The sampled algorithm reads `LIMIT N` rows ordered by primary key (or first column). Changes outside that window appear as no-change unless the row count itself differs; `sampling_window.coverage_warning = true` flags that risk. **Bisection** walks the chunk lattice exhaustively over a single-column primary key; `bisection_stats.depth_capped = true` flags the rare case where the recursion bottomed out at the depth cap before reaching leaf size. `summary.any_coverage_warning` rolls both signals up.",
         "properties": {
+          "base_note": {
+            "description": "Present when no run was recorded on `base_ref` and the diff fell back to the newest run on another named branch — names the stand-in so the diff never silently wears the base's name (#1345).",
+            "type": [
+              "string",
+              "null"
+            ]
+          },
           "base_ref": {
             "type": "string"
           },
--- a/editors/vscode/src/commands/preview.ts
+++ b/editors/vscode/src/commands/preview.ts
@@ -83,7 +83,11 @@ export async function previewDiff(): Promise<void> {
     const summary = result.summary;
     const message = `Rocky preview diff: ${summary?.models_with_changes ?? 0} changed, ${summary?.models_unchanged ?? 0} unchanged, +${summary?.total_rows_added ?? 0}/-${summary?.total_rows_removed ?? 0}/~${summary?.total_rows_changed ?? 0} rows`;
 
-    if (summary?.any_coverage_warning) {
+    if (result.base_note) {
+      // A refused comparison (no run recorded for the named base) must not
+      // read as a clean zero diff (#1345).
+      vscode.window.showWarningMessage(`Rocky preview diff: ${result.base_note}`);
+    } else if (summary?.any_coverage_warning) {
       vscode.window.showWarningMessage(`${message} (coverage warning)`);
     } else {
       vscode.window.showInformationMessage(message);
--- a/editors/vscode/src/types/generated/preview_diff.ts
+++ b/editors/vscode/src/types/generated/preview_diff.ts
@@ -31,7 +31,7 @@ export type PreviewModelDiffAlgorithm =
  */
 export interface PreviewDiffOutput {
   /**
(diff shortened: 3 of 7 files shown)


hugocorreia90:
Stop-time review escalation accepted before the main round: a stand-in comparison, however clearly labeled, is still a different diff that a CI pipeline would publish under the base's name. A missing named base now **refuses the comparison** — empty summary, `base_note` explaining the absence and the remedy, markdown carrying it — instead of falling back to any other run. The unnamed selection used by the cost preview keeps its newest-other-named-branch semantics with detached runs ineligible (same test pins both). This is also the direction #1345 itself called "arguably better".

hugocorreia90:
## Round-1 disposition

All five findings fixed on the branch (single amended commit on top of the refusal escalation):

1. **Bisection bypassed the refusal** — gated on both runs; a refusal is a zero-exit no-op, no adapter is opened.
2. **VS Code rendered refusal as a clean zero diff** — the command now surfaces `base_note` as a warning.
3. **The ⚠️ callout was unreachable on refusal** — the empty-models branch renders it; the PR-comment action posts it verbatim.
4. *(No action — the reviewer confirmed the sampled path's zero-exit refusal is sane.)*
5. **Cost preview regression** — the unnamed selection is restored byte-for-byte to main; detached CI cost baselines are deliberately eligible again, pinned in-test.
6. **SHA/tag bases always refused, breaking the checked-in workflow** — the primary match now covers recorded commit shas (exact or hex prefix ≥7); the workflow's `base.sha` resolves. Remedy text reworded to name both identities; the PR body no longer describes the abandoned stand-in fallback.
9. **Self-comparison false-clean** — `--name X --base X` refuses up front, pinned by an end-to-end test written while the guard was mutated (red), then reverted (green).

Gates: 5955/5955 workspace tests, clippy `-D warnings` clean, fmt clean. On the speculative detached-cost policy note: unchanged from main by point 5, so no policy is being set here.

hugocorreia90:
## Round-2 disposition

All three identity findings fixed and pinned:

- **(7) Precedence** — identity resolution is staged: branch equality first (a branch literally named like a hex string always wins), then exact commit (case-insensitive, closing the uppercase-sha note too), then hex prefix. Pinned by a test where the hex-named branch beats a NEWER commit-prefix match.
- **(9) Ambiguous prefixes** — a prefix matching two-plus distinct recorded commits refuses with a disambiguation note instead of silently taking the newest; mutation-checked (silent-resolution regression goes red).
- **(11) Sha aliases** — a named base that resolves to the branch's own newest run refuses the self-diff (`finish_named` compares selected run ids — the check the string-inequality guard cannot make); mutation-checked red.
- Non-blocking wording: the bisection skip log no longer claims the branch run is missing when the base is.

Gates green on the new head; clippy `-D warnings` clean; fmt clean.

hugocorreia90:
## Round-3 disposition

Both blockers fixed and mutation-verified:

- **Full-SHA branch hijack** — a 40-hex ref states commit intent (the workflow passes `base.sha` verbatim) and now resolves as a commit **before** any branch wearing the same 40-hex name; shorter refs keep branch-first so human-chosen names never lose to prefix coincidences. Pinned by the hijack construction (newer same-name branch loses); stage-removal mutation red.
- **Capped prefix concealment** — uniqueness is proven over **all** matching runs (the store iterates the table regardless; the 64 cap only limited the returned page and let 64 newer runs of one commit hide an older distinct sha). Pinned by the reviewer's own concealment construction (70 runs of one commit + one older distinct → refusal); cap-restoration mutation red.
- Test-binding gaps closed: case-insensitive exact matching pinned; the precedence and alias tests keep their earlier pins.

On the noted policy question (typed namespace for colliding identities): the staged heuristic — full-sha ⇒ commit, else branch, else commit, ambiguity refuses — is now stated in the resolver's comment as the contract; a `--base-commit`/`--base-branch` split would be the explicit follow-up if collisions prove common in practice.

Gates green; the unrelated `run_watch` CI failure was a runner flake (SIGINT/file-watch timing, untouched by this diff) and passed on rerun.

hugocorreia90:
One more stop-time catch closed before the confirmation round: commit intent now holds on an exact **miss** too — a full sha with no recorded commit refuses with commit-intent wording instead of falling through to the branch stage, where a branch wearing the same 40-hex name (recording different commits) could have hijacked the base after all. Pinned and mutation-checked red.

hugocorreia90:
Closing-round finding fixed: prefix uniqueness is now proven with **two limit-1 probes** (newest match, then any second *distinct* sha under the prefix) — exhaustive over the table with O(1) kept rows, so unbounded run retention cannot balloon the path. The stale 'branch always wins' comment now states the full-sha commit-intent stage. Both ambiguity tests (visible pair and the 70-run concealment) pass against the probe design; the second-probe guard is mutation-checked red.

PR #1362, opened 2026-08-02
feat(engine/rocky-bigquery): classify statement.kind on execute spans

hugocorreia90:
Closes #897. BigQuery was the lone warehouse adapter hardcoding `statement.kind = "query"` on its `statement.execute` spans (its `jobs.query` endpoint takes DDL/DML/SELECT alike, so the kind must come from the SQL text — the old TODO said as much). This ports the same keyword matcher the Snowflake (#540, the issue's stated template) and Databricks connectors use, applied to both the bespoke `statement.kind` and canonical `rocky.statement.kind` attributes, so dashboards filter ddl/dml/query traffic uniformly across every adapter.

Tests mirror the Snowflake suite (query/dml/ddl/other + leading-comment stripping), exactly the issue's acceptance bar. Evidence note, stated plainly: the tests bind the **classifier**; the two-line span wiring (`let kind = classify_statement_kind(sql)` into the two attributes) is not exercised by a tracing-capture harness — the same evidence shape as the merged Snowflake template. `cargo test -p rocky-bigquery` green; clippy `-D warnings` clean.

PR #1364, opened 2026-08-02
fix(engine): Snowflake and Databricks statement-kind strippers handle block comments

### LINKED ISSUE #1363: Snowflake/Databricks statement-kind strippers miss block comments
Found in #1362's review. The `classify_statement_kind` comment-strippers in `rocky-snowflake` (`connector.rs` ~623) and `rocky-databricks` (`connector.rs` ~1211) handle only `--` line comments; a statement led by a `/* ... */` block comment classifies as `other` on both adapters. The BigQuery port (#1362) handles `--`, `#`, and `/* */`, so the three matchers now diverge on this input class — this issue tracks bringing the two siblings up to the block-comment behavior (BigQuery's `#` form is GoogleSQL-specific and stays BQ-only). Mirror #1362's per-form tests.

hugocorreia90:
Closes #1363 (filed from #1362's review). A statement led by a `/* ... */` block comment classified `other` on the Snowflake and Databricks `statement.execute` spans while the BigQuery matcher (#1362) stripped it — the three adapters' matchers now agree on the shared comment forms (`--` and block; BigQuery's `#` form stays dialect-specific, as documented on its matcher). Unterminated blocks classify `other`, same as BigQuery.

Per-form tests on both connectors (Databricks previously had no classifier tests at all); the block-comment arm is mutation-checked red on Snowflake. Same evidence shape as #1362: the classifier is bound by tests, the span wiring is the pre-existing call sites, unchanged.

### DIFF
--- a/engine/crates/rocky-bigquery/src/connector.rs
+++ b/engine/crates/rocky-bigquery/src/connector.rs
@@ -2210,7 +2210,9 @@ mod tests {
 /// (#897). Divergence, on purpose: this stripper also handles BigQuery's
 /// `#` line comments; all three handle `/* */` blocks, and only the
 /// Databricks stripper tracks nesting depth (Spark SQL block comments
-/// nest; here and on Snowflake a nested opener is a syntax error).
+/// nest; here and on Snowflake the scan is flat — the comment ends at
+/// the first `*/` and an inner `/*` is inert text, live-verified on
+/// both, so this flat find matches the server's own lexer).
 fn classify_statement_kind(sql: &str) -> &'static str {
     let sql = strip_leading_sql_comments_and_whitespace(sql);
     let keyword = sql
--- a/engine/crates/rocky-bigquery/src/connector.rs
+++ b/engine/crates/rocky-bigquery/src/connector.rs
@@ -2208,9 +2208,9 @@ mod tests {
 /// `query` / `dml` / `ddl` / `other`. Mirrors the Snowflake and Databricks
 /// connectors' matcher so dashboards filter uniformly across adapters
 /// (#897). Divergence, on purpose: this stripper also handles BigQuery's
-/// `#` line comments and `/* */` blocks; the sibling strippers handle only
-/// `--` and share the `/* */` gap — tracked separately so the parity story
-/// stays explicit rather than silently uneven.
+/// `#` line comments; all three handle `/* */` blocks, and only the
+/// Databricks stripper tracks nesting depth (Spark SQL block comments
+/// nest; here and on Snowflake a nested opener is a syntax error).
 fn classify_statement_kind(sql: &str) -> &'static str {
     let sql = strip_leading_sql_comments_and_whitespace(sql);
     let keyword = sql
--- a/engine/crates/rocky-databricks/src/connector.rs
+++ b/engine/crates/rocky-databricks/src/connector.rs
@@ -1366,10 +1366,12 @@ fn strip_leading_sql_comments_and_whitespace(mut sql: &str) -> &str {
             continue;
         }
         if let Some(body) = trimmed.strip_prefix("/*") {
-            // Spark SQL block comments NEST (unlike Snowflake's and
-            // BigQuery's, where a nested opener is a syntax error) — track
-            // depth so `/* outer /* inner */ still-comment */ SELECT 1`
-            // classifies by its real first keyword.
+            // Spark SQL block comments NEST. Snowflake and BigQuery scan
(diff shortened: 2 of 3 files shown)


hugocorreia90:
Stop-time catch fixed: Spark SQL block comments **nest** (a nested opener is a syntax error on Snowflake and BigQuery, so their flat handling is dialect-correct); the Databricks stripper now tracks depth, with nested and nested-unterminated cases pinned. The per-dialect divergence stays documented at each matcher.

hugocorreia90:
## Review round 1 — disposition (verdict: BLOCK at freeze `fe837a92`; all findings addressed on the live head)

1. **Databricks nested block comments (CONFIRMED, the blocker)** — **fixed** in `7a8aa169`/`d5cacf3b` (pushed after the review froze): the Databricks stripper now tracks nesting depth in a single O(n) byte pass. Both reviewer cases are pinned: `/* outer /* inner */ still-comment */ SELECT 1` → `query`, and (added in `e330e779`) the exact fail-closed case `/* outer /* inner */ SELECT 1` → `other` (inner closes, outer never does).
2. **Snowflake `count_statements` ignores block comments (CONFIRMED, pre-existing)** — filed as #1365; it predates this telemetry-only diff (separate scanner from the classifier).
3. **Shared-input parity (CONFIRMED clean)** — no action; `#` remains BigQuery's sole intended line-comment divergence.
4. **Both new tests bind (CONFIRMED)** — matches the local mutation checks (Snowflake arm neutered → exit 100).
5. **Snowflake `//` line comments (SPECULATIVE, non-blocking)** — noted, not filed, per the reviewer's own bar: needs a live SQL API repro first. Tracked as a candidate live-verify.
6. **Stale BigQuery divergence note (CONFIRMED, non-blocking)** — **fixed** in `e330e779`: the doc now says all three handle `/* */`, only Databricks tracks nesting, and a nested opener is a syntax error on Snowflake/BigQuery.

The reviewer flagged that the post-freeze commits need their own pass — a scoped delta review of `fe837a92..HEAD` is dispatched.

hugocorreia90:
## Live grounding of the per-dialect block-comment semantics

The divergence notes previously claimed a nested `/*` opener is a **syntax error** on Snowflake and BigQuery. That mechanism claim was never grounded, so I ran parse-level probes against both live servers (Snowflake SQL API; BigQuery dry-run — both parse-only, no compute):

| Probe | Snowflake | BigQuery |
|---|---|---|
| `/* flat */ SELECT 1` (control) | success | dry-run OK |
| `/* outer /* inner */ still-comment */ SELECT 1` | `syntax error … unexpected 'still'` | `Syntax error: Unexpected identifier "still" at [1:22]` |
| `/* a /* b */ SELECT 1` | **success** | **dry-run OK** |

Both servers scan **flat**: the comment ends at the *first* `*/`, and an inner `/*` is inert text — probe 3 would be an unterminated-comment error under nesting, and it parses clean on both. So the Snowflake and BigQuery flat strippers match their servers' own lexers exactly, and only Databricks (Spark SQL, where comments genuinely nest per its SQL reference) needs the depth scanner. The code comments now state the flat mechanism instead of the refuted syntax-error one (`f8b85b6b`).

Consequence for #1365 (the pre-existing `count_statements` bug filed from round 1): the flat-scan fix sketched there is confirmed the right shape for Snowflake's lexer.

hugocorreia90:
## Delta review round — disposition (verdict: APPROVE at freeze `e330e779`)

The scoped delta review of the post-freeze commits confirmed the depth scanner is linear, boundary-safe, UTF-8-safe, and fail-closed (probing overlap sequences, final-byte closers, sub-2-byte bodies, and multibyte comment content directly against the compiled functions). Two non-blocking gaps, both now closed:

1. **Doc wording** — "a nested opener is a syntax error" was too absolute for Snowflake/BigQuery. Already replaced in `f8b85b6b` (post-dispatch) with the live-verified flat-scan mechanism; the reviewer's doc citations (Snowflake: `/*/` terminates a block comment; BigQuery: nested multiline comments unsupported) corroborate the probe table above.
2. **Unbound delimiter-consumption mutation** — an `i += 1` consumption mutant survived the nesting pins. Fixed in `034f4ab6`: two overlap inputs discriminate the arms (`/* /*/ */ SELECT 1` → `other` binds the opener; `/* /* */*/ SELECT 1` → `query` binds the closer). Both arms mutation-checked red locally (exit 100) and restored.

With round 1's blocker fixed and dispositioned, both independent rounds now approve the live head's substance. Merging on green CI.

Issue #897, opened 2026-06-14
Classify SQL statement kind in BigQuery connector tracing spans


hugocorreia90:
The Snowflake and Databricks connectors both classify each `statement.execute` tracing span's `statement.kind` / `rocky.statement.kind` attribute as `query` / `dml` / `ddl`, via a `classify_statement_kind(sql)` matcher. BigQuery is the lone warehouse-adapter holdout: it hardcodes `statement.kind = "query"` and carries an explicit `TODO`.

This is a **cross-adapter parity / observability-consistency** task, not a bug fix: with BigQuery classifying like the others, dashboards can filter ddl/dml/query traffic uniformly across every warehouse adapter. (The existing TODO hedges "if downstream consumers need to filter" — the justification is span-attribute consistency across adapters.)

> Mirror the merged Snowflake work in PR #540 (closed issue #481) — it's the template; this issue is the BigQuery twin.

## Scope

Port a `classify_statement_kind(sql) -> &'static str` matcher into `rocky-bigquery` and use it for both the `statement.kind` and `rocky.statement.kind` span attributes in `run_query`, replacing the hardcoded `"query"`.

## Code refs

- `engine/crates/rocky-bigquery/src/connector.rs` around line 287 — the `// TODO: classify via rocky-sql` marker and the hardcoded `statement.kind = "query"` in `run_query`
- `engine/crates/rocky-snowflake/src/connector.rs:623` — `fn classify_statement_kind` to mirror
- `engine/crates/rocky-snowflake/src/connector.rs:788+` — the query/dml/ddl unit tests to mirror
- `engine/crates/rocky-databricks/src/connector.rs:1211` — second in-repo reference implementation
- PR #540 — the merged Snowflake template

## Acceptance criteria

- Both `statement.kind` and `rocky.statement.kind` on the BigQuery `statement.execute` span reflect the classified kind (`query` / `dml` / `ddl`)
- Unit tests cover query, dml, and ddl classification (mirror the Snowflake tests)
- `cargo test -p rocky-bigquery` is green

## Effort

2 hours.

PR #1366, opened 2026-08-02
fix(engine/rocky-snowflake): comment-aware statement counting; strip // line comments

### LINKED ISSUE #1365: Snowflake count_statements miscounts on block comments containing semicolons — server rejects MULTI_STATEMENT_COUNT
## Symptom

A raw-SQL model whose text contains a **block comment with a semicolon in it** — e.g.

\`\`\`sql
/* note; kept on purpose */
SELECT 1 AS id
\`\`\`

fails at submission time on Snowflake: the server rejects the request with a MULTI_STATEMENT_COUNT mismatch.

## Cause

\`count_statements\` in \`engine/crates/rocky-snowflake/src/connector.rs\` (used to set Snowflake's \`MULTI_STATEMENT_COUNT\` session parameter) skips semicolons inside string literals and \`--\` line comments, but **not inside \`/* */\` block comments**. The comment text is preserved verbatim into the CTAS wrapper by \`engine/crates/rocky-core/src/sql_gen.rs\`, so the semicolon inside the comment is counted as a statement terminator. Snowflake then sees fewer real statements than the declared count and rejects the mismatch (its SQL API requires the exact count).

## Fix sketch

Teach \`count_statements\` to skip \`/* */\` spans the same way it skips \`--\` lines (flat scan is correct for Snowflake — nested block comments are a syntax error there). Pin with a test: \`/* note; */ SELECT 1\` counts as **1**.

Found during review of #1364 (statement-kind span parity); this defect predates that diff — the classifier and the counter are separate scanners.

hugocorreia90:
Closes #1365.

## What

Two scanners in the Snowflake connector mishandled comments:

1. **`count_statements`** (feeds `MULTI_STATEMENT_COUNT`) skipped semicolons inside string literals and `--` line comments, but counted them inside `/* */` block comments and `// ...` line comments. A raw model containing `/* note; */ SELECT 1` was counted as 2 statements, so the declared count overshot what Snowflake's lexer sees and the server rejected the whole submission (its SQL API requires the exact count). Not cosmetic — a valid model fails to run.
2. **`classify_statement_kind`'s stripper** (the `statement.kind` span attribute, #897) didn't strip `//` line comments, so `// note\nSELECT 1` classified as `other`.

## Grounding

- `//` line comments: **live-verified accepted by the Snowflake SQL API itself** (`// note\nSELECT 1` executes successfully), not just by the Python connector's client-side splitter. Neither Spark SQL nor GoogleSQL documents `//`, so this stays Snowflake-only — the sibling strippers are untouched.
- Block-comment scan is **flat** (close at first `*/`, inner `/*` inert): live-verified on both Snowflake and BigQuery during #1364 (nested probe → `unexpected 'still'`; `/* a /* b */ SELECT 1` parses clean). The counter uses the same flat close, so counted statements match what the server lexer sees, including the `/* a /* b */ SELECT 1; SELECT 2` → 2 case.

## Tests

- `#1365` shape pinned: `/* note; kept */ SELECT 1` → 1; multi-statement, multi-line, flat-close, unterminated-swallows-rest, and string-literal-lookalike cases.
- `//`: counter and classifier cases, including a comment swallowing the statement's own terminator.
- All three new arms mutation-checked red locally (neutered guard / dropped `or_else` → exit 100) and restored.

### DIFF
--- a/engine/crates/rocky-snowflake/src/connector.rs
+++ b/engine/crates/rocky-snowflake/src/connector.rs
@@ -1062,6 +1062,24 @@ mod tests {
         assert_eq!(count_statements(r#"SELECT 1 AS "a;b"; SELECT 2"#), 2);
         assert_eq!(count_statements(r#"SELECT 1 AS "a$$b"; SELECT 2"#), 2);
         assert_eq!(count_statements(r#"SELECT 1 AS "a;b$$c""d"; SELECT 2"#), 2);
+        // Comment markers inside a quoted identifier are literal too
+        // (live-verified: SELECT 1 AS "/*x*/" executes).
+        assert_eq!(count_statements(r#"SELECT 1 AS "/*"; SELECT 2"#), 2);
+        assert_eq!(count_statements(r#"SELECT 1 AS "//"; SELECT 2"#), 2);
+    }
+
+    #[test]
+    fn test_count_statements_wrapped_transaction_with_protected_regions() {
+        // The submission path prepends ALTER SESSION … and RECOUNTS the
+        // wrapped SQL (see submit above): for a body whose statements
+        // carry protected lexical regions, the declared
+        // MULTI_STATEMENT_COUNT must be the real count + 1 — neither
+        // swallowed (comment arms firing inside $$/"…") nor inflated
+        // (semicolons inside them counted).
+        let body = "BEGIN;\nCREATE FUNCTION f() RETURNS INT AS $$ SELECT 1; -- x\n $$;\nSELECT 1 AS \"/*\";\nCOMMIT;";
+        assert_eq!(count_statements(body), 4);
+        let wrapped = format!("ALTER SESSION SET TRANSACTION_ABORT_ON_ERROR = TRUE;\n{body}");
+        assert_eq!(count_statements(&wrapped), 5);
     }
 
     #[test]
--- a/engine/crates/rocky-snowflake/src/connector.rs
+++ b/engine/crates/rocky-snowflake/src/connector.rs
@@ -568,16 +568,21 @@ impl SnowflakeConnector {
 ///
 /// Counts semicolons that terminate distinct statements, then adds 1 for the
 /// final un-terminated statement. Skips semicolons inside SQL string literals
-/// (`'...'`), single-line comments (`-- ...` and `// ...` — Snowflake accepts
-/// both), and `/* ... */` block comments (flat scan: Snowflake's lexer ends a
-/// block comment at the FIRST `*/`; an inner `/*` is inert text —
-/// live-verified). A miscount here is not cosmetic: Snowflake rejects a
-/// submission whose real statement count differs from `MULTI_STATEMENT_COUNT`
-/// (#1365). Dollar-quoted string constants (`$$ ... $$` — Snowflake supports
-/// only the bare form; a `$tag$` opener is a syntax error, live-verified) are
-/// opaque: comment markers, quotes, and semicolons inside them are literal
-/// text. Rocky-generated SQL never embeds semicolons in identifiers, so
-/// identifier-quoting is not considered.
+/// (`'...'`), double-quoted identifiers (`"..."`, `""` escape), single-line
+/// comments (`-- ...` and `// ...` — Snowflake accepts both), and `/* ... */`
+/// block comments (flat scan: Snowflake's lexer ends a block comment at the
+/// FIRST `*/`; an inner `/*` is inert text — live-verified). A miscount here
+/// is not cosmetic: Snowflake rejects a submission whose real statement count
+/// differs from `MULTI_STATEMENT_COUNT` (#1365).
+///
+/// Dollar-quoted string constants (`$$ ... $$` — bare form only; a `$tag$`
+/// opener is a syntax error, live-verified) are opaque: comment markers,
+/// quotes, and semicolons inside them are literal text. `$$` opens a string
+/// only at a TOKEN BOUNDARY: `$` is a valid identifier character, so
+/// `a$$b` and `a$$` are single identifiers (live-verified), while after a
+/// bare numeric token `1$$` the server DOES open a string (live-verified:
+/// unterminated parse error) — tracked via which character started the
+/// current word token.
 ///
 /// Trailing-only whitespace after the last semicolon counts the same as a
 /// terminator (one fewer statement). Empty / whitespace-only input returns 0.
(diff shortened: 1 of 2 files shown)


hugocorreia90:
Stop-time review catch, fixed in `6ba05cbb`: **dollar-quoted string bodies are opaque** — comment markers inside `$$ … $$` are literal text, but the counter's comment arms (the new `//` and `/* */` ones *and* the pre-existing `--` arm) swallowed from inside the body, past the closer and any real terminator, undercounting valid SQL such as `CREATE FUNCTION … AS $$ … $$` bodies.

Live-verified on the SQL API before implementing: `--`, `//`, `/* */`, `'`, and `;` inside `$$ … $$` all execute as literal content (each probe would fail if the server commented inside the body), `$$` inside a normal string stays inert, and a `$tag$` opener is a **syntax error** — Snowflake's grammar is the bare form only, so the scanner needs no tag machinery. The new `$` arm is mutation-checked red (neutered guard → exit 100) and pinned by seven cases including the undercount shape (`SELECT $$ a // b $$; SELECT 2` → 2).

hugocorreia90:
Second stop-time catch, fixed in `a07ca7bf`: the dollar-quote opener was **token-boundary blind** — `$` is a valid identifier character, so `a$$b` and `a$$` are single identifiers and the previous commit's arm swallowed from mid-identifier past real terminators (undercount → server rejection of valid SQL).

Live probes drove the exact rule:

| Probe | Result | Lexer conclusion |
|---|---|---|
| `SELECT 1 AS a$$b` / `AS a$$` | OK | `$$` continues a word token begun by a letter/underscore |
| `SELECT 1$$ AS x` | unterminated-string parse error | after a bare numeric token `$$` **does** open a string |
| `SELECT ($$x$$) AS c` | OK | punctuation is a token boundary — opens |
| `SELECT 1 AS "a;b$$c""d"` | OK | quoted identifiers may contain `;`, `$$`, `""`-escape |

The counter now opens `$$` only at token boundaries (tracking whether the current word began with a letter/underscore, so the numeric case matches the server), and double-quoted identifiers are opaque — closing the doc's previously stated "identifier-quoting is not considered" gap rather than letting it interact with the `$` state. Three new mutation checks red (guard dropped, absorb-rule widened, dquote arm neutered); 174/174 crate tests green.

hugocorreia90:
Third stop-time claim — **refuted by live probes, pinned in `22e30973`**: the claim was that `$$`-suffixed session variables still undercount. The server's own lexing says otherwise:

| Probe | Server result |
|---|---|
| `SELECT $a` (control) | resolution error: *Session variable '$A' does not exist* |
| `SELECT $a$$ x $$` | syntax error: *unexpected '$$ x $$'* — i.e. `$a` followed by a **lexed string token** |
| `SELECT $a$$b` | *parse error near '<EOF>'* — the `$$` opened a string that never closes |

So a `$`-started variable token does **not** absorb a following `$$` — the server opens a string at exactly that boundary, which is what the scanner already does (`word_absorbs_dollar` is true only for letter/underscore-started words). Scanner and lexer agree on every probed shape, and both juxtaposition shapes are syntax errors server-side regardless of count, so no valid SQL diverges. Two pins added so a future "fix" can't reintroduce absorption.

hugocorreia90:
## Round 1 — disposition (verdict: BLOCK at freeze `4d07a6ba`, the PR's first commit; all findings closed on the live head)

The blocker — comment arms firing inside `$$…$$` literals and `"…"` identifiers — is exactly what the incremental stop-time reviews caught while round 1 ran, and the protected-region commits close it:

- `6ba05cbb` made dollar-quoted bodies opaque; the token-boundary commit corrected the opener (`$` is an identifier character) and made double-quoted identifiers opaque with `""` escape.
- Every probe in the finding now counts correctly and is pinned verbatim: `SELECT $$/*$$; SELECT 2` → 2 (dollar-opaque tests), `SELECT 1 AS "/*"; SELECT 2` and `AS "//"` → 2 (added in `1f723ca6`, with `SELECT 1 AS "/*x*/"` live-verified executable).
- The wrapped-transaction concern is pinned at the recount seam (`1f723ca6`): a protected-region body declares real count + 1 after the `ALTER SESSION` prefix — the reviewer's 4-vs-5 shape.
- The reviewer's "least confident" item (live API behavior of adversarial payloads) was already settled ahead of the verdict: comment markers inside `$$` bodies and quoted identifiers execute successfully (probe tables above).

The post-freeze commits get their own scoped delta review before merge, as with #1364.

hugocorreia90:
## Delta round — disposition (verdict: BLOCK at freeze `1f723ca6`; blocker fixed in `f53dbb5b`)

The delta round confirmed the protected-region machinery (states exclusive, token-boundary opener matching the live lexer on `a$$b`/`1$$`/punctuation, no behavior change on unprotected inputs, wrapped recount arithmetic sound) and surfaced one new blocker plus one gap, both closed:

1. **Backslash escapes in string literals (CONFIRMED blocker)** — the `in_string` scanner knew only `''`-doubling, so `\'` closed the string early and the new dollar opener fired inside it: `SELECT 'a\'$$; still string'; SELECT 2` counted 1. Live-verified the premise before fixing (`SELECT 'a\'$$; still string'` executes as one string; `SELECT 'a\\'` closes; and in **quoted identifiers** backslash is literal — `SELECT 1 AS "a\"` executes — so the dquote handler stays doubling-only, now pinned). The backslash arm consumes the escaped char; mutation-checked red (the new test fails with the arm neutered, exit 100) and restored.
2. **Wire-level assertion (the round's "most important missing")** — added: a three-statement body carrying all trap classes (backslash escape before a phantom `$$`, dollar body, quoted identifier) submits `MULTI_STATEMENT_COUNT=4` with the `ALTER SESSION` prefix intact, asserted on the captured wiremock request.

The reviewer's summary nit (my prompt described the dquote pins as `--` + full `/* */` while the tests pin `"/*"` and `"//"`) is noted — the tests are the source of truth. 206/206 crate tests green with all features.

hugocorreia90:
## Scoped round on the backslash commit — disposition (verdict: BLOCK, sole blocker fixed in `fb39343f`)

The round confirmed the backslash arm itself (escape interactions with `''`-doubling walked, terminal-backslash safe, a ~600k-input differential over backslash-free strings found zero divergences vs the previous head, wiremock arithmetic recomputed) and one real finding: **the quoted-identifier pin was vacuous** — an escape-layer slip in my patch tooling dropped the backslash byte, so the committed test was the trivial `"a"` case. The reviewer's walk of the *intended* text matches the code's behavior; `fb39343f` puts the actual byte in the test, which now binds the backslash-is-literal-in-quoted-identifiers semantics (adding wrong backslash-escaping to the dquote handler flips it red).

The round's sandbox couldn't bind the wiremock port; the wire test runs green in CI (all-features lane) and locally.

With that closed, every finding across all three rounds is dispositioned: merging on green CI with the base-currency check.

Issue #1242, opened 2026-07-26
Migrate the gc and apply ledger seams to compare-and-swap (finish #1228)


hugocorreia90:
Follow-up to #1241, which added `LedgerSeamSession` and migrated the `rocky policy` freeze/unfreeze seam. Two seams still publish the shared state object unconditionally, so #1228's data-loss path remains open.

## Remaining writers

Both reach `RemoteStateSession::upload_only_fail_closed`, which forces upload errors to propagate and then delegates to the unconditional `upload_state`:

- `rocky gc` — `commands/gc.rs`
- the shared apply helper `upload_remote_ledger_fail_closed` — `commands/apply.rs`, used by the governed run-plan pre-decision and post-custody, the AI-authored run-plan pre-decision, and restore

## Why these are not a repeat of the policy migration

The policy seam's ledger records have trivial preconditions, so replaying them verbatim onto the winner is correct. Neither of these does.

**`gc apply`** is replayable only by re-running `execute_gc_apply` end to end. Eviction is redb-only — `physical_delete = true` is rejected, and `evict_artifact` atomically removes the exact hash-matched row and inserts its tombstone — but the decision is authorized by an external proof. A winner that re-adds the Delta path legitimately flips a tombstone into a hold, so each attempt must redo fresh candidate and refcount derivation and both liveness reads. That is expensive: the oracle lists `_delta_log` and GETs every JSON commit, twice per eligible candidate, per attempt.

**Restore** needs its own transition: fresh download, re-resolve the live tombstone, confirm its identity still matches the captured artifact hash, re-hash the object already present, then re-run `restore_artifact`. No warehouse re-query is needed if the existing bytes remain hash-exact for the still-live tombstone; if that proof cannot be repeated, refuse rather than retry.

**Budget-burn custody** does not double-burn — consumption counts distinct failing `plan_id`s, not decision rows — but the verify-after constructor calls `Utc::now()`, so a reconstructed record would duplicate the audit row. Replay must preserve the exact constructed record.

## Also required

- **CAS does not close the gate-to-mutation TOCTOU.** Freeze markers are written under their own keys and never advance the shared object's generation, so a marker landing after a seam's LIST raises no conflict. Every irreversible mutation needs its own fresh pre-mutation marker fence.
- **The pre-run governed decision must re-evaluate after a conflict.** If the winner added a budget failure, replaying the original `Allow` verbatim would authorize a run against a budget view that no longer exists. Post-run custody can be replayed verbatim.
[...]

PR #1372, opened 2026-08-03
feat(engine): route the gc apply seam through the CAS ledger session (#1242 PR-1)

hugocorreia90:
First seam of #1242 (finishing #1228's data-loss class; ADR-CONCURRENCY D1, seam class = RETRY). Migrates `rocky gc` apply — the seam the issue's failure walkthrough uses — off the unconditional `upload_only_fail_closed` half-seam.

## What

Under effective CAS (`concurrency_control = "cas"` on a CAS-capable backend), the gc publish goes through `LedgerSeamSession`: the **whole transition replays per attempt** against the freshly downloaded winner. `execute_gc_apply` re-derives candidates, refcounts, and both liveness reads from the fresh store, so a winner that re-added the Delta path legitimately flips that attempt's eviction into a refusal. Exhaustion (3 conflicts) exits nonzero with the typed `LedgerSeamConflict` and preserves the winner. Without effective CAS the legacy fail-closed half-seam is byte-for-byte unchanged (#1228's residual exposure — one writer per `[state]` prefix — stands until fleet-wide enablement; landing this closes nothing operationally by itself).

**Freeze-marker fence**: markers live under their own keys, where blob CAS cannot see one landing mid-seam (the issue's TOCTOU paragraph). The seam re-LISTs at attempt start and again after the transition (so the unfenced window shrinks to the CAS upload itself) and **refuses** on any marker the pre-gate set never evaluated — over-enforcement is the monotone-safe direction. gc's fence can be per-attempt rather than per-row because gc's mutations are redb rows inside the blob being CAS-published: `physical_delete = true` is refused at entry, so no per-candidate side effect outlives a discarded attempt (restore is different and gets per-mutation fences in its own PR).

Supporting changes: `StateSyncError::SeamTransition` (a session caller's domain refusal aborts without publishing; non-retryable), and the oracle parameter becomes `Arc<dyn LivenessOracle>` (the session's `for<'a>` bound requires the attempt future to own its captures).

## Tests

- **Mid-seam interleaving (the #1228 schedule)**: the run winner publishes between gc's download and its CAS put — event-keyed on the armed conflict's put count, deadline-bounded so the legacy path fails fast instead of hanging. Asserts the published blob holds BOTH effects and a counting oracle pins per-attempt liveness re-derivation (6 reads = 3 × 2 attempts). 10/10 repeated runs green.
- **Replay mechanics**: two armed conflicts → exactly three CAS attempts, zero unconditional puts, eviction present once.
- **Exhaustion**: three conflicts → nonzero with `LedgerSeamConflict`, remote winner untouched.
- **Fence**: new-marker refusal (names the marker), already-gated pass, LIST transport failure refuses fail-closed.
- Mutation checks (both red, restores verified): forcing `seam_cas = false` fails the mid-seam test on the watcher-fired assertion; neutering the fence's refusal predicate fails the fence test.

Remaining #1242 seams (restore; apply pre-decision + custody; structural removal of `upload_only_fail_closed` + doctor flip) land as follow-up PRs per the reviewed migration plan.

### DIFF
--- a/engine/crates/rocky-cli/src/commands/apply.rs
+++ b/engine/crates/rocky-cli/src/commands/apply.rs
@@ -1453,7 +1453,7 @@ pub(crate) fn evaluate_apply_policy_with_store(
 /// Given an ALREADY-RESOLVED `[policy]` block, compile the per-model attributes,
 /// or return an early [`PolicyGate`] — `NotConfigured` when there is no policy,
 /// `Allow` when `touched` is empty (a genuine no-op executes nothing).
-fn resolve_policy_and_attrs(
+pub(crate) fn resolve_policy_and_attrs(
     policy: Option<&rocky_core::config::PolicyConfig>,
     touched: &BTreeMap<String, PolicyCapability>,
     models_dir: &Path,
@@ -1494,7 +1494,7 @@ fn resolve_policy_and_attrs(
 /// supplied by the caller so the two variants differ only in how they reach the
 /// ledger (fresh open vs. a held handle).
 #[allow(clippy::too_many_arguments)]
-fn evaluate_apply_policy_core(
+pub(crate) fn evaluate_apply_policy_core(
     policy: &rocky_core::config::PolicyConfig,
     plan_id: &str,
     principal: PolicyPrincipal,
--- a/engine/crates/rocky-cli/src/commands/gc.rs
+++ b/engine/crates/rocky-cli/src/commands/gc.rs
@@ -1518,6 +1518,7 @@ async fn gc_seam_regate(
     principal: rocky_core::config::PolicyPrincipal,
     touched: &BTreeMap<String, PolicyCapability>,
     models_dir: &Path,
+    models_glob: Option<&str>,
     prior_decisions: &[rocky_core::state::PolicyDecisionRecord],
     fresh_store: Option<&StateStore>,
     stage: &str,
@@ -1540,7 +1541,7 @@ async fn gc_seam_regate(
         Some(policy),
         touched,
         models_dir,
-        None,
+        models_glob,
     ) {
         Ok(pair) => pair,
         // The resolver's error IS a gate outcome. Mirror the pre-seam
@@ -1820,6 +1821,7 @@ pub(crate) async fn run_gc_apply_in_with(
         let seam_cfg = loaded_cfg.clone();
         let seam_touched = touched.clone();
         let seam_models_dir = models_dir.clone();
+        let seam_models_glob = models_glob.clone();
         let seam_principal = plan_record.enforcement_principal(runtime_principal);
         let seam_plan = plan.clone();
         let seam_plan_id = plan_id.to_string();
@@ -1829,6 +1831,7 @@ pub(crate) async fn run_gc_apply_in_with(
                 let cfg = seam_cfg.clone();
                 let touched = seam_touched.clone();
                 let models_dir = seam_models_dir.clone();
+                let models_glob = seam_models_glob.clone();
                 let principal = seam_principal;
                 let plan = seam_plan.clone();
                 let plan_id = seam_plan_id.clone();
@@ -1848,6 +1851,7 @@ pub(crate) async fn run_gc_apply_in_with(
                         principal,
                         &touched,
                         &models_dir,
+                        models_glob.as_deref(),
                         &prior_decisions,
                         Some(fresh_store),
                         "during this gc apply",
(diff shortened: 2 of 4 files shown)


hugocorreia90:
## Round 1 — disposition (verdict: BLOCK; all four blockers closed on the live head)

1. **Terminal-failure ghost state (CONFIRMED)** — fixed in the session: `LedgerSeamSession` now restores the remote winner on **every** terminal failure (transition error, non-conflict upload error, and the pre-existing conflict-exhaustion arm, factored into one helper). Bound by a new deterministic test: the liveness oracle itself writes a freeze marker mid-transition (it runs between the attempt-start gate and the pre-publish recheck — no timing), so the deny lands after the attempt evicted locally; the ghost tombstone must be rolled back. Mutation-checked red (restore neutered → \"a refused post-eviction attempt must not leave a local ghost tombstone\").
2. **Ledger-only freeze bypass / no principal-scope matching (CONFIRMED)** — the marker-ID fence is replaced by the **real gate per attempt**: `evaluate_apply_policy_core` over the fresh store's decision snapshot + a fresh marker LIST, with principal/scope matching intact (an Agent-scoped marker no longer refuses a Human gc — your availability point), plus an enforcement-only pre-publish recheck using the pre-attempt snapshot (an attempt must not read its own rows back as budget history). Integration-bound: the mid-seam winner publishes a **ledger-only** freeze row (no marker anywhere) and the replay refuses with the freeze's own reason; regate mutation red.
3. **Audit-row erasure (CONFIRMED)** — the per-attempt gate records its decision rows into the attempt's fresh store, so the trail publishes atomically with the winning attempt; the pre-seam gate remains fail-fast only (its local rows are overwritten by each attempt's download, by design — now documented).
4. **Statically incorrect test (CONFIRMED, corrected with evidence)** — you were right: a single-attempt probe measured **2 oracle reads per attempt** (your static count), and the observed 6 is 2 × **three attempts** — the armed conflict plus a *genuine* winner race in attempt 2's window. The test now pins the attempt structure explicitly (CAS put count == 3) and the comment states the real decomposition. The \"≥20ms\" jitter claim is corrected to 18–22ms.

Also per your finding 5: the fidelity/oracle analysis is accepted as-is (no action); the CAS-vs-retry_transient wrapper asymmetry is noted as pre-existing session design, not introduced here.

Full workspace suite green; scoped delta review of `1ba3ceb3`'s changes dispatched next.

hugocorreia90:
## Delta round — disposition (verdict: BLOCK; all three blockers closed, plus the CI evidence)

1. **`models_glob` dropped in the re-gate (CONFIRMED)** — real regression, fixed: the configured glob now threads through the closure into both re-gates, so replay attrs resolution matches the command gate.
2. **Terminal rollback could still expose ghost state when the restore download fails (CONFIRMED)** — the failure contract is now explicit: if the winner cannot be re-downloaded, the local file is **quarantined aside** (`.redb.unpublished-<pid>`) — path-opening readers see absence, never uncommitted rows; evidence preserved; the next start-download recreates the file. Bound by a fully sequenced test (the oracle writes the denying marker AND arms a total GET outage), mutation-checked red (rename neutered → "the unpublishable local file must be quarantined aside").
3. **Scheduler-dependent race test (CONFIRMED — and CI proved it)** — the watcher-based ledger-freeze test lost its race on the CI runner in the very run that carried your verdict (the predicted 2-attempt schedule). Both interleaved tests now use **oracle-sequenced injection**: the winner (or freeze row) is published from inside attempt 2's first liveness read — after its download, before its CAS put — so the genuine conflict and the third attempt are structural. No tasks, no deadlines, no timing. 5× repeated runs green; attempt counts pinned (3 puts / 6 reads for the preserves test; 4 reads for the freeze test, whose denied attempt 3 must not consult the oracle).

Your round-1 closure notes are also reflected: (a) restore now covers the restore-failure arm via quarantine; (c) best-effort audit recording stands for gc (no custody row exists to under-count — your pre-existing-ambiguity read), documented at the sink.

Full workspace suite + clippy green on the head; base-currency check and merge follow CI settle.

hugocorreia90:
Follow-up to the delta disposition, from the stop-time reviewer: the previous rollback shape left the uncommitted file readable during the restore download. `restore_remote_winner` now quarantines FIRST (atomic rename — the recovery window closes at that instant), then downloads the winner into a fresh file; the quarantined copy is deleted on success, kept for forensics on failure. Both end-state tests (ledger-freeze restore, GET-outage quarantine) still bind; the ordering is the delta.

PR #1379, opened 2026-08-04
fix(scripts): soak harness runs under jq 1.6; meta write fails loud

hugocorreia90:
Two portability defects in `scripts/soak-scheduler.sh`, found by the containerized aarch64-linux preflight arm for the soak report's carried Linux caveat:

1. **The meta.json jq filter breaks under jq 1.6.** It bound a jq variable named `$end` — `end` is a jq keyword; 1.7 (macOS Homebrew) tolerates keyword-named variables, 1.6 (Debian bookworm) rejects the filter at parse time (`syntax error, unexpected end, expecting IDENT`). Renamed to `$end_ts`; the emitted `planned_end` key is unchanged, so `soak_verdict.py` reads the same shape.
2. **The failure was silent.** The redirect had already truncated `meta.json` to zero bytes, the harness soaked the full arm, and only the verdict — 1h (or 24h) later — surfaced INVALID. The write now `die`s immediately on a jq failure or an empty output file: never soak against a broken instrument.

Evidence: the same 1h Linux arm otherwise ran clean end-to-end (preflight checks all OK, 120/120 samples, flat ~38 MB RSS, stable fd counts, zero zombies, DONE sentinel written); the macOS PASS run's `serve.exit` shows the identical `{"exit":127,"expected":true}` shutdown artifact, confirming that part is platform-independent and pre-existing.

### DIFF
--- a/scripts/soak-scheduler.sh
+++ b/scripts/soak-scheduler.sh
@@ -310,18 +310,21 @@ jq -n \
   --arg sha "$(git -C "$REPO" rev-parse HEAD 2>/dev/null)" \
   --arg branch "$(git -C "$REPO" rev-parse --abbrev-ref HEAD 2>/dev/null)" \
   --arg state_path "$STATE_PATH" \
-  --argjson t0 "$T0" --argjson end "$END" --argjson pid "$SERVE_PID" \
+  --argjson t0 "$T0" --argjson end_ts "$END" --argjson pid "$SERVE_PID" \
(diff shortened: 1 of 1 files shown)


hugocorreia90:
Review disposition (verdict: APPROVE): both speculative items are settled — (a) the exact metadata jq invocation was executed under jq 1.6 in the Debian bookworm container and exits 0 (that reproduction is what root-caused the bug); (b) the PR head is `c80504e3`, matching the reviewed tree. The pre-existing harvest-status qualification is noted and out of scope. Merging on the green CI.

Issue #1282, opened 2026-07-29
Snowflake: unquoted references resolve uppercase, so text-equal names can be different objects


hugocorreia90:
Upstream-reference matching compares the **spelled text** of a reference against a rename key. On Snowflake that is not sufficient to establish identity, because quoting is a second identity axis independent of case.

## The defect

- Rocky renders Snowflake targets with every component double-quoted (`SnowflakeSqlDialect::format_table_ref`). A configured target `orders` is therefore the object `orders`, spelled exactly.
- Snowflake resolves an **unquoted** identifier by upper-casing it. `FROM main.orders` names `MAIN.ORDERS`.
- Those are different objects, but their spelled text is identical, so the matcher treats the reference as naming the routed upstream and rewrites the read to the shadow target — silently changing what the query reads.

The same reasoning implies a lowercase-configured Snowflake model could not read its upstream in production either (Rocky created `"orders"`; the model asks for `ORDERS`). That has **not** been verified against a live account, and it is the crux — see "Open question" below.

## Status

Pre-existing and unchanged: the matcher has always compared text and discarded `Ident.quote_style`.

The resolution is already implemented and unit-tested behind `IdentifierCaseRules::uniform_uppercasing` — see `snowflake_resolves_unquoted_identifiers_before_matching` in `engine/crates/rocky-sql/src/defer.rs`, which pins the full matrix:

| configured target | reference | outcome |
|---|---|---|
| `CAT.RAW.ORDERS` (idiomatic) | unquoted, any case | routes normally |
| `cat.raw.orders` | unquoted | reported as a near-miss, run refused |
| `cat.raw.orders` | `"cat"."raw"."orders"` | routes normally |

Enabling it is a one-line change in `dialect_case_rules` (`engine/crates/rocky-cli/src/commands/run.rs`), which currently returns `uniform(true)` for Snowflake with a comment pointing here.

## Why it is not enabled

Turning it on refuses every lowercase-configured Snowflake project, including this repository's own test fixtures. The idiomatic uppercase-target shape is unaffected. Trading a known, pre-existing, unchanged hazard for an unmeasured break needs evidence first.

## Open question to settle first

Against a live Snowflake account, with a lowercase configured target and an unquoted upstream reference: does a plain `rocky run` succeed today?

- If it **fails** ("table does not exist"), the analysis holds, such projects are already broken, and the rule should simply be enabled.
[...]

Issue #1240, opened 2026-07-26
Decimal-input AVG columns are contract-unchecked (inferred Unknown reads as "compatible" at the E011 gate)


hugocorreia90:
## Context

#1239 (squash `81c72bf8`, closes #1238) made `AVG` of a `Decimal` input infer as `RockyType::Unknown`, because the result type is dialect-dependent and the compiler layer has no adapter context — DuckDB returns `DOUBLE`, Databricks returns `DECIMAL(p + 4, s + 4)`, BigQuery preserves `NUMERIC`/`BIGNUMERIC`. That was the right call for *inference*: the previous unconditional `Float64` was a fabricated concrete type that rejected a valid Decimal contract with a false `E011`.

It leaves a residual at the *validation* gate, recorded here so it is not lost.

## The residual

`validate_contract` skips type checking entirely when the inferred type is `Unknown`:

```rust
// engine/crates/rocky-compiler/src/contracts.rs:144
if !matches && col.data_type != RockyType::Unknown {   // ← Unknown ⇒ E011 never fires
```

So a `Decimal`-input `AVG` column is now **not contract-checked at all**. Every declared type passes — `Decimal`, `Float64`, `String`, `Boolean`. Before #1239 a wrong declaration was rejected (for the wrong reason, against a fabricated type, but rejected).

`Unknown` is conservative for inference and permissive for validation. Withholding a type is correct; having the gate read the withheld type as "compatible" is not.

## Reproduce

```
SELECT AVG(amount) AS avg_amount FROM source.raw.orders
```

with `amount` typed `DECIMAL(10,2)` in the source schema, and a contract declaring `type_name = "Boolean"` for `avg_amount`. `rocky compile` emits no diagnostic.

## Scope

Bounded and currently latent in-repo: the models that use `AVG` and the projects that ship a `*.contract.toml` are disjoint, so no in-repo contract is silenced today.

Not affected: `breaking_change.rs`. Every consumer calls `diff_project_ir(&base_ir, &head_ir)` with both sides compiled by the same binary, and `main` mapped every `AVG` to a single type — so there is no type difference for this change to erase.

## Proposed resolution

[...]

PR #1390, opened 2026-08-06
feat(engine): read the live identifier case-sensitivity setting instead of assuming it

hugocorreia90:
Refs #1281 — **does not close it**, see "Deferred" below.

Shadow execution assumed how a warehouse treats identifier case, because it had no way to ask. It can now ask.

## The capability

`WarehouseAdapter::identifier_case_significance` returns whether identifier case is part of object identity **on this connection**. Not a dialect question: Snowflake's `QUOTED_IDENTIFIERS_IGNORE_CASE` is an account/session parameter, so two connections to one dialect can legitimately disagree.

- **Snowflake** reads the parameter live via `SHOW PARAMETERS`.
- **DuckDB / Databricks / Trino** answer with their dialect constant, no round trip.
- **BigQuery** is deliberately left on the default (see Deferred).

Two states, and **no `Unknown` variant** — "could not determine" rides on the `Err` arm. That is deliberate: a third variant is something a caller can match and let fall through a gate, which is the #1240 shape where an `Unknown` type satisfied a compatibility check unconditionally and the gate read as fail-open. The trait default is `Err`, so an ungrounded adapter inherits today's conservative behaviour rather than silently acquiring a relaxation nobody verified.

## The two refusals become decisions

They fail closed in **opposite** directions, so each consults the observation separately rather than sharing one answer:

| question | observed case-**sensitive** | observed case-**insensitive** | could not ask |
|---|---|---|---|
| could two targets be the same object? | route independently — they are two objects | refuse — they are one object | **fold and refuse** (a wrong "distinct" is silent: two models, one shadow table, no error) |
| does this reference name that routed upstream? | leave it — it names a different object | redirect it — it does name the upstream | **refuse** (both guesses unsafe) |

`None` reproduces the pre-#1281 behaviour exactly. The probe failure is also non-fatal at the call site: an unreachable adapter refuses what it refused before rather than failing the run over a probe.

`CollisionIdentity::of_observed_case_sensitive` is a **named constructor** rather than a `rules` parameter on `of`. Folding is that type's whole reason to exist, so making it configurable through the main constructor would put the fail-open choice one argument away at every call site. A named constructor makes every relaxation greppable and argued for where it is written.

## Verification

**The Snowflake probe is live-verified** against a real account:

```
live: QUOTED_IDENTIFIERS_IGNORE_CASE=false -> Significant
test probe_reads_a_definite_answer_from_the_live_session ... ok
```

Worth noting *how* that was possible: the trial on that account has lapsed and all DDL/compute is refused with `000666 / 57014 "account is suspended"`. `SHOW PARAMETERS` is metadata, so it still answers — which is why this probe could be live-verified when the rest of the Snowflake live suite could not.

**Only the `false` state is live-observable.** Flipping the parameter needs `ALTER SESSION`, which needs compute, which the lapsed trial refuses; the SQL API also rejects it as a request-level parameter (`391917`). So the `true` branch is covered by in-memory tests against an extracted `case_significance_from_parameter_cell`, not live. Stating that rather than implying both states were live-verified.

The live test asserts the probe **agrees with a second independent read** of the parameter, not that it equals `Significant` — pinning the latter would assert the account's configuration rather than the probe's behaviour, and would fail if the parameter were ever flipped, which is the very thing the probe exists to notice.

Both relaxations are mutation-checked:

| mutation | caught by |
|---|---|
| any definite answer stops folding (not just `Significant`) | `observed_case_insensitive_still_refuses_the_shared_target` |
| an unknown stops folding | the pre-existing `case_differing_targets_collide_on_every_dialect` |

The second matters most: the existing suite guards the `None` path, so the conservative behaviour cannot regress silently.

## Deferred, deliberately

**BigQuery's per-dataset `is_case_insensitive`.** The issue asks for it, and I did not implement it, for two reasons worth stating rather than burying:

1. Its answer varies **per dataset**, while `IdentifierCaseRules` is one value for the run — that is a real design change, not a fifth adapter impl.
2. The BigQuery trial has expired, so it cannot be live-verified. This repo's own record is that BigQuery stub-JSON tests shipped **five structural bugs** that only live execution caught. An unverifiable probe feeding a fail-open seam is the worst possible combination.

BigQuery therefore keeps today's conservative behaviour, and the docs say so.

**#1282 is deliberately out.** `unquoted_uppercases` is a different identity axis — Snowflake folds unquoted identifiers regardless of this parameter — and enabling it refuses every lowercase-configured Snowflake project, including this repo's own fixtures.

## What I'm least confident about

The reference-matching relaxation under observed case-**sensitive**, where the refusal is dropped and the reference is left unrewritten. The argument is that such a reference names a genuinely different object, so reading it is correct. That follows from the setting, but unlike the collision half I could not exercise it end-to-end against a live warehouse — the fixture proves the routing decision, not that Snowflake resolves the two spellings to different objects at execution. The banked Snowflake live suite is from engine ~1.47.x and we are on 1.68.0.

## What I'm most likely missing

Whether `replay.rs` should re-probe. It calls `dialect_case_rules` independently, and I left it untouched, so a replayed plan keeps the static answer. That is the conservative choice, but it means a plan reviewed under one identity model could execute under another if the account parameter changed in between — the guard-at-the-replay-seam shape. Worth a follow-up rather than a silent decision here.

### DIFF
--- a/engine/crates/rocky-core/src/traits.rs
+++ b/engine/crates/rocky-core/src/traits.rs
@@ -305,11 +305,81 @@ pub enum SplitStrategy {
 // Warehouse
 // ---------------------------------------------------------------------------
 
+/// Whether the case of a **quoted** identifier is part of object identity.
+///
+/// ‼️ Read the scope narrowly; it is one of two axes, not the whole answer.
+/// This describes identifiers as Rocky renders its own targets — quoted. It
+/// says nothing about *unquoted* references, which several warehouses fold
+/// independently of this setting: Snowflake upper-cases them whatever
+/// `QUOTED_IDENTIFIERS_IGNORE_CASE` says, so under `Significant` the quoted
+/// `"orders"` and `"Orders"` are two objects while the unquoted `orders` and
+/// `ORDERS` remain one. A consumer deciding whether an arbitrary SQL reference
+/// names a routed target needs BOTH axes; the second is #1282. Treating this
+/// value alone as "does case separate these two names" is the exact mistake an
+/// earlier revision of #1281 made, and it produced a silent isolation break.
+///
+/// Deliberately two states and no `Unknown`. "Could not determine" is carried
+/// by the `Err` arm of [`WarehouseAdapter::identifier_case_significance`], so
+/// a caller cannot pattern-match a third variant and let it fall through a
+/// gate — the failure mode #1240 records, where an `Unknown` type satisfied a
+/// compatibility check unconditionally and the gate read as fail-open.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CaseSignificance {
+    /// Two QUOTED spellings differing only by case name TWO objects.
+    Significant,
+    /// Two QUOTED spellings differing only by case name ONE object.
+    Insignificant,
+}
+
 /// Executes SQL against a warehouse and provides dialect information.
 #[async_trait]
 pub trait WarehouseAdapter: Send + Sync {
     /// Returns the SQL dialect for this warehouse.
     fn dialect(&self) -> &dyn SqlDialect;
+
+    /// Does the warehouse treat a QUOTED identifier's case as part of object
+    /// identity, as observed right now?
+    ///
+    /// Not a dialect question. Snowflake's `QUOTED_IDENTIFIERS_IGNORE_CASE` is
+    /// account/session state, and a BigQuery dataset carries its own
+    /// `is_case_insensitive` — so two connections to the same dialect can
+    /// legitimately disagree, and the answer has to be asked for rather than
+    /// looked up (#1281).
+    ///
+    /// **The answer describes the probe's own request, and does not extend to
+    /// a later one.** It is not durable connection state a caller may cache:
+    /// Snowflake's SQL API opens an independent session per POST, so a
+    /// subsequent statement can execute under a different value, and the
+    /// account-level setting can change between the two regardless. A consumer
+    /// that relaxes a safety decision on this answer must pin the setting for
+    /// the statement it is deciding about, or accept that it is deciding on
+    /// state that may no longer hold. This is not hypothetical caution — it is
+    /// why #1281 ships the probe without a consumer: the routing relaxation it
+    /// was written for could not honour it.
+    ///
+    /// **Nothing in the engine calls this yet.** #1281 lands the capability and
+    /// its Snowflake grounding; #1282 is the intended consumer, because routing
+    /// a reference correctly also needs the *unquoted*-uppercasing axis this
+    /// answer does not carry. Shipping the probe unwired is deliberate: it is
+    /// the half that is safe in every state, and it was live-verifiable while
+    /// the Snowflake trial still answered metadata queries.
+    ///
+    /// **The default is `Err`, and that is load-bearing** for whoever wires it.
+    /// An adapter that has not been grounded must not be read as reporting a
+    /// definite answer, so a new adapter inherits today's refusals rather than
+    /// silently acquiring a relaxation nobody verified. Adapters whose answer is
+    /// a dialect constant (DuckDB, Databricks, Trino) override with that
+    /// constant and issue no round trip.
+    ///
+    /// Callers MUST NOT relax a safety decision on `Err`. In particular a
+    /// transport failure, a permissions error, and "this adapter does not
+    /// implement the probe" are indistinguishable here **on purpose** — all
+    /// three mean the same thing to a gate.
+    async fn identifier_case_significance(&self) -> AdapterResult<CaseSignificance> {
+        Err(AdapterError::msg(
+            "this adapter does not report whether identifier case is part of object identity",
+        ))
+    }
 
     /// Execute a SQL statement (DDL/DML) without returning rows.
     async fn execute_statement(&self, sql: &str) -> AdapterResult<()>;
--- a/engine/crates/rocky-databricks/src/adapter.rs
+++ b/engine/crates/rocky-databricks/src/adapter.rs
@@ -134,6 +134,18 @@ impl WarehouseAdapter for DatabricksWarehouseAdapter {
         &self.dialect
     }
 
+    /// Databricks folds identifier case, so this is a dialect constant and needs
+    /// no round trip (#1281).
+    ///
+    /// Identifiers are case-insensitive, delimited ones included; Unity Catalog
+    /// stores names lower-cased —
+    /// docs.databricks.com/aws/en/sql/language-manual/sql-ref-identifiers
+    async fn identifier_case_significance(
+        &self,
+    ) -> AdapterResult<rocky_core::traits::CaseSignificance> {
+        Ok(rocky_core::traits::CaseSignificance::Insignificant)
+    }
+
     async fn execute_statement(&self, sql: &str) -> AdapterResult<()> {
         self.connector
             .execute_statement(sql)
--- a/engine/crates/rocky-duckdb/src/adapter.rs
+++ b/engine/crates/rocky-duckdb/src/adapter.rs
@@ -117,6 +117,18 @@ impl WarehouseAdapter for DuckDbWarehouseAdapter {
         &self.dialect
     }
 
+    /// DuckDB folds identifier case, so this is a dialect constant and needs
+    /// no round trip (#1281).
+    ///
+    /// "Identifiers … are always case-insensitive … DuckDB also treats quoted
+    /// identifiers as case-insensitive" (case is preserved for display only) —
+    /// duckdb.org/docs/stable/sql/dialect/keywords_and_identifiers
+    async fn identifier_case_significance(
+        &self,
+    ) -> AdapterResult<rocky_core::traits::CaseSignificance> {
+        Ok(rocky_core::traits::CaseSignificance::Insignificant)
+    }
+
     async fn execute_statement(&self, sql: &str) -> AdapterResult<()> {
         let conn = Arc::clone(&self.connector);
         let sql = sql.to_string();
--- a/engine/crates/rocky-snowflake/src/adapter.rs
+++ b/engine/crates/rocky-snowflake/src/adapter.rs
@@ -62,6 +62,29 @@ impl WarehouseAdapter for SnowflakeWarehouseAdapter {
         Some(self.connector.warehouse())
     }
 
+    /// Read `QUOTED_IDENTIFIERS_IGNORE_CASE` off the live session (#1281).
+    ///
+    /// Rocky renders Snowflake targets DOUBLE-QUOTED (see this crate's
+    /// `dialect::format_table_ref`), so this parameter — and not the unquoted
+    /// upper-casing rule — is what decides whether two of Rocky's targets
+    /// differing only by case are one object or two. `TRUE` makes quoted
+    /// identifiers case-insensitive, i.e. case is NOT part of identity.
+    ///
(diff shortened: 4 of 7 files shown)


hugocorreia90:
**Held as draft pending independent review — do not merge yet.** CI is green and the branch is ready; this is a deliberate hold, not an unfinished PR.

Everything *enumerable* here is verified. Both relaxations are mutation-checked across all three states, the `None` path is guarded by the pre-existing `case_differing_targets_collide_on_every_dialect` (letting an unknown stop folding fails it), and the claim that the folding dialects are unaffected is now pinned by a test rather than asserted in prose.

What is **not** verified is the design judgment: whether relaxing two deliberately fail-closed seams is correct. The collision half is the reason for caution — it is currently unconditionally safe (it always folds), and this PR makes it conditional on an observation. A wrong relaxation there is silent: two models routed to one shadow table with no error at all.

Independent red-team capacity (a genuinely different model) returns **2026-08-08**. Re-gate and merge after that.

Two specific things worth pointing the reviewer at:

1. The reference-matching relaxation under observed case-**sensitive**, where the refusal is dropped and the reference is left unrewritten. The fixture proves the routing decision; it does not prove Snowflake resolves the two spellings to different objects at execution.
2. `replay.rs` calls `dialect_case_rules` independently and is untouched, so a replayed plan keeps the static answer. Conservative, but a plan reviewed under one identity model could execute under another.

hugocorreia90:
## Independent red team: **UNSAFE — do not merge as-is**

Reviewer: **Codex (GPT-5 family, `gpt-5.6-terra` for the confirming pass)** — a genuinely different model, per the standing rule that `/code-review` does not satisfy independence. It found a **deterministic** regression, not a hypothetical one, and I have confirmed the decisive step against evidence I gathered independently.

### The finding I accept without reservation

**Seam 2 (reference matching) is a deterministic isolation break under `Significant`.**

Setup — the *idiomatic* Snowflake configuration:
- upstream target configured `CAT.RAW.ORDERS` (uppercase)
- downstream SQL references it **unquoted**: `FROM cat.raw.orders`
- account has `QUOTED_IDENTIFIERS_IGNORE_CASE = FALSE`, so the probe returns `Significant`

What happens: `dialect_case_rules` returns `uniform(true)`, whose `unquoted_uppercases` is **false**, so the matcher sees a case-fold-only near-miss rather than a match. This PR then suppresses the refusal for *every* `Some(...)` (`run.rs:6461`), so the reference is left unrewritten — and the model **writes its shadow table while reading production**, exit 0. On `main` today this exact input is **refused**.

### Why my reasoning was wrong

I justified the `Significant` branch as: "the reference names a genuinely different object, so leaving it alone is correct." That is false, and my own earlier probe already disproved it. `QUOTED_IDENTIFIERS_IGNORE_CASE` governs **quoted** identifiers; **unquoted** ones are folded to uppercase first, independently of it. I demonstrated exactly that live on this account: `SELECT 1 AS Foo, 2 AS "Bar"` returns column names `['FOO', 'Bar']`.

So an unquoted `cat.raw.orders` and a quoted target `"CAT"."RAW"."ORDERS"` **are the same object** even when case is significant. Those are two identity axes, and I asserted they were independent when I deliberately scoped #1282 (`unquoted_uppercases`) out of this PR. Scoping it out is precisely what makes the relaxation unsafe — the two are coupled.

### Also found

- **Seam 1 (collision) — cross-session TOCTOU.** The probe runs in its own SQL API request; every later statement opens an independent session (`rocky-snowflake/src/connector.rs:420-476`). A probe returning `Significant` stops the folding, and if the account parameter flips before the DDL runs, two case-only targets alias to one object. `main` has no such race, because it always folds.
- **Plan/replay gap.** A `RunPlan` persists shadow flags but not the observed case identity, and the fingerprint check runs *before* the live probe, so an approved plan can execute under a different routing decision without a fingerprint mismatch.
- **Pre-existing, not from this PR:** `rocky compare`'s `get_row_count` interpolates `TargetRef::validated_full_name()` (bare `catalog.schema.table`) straight into `SELECT COUNT(*)` (`compare.rs:339-347`, `rocky-ir/src/ir.rs:175-186`) instead of dialect-formatting it — so on Snowflake it can count the wrong object. Worth its own issue.

### What confirmed and what did not

Codex **could not refute** two things, and said so explicitly rather than staying silent: the probe does belong to the warehouse that executes the DDL (it traced the explicit-pipeline, replication, and `--dag --shadow` routes), and `Err`/`None` does reproduce the pre-change behaviour exactly.

Its own stated least-confidence: `gh` was unreachable from its sandbox, so it reviewed the local range `586500c4..c34149bd` and could not confirm that is still the remote head. It is — that range is this PR.

### Next

Revising rather than merging. Holding this PR was the right call, and it was `/loop`'s stop-review discipline that produced the finding rather than anything in my own testing — my mutation checks all passed because they tested the rule I had written, not the rule the warehouse enforces.

hugocorreia90:
## Revised against the red-team finding

Both of Codex's seam findings are addressed, both by moving toward the conservative side rather than by trying to be clever.

### 1. `Significant` no longer drops the case-fold-only refusal — the deterministic break

Accepted in full. The refusal now requires an observed **Insignificant** connection, where dropping it is vacuous anyway (the matcher has already folded, so such a reference was redirected and cannot reach the check). `Significant` refuses exactly as `None` does.

Resolving that reference *correctly* needs #1282's unquoted-uppercasing axis, which stays out of scope. What this PR should never have done is treat "case is significant" as licence to route it.

### 2. Collision relaxation reverted entirely — the TOCTOU

Collision identity always folds again, including under `Significant`. The probe runs in its own SQL API request and every later statement opens an independent session, so an account flipping the parameter between probe and DDL aliases a pair judged distinct, and `CREATE OR REPLACE` clobbers silently. `main` has no such race.

The trade is lopsided and I should have weighed it that way originally: folding costs at most an unnecessary refusal on a case-differing target pair, which an operator resolves by renaming one. Not folding costs two models writing one shadow table with no error at all.

### What survives

The half that is safe in every state: the probe itself, live-verified against a real Snowflake account, and routing a reference when the connection is observed to **fold** case — which turns a previously unresolvable near-miss into a correct redirect.

So the PR is now **strictly safer than `main` in every state**, rather than safer in one and more dangerous in another.

### Tests

- `observed_case_sensitive_still_refuses_a_case_fold_only_reference` — reproduces the red team's exact input (uppercase target `MAIN.ORDERS`, unquoted `FROM main.orders`, probe `Significant`). **Mutation-checked:** restoring the any-`Some` suppression fails it.
- `collision_folds_even_when_the_connection_is_observed_case_sensitive` — replaces the test that asserted the now-reverted relaxation, and records the TOCTOU reason.

`fmt` clean · `clippy --all-targets --all-features -D warnings` clean · `rocky-cli` lib suite **1422 passed, 0 failed**.

### Still open from the review, not addressed here

- **Plan/replay gap** — a `RunPlan` persists shadow flags but not the observed case identity, and the fingerprint check runs before the probe. Less acute now that neither relaxation is unlocked by `Significant`, but still real. Follow-up.
- **`rocky compare` interpolates an unquoted identifier** — pre-existing, filed separately as #1396.

### On the process

Worth recording that my own mutation checks all passed on the broken revision. They tested the rule I had written, not the rule the warehouse enforces — and the decisive evidence (`SELECT 1 AS Foo` → `FOO`) was something I had already gathered and failed to connect. The stop-review is what caught it.

Leaving this as a draft pending your call on whether the reduced scope is still worth landing.

hugocorreia90:
## Red team on the fix delta (`3cd4a6c5`) — and what the PR actually still changes

Ran an independent review scoped to `3cd4a6c5` alone, since that commit was written *in response* to the previous red team and had never itself been reviewed. Two reviewers disagreed, so I verified the disputed claim from the source. The result changes how big this PR is.

### After `3cd4a6c5`, two of the three changes at this seam are no-ops

**The `ensure!` relaxation is unreachable.** `Some(Insignificant)` sets `case_rules = all_insensitive()` (`run.rs:6171`). The candidate matcher filters with `tail_matches_with_case(key, original, self.case_rules)` (`defer.rs:749`), and the near-miss check in the no-candidate arm uses the *same predicate* with `all_insensitive()` (`defer.rs:774`). So under `Insignificant`, "no candidate under the rules" already implies "no candidate under folding" — `case_fold_only_refs` is provably empty and the new disjunct can never be the reason the `ensure!` passes. `defer.rs:770` says as much in its own words: "Inert wherever `case_rules` is already all-insensitive... this can never fire."

**`collision_identity` is now byte-identical to main.** The revert restored `CollisionIdentity::of`.

So the surviving behavioural change is one line:

```rust
Some(Insignificant) => IdentifierCaseRules::all_insensitive(),
_                   => dialect_case_rules(dialect)?,   // == main
```

and it bites on **Snowflake only** — `dialect_case_rules` already returns `uniform(false)` for DuckDB/Databricks/Trino (`run.rs:5973`) and the PR pins `uniform(false) == all_insensitive()`. Net: when a Snowflake connection reports `QUOTED_IDENTIFIERS_IGNORE_CASE = TRUE`, the matcher folds where main would have refused.

That is worth stating plainly because the diff reads as 683 lines and the decision surface is one match arm plus the probe.

### The finding: the argument that reverted one relaxation was not applied to the other

`3cd4a6c5` reverts `collision_identity` because the probe runs in its own SQL API request and every later statement opens an independent session, so an account flipping the parameter between probe and DDL aliases two targets judged distinct.

The same instability applies to the surviving `case_rules` line, and the commit does not address it. Under an observed `TRUE`, a reference like `FROM raw.Orders` folds and is rewritten to the routed upstream's shadow. If the parameter is `FALSE` at execution, `raw.Orders` and `RAW.ORDERS` are two objects — the model reads the wrong table's shadow and exits 0. Main refuses that input.

**The counter-argument, stated fairly:** it is less severe than the collision case. Collision aliasing means `CREATE OR REPLACE` destroys data; this only produces a wrong read *inside a shadow run*, and because the data differs, `rocky compare` should surface it rather than reporting clean. So the lopsided-trade reasoning that justified reverting collisions does not automatically justify reverting this one. It does mean the commit's stated principle — "nothing but a definite observation may unlock one", where the observation is not pinned to the executing session — is currently applied to one relaxation and not the other, with no reason given for the split.

### Reviewer disposition (recorded explicitly)

- Independent `gpt-5.6-terra` pass: **SHIP**, no concrete repro against the delta. Its clean checks on the replay seam (`replay.rs:1742`, `:1796` — refuses case-only near-misses, so the unresolved axis yields non-replayable rather than a production read) and on refusal-before-mutation (`run.rs:7648`, `:7664`) I spot-checked and agree with.
- The parent thread **overruled** it on exactly the symmetry above, then stopped before recording a verdict. Its argument is right; it located the problem at the `ensure!`, where it is unreachable. The real location is `run.rs:6171`.

### What this leaves for the keep-or-drop call

Three options, all cheap now that the surface is one line:

1. **Land as-is** — accept the cross-session risk for the `case_rules` fold on the grounds that it is a shadow-only wrong read that `compare` catches.
2. **Land the probe, drop the `case_rules` fold** — keep `identifier_case_significance` and its Snowflake implementation as the foundation #1282 needs, and leave routing on the static rule. This makes the PR a pure capability addition with no behavioural change at all, which is trivially safe to merge and unblocks #1282.
3. **Drop** — nothing in the PR is load-bearing for anything else today.

I would take (2). It banks the live-verified probe (the part the suspended Snowflake trial made expensive to get) and defers every routing decision to #1282, which is where the unquoted-uppercasing axis actually gets resolved.

Also note the PR is `BEHIND`; none of the two commits on main since its base touch any of its nine files, so the rebase is clean whenever this is decided.

PR #1395, opened 2026-08-07
chore(deps): bump the release actions and their allow-list entries together (absorbs #1414)

hugocorreia90:
Supersedes #1369.

> **Verification re-done 2026-08-16.** Dependabot rebased this branch after the original review, which
> re-resolves to the newest release rather than the one that was reviewed — the `install-action` pin moved
> `6a1bd70e` (v2.85.5) → `6c6fd71f` (v2.85.10) without the table being updated. The row above now describes
> the SHA actually in the diff, re-checked against upstream. `pypa/gh-action-pypi-publish@dc37677b` was
> re-checked too and is still the live HEAD of `release/v1`.
>
> **This PR is not mergeable as it stands:** the containment gate reports 10 real violations on the current
> head (`e2dfecfd`), not a stale run. The branch is 9 days behind `main`, so the allow-list constants and the
> workflow pins have drifted out of agreement. It needs a rebase onto `main` and a re-reconcile of the two
> halves in one commit — the same all-or-nothing property this PR was opened to satisfy.

## Why #1369 could never go green

`check_credential_containment.py` pins every release-job action source by exact SHA in `RELEASE_ACTION_SOURCES`. Dependabot bumped the four workflow files but cannot know about that allow-list, so #1369 fails with three violations by construction:

```
.github/workflows/dagster-release.yml: release step action source is not allow-listed
.github/workflows/engine-release.yml:  release step action source is not allow-listed
.github/workflows/sdk-release.yml:     release step action source is not allow-listed
```

The reverse split is equally broken — updating the allow-list first would un-allow the SHAs `main`'s own workflows still use. **The two halves have to land in one commit**, which is why this replaces the dependabot PR rather than adding to it.

## Both SHAs verified against upstream

That allow-list exists precisely to catch a SHA that is not what it claims to be, so I checked rather than trusted the bump:

| Action | Old → New | Verification |
|---|---|---|
| `pypa/gh-action-pypi-publish` | `ba38be9e` → `dc37677b` | real commit, 2026-07-28; **is the current HEAD of `release/v1`** — the ref the pin's own comment claims to track |
| `taiki-e/install-action` | `41049aa5` → `6c6fd71f` | real commit, message `Release 2.85.10`; **tag `v2.85.10` resolves to exactly that SHA** |

This matters more than usual: `gh-action-pypi-publish` is the step holding PyPI OIDC, so an unreviewed source there is a direct supply-chain path into published wheels. That is the checker's own stated rationale.

## Verification

- Containment suite green: `python3 -m unittest discover -s .github/tests` → **110 tests, OK** (the exact CI invocation).
- Policy checker passes on this tree: `credential-containment policy: passed`.
- **Mutation-checked in the direction that matters.** Reverting only the two allow-list constants — precisely #1369's state — reproduces its three violations, then restoring them passes again. So the allow-list half is demonstrably the necessary part, not incidental.

## Note

`engine-ci.yml`'s `taiki-e/install-action` is a pull-request-reachable job and so is *not* governed by `RELEASE_ACTION_SOURCES`. Bumped in step anyway to keep one version of the action across the repo.

This touches only `.github/`, so the engine-path-filtered required checks will not run and it will sit `BLOCKED` — the documented `--admin` case.

### DIFF
--- a/.github/scripts/check_credential_containment.py
+++ b/.github/scripts/check_credential_containment.py
@@ -37,9 +37,9 @@ RUST_TOOLCHAIN_SOURCE = (
     "dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8"
 )
 RUST_CACHE_SOURCE = (
-    "Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4"
+    "Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6"
 )
-SETUP_UV_SOURCE = "astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9"
+SETUP_UV_SOURCE = "astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d"
 SETUP_NODE_SOURCE = (
     "actions/setup-node@820762786026740c76f36085b0efc47a31fe5020"
 )
@@ -55,13 +55,13 @@ WASM_PACK_SOURCE = (
 )
 SETUP_ZIG_SOURCE = "mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29"
 PYPI_PUBLISH_SOURCE = (
-    "pypa/gh-action-pypi-publish@ba38be9e461d3875417946c167d0b5f3d385a247"
+    "pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33"
 )
 GH_RELEASE_SOURCE = (
     "softprops/action-gh-release@3d0d9888cb7fd7b750713d6e236d1fcb99157228"
 )
 TAIKI_INSTALL_SOURCE = (
-    "taiki-e/install-action@41049aa56687c35e0afa74eed4f09cec4f9afabf"
+    "taiki-e/install-action@82cd3e7658a6f96c86c0234aeeda1748937cb0a1"
 )
 # Actions a job may run when it is not reachable from a pull request. Release,
 # publish and deployment workflows never execute candidate code, so the
--- a/.github/workflows/codegen-drift.yml
+++ b/.github/workflows/codegen-drift.yml
@@ -65,7 +65,7 @@ jobs:
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
 
-      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2
+      - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2
         with:
           workspaces: engine
 
@@ -78,7 +78,7 @@ jobs:
           sudo fallocate -l 4G /mnt/swapfile && sudo chmod 600 /mnt/swapfile && sudo mkswap /mnt/swapfile && sudo swapon /mnt/swapfile || true
 
       - name: Install uv
-        uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
 
       - name: Install Node.js
         uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
--- a/.github/workflows/dagster-ci.yml
+++ b/.github/workflows/dagster-ci.yml
@@ -39,7 +39,7 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
-      - uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
+      - uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
       - run: uv sync --dev
       - run: uv run python -m pytest -v
       - run: uv run python -m ruff check src/ tests/
--- a/.github/workflows/dagster-release.yml
+++ b/.github/workflows/dagster-release.yml
@@ -57,10 +57,10 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
-      - uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
+      - uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
       - run: uv build
       - name: Publish to PyPI
-        uses: pypa/gh-action-pypi-publish@ba38be9e461d3875417946c167d0b5f3d385a247 # release/v1
+        uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # release/v1
         with:
           packages-dir: integrations/dagster/dist/
 
--- a/.github/workflows/engine-bench.yml
+++ b/.github/workflows/engine-bench.yml
@@ -32,7 +32,7 @@ jobs:
 
       - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
 
-      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2
+      - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2
         with:
           workspaces: engine
           key: bench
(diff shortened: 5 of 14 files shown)


hugocorreia90:
## This is a frozen-trust-root bootstrap — the red gate is expected

`Credential containment policy` **failed**, and per `.github/SECURITY_ENVIRONMENTS.md` § "Updating frozen trust roots" that is the designed behaviour for this change shape, not a defect. Recording the procedure evidence rather than asking for a bypass on trust.

### Why it fails, and why it cannot be split

The gate runs **main's** copy of the checker (`trusted/.github/scripts/check_credential_containment.py`) against the candidate, and `FROZEN_TRUST_ROOTS` byte-compares that file against the trusted base (`check_credential_containment.py:1748`). So there are exactly two failure modes and this PR necessarily hits both:

```
check_credential_containment.py: frozen trust root differs from trusted base
.github/workflows/{dagster,engine,sdk}-release.yml: release step action source is not allow-listed
4 violation(s)
```

- Bump the workflows alone (#1369): trusted checker still holds the old SHAs → 3 violations. **Dependabot cannot ever produce a passing PR here.**
- Update the allow-list alone: the trust root differs → 1 violation, *and* main's own workflows would be un-allowed.
- Both together: 4 violations, which is this PR.

Precedent is exact: #1265 ("bump pinned GitHub Actions trust roots") stated the same expectation and merged the same way.

### Procedure steps 1–3, completed

| Step | Status |
|---|---|
| 1. Dedicated PR, only reviewed trust-root changes | 5 files, nothing but the two SHA substitutions + version comments |
| 1. Independent security review | **in progress** — dispatched to Codex (a different model), asked to verify provenance, confirm the diff widens nothing, and challenge the atomicity argument. Verdict will be posted here. |
| 2. Unprivileged policy-test workflow | `Credential-containment regression tests` = **pass** on CI |
| 2. Full local unit suite | `python3 -m unittest discover -s .github/tests` → **110 tests, OK** |
| 2. `actionlint` | exit 0, **0 findings** |
| 2. `shellcheck` (via `actionlint -shellcheck`) | exit 0, **0 findings** |
| 2. `git diff --check` | clean |
| 3. No privileged job executes candidate code | **confirmed** — `ci-security-policy.yml` checks out `github.workflow_sha` (the trusted revision) with `persist-credentials: false`, and pulls the candidate in as *data* via `fetch_policy_candidate.py`. It never executes or imports candidate code. |

Provenance, verified independently of dependabot:

- `pypa/gh-action-pypi-publish` `dc37677b` — real commit (2026-07-28) and **the current HEAD of `release/v1`**, the ref the pin's own comment claims to track.
- `taiki-e/install-action` `6a1bd70e` — real commit, `Release 2.85.5`, and tag `v2.85.5` resolves to exactly it.

The allow-list is **not broadened** — each entry is replaced one-for-one, per step 4's "Do not disable the policy or broaden its allowlist."

### Steps 4 and 5 are yours

> 4. Have an **authorized maintainer** use the narrow branch-protection override to merge that reviewed SHA.

I am deliberately **not** admin-merging this. That differs from #1387/#1388, where the required checks were *absent* because of path filtering; here a required security check is *failing and correct*, and the procedure explicitly reserves the override for a maintainer.

> 5. The confirmation run is **the next pull request opened after the bootstrap merge** — one that does not itself touch a trust root.

#1390 or the `rocky doctor` fix will serve; both are engine-only. Its `Credential containment policy` must be green before anything else merges.

hugocorreia90:
## Independent security review — round 1: REJECT on missing provenance

Reviewer: **Codex (GPT-5 family)**, with a confirming pass by `gpt-5.6 Terra`.

> **VERDICT: REJECT** — both newly trusted action SHAs and their upstream source changes are unverified because GitHub was unreachable. The local scope and privilege-separation checks pass, but they cannot compensate for missing provenance on code receiving PyPI OIDC and release credentials.

That is a reject on **evidence it could not obtain from its sandbox**, not on anything found wrong — and it is the right call for a supply-chain review. What it *did* confirm locally:

- The diff widens nothing: entries are replaced one-for-one, `FROZEN_TRUST_ROOTS` untouched, no rule weakened.
- Procedure step 3 passes, with a nuance worth recording: `ai-review.yml` and `preview-comment.yml` **do not wait** for `ci-security-policy` and may run concurrently. Safety comes from privilege separation and trusted-data consumers, not execution ordering. It found no bypass in that separation.

It also named the gap in **my** verification, correctly: I had confirmed the SHAs are *who they claim to be*, but never inspected *what changed inside them*.

## Provenance, now supplied

### `pypa/gh-action-pypi-publish` — `ba38be9e...dc37677b`

`dc37677b` is a real commit (2026-07-28), and `git/ref/heads/release/v1` resolves to exactly it.

**4 commits, 2 files, +20/-13 — both are dependency pins.** No `action.yml`, no entrypoint, no networking code, no credential handling.

```diff
-# NOTE: v6.1 is needed to support metadata v2.4 including PEP 639
-twine >= 6.1
+# NOTE: v7 is needed to support metadata v2.5 including PEP 794
+twine >= 7
-pypi-attestations ~= 0.0.29
-sigstore ~= 4.1
+pypi-attestations ~= 0.0.30
+sigstore ~= 4.5
```

Authored by Thomas Kluyver and Facundo Tuesca (Trail of Bits), merged by Sviatoslav Sydorenko — the project maintainer.

### `taiki-e/install-action` — `41049aa5...6a1bd70e` (v2.85.2 → v2.85.5)

`6a1bd70e` is a real commit, message `Release 2.85.5`, and tag `v2.85.5` resolves to exactly it.

**21 commits, 12 files, +454/-36 — every changed file is a tool-version manifest or the changelog** (`CHANGELOG.md`, `manifests/{biome,cargo-binstall,cargo-neat,cargo-shear,martin,mise,sccache,syft,ubi,uv,xh}.json`). No `action.yml`, no JS entrypoint, no shell script. Sole author across all 21 commits: Taiki Endo, the repo owner.

So neither action changed executable surface, network behaviour, or credential handling — only pinned data.

Sent back for a revised verdict rather than overriding the reject myself. Round 2 will be posted here.

hugocorreia90:
Absorbed #1414's pins so one PR carries both halves of the change:

| action | was here | now |
|---|---|---|
| `taiki-e/install-action` | v2.85.5 | **v2.85.10** (`6c6fd71f`) |
| `Swatinem/rust-cache` | not bumped | **`6323deb1`** |
| `pypa/gh-action-pypi-publish` | `dc37677b` | unchanged |

**This is now blocking a release, not just hygiene.** `dc37677b` carries twine 7.0.0. The currently allow-listed pin carries twine 6.1.0, which rejects the `Metadata-Version: 2.5` that hatchling 1.32.0 began emitting on 2026-08-11 — that is what failed `sdk-release` today. Tested directly: twine 6.1.0 and 6.2.0 reject a 2.5 wheel; 7.0.0 accepts it. #1420 works around it by bounding hatchling to `<1.32`; merging this is what lets that ceiling be lifted.

Verification:
- all three pinned SHAs resolve upstream
- running **this branch's own** `check_credential_containment.py` against this tree exits **0** — the allow-list and the workflows are internally consistent

The remaining red check is the frozen-trust-root comparison: CI runs the *base's* copy of the script, which cannot know about an allow-list entry the PR itself introduces. No PR that adds an entry and uses it can go green. That is the control working as intended, and it is why this needs a maintainer narrow override rather than a fix from me.

hugocorreia90:
Rebased onto `main` (now `ff7b94fe`). The containment gate will still go red, and I can now explain exactly why — it is **structural, not a defect in this PR**.

## The gate evaluates this PR's workflows against `main`'s allow-list

Reproduced locally, with a calibrated baseline first so the result is not just "my script says red":

| What was checked | Result |
|---|---|
| `main`'s checker on `main`'s own tree | **passes** (exit 0) |
| `main`'s checker + `main`'s allow-list, with **only this PR's workflow files** overlaid | **9 violations** — same files CI names |
| This PR's own tree, checker and allow-list together | **passes** (exit 0) |

That third row is the post-merge state: once this lands, `main` is self-consistent again.

The mechanism is deliberate and correct. `ci-security-policy.yml` runs on `pull_request_target` against a **trusted** checkout, so the policy that judges a PR is the one already on `main` — otherwise a PR could whitelist a malicious action by editing `RELEASE_ACTION_SOURCES` in the same commit. That is exactly the attack the allow-list exists to stop.

## The consequence is worth writing down

**No PR that bumps a pinned action SHA can pass this gate.** The matching allow-list entry only becomes trusted after it merges. The PR body says the two halves must land in one commit — true, but insufficient: even the combined commit cannot go green, because the trusted allow-list it needs is the one it is trying to introduce.

That also explains #1369, which this supersedes: it was never fixable by splitting or reordering.

So this is the documented structurally-blocked case, not a red flag to chase. The gate is doing its job; the job just has no green path for a SHA bump.

## Contents re-verified against upstream, again

Because a red supply-chain gate is the worst place to take a pin on trust:

| Action | Pinned here | Upstream check |
|---|---|---|
| `pypa/gh-action-pypi-publish` | `dc37677b` | still the live HEAD of `release/v1`, the ref the pin's comment claims to track |
| `taiki-e/install-action` | `6c6fd71f` | tag `v2.85.10` resolves to exactly this SHA |

## Why this is now the keystone

It gates three other PRs. The `pypa/gh-action-pypi-publish` pin on `main` carries `twine==6.1.0`, which **rejects** `Metadata-Version: 2.5`; `dc37677b` carries `twine==7.0.0`, which accepts it (read from `requirements/runtime.txt` at both SHAs). So #1452 and #1457 — the dependabot PRs relaxing `hatchling<1.32` — cannot safely land until this does, and I have held both with that reasoning. Nothing in PR CI publishes, so those two would go green and fail only at release time, on a tag.

**Merging needs an admin override**, and given this is the step holding PyPI OIDC I am not going to do that unilaterally — flagging it for a decision rather than forcing it through.

PR #1402, opened 2026-08-08
fix(dagster): forward pipeline in streaming and pipes

### LINKED ISSUE #1401: bug(dagster): run_streaming and run_pipes cannot select a pipeline
## Description

`RockyResource.run()` accepts `pipeline=` and forwards it as `--pipeline`, but the two context-aware whole-run methods do not:

- `RockyResource.run_streaming()` has no `pipeline` keyword and cannot forward it to `RockyClient.run()`.
- `RockyResource.run_pipes()` has no `pipeline` keyword and cannot add it to the persisted `rocky plan` invocation.

Both methods are documented as having the same run semantics/arguments as `run()`. The engine requires `--pipeline <name>` whenever more than one pipeline is configured, so both methods are unusable for multi-pipeline projects.

## Steps to reproduce

```python
from dagster_rocky import RockyResource

RockyResource.run_streaming(None, None, "tenant=acme", pipeline="bronze")
RockyResource.run_pipes(None, None, "tenant=acme", pipeline="bronze")
```

Both calls fail during Python argument binding:

```text
TypeError: RockyResource.run_streaming() got an unexpected keyword argument 'pipeline'
TypeError: RockyResource.run_pipes() got an unexpected keyword argument 'pipeline'
```

Omitting the keyword does not provide a workaround for a multi-pipeline config: the engine rejects the unscoped command with `multiple pipelines defined (...). Use --pipeline <name> to select one.`

## Expected behavior

- `run_streaming(..., pipeline="bronze")` invokes the fused run with `--pipeline bronze`.
- `run_pipes(..., pipeline="bronze")` includes `--pipeline bronze` on the plan command; the apply command remains `rocky apply <plan-id>`.
- `pipeline=None` preserves the existing argv.

## Scope

Add the keyword-only parameter to both methods, forward it through the existing shared argv builders, and add focused positive and `None` control tests. Update the public method signatures in the Dagster resource docs.

This is distinct from #1292/#1347 (`run_model` transformation-asset routing) and #1348 (DAG discovery's unconditional `--models` override). None of those changes `run_streaming` or `run_pipes`.

- Exhausted the current open issue/PR list and searched titles, bodies, and comments for both method names and pipeline-selection variants; no existing item covers this defect.
- Reproduced both `TypeError`s on current `main` (`fef1674a`).
- Current suites are green before the fix: SDK 212 passed; Dagster 710 passed.

## Independent red team

An independent GPT-5.6-terra review confirmed this as a correctness defect and confirmed that #1292, #1347, and #1348 do not cover it.

**What I am least confident about:** severity calibration because buffered `run()` remains a workaround; the correctness failure itself is deterministic.

**What's the most important thing I may be missing:** production prevalence of multi-pipeline users choosing streaming or Pipes; this affects prioritization, not the fix or reproduction.

mattfaltyn:
## Summary

Closes #1401.

`RockyResource.run()` already accepts `pipeline=`, but its two context-aware whole-run variants did not. Multi-pipeline callers therefore could not use live stderr streaming or Dagster Pipes: Python rejected the selector, and an unscoped command is rejected by the engine when several pipelines exist.

## Subproject(s) touched

- [x] Documentation

## Changes

- Add the optional, keyword-only `pipeline` argument to `run_streaming()` and forward it to `RockyClient.run()`.
- Add the same argument to `run_pipes()` and persist it on the plan-phase argv; apply remains `rocky apply <plan-id>`.
- Pin both positive `--pipeline <name>` paths and the `pipeline=None` compatibility controls.
- Correct all three public resource signatures in the Dagster docs and record the fix in the changelog.

No SDK or engine change is needed: both shared SDK argv builders already accept `pipeline` and omit the flag for `None`.

## Test Plan

- `UV_CACHE_DIR=/tmp/rocky-uv-cache uv run pytest -q` — **711 passed**
- Focused mutation boundary — **4 passed**
- `uv run ruff check src/ tests/` — clean
- `uv run ruff format --check src/ tests/` — 52 files formatted
- `uv build` — wheel and sdist built
- `npm run build` in `docs/` — 103 pages built

## Independent red team

An independent GPT-5.6-terra review traced the streaming and plan/apply paths and returned **Approve, no findings**. It confirmed that component callers do not carry pipeline identity and are outside this public-resource API fix; changing component discovery/grouping semantics would be separate work.

**What I am least confident about:** no real-engine multi-pipeline Pipes fixture was run; the changed seam is covered at the exact argv boundary.

**What's the most important thing I may be missing:** an end-to-end fixture with two replication pipelines through a real `PipesSubprocessClient`. Existing plan persistence plus the new argv assertions cover the behavior changed here.

## Checklist

- [x] Conventional commit, scoped to Dagster
- [x] No `Co-Authored-By` trailer
- [x] Public docs and changelog updated
- [x] No CLI JSON schema or DSL change

> **ModelIr / contract semantics preserved: yes** — this only restores pipeline selection before execution.

### DIFF
--- a/docs/src/content/docs/dagster/resource.md
+++ b/docs/src/content/docs/dagster/resource.md
@@ -60,7 +60,7 @@ Runs `rocky plan` and returns the planned SQL statements without executing them.
 | `pipeline` | `str \| None` | `None` | Pipeline name (required when multiple pipelines are defined) |
 | `env` | `str \| None` | `None` | Optional environment name |
 
-### `run(filter, governance_override=None, *, run_models=False, partition=None, partition_from=None, partition_to=None, latest=False, missing=False, lookback=None, parallel=None) -> RunResult`
+### `run(filter, governance_override=None, *, pipeline=None, run_models=False, partition=None, partition_from=None, partition_to=None, latest=False, missing=False, lookback=None, parallel=None) -> RunResult`
 
 Runs Rocky in buffered mode (`subprocess.run`) and returns the full execution result including materializations, check results, drift detection, and permission changes.
 
@@ -70,6 +70,7 @@ Runs Rocky in buffered mode (`subprocess.run`) and returns the full execution re
 |---|---|---|---|
 | `filter` | `str` | required | Component filter (e.g. `"tenant=acme"`) |
 | `governance_override` | `dict \| None` | `None` | Per-run governance config (workspace_ids, grants), merged with `rocky.toml` defaults |
+| `pipeline` | `str \| None` | `None` | Pipeline name (required when multiple pipelines are defined) |
 | `run_models` | `bool` | `False` | Also execute compiled models (passes `--models` and `--all`) |
 | `partition` | `str \| None` | `None` | Single partition key (e.g. `"2026-04-07"`) |
 | `partition_from` | `str \| None` | `None` | Lower bound of a partition range (requires `partition_to`) |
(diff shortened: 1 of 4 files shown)


hugocorreia90 (approved):
Approving.

I traced this past the argv boundary the unit tests stop at, since `run_pipes` puts `--pipeline` only on the plan command and leaves apply bare. `rocky plan --pipeline` exists (`main.rs:714`), is persisted as `RunPlan.pipeline` / `ReplicationPlan.pipeline` (`plan.rs:1014`, `plan.rs:1323`), and apply replays it (`apply.rs:762`, `:3451`, `:3554`) — so keeping apply as `rocky apply <plan-id>` is correct, not just convenient. The one arm that drops the persisted pipeline is `--dag` (`apply.rs:629`), and nothing in the SDK or this resource can emit `--dag`, so it is unreachable from here.

Mutation-checked both forwards with the signatures left in place, so the failures are real rather than `TypeError`s: dropping `pipeline=pipeline` fails `test_run_streaming_threads_all_routing_flags`, and dropping the `build_kwargs` entry fails `test_run_pipes_threads_all_routing_flags`. The two `pipeline=None` assertions pass either way, which is the right role for a regression control.

One boundary worth recording rather than changing: `rocky plan` resolves a **replication** pipeline (`plan.rs:105` → `resolve_replication_pipeline`), so `run_pipes(pipeline=<a transformation pipeline>)` still refuses — loudly, naming the pipeline and its type. That limit is independent of pipeline count and predates this change, so "restores both execution modes" reads correctly as parity with single-pipeline projects. `pipes.md` already claimed `run_streaming` takes the same keywords as `run()`; that claim was false and is now true.

Independent red team run on the diff: it found no defect in this change. Its other findings are pre-existing engine gaps reachable through kwargs this PR did not touch — I'm filing those separately. Local on the PR head: 711 passed, `ruff check` and `ruff format --check` clean.

Issue #1403, opened 2026-08-08
bug(engine): rocky plan --shadow is silently dropped from a replication-only plan, so apply writes production and reports Success


hugocorreia90:
## Description

`rocky plan --shadow` is accepted on a replication-only project, silently discarded when the plan is persisted, and `rocky apply <plan-id>` then writes **production** targets and reports `Success`.

`PlanRunOptions` carries `shadow` / `shadow_suffix` / `shadow_schema` (`plan.rs:44-46`). The `RunPlan` branch persists all three (`plan.rs:1032-1034`), and `apply.rs` documents at length why it must — dropping them "is the same defect as #1272 one level up: `rocky plan --dag --shadow` followed by `rocky apply` would write production while reporting success."

The replication branch does exactly that. `build_and_persist_replication_plan` (`plan.rs:1309`) receives the same `PlanRunOptions` and never reads the shadow fields; `ReplicationPlan` (`output.rs:3711`) has no shadow fields to read them into; and the replication apply arm passes `None` for `shadow_config` (`apply.rs:3566`) under a comment asserting that "shadow paths are independent of replication-plan replay." That branch fires whenever the project has no `models/` directory or compile produced zero models — i.e. every replication-only project.

Reachable from the Python SDK and the Dagster resource: `RockyClient.plan(..., shadow_suffix=...)` and `RockyResource.run_pipes(context, filter=..., shadow_suffix="_pr")` both route through `_build_plan_args`.

## Steps to reproduce

Replication-only DuckDB project, two source tables under `raw__acme`, no `models/` directory:

```toml
[adapter.default]
type = "duckdb"
path = "wh.duckdb"

[pipeline.bronze]
type = "replication"
strategy = "full_refresh"

[pipeline.bronze.source]
discovery = { adapter = "default" }

[pipeline.bronze.source.schema_pattern]
prefix = "raw__"
separator = "__"
components = ["source"]

[pipeline.bronze.target]
adapter = "default"
catalog_template = "wh"
schema_template = "bronze__{source}"

[pipeline.bronze.target.governance]
auto_create_schemas = true
auto_create_catalogs = false
```

**Control — the fused path, same flags.** Shadow routing is active; nothing is written to production:

```
$ rocky run --pipeline bronze --shadow --shadow-suffix _pr
Error: 2 table(s) failed during parallel execution
$ # tables matching bronze%: none
```

**Test — plan then apply, same flags:**

```
[...]

Issue #1280, opened 2026-07-29
Split `TableTask` source and target table names so replication shadow suffix mode works


hugocorreia90:
`rocky run --shadow` (suffix mode) currently **refuses** replication pipelines. This issue tracks the underlying defect that forces the refusal.

## The defect

`TableTask` carries a **single** `table_name` used for both sides of a replication copy:

```rust
struct TableTask {
    source_catalog: String,
    source_schema: String,
    target_catalog: String,
    target_schema: String,
    table_name: String,   // <-- shared by source AND target
    ...
}
```

The connector loop stores the shadow-**suffixed** name in that field, and `process_table` then builds both refs from it:

```rust
let source_table = TableRef { schema: task.source_schema.clone(), table: task.table_name.clone(), .. };
let target_table = TableRef { schema: task.target_schema.clone(), table: task.table_name.clone(), .. };
```

So a copy of `raw.orders` reads `raw.orders_rocky_shadow`. Normally that table does not exist and the run fails with a misleading "table not found"; if a table by that name *does* exist, the run silently copies the wrong one.

`schema_override` mode is unaffected — no suffix is applied and only `target_schema` moves, so source and target legitimately share the table name. That mode is verified working and remains supported.

## Why this is not a two-line fix

`table_name` is not only the copy's source/target identity. It also feeds:

- the asset key (`asset_key.push(task.table_name.clone())`)
- drift comparison
- checks
- state / watermark keys

Splitting it into `source_table_name` and `target_table_name` therefore needs a decision per consumer about *which* name is the right one under shadow — in particular whether an asset key and a watermark should follow the production identity or the shadow one. Getting that wrong would either collide shadow and production state or strand watermarks.

## Acceptance

- `rocky run --shadow --shadow-suffix _s` on a replication pipeline reads the production source and writes the suffixed target.
- Shadow and production runs of the same pipeline do not share watermark/state keys in a way that lets one corrupt the other.
- The refusal added alongside this issue is removed, and a test covers the source-read identity directly.

PR #1406, opened 2026-08-09
fix(engine): a shadow replication run must read production, not its own target

hugocorreia90:
## Summary

`ModelIr::replication`'s `SourceRef` was built from `task.target_table_name`, so a shadow-suffix replication run emitted `FROM <source_schema>.<table><suffix>` — it read the table it was about to write.

Normally no such table exists at the source and the run fails with a misleading "table not found". Where one does exist, the run copies **that** table into the shadow target and exits 0 reporting `Success`.

**Shipped in engine-v1.69.0.**

## Reproduction

On the released 1.69.0 binary, replication-only DuckDB project, sources under `raw__acme`:

```
$ rocky run --pipeline bronze --shadow --shadow-suffix _pr
DuckDB error: Catalog Error: Table with name payments_pr does not exist!
  LINE 3: FROM raw__acme.payments_pr
Error: 2 table(s) failed during parallel execution
```

`--shadow-schema` is unaffected, and that split is the proof it is this line rather than a fixture problem: under a schema override `target_table_name == table.name` (`run.rs:3097-3105`), so the wrong field happens to hold the right value. Only suffix mode is broken — exactly what the defect predicts and not something a bad fixture produces.

The row-count check *did* notice — it reads the source through a different, correct seam and reported source 2 vs target 1 — but a failed check does not gate exit status, so the run still returned `Success`.

## Why the existing test did not catch it

`TableTask::source_table_name` exists precisely for this, and its doc comment states the failure verbatim:

> a single shared field made the copy read its own shadow target: normally absent, so the run failed with a misleading "table not found", and if a table by that name did exist it silently copied the wrong one.

#1280 split the field for that reason. #1383 then applied the split to `copy_endpoints` and to this constructor's `TargetRef`, and left the `SourceRef` on the target name — both lines are in the same commit (`063ad15d`).

Its regression test, `shadow_suffix_reads_production_and_writes_the_suffixed_target`, asserts on `copy_endpoints`, and its stated mutation contract ("build `source_table` from `task.target_table_name`") does turn it red. But **nothing derived from `copy_endpoints` reaches the copy statement**: its `source_table` feeds only the change marker, the describe, and the row-count check. The statement is generated from the `ModelIr`.

So the property was asserted at one seam and violated at the other, and a test whose own comment says *"asserting the SOURCE side is the point. A test that only checked the target would have passed throughout the bug's entire lifetime"* passed throughout the bug's entire lifetime.

## Changes

- `SourceRef.table` now uses `task.source_table_name`.
- Extracted `replication_model_ir` so that seam is testable at all — the bug existed because the construction was inline in a large async fn with no unit-level access.
- New test asserts on the **generated SQL**, not on a struct one layer up: the shadow name must appear before `FROM` and must not appear after it. It fails on precisely the mutation that shipped.
- Dropped the now-false clause from the `rocky run --dag --shadow` refusal, which cited "replication's suffix mode rewrites the source it reads" as one of its grounds. That ground disappears with this fix; the refusal stands on its main one.

## Test Plan

- New test `shadow_suffix_copy_sql_reads_production_not_the_shadow_target`, mutation-checked against the exact shipped line.
- Full suite via CI. Local verification was abandoned deliberately: the `rocky-cli` lib-test target OOM-killed this machine twice (swap 4.8G/6.1G, rustc at 0% CPU), which is a laptop limit and not a signal about the change.

## Blast radius

`ModelIr::recipe_hash` covers `source`, so this changes that hash — but only where the two names differ, and they differ **only** under shadow-suffix (`run.rs:3097-3113`). No run that works today changes identity; the only hashes that move belong to runs that currently fail or copy the wrong table. On this path the hash reaches `recipe_identity` on the materialization output — reporting, not a reuse or skip gate.

## Independent red team

Found by a parallel investigation of the plan/apply seam for #1403, then verified independently against `git blame` and a live reproduction before being acted on. Two of the investigating lenses **contradicted each other** on whether the `--dag` refusal's claim was stale: one read #1383 and concluded the split had landed, the other ran the binary and found it had not. The lens that was wrong was the one that trusted the fix commit without executing it.

**What I am least confident about:** whether any deployment depends on the current (broken) behaviour — I cannot see how, since suffix-mode shadow replication either errors or silently copies the wrong table, but the hash change is the observable difference.

**What I am most likely missing:** whether other consumers of the replication `ModelIr` beyond SQL generation and `recipe_identity` care about `source.table`. I traced those two; a third would show up as a test failure rather than silently.

> **ModelIr / contract semantics preserved: yes** — the IR now describes what the run actually does.

Refs #1383

### DIFF
--- a/docs/src/content/docs/concepts/shadow-mode.md
+++ b/docs/src/content/docs/concepts/shadow-mode.md
@@ -16,8 +16,9 @@ Shadow mode writes pipeline output to shadow tables instead of (or alongside) pr
 
 :::caution[Where shadow isolation applies]
 Isolation covers a plain `rocky run` over **transformation** pipelines, and
-**replication** pipelines under `--shadow-schema` (or a branch). Everywhere else
-Rocky now refuses the flag instead of running without isolation:
+**replication** pipelines in either mode — `--shadow-suffix` or `--shadow-schema`
+(or a branch). Everywhere else Rocky refuses the flag instead of running without
+isolation:
 
 - **`rocky run --dag`** refuses `--shadow` / `--branch` outright. The DAG runs
   each model as its own sub-run, so a model's reads of an upstream built by the
@@ -25,10 +26,12 @@ Rocky now refuses the flag instead of running without isolation:
   shadow table would be built from production data while the run reported
   success. Run the shadow pipeline without `--dag`.
 - **Snapshot and load** pipelines refuse it: their targets are not rewritten.
-- **Replication in suffix mode** refuses it. The suffix would be applied to the
-  table name that the source read and the target write share, so the run would
-  read `<source_schema>.<table>_rocky_shadow`. Use `--shadow-schema` instead,
-  which moves only the target schema and leaves the source alone.
+Replication supports both shadow modes, and they isolate differently. Under
+`--shadow-suffix` the copy writes `<table><suffix>` **in the pipeline's own
+target schema**, reading the unsuffixed production source. Under
+`--shadow-schema` the whole target schema moves and table names are untouched.
+Prefer the schema override when you want the shadow objects kept away from
+production tables rather than sitting beside them.
 - **Seeds** cause a `--dag` run to be refused along with the rest; `rocky seed`
   itself has no shadow mode and always writes its configured target.
 
--- a/engine/crates/rocky-cli/src/commands/run.rs
+++ b/engine/crates/rocky-cli/src/commands/run.rs
@@ -333,6 +333,44 @@ fn copy_endpoints(task: &TableTask) -> (TableRef, TableRef) {
     )
 }
 
+/// Build the replication [`ModelIr`] the copy statement is generated from.
+///
+/// Extracted from `process_table` so the source/target split is testable at the
+/// seam that actually emits SQL. [`copy_endpoints`] above splits the two names
+/// correctly and has its own test, but **nothing derived from it reaches the
+/// copy statement** — its `source_table` feeds only the change marker, the
+/// describe, and the row-count check. The statement itself is built from this
+/// `ModelIr`, which is where #1280's split has to hold and where it did not.
+fn replication_model_ir(
+    task: &TableTask,
+    strategy: MaterializationStrategy,
+    governance: GovernanceConfig,
+) -> ModelIr {
+    ModelIr::replication(
+        TargetRef {
+            catalog: task.target_catalog.clone(),
+            schema: task.target_schema.clone(),
+            table: task.target_table_name.clone(),
+        },
+        strategy,
+        SourceRef {
+            catalog: task.source_catalog.clone(),
+            schema: task.source_schema.clone(),
+            // The read keeps the PRODUCTION name — see `TableTask::source_table_name`,
+            // whose doc comment specifies exactly this. #1280 split the field for it;
+            // #1383 then applied the split to `copy_endpoints` and to the target above
+            // but left this one on `target_table_name`, so a shadow-suffix run built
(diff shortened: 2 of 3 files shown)


hugocorreia90:
## Mutation check — and it demonstrates the seam gap rather than just asserting it

Ran both shadow tests against the fix, then against the exact line that shipped in 1.69.0 (`SourceRef.table: task.target_table_name.clone()`), then restored:

```
### FIXED
  PASS  shadow_suffix_reads_production_and_writes_the_suffixed_target
  PASS  shadow_suffix_copy_sql_reads_production_not_the_shadow_target
  2 tests run: 2 passed

### MUTATED (the shipped bug)
  FAIL  shadow_suffix_copy_sql_reads_production_not_the_shadow_target
  PASS  shadow_suffix_reads_production_and_writes_the_suffixed_target
  2 tests run: 1 passed, 1 failed
```

The second line of the mutated run is the point. **The pre-existing #1383 test passes against the buggy code.** That was the claim in the PR description — that it asserts the right property at a seam the copy statement never touches — and this is the demonstration rather than the argument. A test whose own comment reads *"asserting the SOURCE side is the point. A test that only checked the target would have passed throughout the bug's entire lifetime"* does in fact pass throughout the bug's lifetime, because `copy_endpoints` is not where the statement is built.

The new test fails on that mutation and passes with the fix, so it is genuinely fix-sensitive rather than vacuous.

CI is separately green on the full suite: **6007/6007 passed**, with both tests explicitly reported PASS.

For the record, the local run was worth insisting on: CI green only shows the test passes *with* the fix. Nothing in CI can show it fails *without* it, and that is the property that distinguishes this test from the one it replaces.

hugocorreia90:
## Red team: two models, both approve — and both caught the same half-done removal

All five attack points came back clean. The independent reviewer traced the one thing I most wanted checked, that `source_table_name` is genuinely the right field in every configuration: overrides cannot rename tables (the whole override surface is strategy/keys/timestamp/enabled), `--shadow-schema` changes only the target schema, Fivetran resolves `name_in_destination` *before* producing `table.name`, and Airbyte has no destination-name field. It also confirmed the extracted helper preserves every former argument, that the replication IR reaches only SQL generation and recipe identity, and that the operational identities (watermark, drift, asset key, checks, Pipes) correctly follow the target.

Both flagged that my removal was **half-done**, and they were right:

- `run_dag_exec.rs` still carried "Replication suffix mode corrupts the SOURCE" as reason 3 of four in the comment above the refusal, even though I had removed it from the error text. Its stated mechanism — *"`TableTask` carries one `table_name` for both sides"* — had been false since #1280 split the field; only its consequence stayed true, because the `SourceRef` kept the target name. Now three reasons, with a note on why the fourth outlived the field it described.

- `docs/concepts/shadow-mode.md` was worse than stale. It told users replication **refuses** `--shadow-suffix`, and to use `--shadow-schema` instead. Replication has honoured both modes since #1280, so the documented refusal does not exist — the suffix run is accepted and then reads the wrong table, which is this bug. A user following that page would have concluded the tool was protecting them. Rewritten to describe both modes and how they differ.

Worth naming the pattern, since it is the third instance in this PR's history: a justification outlived the code it justified, and nothing failed. `TableTask::source_table_name`'s doc comment specified the correct behaviour while the constructor violated it; #1383's test asserted the right property at a seam the statement never reaches; and the `--dag` comment and the docs both kept a reason that had stopped being true. Compiling, passing CI, and reading plausibly are all compatible with a claim being false — only re-deriving it from the code catches that.

Residual, from the reviewer and worth stating rather than burying: the mutation-sensitive test starts from a constructed `TableTask`, so it does not cover discovery → suffix-shadow end to end. The defect itself is covered; the construction seam is not.

PR #1407, opened 2026-08-09
perf(engine): cap the redb cache for the request-local run-history open (#1399)

### LINKED ISSUE #1399: perf(engine/serve): Linux/glibc RSS tracks RUN_HISTORY size while /api/v1/runs is polled — request-local redb cache transients retained
## Summary

On Linux/glibc, a long-running `rocky serve --scheduler` shows RSS growth that tracks `RUN_HISTORY` size **whenever `GET /api/v1/runs` is being polled** (any dashboard or monitoring harness). A 24h scheduler soak (aarch64 Linux container, glibc, 1 run/min, `/runs` polled every 600s) grew at **+446 KB/h** (full-run OLS, 32→47.7 MB). A paired 4h control arm that was never polled on `/runs` held a flat fitted slope. The same harness on macOS holds flat — this is allocator-/platform-specific retention, not a logical leak.

This is **not** #1304 returning: the #1331 two-pass `list_runs` is present in the tested binary and working as designed.

## Mechanism (measured; confidence labeled)

Each `/runs` request opens a **request-local** `StateStore` (`engine/crates/rocky-cli/src/commands/history.rs`, via `open_read_only`) and drops it when the response returns. Frozen-history single-request probes (three sizes × three replicates each, gap-free sampling of `/proc/<pid>/smaps_rollup`, every request verified HTTP 200 with a full 50-row page) measured the RSS retained after **one** request against a quiesced baseline:

| rows in RUN_HISTORY | retained step after one `/runs` | per row |
|---|---|---|
| 216 | +1.0 MB | 4.8 KB |
| 1,435 | +8.6 MB | 6.0 KB |
| 5,760 | +18.8 MB | 3.3 KB |

- **Step-then-flat (confirmed):** 2,100 further requests at a frozen size add ~nothing — this is a high-water mark, not a per-call leak.
- **Cache attribution (strong but confounded):** a measurement build capping redb's cache to 1 MiB collapsed the 1,435-row step to +1.0 MB (request latency unchanged, 45→48 ms). Confound disclosed: the cap was applied globally in the measurement build, so it also affected startup opens (capped baselines sat ~2 MB higher), meaning the headline ratio overstates the request-only effect; the absolute post-request means still differ by ~5.6 MB. redb source confirms `set_cache_size` changes cache budgets/eviction only, not rows visited.
- **Interpretation:** pass 1 of `list_runs` must visit every row (`RUN_HISTORY` is keyed by `run_id`), pulling table pages through a redb page cache that a request-local store **can never reuse** — it is dropped one response later. The freed transients are retained by glibc arenas as high-water; macOS returns them. Exact allocator mechanics unproven (`mallinfo2`/`malloc_trim` probes would settle it).
- **Scaling caveat:** per-row cost is sublinear over the measured 1,435→5,760 interval but *rose* over 216→1,435; behavior beyond 5,760 rows (B-tree depth, allocator size classes) is unmeasured. The default retention horizon (`max_age_days = 365`, 1 run/min) is ~525k rows — ~91× the largest measured size.

## Why it matters

A monitored Linux deployment ratchets RSS with history size for as long as retention allows growth. With defaults, that is a year of runs. Memory-gated environments (containers with limits) will see steady growth attributable to observability itself.

## Proposed direction (scoped)

Give the **request-local history open** an explicit small cache budget, rather than the default. Explicitly **not** proposed: changing `open_read_only` globally (other users — model history, metrics, LSP schema-cache reads, durable-job fallback — are unmeasured), or holding a resident `StateStore` in the server (a live redb handle would hold the exclusive file lock against scheduler children and `persist_job`; the path-not-handle design is deliberate).

Related observation for whoever picks this up: the nominally read-only open currently begins and commits a **write** transaction to eagerly open all tables — worth revisiting alongside the cache policy.

Risks to measure in a prototype before merging: extra I/O and full-scan latency at large histories, lock occupancy / `Busy` rates. Validation gate: a repeated-poll scheduler soak on a real Linux host with the scoped prototype, plus at least one substantially larger frozen history, with p95/p99 `/runs` latency and `Busy`-rate accounting. Whether a `started_at` secondary index eventually becomes necessary at horizon scale is an open question — the measurements here neither require nor rule it out.

## Environment

rocky 1.68.0-dev at `ab473ac3`, aarch64 Linux (Debian bookworm container), glibc malloc, redb 2.6.3. Probe protocol: fresh `serve` (no `--scheduler`) per probe on a pristine state copy → 60s settle → 5 min baseline @10s → one verified `/runs` → 30s @1s → 10 min @10s; step = mean(last 5 min) − mean(last 3 min of baseline).

hugocorreia90:
Closes #1399.

## What

`GET /api/v1/runs`, `rocky history`, and the MCP `history` tool all open a **request-local** `StateStore` — open, read one 50-row page, drop. redb's default cache capacity (~1 GiB) is sized for a long-lived handle that re-reads pages; a request-local store can never reuse it, yet `list_runs` pass 1 must visit every `RUN_HISTORY` row (the table is keyed by `run_id`, so the newest N cannot be found without a full scan). The scan fills that cache with pages the process frees one response later, and on Linux/glibc the RSS is not returned — so a polled `/runs` ratchets `serve`'s RSS with history size.

This adds `StateStore::open_read_only_with_cache`, threading an explicit redb cache budget through `open_inner`, and applies a 1 MiB budget to the run-history path only.

## Evidence

Frozen-history probes on Linux/aarch64 (glibc, redb 2.6.3). Each: fresh `serve`, pristine state copy, 60s settle, 5-min baseline, **one** verified 200/50-row `/runs`, 10-min post window; step = mean(last 5 min) − mean(last 3 min of baseline), `/proc/<pid>/smaps_rollup`. 3 accepted replicates per cell; probes with any >30s sampling gap were rejected and re-run.

**Retained RSS after one request:**

| rows | default cache | 1 MiB budget |
|---|---|---|
| 216 | +1.0 MB | +2.8 MB |
| 1,435 | +8.6 MB | +1.6 MB |
| 5,760 | +18.8 MB | +1.4 MB |
| 50,000 | **+211 MB** | **+4.0 MB** (−98.1%) |

**Latency** (120 requests/probe, accepted attempts only, n=360): at 50,000 rows p50 165ms / p95 167ms / p99 187ms with the budget vs 170/179/205ms without. No regression — both builds re-open per request, so there was no cross-request cache reuse to lose.

**MCP `history` with a trigger filter** (the worst case for a small cache — `list_runs_matching` decodes candidates during the scan), 50,000 rows, peak RSS (`VmHWM`), 5 reps: no-match 27 MB vs 235 MB, match-all 28 MB vs 235 MB. (This probe holds stdin open for a fixed interval, so its wall-clock is floored and says nothing about latency; the HTTP numbers above are the latency evidence.)

## What this does NOT claim

- **The pre-registered acceptance rule failed and is recorded as a failure, not relabelled.** It required steps "size-independent and ≤1.5 MB at every size"; the worst was 4.0 MB. The threshold looks mis-specified rather than the fix failing — it is violated at 216 rows (2.8 MB), where table size cannot be the cause, and the budgeted arm's within-cell spread (0.7–2.8 MB for identical configs) is comparable to its between-size differences, while the default-cache arm replicates within ±1%. That reading is an inference, not a demonstration.
- No claim of size independence, of a proven allocator mechanism, of safety at the 365-day retention horizon (~525k rows — 50k is under 10% of it), or of repeated-poll flatness. Whether a `started_at` secondary index is eventually needed at horizon scale remains open and is not answered here.
- Not a hard RSS ceiling: cache bookkeeping is extra, and a single oversized `RunRecord` can transiently exceed the budget.

## Scope

Only `history_runs_output_filtered` takes the budget. `open_read_only`'s other users — `model_history_output`, recipe history, the `--audit` re-open, LSP schema-cache reads, metrics, durable-job fallback — are unmeasured and unchanged. Model history is also reachable over HTTP and MCP, so polling *those* surfaces retains the original behaviour; that is a known residual, not a regression. Separately worth revisiting: this nominally read-only open still begins and commits a write transaction to eagerly open its tables.

## Review

Reviewed adversarially by Codex (GPT-5.4) with GPT-5.6 Terra as an independent second reviewer.

- Confirmed by source inspection: the `None` path is byte-identical to the previous behaviour (`Database::create` is `builder().create(...)`); 1 MiB splits to 943,713 B read / 104,857 B write with no minimum or page rounding; the open-time write transaction is bounded by table count, not history size, so the budget cannot endanger it; a budget combined with the recreate policy is currently unreachable (`ReadOnly + Fail` only).
- **Dissent, recorded:** Terra voted to block on the ground that the unit test is a parity test — it passes if the cap is removed — and separately that the pre-registered gate failed. Codex agreed with the diagnosis but not the severity, since the endpoint ladder tests the actual behaviour and gates the merge. Follow-up, not done here: assert on redb `cache_metrics` evictions so a removed cap fails a unit test.
- Terra also caught that the first latency figures included gap-rejected attempts; the numbers above are accepted-only (that correction moved p99 from 199ms to 187ms).

## Merge gate

The 24h repeated-poll scheduler soak has run: **correctness PASS (1,440/1,440 fires, zero failures, fds flat), memory gate FAIL** — final-quarter slope +228.1 KB/h against a 200 KB/h ceiling, grid-validated over 720 readings with no gaps. Per the rule registered before the run, that does not falsify the per-request evidence above; it indicates a second, unattributed growth source in the resident scheduler path at soak scale, tracked separately. See the verdict comment below. The merge decision is the maintainer's.

`perf(engine)` commit carries the change; the follow-up commit is documentation wording only, and the soak ran against the former.


hugocorreia90:
Checked this against the two failure modes the #1399 attribution turned up, since those are the ways this fix could have gone wrong:

- **Not the resident store.** That option was unsound — `flock` versus child processes — and this doesn't revive it. It caps the cache on the request-local open and leaves the handle lifecycle alone.
- **The knob doesn't leak.** `open_redb_with_retry(path, None)` preserves redb's default for every pre-existing caller, and the budget is threaded from exactly one site. That was the specific way my own earlier measurement went wrong, so it's worth stating explicitly rather than assuming.

Also verified the coverage claim rather than taking the description's word for it: `GET /api/v1/runs` (`api.rs:854`) calls `history_runs_output`, which is a straight delegation to `history_runs_output_filtered` (`history.rs:144`) — so all three named consumers do reach the budgeted open. I went looking for a gap there and there isn't one. The remaining `open_read_only` calls in that file (`:197`, `:264`, `:408`) are the model/recipe history paths, which the comment already scopes out honestly.

Two things:

**The latency claim doesn't match its own numbers.** The doc says "with no latency cost (p99 187ms vs 205ms at 50,000 rows)". Those two differ by ~10%, and the ordering isn't stated — so as written it either reports a ~10% p99 regression while calling it free, or a ~10% improvement, and a reader can't tell which. Worth naming the arms (`default 187ms → budgeted 205ms`, or the reverse) and, if the budgeted side is the slower one, saying that 10% at p99 is an accepted cost rather than no cost. The memory result is strong enough on its own — 211 MB to 4.0 MB — that it doesn't need the latency line to be free.

**Nothing fails if the budget stops being applied.** `open_read_only_with_cache_reads_identically` is the right test for the property it targets — a resource limit must not change read semantics — but it passes just as well with `set_cache_size` deleted, so no test distinguishes "budget applied" from "budget silently dropped". That may be untestable without measuring RSS, in which case it's worth one line in the doc comment saying the guarantee rests on the soak rather than on CI; a future refactor that drops the builder call would otherwise be invisible.

Neither blocks. The scoping is right and it's the narrow fix the issue asked for.

hugocorreia90:
## 24h soak result: **correctness PASS, memory gate FAIL** (G3b)

The merge-gate soak has run to completion on the budgeted build (pinned `0fd0e176`; the only delta to this PR's head is documentation wording). Full 24h, **zero sampling gaps**, so unlike three earlier attempts this is a valid, judgeable run.

**Correctness — clean on every axis:**

```
ticks 5745   executions 1440   harvested_fires 1440   (exactly 1/min for 24h)
sum_failed 0   child_failures 0   state_busy 0   lock_overridden 0   drained 0   max_zombies 0
fd 11 -> 11 (every class delta 0)
```

**Memory — G3b FAIL, and the measurement is sound:**

```
final quarter: +228.1 KB/h  (ceiling 200)   over 6.00h
               720 readings, 0 dropped, 0 off-grid spacings, worst spacing 31s (grid 30s)
full run:      +335.8 KB/h,  +7.77 MB across 24h
```

This is not an instrument artifact: the metrics are populated and grid-validated, which is precisely what the earlier `INVALID` runs could not produce.

## What this does and does not mean

Written down **before** the run, in the soak directory's reading rule, so it isn't a post-hoc reinterpretation:

- **It does not falsify this PR's claim.** The per-request result — 211 MB → 4.0 MB retained at 50,000 rows, plus 235 MB → 27 MB on the MCP path — was measured independently at four sizes with three replicates each and reviewed twice. That evidence stands on the ladder, not on this soak.
- **It does say the soak-scale growth is not dominated by the path this PR fixes.** The strongest evidence is a comparison that runs the wrong way for the observer hypothesis: an **unfixed** 4h arm polled at the identical 600s cadence measured **+168 KB/h**, *lower* than this **fixed** build's +228 KB/h. Different durations and history sizes make that not a controlled comparison — but it rules out reading +228 as leftover `/runs` cost.
- **Conclusion:** there is a second, unattributed source of growth in the resident scheduler path at soak scale (~1,440 rows), separate from the request-local cache behaviour fixed here. Filed separately rather than re-opening #1399.

## Ask

The merge decision is the maintainer's, not mine. Options as I see them:

1. **Merge this** on the strength of the per-request evidence and track the residual separately — my recommendation, since the change is narrow, reviewed, CI-green, and strictly reduces retained memory wherever it applies.
2. **Hold** until the residual is attributed, if the operational flat-memory claim is what matters for the release decision.

Either way the GA question for the resident scheduler now carries a documented, quantified caveat instead of an unknown: ~228 KB/h at the final quarter, ~7.8 MB/day, with correctness perfect throughout.

hugocorreia90:
Adding the two follow-ups before merge, per tonight's triage.

**1. The test now proves the cap binds.** The existing test was parity-only — same rows in, same rows out — so it passed with `set_cache_size` deleted. It could not detect the fix being removed.

Mutation check, cap deleted:

```
new test     a_cache_budget_actually_constrains_redb   FAILS
             "got 0, which means the cap is not in force"
old test     open_read_only_with_cache_reads_identically  PASSES
```

It asserts on redb's eviction counter: evictions only occur when a cache is small enough to fill, and the default-cache arm asserts **zero** so the fixture stays discriminating.

One thing this surfaced, worth recording: redb gates that counter behind its `cache_metrics` feature. My first cut read 0 and looked like proof the cap was inert — it was measuring a disabled counter. The feature is now enabled for `rocky-core` **test builds only** (dev-dependencies), so production builds are unchanged.

**2. Latency wording — I am rejecting half of this one, with evidence.** The review asked to stop claiming "no latency cost" over a ~10% p99 delta. The delta runs the other way: 187ms is the **budgeted** arm, 205ms unbudgeted. The cap is not a trade, and the numbers were already in the comment.

The real defect was that the sentence never said which number belonged to which arm. That is now explicit. If you disagree with keeping "no latency cost", say so and I will soften it — but the measurement supports it as written.

Issue #1272, opened 2026-07-27
bug(engine): --dag, snapshot, and load accept --shadow/--branch but write production targets


hugocorreia90:
## Description

`rocky run --dag`, and the snapshot and load pipeline kinds, all accept `--shadow` / `--shadow-schema` / `--branch` and then write to **production** targets. This is the same defect class as #1269, on entrypoints that issue did not cover.

#1270 threads `ShadowConfig` into the transformation and model-only paths. These three remain.

## Steps to reproduce

`--dag` is not declared as conflicting with the shadow flags, so the combination parses:

```console
$ rocky --config rocky.toml run --dag --defer
error: the argument '--dag' cannot be used with '--defer'

$ rocky --config rocky.toml run --dag --shadow
# accepted — proceeds to config load and executes

$ rocky --config rocky.toml run --dag --branch some-branch
# accepted — proceeds to branch lookup and executes
```

The `--defer` case is the control: `conflicts_with` is enforced where it is declared, and it is not declared for the shadow flags.

## Root cause

**`--dag`** — `engine/rocky/src/main.rs` builds `shadow_config`, then the `else if dag` arm calls:

```rust
rocky_cli::commands::run_with_dag(&cli.config, &state_path, json, &skip_opts)
```

`run_with_dag` has no `ShadowConfig` parameter, so the resolved config is dropped. Each DAG sub-run in `commands/run_dag_exec.rs` then passes `None` for shadow into `run()`. A transformation node writes its configured production target and the run exits 0.

**Snapshot and load** — both are dispatched from a `run()` that holds a `ShadowConfig`, but neither callee accepts one:

- `commands/snapshot.rs`: `run_snapshot(config_path, pipeline_name, dry_run, json)`
- `commands/load.rs`: `run_load(config_path, loaded, state_path, cli_source_dir, format, target_table, pipeline, truncate, durability, json)`

## Expected behavior

A run that accepts a shadow or branch flag must never write a configured production target. Either the flag routes targets, or the combination is refused before any execution.

## Suggested fix

Fail closed first, implement routing second:

1. Add `conflicts_with_all = ["shadow", "shadow_schema", "branch"]` to the `dag` argument on `rocky run`, mirroring how `defer` and `watch` already declare their conflicts.
2. Bail early in `run()` for the snapshot and load arms when `shadow_config.is_some()`, naming the unsupported pipeline kind.
[...]

PR #1408, opened 2026-08-09
fix(engine): a replication plan must persist and replay its shadow routing

### LINKED ISSUE #1403: bug(engine): rocky plan --shadow is silently dropped from a replication-only plan, so apply writes production and reports Success

hugocorreia90:
## Summary

`rocky plan --shadow` on a replication-only project is accepted, silently discarded, and `rocky apply <plan-id>` then writes the **production** targets and reports `"status": "Success"`.

Closes #1403.

## Reproduction (released 1.69.0)

Replication-only DuckDB project, sources under `raw__acme`:

```
$ rocky plan --pipeline bronze --shadow --shadow-suffix _pr --output json
exit=0
$ # .rocky/plans/<id>.json payload keys:
['config_snapshot', 'pipeline', 'resume_latest', 'source_state_snapshot']     # no shadow key at all

$ rocky apply <plan-id> --output json
"status": "Success", "tables_copied": 2
$ # tables matching bronze%:  bronze__acme.orders, bronze__acme.payments      # production
```

The only signal the caller receives is a successful production write.

This is the #1272 shape one plan kind over. `apply.rs` already spells out why the `RunPlan` branch must not do this — *"Dropping it here is the same defect as #1272 one level up — `rocky plan --dag --shadow` followed by `rocky apply` would write production while reporting success"* — and the replication branch did exactly that, via a literal `None`.

## Why persist rather than refuse

Refusing `--shadow` at plan time was the obvious alternative and is wrong on two counts:

- `reject_unsupported_shadow` is the repo's refusal helper for kinds whose targets are not rewritten, and its own doc comment puts replication on the **routed** side: *"Transformation and replication rewrite their targets for a shadow run; snapshot and load do not."*
- The fused path has honoured `--shadow` for replication in both modes since #1280. Refusing here would mean `rocky run --pipeline X --shadow` allowed while `rocky plan --pipeline X --shadow` refused, for the same warehouse effect — the exact asymmetry the isolate-or-refuse rule exists to prevent.

The executor was already shadow-aware end to end (policy gate widened, target schema override, suffix on the target name, freeze fence checking both names). The missing links were the plan fields and the `None`.

## `--branch` is a fourth field, and the same bug

`--branch` is internally `--shadow --shadow-schema <branch.schema_prefix>`, and clap rejects it alongside the shadow flags — so a branch plan carries `shadow == false`. Persisting only the three shadow fields would have left every branch plan replaying against production: the identical defect wearing a different flag. Apply does the branch lookup, mirroring the `RunPlan` arm.

## plan_id stability

`plan_id` is `blake3({kind, payload})`, so a field that always serializes hands every unchanged replication project a new plan id on upgrade for a feature it does not use. `RunPlan.shadow` emits unconditionally; this deliberately does not, using the `skip_serializing_if = "std::ops::Not::not"` idiom already present in `output.rs`. A non-shadow plan's payload is byte-identical to before, and the pre-existing `replication_plan_same_payload_same_plan_id` still passes.

## No codegen cascade — checked, not assumed

`ReplicationPlan` derives `JsonSchema`, which normally implies a published-API change. It is **not** in the exported set: it appears in none of `schemas/`, `docs/public/openapi.json`, the Python SDK types, or the VS Code types. I ran `just codegen` and `just regen-fixtures` anyway; both produced zero drift. The derive being present does not make the type exported.

## Test Plan

The reconstruction is extracted into `replication_shadow_config` rather than left inline. Not tidying — the defect was invisible *because* the decision lived inside a long async function as a literal `None`, unreachable from any test. Same shape as the `SourceRef` fix in #1406.

- `a_non_shadow_replication_plan_serializes_without_the_shadow_keys` — fails if `skip_serializing_if` is dropped from any of the four fields, which nothing else in the type system would catch.
- `a_shadow_replication_plan_replays_its_routing` — fails if the reconstruction returns `None`, i.e. the shipped behaviour. Carries the control that a non-shadow plan must **not** acquire a shadow config: a fix that shadowed every replication apply would be worse than the bug.

Locally: 7/7 in the replication-plan group including the pre-existing `plan_id` tests; `cargo fmt --check` and `cargo check --all-targets` clean. Full suite via CI.

**What I am least confident about:** whether a shadow replication run's objects can later be re-discovered as *sources* on a subsequent plan, if a user's `--shadow-schema` value happens to match the pipeline's `schema_pattern.prefix`. That is a cross-cycle hazard the fused path already has, not something this change introduces, but this makes the plan/apply path reach it too. Untested — I built no fixture for it.

**What I am most likely missing:** whether `env` and the partition flags are dropped by the same function for the same reason. I fixed the field I could demonstrate writing production; I did not audit the rest of `PlanRunOptions` against `ReplicationPlan`.

> **ModelIr / contract semantics preserved: yes** — no IR or contract path changes; this is plan payload plus apply-time routing.

### DIFF
--- a/engine/crates/rocky-cli/src/commands/apply.rs
+++ b/engine/crates/rocky-cli/src/commands/apply.rs
@@ -3415,6 +3415,55 @@ async fn run_apply_backfill_plan(
 /// the adapter surfaces them) is treated as drift and aborts the apply with
 /// a clear "re-plan and re-apply" error. The check happens BEFORE any SQL
 /// is emitted to the warehouse.
+/// Rebuild the shadow routing a [`ReplicationPlan`] captured (#1403).
+///
+/// Deliberately the same shape as the `RunPlan` arm's reconstruction, branch
+/// lookup included. `--branch` is internally `--shadow --shadow-schema
+/// <branch.schema_prefix>` and clap rejects it alongside the shadow flags, so a
+/// branch plan carries `shadow == false` — consulting only `shadow` would
+/// replay it as a PRODUCTION run.
+///
+/// Extracted rather than inlined so the mapping is testable without standing up
+/// an apply. The defect this fixes was invisible precisely because the decision
+/// lived inside a long async function as a literal `None`.
+///
+/// `state_path` is read ONLY on the branch arm; the other arms never open the
+/// store, so a caller with no branch cannot fail here on an unreadable path.
+fn replication_shadow_config(
+    replication_plan: &ReplicationPlan,
+    state_path: &Path,
+) -> Result<Option<rocky_core::shadow::ShadowConfig>> {
+    // Mirrors the `RunPlan` arm: clap gives `--shadow-suffix` a default, and
+    // the CLI drops it before it reaches a plan unless shadow was requested, so
+    // a `None` here means "not a shadow plan" rather than "defaulted".
+    let suffix = replication_plan
+        .shadow_suffix
+        .clone()
+        .unwrap_or_else(|| "_rocky_shadow".to_string());
+
+    if let Some(ref name) = replication_plan.branch {
+        let store = rocky_core::state::StateStore::open_read_only(state_path)
+            .with_context(|| format!("failed to open state store at {}", state_path.display()))?;
+        let record = store.get_branch(name)?.with_context(|| {
+            format!("branch '{name}' not found — create it with `rocky branch create {name}`")
+        })?;
+        return Ok(Some(rocky_core::shadow::ShadowConfig {
+            suffix,
+            schema_override: Some(record.schema_prefix),
+            cleanup_after: false,
+        }));
+    }
+    Ok(if replication_plan.shadow {
+        Some(rocky_core::shadow::ShadowConfig {
+            suffix,
+            schema_override: replication_plan.shadow_schema.clone(),
+            cleanup_after: false,
+        })
+    } else {
+        None
+    })
+}
+
 async fn run_apply_replication_plan(
     root: &Path,
     config_path: &Path,
@@ -3546,6 +3595,8 @@ async fn run_apply_replication_plan(
     // plan's policy checks.
     let apply_run_id = new_apply_run_id();
 
+    let replication_shadow_config = replication_shadow_config(&replication_plan, state_path)?;
+
     crate::commands::run::run(
         config_path,
         // Execute-from-owned: the SAME snapshot the drift check read (#1120).
@@ -3563,9 +3614,10 @@ async fn run_apply_replication_plan(
         false,
         replication_plan.resume.as_deref(),
         replication_plan.resume_latest,
-        // No shadow config — branch promote and shadow paths are
-        // independent of replication-plan replay.
-        None,
+        // Shadow routing replays, exactly as it does for a `RunPlan` (#1403).
+        // Passing `None` here meant `rocky plan --shadow` was accepted and then
+        // applied against PRODUCTION, reporting Success — the #1272 shape.
+        replication_shadow_config.as_ref(),
         &partition_opts,
         // No model filter — replication runs every discovered table.
         None,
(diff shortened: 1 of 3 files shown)


hugocorreia90:
## Red team: REQUEST-CHANGES, both findings real, both fixed

Correcting the record first: **the PR text above claimed a non-shadow plan's payload is byte-identical to before, and that was false.**

`--shadow-schema` and `--shadow-suffix` are accepted WITHOUT `--shadow` — only `--branch` conflicts with them — and such a run is still a production run. The first revision persisted the descriptor anyway, so a plan whose behaviour is completely unchanged acquired a payload key and a new `plan_id`. Verified against the released binary: `rocky plan --pipeline bronze --shadow-schema shadow_s` exits 0 today and persists no shadow keys at all.

`shadow_descriptor` now drops inert descriptors. Worth noting `main.rs` already applies exactly this normalisation to `shadow_suffix` before it reaches `PlanRunOptions`, and for the same stated `plan_id` reason — but not to `shadow_schema`. I fixed it in the replication builder rather than widening `main.rs`, since that path also feeds `RunPlan` and would have shifted those ids too.

**The branch arm had no test**, which I had flagged as a gap and the reviewer confirmed the consequence of: persisting `branch: None`, or returning `None` from the branch arm, leaves the other two tests green while silently restoring production routing for *every* branch plan. So the field added specifically because it was this same bug in disguise was the one field nothing proved.

`a_branch_replication_plan_replays_the_branch_schema` now covers it against a real state store — the schema prefix lives in the branch record rather than the plan, so the arm is unreachable without one — and additionally pins that an unknown branch fails loudly rather than degrading to a production run, which is the exact failure mode this PR exists to remove.

## Mutation-checked, against the mutations the reviewer named

A test written in response to review is the kind most likely to be written to pass rather than to catch, so both new guards were checked against the specific mutation each claims to detect:

```
MUTATION A  shadow_descriptor returns value.cloned() unconditionally
  FAIL  an_inert_shadow_descriptor_is_not_persisted

MUTATION B  branch arm returns None
  FAIL  a_branch_replication_plan_replays_the_branch_schema
```

Locally 9/9 in the replication-plan group, `fmt` and `check --all-targets` clean.

## What is clean

Points 1, 2 and 4 came back clean, and 2 is the one I most wanted checked: a `ReplicationPlan` written by 1.69.0 carries none of the four new fields, and still deserializes — `#[serde(default)]` on the bool, `Option` on the rest. A plan that stopped being readable would have been a worse regression than the bug.

## Note on the previous revision

CI was fully green on it — 6016/6016, both tests reported PASS by name — and it was the revision the reviewer falsified. Green CI and correct are different claims; the `plan_id` shift was invisible to every test in the suite because no test asserted on a flag combination nobody had thought to write down.

PR #1411, opened 2026-08-10
fix(engine/rocky-cli): persist successful watermarks on partial failure

### LINKED ISSUE #1410: bug(engine): fail_fast recovery duplicates committed incremental rows
## Description

With `fail_fast = true`, a replication run can commit an incremental table's warehouse write, then suppress that table's deferred watermark because an unrelated sibling table failed. The next run reads the old watermark and appends the already-committed delta again.

This silently duplicates rows in append-only incremental targets during ordinary failure recovery.

## Steps to reproduce

1. Configure a DuckDB replication pipeline with two incremental source tables, `concurrency = 1`, and `fail_fast = true`.
2. Run once with both tables containing `(id=1, ts=2026-01-01)` so both targets and watermarks are established.
3. Add `(id=2, ts=2026-01-02)` to `a_good`; remove the configured timestamp column from sibling `z_bad` so its incremental statement fails.
4. Run again. `a_good` succeeds and appends `id=2`; `z_bad` fails. Rocky exits 2 and logs that it is skipping one pending watermark commit.
5. Restore `z_bad` and run again.
6. Query `a_good` grouped by `id`.

Confirmed on `main` at `fa33ceb62895a8ec24d457a5db259caad6e4c3a4` / Rocky 1.70.0: `id=2` has two copies, producing 3 total rows for 2 distinct IDs.

The identical control with `fail_fast = false` leaves 2 total rows and 2 distinct IDs.

## Expected behavior

A successfully materialized table's watermark must advance to match its already-committed warehouse write, even if a sibling table fails. Recovery should not append that successful delta again.

## Actual behavior

The warehouse write commits in `process_table`, which returns the new watermark as deferred state. After all table tasks finish, `run.rs` discards every deferred watermark when `fail_fast && !table_errors.is_empty()`. Because warehouse writes are not transactionally rolled back across sibling tables, the state store remains behind the target and recovery duplicates the delta.

## Suggested fix

Commit deferred watermarks for successfully materialized tables regardless of sibling failures. Only successful tables produce deferred watermarks, so this keeps redb state aligned with warehouse writes that have already committed.

Add an inline DuckDB regression that runs bootstrap, partial failure, and recovery, then asserts the successful table's delta appears exactly once. Update the execution-flow documentation, which currently says withholding all layer watermarks makes partial failure fully safe to rerun.

## Environment

- Rocky 1.70.0
- `main` at `fa33ceb62895a8ec24d457a5db259caad6e4c3a4`
- macOS 26.6.1 arm64
- DuckDB 1.5.2

mattfaltyn:
## Summary

Closes #1410.

Persist table-specific watermarks for successful incremental writes even when a sibling table fails under `fail_fast`. This keeps redb state aligned with warehouse data that has already committed and prevents the recovery run from appending the survivor's last delta twice.

## Subproject(s) touched

## Changes

- Remove the `fail_fast` gate that discarded every deferred watermark after a partial table failure. Failed tables already enqueue no watermark, so the existing batch contains only successful durable writes.
- Add a DuckDB regression covering bootstrap, one successful incremental write plus one sibling failure, state inspection, repair, and recovery without duplicate rows.
- Correct the execution-flow documentation and record the fix in the engine changelog.

## Test Plan

- Mutation proof: the new regression failed before the production change because `a_good` remained at `2026-01-01T00:00:00+00:00` instead of advancing to `2026-01-02T00:00:00+00:00`; it passes with the fix.
- `cargo test -p rocky-cli --lib watermark -- --nocapture` — 6 passed, 0 failed.
- `npm run build` in `docs/` — passed; 103 pages built.
- `cargo test --all-targets` — all completed targets passed, but the unrelated existing `run_watch_reruns_on_file_change_and_exits_clean_on_sigint` test exceeded its configured phase budgets and then blocked in cleanup locally; the run was interrupted and no child process remained.
- `cargo nextest run --all-features` — the entire all-feature test profile compiled successfully; local `cargo-nextest` then remained idle without launching child tests on two attempts, so the harness was interrupted.

## Checklist

- [x] If this is a CLI JSON schema or DSL syntax change, all consumers (`integrations/dagster/`, `editors/vscode/`) are updated in this same PR (not applicable; no schema or DSL change)

### Engine (`engine/`)
- [ ] `cargo test --all-targets` passes (see unrelated local watch-test limitation above)
- [x] `cargo clippy --all-targets -- -D warnings` is clean (also passed with `--all-features`)

- [ ] `uv run pytest` passes

- [ ] `npm run lint` is clean

### DIFF
--- a/docs/src/content/docs/concepts/execution-flow.md
+++ b/docs/src/content/docs/concepts/execution-flow.md
@@ -168,13 +168,13 @@ Each check runs a `SELECT` against the freshly written target table. Failed chec
 
 ### 6g. Defer watermark write
 
-Rocky does **not** immediately write the watermark to the state store after a model succeeds. Instead, it queues the write. Only when the entire layer completes successfully does Rocky commit all watermarks in that layer in a single transaction.
+Rocky does **not** immediately write the watermark to the state store after a model succeeds. Instead, it queues the write. After the layer drains, Rocky commits the watermarks for every successfully written model in a single transaction; failed models never enqueue one.
 
-If any model in the layer fails, no watermarks are committed for that layer. This means a partial layer failure is fully safe to re-run: every model in the layer will start from its previous watermark.
+Successful model watermarks are committed even when a sibling fails. Warehouse writes commit independently per model, so advancing each survivor's table-specific watermark keeps state aligned with data that is already durable and prevents a retry from appending the same incremental delta again.
 
 ## Step 7: Batch-commit watermarks
 
-After a layer completes (all models succeeded, or the run is in partial mode), Rocky commits the deferred watermarks in one redb transaction. Atomic: all-or-nothing.
+After a layer completes, Rocky commits all successful deferred watermarks in one redb transaction. The state update is atomic across those successful models; failed models retain their prior watermarks.
 
 ## Step 8: Fire post-run hooks
 
--- a/engine/CHANGELOG.md
+++ b/engine/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`fail_fast` recovery no longer duplicates a successful incremental table's last delta.** Warehouse writes commit independently per table, but a sibling failure previously suppressed every deferred watermark in the run. Recovery then read the successful table's old watermark and appended its already-committed rows again. Successful table watermarks now advance after the fan-out drains; failed tables retain their prior state. (#1410)
+
 ## [1.70.0] - 2026-08-10
 
 ### Fixed
--- a/engine/crates/rocky-cli/src/commands/run.rs
+++ b/engine/crates/rocky-cli/src/commands/run.rs
@@ -4144,50 +4144,41 @@ pub async fn run(
 
     // --- Batch watermark updates (no lock contention — sequential post-run) ---
     //
-    // Under `fail_fast = true` semantics, any table failure should prevent
-    // the surviving tables from advancing their watermarks — otherwise the
-    // next run resumes from a point that implies siblings succeeded. Under
-    // `fail_fast = false` (the default, partial-success semantics), commit
-    // whatever completed cleanly so forward progress isn't lost.
-    let skip_watermarks_due_to_failure = fail_fast && !table_errors.is_empty();
-    if skip_watermarks_due_to_failure {
-        tracing::warn!(
-            deferred = deferred_watermarks.len(),
-            failed_tables = table_errors.len(),
-            "fail_fast + partial failure: skipping {} pending watermark commits so the next run starts from the same baseline",
-            deferred_watermarks.len(),
-        );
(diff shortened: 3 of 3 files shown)


hugocorreia90 (approved):
Approving. The bug is real and the fix is correct. I verified the load-bearing premise against the code rather than taking it from the description, because the failure mode if it were wrong is silent data *loss* — a watermark advancing past rows that were never written — which is worse than the duplication being fixed.

**"Failed tables never enqueue a watermark" holds.** All three `deferred_watermarks.push` sites (`run.rs:3749`, `:4110`, `:11528`) sit strictly inside `Ok(TableOutcome::Materialized(..))` arms; the failure arms only push `TableError`. The retry path correctly drops a table's error when its retry succeeds, so a table can't be both errored and watermarked.

**It's stronger than the description claims.** The watermark value is `MAX(target.ts)` read back *from the target after the write* (`resolve_new_watermark`, `:11249`), specifically to close a source-side TOCTOU. A deferred watermark therefore cannot describe anything but already-durable rows — durability isn't incidental to the enqueue, it's the documented design at the construction site. That's the sentence I'd put in the changelog.

**The gate was already inconsistent with the engine's other terminal seam.** The SIGINT flush at `:3909-3939` commits every deferred watermark unconditionally, ignoring both `fail_fast` and `table_errors`. So an interrupted run with failures already kept its watermarks while a `fail_fast` run discarded them — same durability facts, opposite conclusions. Removing the gate makes the two agree.

The removed comment's rationale ("the next run resumes from a point that implies siblings succeeded") assumed a cross-table coupling that doesn't exist — watermarks are keyed per `TableRef::state_key()`. And withholding one never rolled back the warehouse write; it only made state disagree with what was durable.

Also checked and clear: quality checks are consumed at `:4759`, *after* the commit site, so check failures never gated watermarks in either version; and `run_dag_exec.rs` has no watermark handling, so there's no `--dag` divergence.

---

Two non-blocking notes.

**1. The docs edit keeps a frame that was already wrong.** §6g and Step 7 sit inside the *transformation-model* narrative — 6e describes `ModelIr`, 6f describes `[[tests]]` sidecar blocks — and speak of "models" and "layers". But this mechanism is the replication table fan-out: `deferred_watermarks` is populated only from `TableOutcome::Materialized`, and transformation models write no watermarks at all. So the page was already misattributing a replication mechanism to the model flow; this PR fixes the claim but keeps "every successfully written model" and "After the layer drains".

Your changelog entry gets it right — "table", "fan-out". Worth using those same words in the docs. Pre-existing, so fine to leave for a follow-up.

**2. The fixture names are load-bearing.** With `concurrency = 1` and `fail_fast`, if `z_bad` were processed first its `abort_all` would kill `a_good` before it materialized and the watermark assertion would fail. It's safe — DuckDB discovery does `ORDER BY table_name` — but that's invisible at the test site. A one-line comment would stop someone renaming the fixtures and getting a confusing failure.

Thanks for the issue write-up — the `fail_fast = false` control run and the pinned commit made this quick to confirm.

Issue #1124, opened 2026-07-16
RockyClient.compile() silently ignores model_filter when server_url is set


mattfaltyn:
## Subproject

`sdk/python/` — `rocky-sdk` 0.8.0

## Description

`RockyClient.compile(model_filter=...)` silently discards `model_filter` when the client is configured with `server_url`.

CLI mode forwards the requested model as `--model <name>`, but server mode returns early and always requests the unfiltered `GET /api/v1/compile` endpoint. The endpoint itself runs compilation with `model_filter = None`.

This produces a successful, valid-looking response containing unfiltered diagnostics even though the caller requested one model.

## Environment

- rocky-sdk: 0.8.0 ([`sdk-v0.8.0`](https://github.com/rocky-data/rocky/releases/tag/sdk-v0.8.0))
- Canonical `main`: `7a03e647f8336d93fcc3f748d82d00c6a02037a4`
- Python: 3.14.4
- Platform: macOS 26.5.2, arm64

The affected implementation is identical in the release tag and current `main`.

## Minimal reproduction

```python
from unittest.mock import Mock

from rocky_sdk import RockyClient

body = (
    '{"version":"1","command":"compile","models":0,'
    '"execution_layers":0,"diagnostics":[],"has_errors":false}'
)

filtered = RockyClient(server_url="http://localhost:8080")
filtered._http_get = Mock(return_value=body)
filtered.compile("orders")

default = RockyClient(server_url="http://localhost:8080")
default._http_get = Mock(return_value=body)
default.compile()

cli = RockyClient()
cli._version_checked = True
cli.run_cli = Mock(return_value=body)
cli.compile("orders")

print("filtered server:", filtered._http_get.call_args.args[0])
print("default server:", default._http_get.call_args.args[0])
print("filtered CLI:", cli.run_cli.call_args.args[0])
```

Actual output, reproduced twice:

```text
filtered server: /api/v1/compile
default server: /api/v1/compile
filtered CLI: ['compile', '--models', 'models', '--model', 'orders']
```

The existing unfiltered HTTP fallback test still passes:

```text
1 passed, 39 deselected
```

## Expected behavior

When `server_url` and `model_filter` are both set, the SDK should either:

1. preserve the requested filter through an API that supports it, or
2. fail explicitly before making the HTTP request.

It should not silently return unfiltered results.

## Actual behavior

Filtered and unfiltered server calls are indistinguishable: both request `/api/v1/compile` and return successfully.

## Root cause

[...]

PR #1423, opened 2026-08-12
fix(engine/rocky-cli): scope compile model output

### LINKED ISSUE #1422: bug(engine): compile --model returns project-wide output and accepts unknown models
## Description

`rocky compile --model <name>` applies the selector only to diagnostics and
expanded SQL. The rest of the command still reports whole-project state:
`models`, `execution_layers`, `models_detail`, text execution order, and
`has_errors` all come from the unfiltered compile result.

This creates two user-visible correctness failures:

1. An unknown model name exits successfully and returns the entire project.
2. A selected healthy model can exit nonzero with `diagnostics: []` because an
   unrelated model has an error that was hidden by the diagnostic filter while
   the global `has_errors` flag remains true.

The public CLI contract says `--model` filters compilation to one model, and
the documented example reports `models: 1` and `execution_layers: 1`.

## Steps to reproduce

Using the `00-playground-default` example on current `main`:

```console
rocky \
  --config examples/playground/pocs/00-foundations/00-playground-default/rocky.toml \
  --output json \
  compile \
  --models examples/playground/pocs/00-foundations/00-playground-default/models \
  --contracts examples/playground/pocs/00-foundations/00-playground-default/contracts \
  --model revenue_summary
```

The result reports all three models (`raw_orders`, `revenue_summary`, and
`customer_orders`) and three execution layers rather than the selected model.

Changing only the selector also succeeds:

```console
rocky \
  --config examples/playground/pocs/00-foundations/00-playground-default/rocky.toml \
  --output json \
  compile \
  --models examples/playground/pocs/00-foundations/00-playground-default/models \
  --contracts examples/playground/pocs/00-foundations/00-playground-default/contracts \
  --model does_not_exist
```

Actual result: exit 0, `has_errors: false`, and the same three-model output.

For the contradictory failure case, add two independent models that target the
same physical table so they receive E036, then select `revenue_summary`.
Compilation exits 1 while returning `diagnostics: []`, `has_errors: true`, and
all project models in `models_detail`.

The behavior was reproduced twice from a clean checkout of canonical `main` at
`96d95b0b2706cc62193c760d1b3981b8abb8f0e1`.

## Expected behavior

- A known `--model` selector returns a self-consistent selected-model view in
  JSON and text output.
- `models`, `execution_layers`, `models_detail`, execution order, diagnostics,
  expanded SQL, and `has_errors` all describe the same selected scope.
- An unknown model is rejected with a clear nonzero error instead of producing
  a false-green whole-project result.
- An error belonging only to an unrelated model does not make the selected
  model result fail without a visible diagnostic.
- Full-project compilation may remain internal so dependency and type context
  are still available; this issue concerns the command's reporting and failure
  contract, not partial compiler execution.

## Actual behavior

- Known selection: exit 0 with `models: 3`, `execution_layers: 3`, and three
  `models_detail` entries.
- Unknown selection: exit 0 with the same whole-project output.
- Unrelated E036 errors: exit 1 with `diagnostics: []` and `has_errors: true`.

## Root cause

In `engine/crates/rocky-cli/src/commands/compile.rs`, `compile_inner` filters
diagnostics at the end of the compile, but constructs model details, text data,
counts, layers, and `has_errors` directly from the global `CompileResult`.
There is no selector-existence check.

## Acceptance criteria

- Validate a supplied model name against the compiled project.
- Derive every selected output field and the exit decision from one consistent
  reporting scope.
- Preserve the current no-selector behavior.
- Preserve the existing JSON field schema; no consumer/codegen cascade should
  be necessary.
- Add regression coverage for:
  - a known model in a multi-model project;
  - an unknown model;
  - an unrelated model error;
  - a selected model error;
  - JSON/text agreement for model counts and failure state.

## Rocky version

`rocky 1.70.1`; canonical `main` at
`96d95b0b2706cc62193c760d1b3981b8abb8f0e1`.

## OS / architecture

macOS 26.6.1, arm64.

mattfaltyn:
## Summary

Closes #1422.

Make `rocky compile --model <name>` return one self-consistent selected-model
view instead of mixing filtered diagnostics with project-wide counts, details,
execution order, layers, and failure state. Unknown model names now fail
explicitly rather than returning the whole project with exit 0.

This continues the model-selector correctness work from #1160 (`run --model`
rejects unknown names) and #1166 (`plan --model` previews the scope that apply
executes), both authored by @mattfaltyn.

## Subproject(s) touched

- [x] Cross-project (reference docs)

## Changes

- Validate a supplied selector against the compiled transformation project and
  return the existing typed `ModelNotFound` error when it is absent.
- Derive diagnostics, `has_errors`, model/layer counts, `models_detail`, expanded
  SQL, text execution order, typed-column counts, and source mappings from one
  exact-model reporting scope.
- Keep full-project compilation internally for dependency and type context, and
  preserve the no-selector path's existing project-wide behavior.
- Map the newly surfaced typed error to MCP's existing `model_not_found` code
  instead of flattening it to `compile_failed`.
- Add real-binary regressions for selected JSON/text output, unknown selectors,
  unrelated and selected errors, and no-selector compatibility, plus an MCP
  taxonomy regression.
- Document the exact-selector reporting contract in the CLI reference and
  unreleased engine changelog.

No CLI JSON field shape, DSL syntax, generated binding, dependency, or persisted
IR contract changes.

## Duplicate coverage check

Before filing #1422 and again immediately before opening this PR, GitHub issue
and PR searches covered `compile --model`, unknown model handling,
`has_errors`, and `execution_layers`. No existing issue or open PR covers this
engine CLI defect. #1124/#1125 concern the Python SDK's server-mode filter, and
#1165/#1166 concern plan/apply preview fidelity; neither fixes `rocky compile`.

## Test Plan

- `cargo test -p rocky-cli -p rocky-mcp -p rocky --all-features --no-run` — passed.
- Ad-hoc-signed focused CLI integration target before the final compatibility
  case was added — 4 passed, 0 failed (selected JSON/text scope, unknown model,
  unrelated error, selected error).
- Ad-hoc-signed `rocky-mcp` focused test
  `compile_unknown_model_is_model_not_found` — 1 passed, 0 failed.
- Updated five-case `compile_model_scope` target — compiled successfully; both
  the test executable and child `rocky` binary pass strict code-signature
  verification. Execution from Codex remained parked in macOS `_dyld_start`,
  so the added no-selector case was compile/clippy verified locally; the final
  target passed in the all-feature Linux Nextest job below.
- A second full local Nextest attempt through macOS Terminal reached
  `cargo-nextest`, then macOS blocked its `cargo metadata` child from reading
  the repository on the external volume. No security controls or target paths
  were bypassed; the exact Linux all-feature suite is covered by CI.
- `cargo clippy -p rocky --test compile_model_scope -- -D warnings` after the
  final compatibility test — passed.
- `npm run build` in `docs/` — passed (103 pages built).
- `bash scripts/lint-adapter-boundary.sh` from the repository root — fails on
  existing imports in unchanged `output.rs`, `compact.rs`, `discover.rs`,
  `export_schemas.rs`, `restore.rs`, `archive.rs`, `replay.rs`, and
  `run_content_addressed.rs`; this change adds no adapter import.
- GitHub `engine-ci / Test` (`cargo nextest run --all-features`) — passed on the
  final head, including the five-case integration target.
- All final-head GitHub checks passed: Test, Clippy, Format, adapter boundary,
  release smoke, codegen drift, both credential-containment checks, and the
  credential-free agent error-contract harness.

## Independent red team

An independent GPT-5.6-terra review inspected the complete diff, issue contract,
CLI/MCP callers, diagnostic producers, and no-selector behavior. It returned
**no P0-P2 findings**. The reviewer identified explicit no-selector compatibility
as the most important missing proof; that regression was added before commit.

**What I am least confident about:** whether maintainers intend `--model` to
report only the exact model or its dependency closure. This implements exact
model scope: it matches the public “single model” contract and Rocky's established
selected-run behavior, while still compiling the whole project internally for
dependency/type context.

**What's the most important thing I may be missing:** execution of the complete
five-case integration target on this macOS host. The final target compiles and
lints, its first four behavioral cases passed in the prior signed build, and the
complete target passes in Linux all-feature CI; macOS Terminal currently lacks
permission to read this external-volume worktree for the final local run.

**ModelIr / contract semantics preserved: yes** — this changes command reporting
and selector validation only; compiler IR construction, types, contracts, and SQL
generation are unchanged.

## Checklist

- [x] If this is a CLI JSON schema or DSL syntax change, all consumers (`integrations/dagster/`, `editors/vscode/`) are updated in this same PR (not applicable; no schema or syntax change)

### Engine (`engine/`)

- [x] `cargo test --all-targets` passes (repository-standard all-feature Nextest job)
- [x] `cargo clippy --all-targets -- -D warnings` is clean (also passed with `--all-features`)

- [ ] `uv run pytest` passes (not touched)
- [ ] `uv run ruff check` is clean (not touched)
- [ ] `uv run ruff format --check` passes (not touched)

- [ ] `npm run compile` succeeds (not touched)
- [ ] `npm run test:unit` passes (not touched)
- [ ] `npm run lint` is clean (not touched)

### DIFF
--- a/docs/src/content/docs/reference/commands/modeling.md
+++ b/docs/src/content/docs/reference/commands/modeling.md
@@ -23,7 +23,7 @@ rocky compile [flags]
 |------|------|---------|-------------|
 | `--models <PATH>` | `PathBuf` | `models` | Directory containing `.sql` and `.toml` model files. |
 | `--contracts <PATH>` | `PathBuf` | | Directory containing data contract definitions. |
-| `--model <NAME>` | `string` | | Filter compilation to a single model by name. |
+| `--model <NAME>` | `string` | | Restrict the reported result and exit status to one exact model name. The full project is still compile-checked for dependency and type context. |
 | `--expand-macros` | `bool` | `false` | Expand macros from `macros/` and include the expanded SQL in the output. |
 | `--target-dialect <DIALECT>` | `dbx` \| `sf` \| `bq` \| `duckdb` | | Run the **P001 dialect-portability lint** against the chosen target. Non-portable constructs emit `error`-severity diagnostics. Precedence: flag > `[portability] target_dialect` in `rocky.toml` > unset. See [Portability linting](/concepts/linters/). |
 | `--with-seed` | `bool` | `false` | Execute `data/seed.sql` against an in-memory DuckDB and use its `information_schema` as the source-of-truth for raw source schemas. Turns leaf `.sql` models from `Unknown` columns into concrete types. Requires the `duckdb` feature (enabled by default in the shipped binary). |
@@ -100,6 +100,11 @@ rocky compile --model fct_revenue --contracts contracts/
   "compile_timings": { "load_ms": 5, "resolve_ms": 1, "typecheck_ms": 12 }
 }
 ```
+
+Model selection is exact: an unknown name is an error. Rocky still loads and
+compile-checks the full project internally so the selected model has dependency
+and type context, but the visible counts, layers, model details, diagnostics,
+and failure state describe only the selected model.
 
 Every diagnostic carries a severity (`"error"`, `"warning"`, `"info"`), a code (`E###` errors, `W###` warnings, `P###` portability lints, or `V###` validation), the owning model, and (when the compiler can locate it) a `span` and `suggestion`.
 
--- a/engine/CHANGELOG.md
+++ b/engine/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`rocky compile --model <name>` now reports exactly that model and rejects unknown names.** The selector previously filtered diagnostics and expanded SQL only: counts, execution layers, model details, text execution order, and `has_errors` still described the whole project. A healthy selected model could therefore exit nonzero with no visible diagnostic because an unrelated model was broken, while a misspelled model name exited 0 and returned the whole project. Rocky still compile-checks the full project internally for dependency and type context, but the returned report and exit status now share one exact-model scope; an unknown selector fails with `model not found`. (#1422)
+
 ## [1.70.1] - 2026-08-11
 
 ### Changed
--- a/engine/crates/rocky-cli/src/commands/compile.rs
+++ b/engine/crates/rocky-cli/src/commands/compile.rs
@@ -18,6 +18,8 @@ use rocky_sql::transpile::Dialect;
 
 use crate::output::{CompileOutput, CostHint, ModelDetail, print_json};
 
+use super::ModelNotFound;
+
 /// Execute `rocky compile`.
 ///
 /// `cache_ttl_override`: optional CLI flag value from the binary's
@@ -145,6 +147,13 @@ fn compile_inner(
 
     let mut result = compile::compile(&config)?;
 
+    if let Some(filter) = model_filter
+        && result.project.model(filter).is_none()
+    {
+        return Err(anyhow::Error::new(ModelNotFound(filter.to_string())));
+    }
+    let in_scope = |name: &str| model_filter.is_none_or(|filter| name == filter);
+
     // Portability lint. Effective target_dialect = CLI flag > [portability]
     // config > unset. Project-wide allow list and per-model `-- rocky-allow:`
     // pragmas suppress matching constructs before they become diagnostics.
@@ -199,9 +208,7 @@ fn compile_inner(
 
         let mut expanded = HashMap::new();
         for model in &result.project.models {
-            if let Some(filter) = model_filter
-                && model.config.name != filter
-            {
+            if !in_scope(&model.config.name) {
                 continue;
             }
             let sql = expand_macros(&model.sql, &macro_defs)?;
@@ -264,21 +271,18 @@ fn compile_inner(
     }
 
     // Re-apply model filter to diagnostics (may now include E027).
-    let diagnostics: Vec<_> = if let Some(filter) = model_filter {
-        result
-            .diagnostics
-            .iter()
-            .filter(|d| d.model == filter)
-            .cloned()
-            .collect()
-    } else {
-        result.diagnostics.clone()
-    };
+    let diagnostics: Vec<_> = result
+        .diagnostics
+        .iter()
+        .filter(|d| in_scope(&d.model))
+        .cloned()
+        .collect();
 
     let models_detail: Vec<ModelDetail> = result
         .project
         .models
         .iter()
+        .filter(|model| in_scope(&model.config.name))
         .map(|model| {
             let typed_cols = result
                 .type_check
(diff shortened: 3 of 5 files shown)


hugocorreia90 (approved):
Approving. I want to record how this review went, because I initially reached the wrong conclusion and the correction is worth more than the verdict.

**I first read this as a fail-open** — that scoping `has_errors` to the exact model would let `compile --model X` exit 0 when a model X depends on is broken, and that the exit code would therefore stop being trustworthy. I built the branch and produced what looked like strong evidence: exit 0 where 1.70.1 gave exit 1, holding at multi-hop depth, with controls (`--model <the broken one>` and no-selector both still exit 1).

**That analysis was wrong, and the probe was the reason.** My "dependent" model read `poc.main.broken_upstream` — a three-part physical name, which `resolve.rs:67` deliberately classifies as an external `RawRef` rather than a model edge. So it never exhibited the compile-time dependency I was claiming to test. Four apparently independent lines of evidence turned out to share one bad premise.

The discriminating experiment settles it. When an upstream failure genuinely prevents the selected model from compiling — an upstream SQL parse error — dependency resolution fails *before* the filter applies and `compile --model down` still **exits 1**. The only case that exits 0 is E020 on the upstream, where per `typecheck.rs:205` the compiler retains and inserts the upstream's typed columns, so the downstream genuinely type-checks. And `run --model X` likewise selects only X and can legitimately build against an already-materialized upstream, which is exactly what `--defer` supports.

So the contract you implemented and documented — `compile --model X` answers *"is X's own source valid"*, not *"is X's input closure freshly rebuildable"* — is the right one. Your PR body flagged that ambiguity as the thing you were least confident about, and you resolved it correctly.

Credit where due: an independent Codex review plus a GPT-5.6-terra second reviewer refuted my analysis with the `typecheck.rs:205` and `resolve.rs:67` citations above.

---

Four non-blocking notes, none of which need to hold up the merge:

1. **MCP `model_count` documentation is now stale.** `result_types.rs:41` still says "Number of models in the project", but it receives `output.models`, which is 1 under a selector. Found by the Codex reviewer, not by me.

2. **`execution_layers` is always 1 under a selector.** `layers.iter().filter(|l| l.iter().any(in_scope)).count()` can only ever match the single layer containing the selected model — I measured it identical for a model with and without dependencies. Your `assert_eq!(json["execution_layers"], 1)` will therefore pass regardless of graph shape. Worth deciding whether the field should mean something under a selector, or be omitted.

3. **One docs sentence invites the mistake I made.** "The full project is still compile-checked for dependency and type context" reads as a promise that dependency problems still fail the command. Saying plainly that the exit status covers the selected model's own validity — and that an upstream which compiles but will not materialize is out of scope — would have saved me this round trip.

4. **A project parse failure supersedes `ModelNotFound`**, since full compilation runs before the selector check. Consistent with the documented full-context precondition; noting it only so it is a known consequence.

I will open a follow-up issue for 1-3 rather than asking you to expand this PR.

Thanks for the issue write-up and the test helper — `write_model(.., depends_on)` made it quick to construct the adversarial cases, which is ultimately what let the analysis be settled rather than argued.

PR #1426, opened 2026-08-13
fix(engine/rocky-cli): test declarative DSL models

### LINKED ISSUE #1425: bug(engine/rocky-cli): declarative tests skip .rocky DSL models
## Description

`rocky test --declarative` silently skips declarative tests attached to `.rocky` DSL models. The same sidecar and materialized target are tested correctly when the model source is `.sql`.

The command exits successfully and reports `total: 0`, so CI can appear green while declared data-quality checks never run.

## Steps to reproduce

Create a DuckDB-backed transformation project:

```toml
# rocky.toml
[adapter.local]
type = "duckdb"
path = "/tmp/rocky-declarative-dsl/repro.duckdb"

[pipeline.main]
type = "transformation"
models = "models"

[pipeline.main.target]
adapter = "local"
```

Seed its materialized target:

```sql
CREATE SCHEMA demo;
CREATE TABLE demo.checked_model (id INTEGER);
INSERT INTO demo.checked_model VALUES (NULL);
```

Add a DSL model:

```rocky
-- models/checked_model.rocky
from raw_orders
select { id }
```

and its sidecar:

```toml
# models/checked_model.toml
name = "checked_model"

[strategy]
type = "full_refresh"

[target]
catalog = "repro"
schema = "demo"
table = "checked_model"

[[tests]]
type = "not_null"
column = "id"
severity = "error"
```

Then run:

```bash
rocky -c rocky.toml test --declarative --models models --output json
```

## Actual behavior

The command exits `0` and reports no declarative tests:

```json
{
  "declarative": {
    "total": 0,
    "passed": 0,
    "failed": 0,
    "results": []
  }
}
```

Repeating the command produces the same result. Replacing `checked_model.rocky` with an equivalent `checked_model.sql` discovers the sidecar test, runs it, reports `1 NULL row(s) found`, and exits non-zero.

## Expected behavior

Declarative tests should be discovered and executed for `.rocky` DSL models exactly as they are for `.sql` models. The reproduction should report one failed `not_null` test and exit non-zero.

## Root cause

`run_declarative_tests` loads models through its private `load_all_models`, which delegates to `rocky_core::models::load_models_from_dir`. That loader discovers `.sql` sources only. Other CLI model-discovery paths use the shared `.sql`- and `.rocky`-aware project loader instead.

Relevant code:

- `engine/crates/rocky-cli/src/commands/test.rs` — `load_all_models` / `run_declarative_tests`
- `engine/crates/rocky-cli/src/models_loader.rs` — shared CLI project loader
- `engine/crates/rocky-compiler/src/project.rs` — `.sql` and `.rocky` loading contract

## Acceptance criteria

- `rocky test --declarative` discovers sidecar tests for a `.rocky` DSL model.
- A failing declarative test on a `.rocky` model produces a failing result and non-zero command status.
- Existing `.sql` declarative-test behavior is unchanged.
- A focused regression test fails on the current implementation and passes with the fix.
- `cargo nextest run --all-features`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo fmt -- --check` pass.

## Environment

- Rocky: `1.70.1`, reproduced from canonical `main` at `fcbfea1db47393f10d40c1204766fc07a721947d`
- OS: macOS 26.6.1 arm64
- DuckDB: 1.5.2

## Related work checked

- #744 added `.rocky` support to model-listing commands but did not update the declarative-test loader.
- #1328 unified recursive project discovery and sidecar loading but did not update `run_declarative_tests`.
- #1307 was closed unmerged and explicitly retained the declarative-test command's separate loader.
- #476 and #738 concern `--model` filtering and result presentation, not DSL model discovery.

mattfaltyn:
## Summary

`rocky test --declarative` used a private SQL-only model loader. As a result, sidecar tests attached to `.rocky` DSL models were silently skipped and the command could report zero tests with a successful exit.

This change reuses Rocky's shared recursive project loader, which already supports both `.sql` and `.rocky` sources, while preserving deterministic model ordering.

Closes #1425.

## Subproject(s) touched

## Changes

- replace the declarative runner's duplicate SQL-only loader with `models_loader::load_project_models`
- add a focused regression test proving a `.rocky` model's sidecar test is loaded
- document the fix in the engine changelog

## Test Plan

- [x] `cargo nextest run --all-features --retries 2 --no-fail-fast --status-level fail --final-status-level fail` (6,026 passed; 111 skipped)
- [x] `cargo fmt -- --check`
- [x] `cargo build --all-features`
- [x] End-to-end `.rocky` fixture: `rocky test --declarative` discovers one test and returns the expected failing `not_null` result instead of `total: 0`

## Checklist

- [x] Commit message follows Conventional Commits and is scoped to `engine/rocky-cli`
- [x] No CLI JSON schema or DSL syntax change; downstream consumer updates are not required

### Engine (`engine/`)

- [x] Full all-feature test suite passes via nextest (6,026 passed; 111 skipped)
- [x] `cargo fmt -- --check` passes

## Review notes

- Least confident area: the end-to-end proof uses DuckDB and validates discovery/execution through the CLI, but it does not repeat the fixture against every warehouse adapter.
- Most important missing coverage: there is no dedicated integration fixture that asserts mixed `.sql` and `.rocky` declarative tests together; the shared loader already has broader recursive-loading coverage, while this regression test targets the missing DSL path.
- Independent different-model red-team pass: not performed in the local environment; the change was manually re-reviewed against the command path, shared loader contract, issue/PR search, and full local suite.

### DIFF
--- a/engine/CHANGELOG.md
+++ b/engine/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`rocky test --declarative` now executes sidecar tests for `.rocky` DSL models.** The command's private model loader discovered `.sql` sources only, so DSL models' declared tests were silently skipped and the command exited successfully with `total: 0`. Declarative testing now uses the shared project loader and treats `.sql` and `.rocky` models consistently. (#1425)
+
 - **`rocky compile --model <name>` now reports exactly that model and rejects unknown names.** The selector previously filtered diagnostics and expanded SQL only: counts, execution layers, model details, text execution order, and `has_errors` still described the whole project. A healthy selected model could therefore exit nonzero with no visible diagnostic because an unrelated model was broken, while a misspelled model name exited 0 and returned the whole project. Rocky still compile-checks the full project internally for dependency and type context, but the returned report and exit status now share one exact-model scope; an unknown selector fails with `model not found`. (#1422)
 
 ## [1.70.1] - 2026-08-11
--- a/engine/crates/rocky-cli/src/commands/test.rs
+++ b/engine/crates/rocky-cli/src/commands/test.rs
@@ -6,7 +6,6 @@ use std::sync::Arc;
 use anyhow::{Context, Result};
 use tracing::{debug, info, warn};
 
-use rocky_core::models;
 use rocky_core::tests::{TestSeverity, TestType, generate_test_sql_with_dialect};
 use rocky_core::traits::WarehouseAdapter;
 
@@ -177,22 +176,9 @@ pub fn run_test(
 // Declarative test runner (`rocky test --declarative`)
 // ---------------------------------------------------------------------------
 
-/// Load all models from a directory including one level of subdirectories.
-fn load_all_models(models_dir: &Path) -> Result<Vec<models::Model>> {
-    let mut all = models::load_models_from_dir(models_dir).context(format!(
-        "failed to load models from {}",
-        models_dir.display()
-    ))?;
-
-    if let Ok(entries) = std::fs::read_dir(models_dir) {
-        for entry in entries.flatten() {
-            if entry.path().is_dir()
-                && let Ok(sub) = models::load_models_from_dir(&entry.path())
-            {
-                all.extend(sub);
-            }
-        }
-    }
+/// Load every `.sql` and `.rocky` model beneath the models directory.
+fn load_all_models(models_dir: &Path) -> Result<Vec<rocky_core::models::Model>> {
+    let mut all = crate::models_loader::load_project_models(models_dir)?;
     all.sort_unstable_by(|a, b| a.config.name.cmp(&b.config.name));
     Ok(all)
 }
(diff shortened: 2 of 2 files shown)


hugocorreia90 (approved):
Approving. The fix targets the right thing — `models_loader::load_project_models` is the canonical entry point, and its traversal deliberately lives in `rocky_core::model_walk` so that "every scanner of a models tree sees the same directory set." Moving the declarative runner onto it makes `rocky test` agree with `run` and `compile` about what a project contains, rather than adding a fourth opinion.

You verified the test plan; I verified the claims:

- The old `models::load_models_from_dir` really is `.sql`-only (its doc: *"models from one directory whose primary `.sql` paths satisfy `include`"*), so the new unit test fails against the previous code rather than passing for free.
- `rocky_compiler::project::load_dir_models` really does support both — *"Supports both `.sql` (with sidecar `.toml`) and `.rocky` files."*

**The PR undersells itself twice.** Both are worth putting in the changelog entry, because they are separate silent-failure classes:

1. **This was also a surviving instance of #1425's cousin, #1262.** The private loader walked the models dir plus *one level* of subdirectories — precisely the defect #1262 describes ("Models below the first subdirectory level are never loaded"), closed on 2026-08-01. The shared walk was adopted everywhere else; `test --declarative` kept its own copy and kept the bug. So this fixes DSL discovery *and* depth, and a `.sql` model nested two levels down was previously untested with no indication.

2. **The old code swallowed subdirectory load errors.** `if let Ok(sub) = models::load_models_from_dir(...)` discarded the `Err` arm, so a subdirectory that failed to load contributed nothing and said nothing. `load_project_models` returns the first error and appends walk errors, which is the same silent-drop family the walk exists to close.

Given the bug is "a test command reported success having run nothing," the fact that the previous implementation could *also* drop a whole directory silently seems worth naming.

---

Two non-blocking notes.

**`total: 0` is still a silent green.** Discovery is fixed, but a project that finds no tests still exits 0 with no signal. That is the exact shape that let #1425 hide — the command was honest about what it found, and what it found was nothing. A one-line "no tests found" warning would make the next instance of this visible rather than invisible. Your call whether that belongs here or in its own change.

**`sort_unstable_by` on name alone.** Now that the walk is recursive, two directories can contribute models with the same name, and `sort_unstable` gives no guarantee about their relative order. The compiler rejects duplicate model names, but the declarative path loads models directly without that check, so ties would order arbitrarily between runs. Either sort by `(name, file_path)` or rely on the duplicate rejection explicitly — the current comment says "preserving deterministic model ordering," which holds only while names are unique.

Nothing here blocks. Thanks for the clear issue write-up on #1425 — "reports zero tests and exits 0" is exactly the kind of failure that is easy to look at and not see.

PR #1430, opened 2026-08-14
docs: guidance accuracy sweep — sync the skills mirror, gate it, fix stale claims

hugocorreia90:
An audit of the agent-guidance stack found that every claim backed by a CI drift check was true, while several claims with no check had silently gone stale. This PR corrects the stale claims and adds the missing check.

## CI changes

- **New `skills-mirror-drift.yml`** — AGENTS.md requires `.agents/skills/` and `.claude/skills/` to stay byte-identical, but no workflow verified it, and the trees had in fact diverged: `.agents/skills/rocky-release/SKILL.md` missed the #1394 corrections and still described 5 release archives + `SHA256SUMS` + immediate publish (the real flow: 10 archives, `checksums.txt`, draft-until-complete). This PR syncs the trees and adds a `diff -r` gate.
- **`codegen-drift.yml` path fix** — it watched `00-playground-default`, but `regen_fixtures.sh` actually reads `01-replication-basics` and `02-performance/03-partition-checksum`. Edits to the real fixture sources never triggered fixture drift checking. The normalizer script is now watched too.

## Guidance corrections

- **AGENTS.md** — "there are no hand-written bindings" was false (`rocky_sdk/types.py` keeps hand-written runtime models, gated by `test_schema_parity.py`); "touching `engine/**` triggers `engine-ci.yml` only" was false (`codegen-drift.yml` matches `engine/**` too); the DSL cascade now includes the docs step the `rocky-dsl-change` skill already required; the release-recipe list now includes `release-sdk`; a new bullet documents that subproject-local skills are currently Claude-only discovery surfaces.
- **Crate counts** — AGENTS.md and CONTRIBUTING.md said 23, `engine/AGENTS.md` said 25; the workspace has 27 members. Replaced with count-free phrasing so the number cannot rot again.
- **`engine/AGENTS.md`** — removed the `rocky drift` row from the JSON output table (no such verb exists; the file's own drift section says so, and `rocky --help` confirms it); aligned the `engine-bench` description with the workflow (raw artifact upload, deliberately no baseline/alert).
- **Skills** — `rocky-dsl-change` said `lower.rs` lowers "to the IR"; it lowers to SQL (`lower_to_sql`), per the module doc and AGENT_REVIEW.md. `rocky-dev` was missing the `sdk/python` subproject entirely and pointed the codegen follow-up at the dagster shim instead of `rocky_sdk`.

One judgment call to flag: the new AGENTS.md bullet *documents* that engine/vscode-local skills are Claude-only rather than mirroring them into `.agents/` or hoisting their rules into the subproject AGENTS.md files. Happy to follow up with either of those if preferred.

A follow-up issue covers the code-side inconsistency this PR deliberately does not touch: `export_schemas.rs` exports a `drift` schema for a CLI verb that does not exist.


hugocorreia90:
Post-review updates (independent red-team pass, Codex, on the full diff):

- **Fixture inputs**: `regen_fixtures.sh` consumes nine POCs, not the two originally added to the filter — the workflow now watches `examples/playground/pocs/**` (so the set cannot rot) plus `scripts/copy_project_schema.py`, which `just codegen` invokes.
- **Untracked output**: the drift step now fails when codegen produces files that are not committed; `git diff --exit-code` alone ignores brand-new files.
- **Path accuracy**: `rocky-dev`'s DSL rows used shorthand paths that do not exist (`engine/rocky-lang` → `engine/crates/rocky-lang`), and its engine row now names both shipped binaries (`rocky` + `rocky-lsp`).
- **Parity-test scope**: AGENTS.md now states precisely what `test_schema_parity.py` guards (top-level shape; not nested or typed parity).
- **Leftover counts**: three engine-local skills still said "23-crate"; a repo-wide sweep now finds zero stale counts. CONTRIBUTING's `rocky-release` reference is vendor-neutral.

Confirmed by the same review: the mirror-gate workflow fails closed for add/modify/delete in either tree (verified against diff exit-code semantics), the release-skill sync direction was correct, and the removed `rocky drift` row matches the clap command enum. The one remaining known hole — a deleted *optional* fixture-source POC can still false-pass — is tracked in the follow-up issue, since it lives in `regen_fixtures.sh` semantics rather than the workflow.

Issue #1433, opened 2026-08-15
import-dbt emits a docs URL on a host the site no longer uses


hugocorreia90:
`rocky import-dbt` writes a `MIGRATION-NOTES.md` next to the imported project, and that file links to the migration guide:

```rust
// engine/crates/rocky-compiler/src/import/emit.rs:1018
out.push_str("4. See the [migration guide](https://rocky-data.github.io/rocky/guides/migrate-from-dbt/) for the long tail.\n");
```

The docs site is `rocky-data.dev` (`docs/astro.config.mjs`), not `rocky-data.github.io/rocky`. Every note file already written to a user's disk, and every future import, points at the older host. Whether that URL still resolves depends on a GitHub Pages redirect that is not owned by this repo's config.

Committed evidence of the emitted output: `examples/playground/pocs/06-developer-experience/19-import-dbt-unit-tests/imported/MIGRATION-NOTES.md:44`.

## Why it matters

The link is compiled into the binary, so a docs edit cannot fix it. It is also the reason the guide page must keep its current path: the in-flight documentation sweep deliberately preserved `guides/migrate-from-dbt/` rather than renaming it, precisely because the binary hardcodes that path. Renaming the page later would break notes files that already exist on disk.

## Fix

Point the string at `https://rocky-data.dev/guides/migrate-from-dbt/` and ship it with the next engine release. Worth checking at the same time whether any other user-facing string carries the old host:

```bash
rg -n 'rocky-data\.github\.io' engine/
```

PR #1434, opened 2026-08-15
docs: plain-language sweep across every page, and remove dbt positioning

hugocorreia90:
Rewrites all public documentation to a single plain-language standard, removes the dbt competitive material, and fixes the factual drift an audit found along the way.

## The standard

New: `docs/STYLE.md`, based on ASD-STE100 (Simplified Technical English). One idea per sentence, 25 words maximum, active voice, plain words, every term of art explained on first use, and an ASCII diagram wherever a page describes a flow, a tree, or a lifecycle. ASCII rather than mermaid so diagrams render everywhere and diff cleanly.

## dbt

The competitive material is gone: `getting-started/rocky-vs-dbt.md` is deleted, and the dbt columns and rows are stripped from the comparison and benchmark pages.

The usage documentation stays, because the commands ship. `rocky import-dbt` and `rocky validate-migration` are both in the clap enum, both have published JSON schemas and generated bindings in three languages, and both are exposed in the VS Code palette. Deleting their docs would leave working commands undocumented.

One constraint shaped this: `engine/crates/rocky-compiler/src/import/emit.rs:1018` compiles the guide's URL into the binary, and every `MIGRATION-NOTES.md` the importer writes points there. So `guides/migrate-from-dbt/` keeps its exact path, and its two deep-linked headings keep their exact text. (The embedded URL also names the wrong host — filed as #1433.)

## Stale claims fixed

Each was verified against the code, not against another doc:

| Claim | Reality |
|---|---|
| `rocky mcp` exposes 28 tools | 30 |
| 82 typed schemas | 87 |
| Diagnostics run E001-E035, W001-W012 | E036 and W031 exist |
| `rocky drift` is a command | It is not, and never was — drift runs inside `run`/`plan` |
| `lexer.rs`, `lowering.rs` | `token.rs`, `lower.rs` |
| `dialect.rs` is Databricks-only | Covers Databricks, Snowflake, BigQuery, DuckDB |
| Partial mask emits `al...om` | `al***om` |
| `client.run(filter={...})` | `filter` is a required positional string |
| `discovery.connectors` | `.sources` |
| SDK watchdog kills after 30s idle | Wall-clock, default 3600s |
| Cross-team contracts are not a Rocky feature | They ship: `publish-ir`, `[imports.*]`, E030-E034 |
| `plan --filter` is required | Optional |
| `emit-sql` renders every model | Ephemeral and live-warehouse strategies are reported on stderr |

Also documents surfaces that had none: `rocky apply` (42 prose mentions, no reference section), nine entirely undocumented commands, the missing `plan` and `discover` flags, and seven real `rocky.toml` sections. Config keys were read out of `config.rs` directly — `deny_unknown_fields` means an invented key is a hard load failure for whoever copies it.

## How it was checked

Each page was rewritten by one agent and then audited by an independent one. That audit found **74 defects in the rewrite** — the rewrite pass alone would not have been safe to ship. The sharpest: a new sentence claiming the benchmark harness measures Rocky only, when the reproduce commands on the same page install and run dbt; and `rocky replay` described as reproducing output byte-for-byte, when a bare invocation only inspects (`--execute --verify` does the reproduction). 77 defects were fixed, 3 rejected with evidence.

Objective gates, all green:

```
astro build ............ PASS (101 pages)
broken links ........... 0
broken anchors ......... 0     <- heading renames held
flags not in engine .... 0
config keys not in code  0
refs to deleted page ... 0
```

The link and anchor checker was calibrated against a **zero baseline on main** first — its initial run reported 20 problems that were all bugs in the checker (GitHub's slugger keeps underscores and does not collapse double spaces, and `docs/public/` assets serve from the site root).

## Also here

`engine-docs.yml` only ran on push to main, so a docs PR got no build at all and a broken page would have failed for the first time on main, mid-deploy. It now builds on docs PRs, with the deploy job and artifact upload skipped for pull requests and a separate concurrency group so a PR can never cancel an in-flight deploy. This PR is the first to exercise it.


hugocorreia90:
## Red-team disposition

Reviewer: **Codex**, independent pass over four risk slices. Verdict was **NOT MERGE-SAFE** with 23 findings. All are addressed in `10c7273c`.

**Two slices came back clean**, which is worth recording: `migrate-from-dbt.md` lost no operating content when the pitch was removed (the reviewer diffed it against `main` section by section), and the three-way deduplication of the trust-failure pitch dropped nothing unique from `index.mdx` or `architecture-of-trust.md`.

### The finding that mattered most

Two `rocky.toml` sections document behavior that **never runs**:

- `[schema_evolution]` promises disappeared columns stay `NULL`, warn during a grace period, then drop. The field parses, but the live drift path detects only additions and type changes; `detect_column_drops` has test callers only.
- Project-level `[freshness]` does not inherit `time_column` or `severity` — the conversion helper has no caller. Compile reduces the block to a boolean suppressing `W005`; scheduling reads only `expected_lag_seconds`.

My deterministic checks could not have caught this. They confirm a config section **exists** in `config.rs`; existence was never in question. Both pages now state plainly that the behavior is inert, and the engine gap is filed as **#1435** — the config still accepts input it ignores, which is the fail-open shape where an operator believes a control is active.

### The predicted failure, confirmed

Plain-language rewriting turned capabilities into guarantees in four places: "Rocky can resume" → "Rocky resumes it" (needs `--resume`/`--resume-latest`); `ai-sync` "maintains the models as upstream schemas change" (upstream detection is not wired, as the same page says 60 lines earlier); `rocky lineage --column` "shows which upstream column drives the cost" (it emits lineage edges, not cost attribution); and "the receiving service verifies" a webhook signature (Rocky adds the header; it cannot make a receiver check it).

Reassuringly, the three most technical pages — `verify-a-run`, `running-without-an-orchestrator`, `content-addressed` — kept every qualifier, scope limit, and exclusion intact.

### Two user traps fixed

- `rocky load --truncate` empties the target **once per input file**. Several files sharing one target leave only the last file's rows. Now a caution block that also explains when files share a target and when they do not.
- The `ROCKY_EXPLAINED` contract example used `[required]`/`[protected]`, which the parser does not recognise. Because that struct lacks `deny_unknown_fields` and both real fields default empty, it deserialized into a contract that **silently enforced nothing**. Replaced with the real `[[columns]]`/`[rules]` shape, plus an explicit warning that a wrong-keyed contract still parses and checks nothing.

Also fixed: `cli.md` contradicting `core-pipeline.md` on `plan --filter`; `discover --no-cache` described far more broadly than it acts; plan-flag persistence; stale-plan refusal scope (human applies are ungated); `profile`, `estimate` and `bench` behavior; the README `emit-sql` caveats and agent-gate diagram; and five `ROCKY_EXPLAINED` claims.

One finding was partly rejected with evidence rather than applied.

**Re-verified after the fixes:** astro build passes; link, anchor, flag and config checks all zero (against a zero baseline on `main`).

PR #1438, opened 2026-08-16
fix(sdk): refuse defer on the plan verb instead of building an argv the CLI rejects

### LINKED ISSUE #1404: bug(sdk/engine): the SDK emits --defer/--defer-to on rocky plan, which has no such flags, so plan() and run_pipes(defer=True) fail at argument parsing
## Description

`RockyClient._build_run_or_plan_args` is shared between the `run` and `plan` verbs on the stated premise that "the engine backfilled every `run` flag onto `plan` so the only difference is the verb" (`client.py:568`). That premise is false for `--defer` / `--defer-to`: both are defined on the `Run` variant only (`main.rs:1045`, `main.rs:1052`; the `Plan` variant spans `main.rs:705-873`).

So every `plan`-verb caller that passes `defer=True` builds an argv the CLI rejects before doing any work:

- `RockyClient.plan(...)` via `_build_plan_args`
- `RockyResource.run_pipes(context, filter=..., defer=True)`, whose docstring says its flags are "same as `run()`"
- the Dagster docs, which describe `run_pipes` as having `run()`'s semantics

## Steps to reproduce

```
$ rocky plan --pipeline bronze --defer
error: unexpected argument '--defer' found

  tip: a similar argument exists: '--filter'

$ rocky run --pipeline bronze --defer
(accepted, exit 0)
```

And the argv the SDK builds for the `plan` verb:

```python
>>> client._build_plan_args("x=y", governance_override=None, pipeline="bronze",
...                         defer=True, defer_to="prod")
['plan', '--filter', 'x=y', '--pipeline', 'bronze', '--defer', '--defer-to', 'prod']
```

Verified on the released `rocky 1.69.0` binary.

## Expected behavior

Either:

1. `rocky plan` gains `--defer` / `--defer-to` and persists them into `RunPlan` so apply replays the deferred resolution — which is what the shared-builder docstring already promises; or
2. `_build_plan_args` stops accepting `defer` / `defer_to`, and the two verbs stop sharing the flag surface for those two flags, so the mismatch is a type error at the call site rather than a CLI parse failure at runtime.

(1) is the better shape if deferred `ref()` resolution is meant to be reviewable before it executes, since that is exactly what plan/apply is for. (2) is correct if defer is deliberately run-only.

## Scope

`sdk/python/src/rocky_sdk/client.py` (`_build_run_or_plan_args`, and its docstring claim), plus `engine/rocky/src/main.rs` and `RunPlan` if option 1 is taken. The Dagster resource needs no change under either option beyond a docstring correction — `run_pipes` already forwards whatever the builder accepts.

While in there: the same shared builder is the reason `--shadow` reaches `rocky plan`, which is a separate and worse problem on the replication branch — see #1403.

- Reproduced both the rejection and the run-verb control on `rocky 1.69.0`.
- Confirmed the `Plan` and `Run` clap variant boundaries so the "missing on plan" claim is about the variant, not about a grep window.
- Confirmed the argv the SDK actually emits rather than inferring it from the source.

**What I am least confident about:** which option is intended. `--defer` is documented as a dbt-style deferral for transformation models, and `rocky plan`'s `RunPlan` blueprint only exists when a `models/` directory is present, so the flag is meaningful on plan — but that is an inference about intent, not a fact I can read off the code.

**What I am most likely missing:** whether any other flag in the shared builder has the same mismatch. I checked `--pipeline` (present on both) and `--defer` (run-only); the remaining flags in `_build_run_or_plan_args` deserve the same one-by-one check rather than trusting the docstring.

hugocorreia90:
Closes #1404.

`--defer` / `--defer-to` are declared on the `Run` clap variant **only**, but the SDK's shared argv builder appended them for both verbs — on the stated premise that *"the engine backfilled every run flag onto plan so the only difference is the verb"*. False for exactly these two flags, so a caller passing defer built `rocky plan … --defer`, which the CLI rejects at argument parsing before doing any work.

Refused in `_build_plan_args`, before a subprocess spawns, matching how the client already handles `env` on `retention_status` (`client.py:1442`).

**`run` is untouched.** The flags are legitimate there, and a fix that scoped wider would break the deferred-resolution workflow it is meant to protect — pinned by `test_run_still_accepts_defer`.

## One correction to the issue

The issue lists `RockyClient.plan(...)` as an affected caller. **It is not.** The public signature is `plan(filter, *, pipeline, env)` and never exposed defer — my first test failed with `TypeError: unexpected keyword argument 'defer'`, which is what surfaced it.

The reachable path is **dagster's `RockyResource.run_pipes`**, which puts `defer` into its build kwargs (`resource.py:1122`) and calls the same builder. So the blast radius is the Pipes path, not every plan caller. The guard still sits in the builder rather than in `run_pipes`, so any future route through it is covered.

## Two false comments corrected

Both asserted the premise that caused this:

- the shared builder's docstring ("the engine backfilled **every** run flag onto plan")
- `run_pipes`' claim that its flags are "Same as `run`" — it now names the exception

## Verification

| | |
|---|---|
| `test_build_plan_args_rejects_defer` | all three shapes: `defer`, `defer_to`, both |
| `test_build_plan_args_without_defer_is_unaffected` | ordinary plan argv unchanged |
| `test_run_still_accepts_defer` | the run verb keeps both flags |
| SDK suite | 215 passed |
| ruff check + format | clean on `sdk/python` and `integrations/dagster` |

### DIFF
--- a/integrations/dagster/src/dagster_rocky/resource.py
+++ b/integrations/dagster/src/dagster_rocky/resource.py
@@ -1080,8 +1080,11 @@ class RockyResource(dg.ConfigurableResource):
             context: Dagster execution context (required for Pipes injection).
             filter: Component filter (e.g. ``"tenant=acme"``).
             governance_override / run_models / partition / shadow_suffix / …:
-                Same as :meth:`run`. Threaded into the rocky CLI via
-                :meth:`_build_plan_args`.
+                Same as :meth:`run`, with one exception: ``defer`` / ``defer_to``
+                are ``run``-only flags, so supplying them here raises
+                ``ValueError`` -- the plan step builds the ``rocky plan`` verb,
+                which the CLI parses without them (#1404). Threaded into the
+                rocky CLI via :meth:`_build_plan_args`.
             pipes_client: Optional pre-configured ``PipesSubprocessClient``. When
                 supplied, ``asset_key_fn`` / ``include_keys`` are ignored.
             asset_key_fn: Optional transform applied to each Pipes event's asset
--- a/integrations/dagster/tests/test_resource.py
+++ b/integrations/dagster/tests/test_resource.py
@@ -1909,6 +1909,16 @@ def test_build_plan_args_mirrors_build_run_args_flag_surface():
     on the Python side — every flag ``_build_run_args`` emits must also
     appear in ``_build_plan_args``'s output for the same kwargs (only the
     leading verb differs).
+
+    The parity is NOT total, and this test used to assert that it was: it
+    passed ``defer=True`` / ``defer_to=...`` and required byte-identical
+    argv tails. Those two flags are declared on the ``Run`` clap variant
+    only, so the argv it pinned as correct is one ``rocky plan`` rejects at
+    argument parsing (``error: unexpected argument '--defer' found``,
+    exit 2). The test therefore encoded the defect in #1404 as the contract,
+    which is part of why it survived. Parity is now asserted over the flags
+    that genuinely mirror, with the run-only pair excluded and covered by
+    its own refusal test below.
     """
     rocky = RockyResource(models_dir="m")
     kwargs: dict[str, Any] = {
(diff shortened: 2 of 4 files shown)


hugocorreia90:
CI caught something worth recording: `test_build_plan_args_mirrors_build_run_args_flag_surface` (dagster) failed against this fix — because **that test pinned the defect as the contract**.

It passed `defer=True` / `defer_to="prod_main"` and asserted `run_args[1:] == plan_args[1:]`, i.e. it required `_build_plan_args` to emit exactly the argv the CLI refuses:

```
$ rocky plan --defer
error: unexpected argument '--defer' found        exit 2
$ rocky run --defer
(accepted)
```

Confirmed both ways — the clap `Run` variant declares `defer` and `defer_to`, `Plan` declares neither, and the released 1.70.1 binary behaves accordingly.

The test cites engine PR #535 for flag-surface parity. That parity is real for every flag *except* this pair, so it is now asserted over the flags that genuinely mirror, plus a second test pinning the refusal and confirming `run` still accepts both.

This is likely why the defect survived: fixing it broke a green test asserting the opposite.

PR #1448, opened 2026-08-16
fix(engine/rocky-cli): fail on unreadable audit run history

### LINKED ISSUE #1447: bug(engine/rocky-cli): audit --for treats unreadable run history as empty
## Description

`rocky audit --for <subject>` silently treats an unreadable run-history page as an empty history.

`compute_audit_for` correctly propagates failures from `list_policy_decisions`, but the adjacent run-history read uses:

```rust
let r = store.list_runs(MAX_HISTORY_SCAN).unwrap_or_default();
```

`StateStore::list_runs` intentionally returns an error when a run row in the requested page cannot be deserialized. Converting that error to an empty vector changes the meaning of the audit result:

- a real run ID is no longer recognized as a run and falls through to `AuditSubjectKind::Model`;
- the runs link reports `no_data` instead of `unavailable` or returning an error;
- the command can report that nothing references the subject even though the state store contains the requested run.

This contradicts the custody-chain contract documented in `commands/audit.rs`: every link should fail closed, and an unavailable signal must not be presented as genuinely absent data. The same computation backs the MCP `audit_query` tool, so both CLI and MCP callers are affected.

This is separate from #1304/#1331 (run-history scan cost) and #1399/#1407 (request-local redb cache retention). Those changes preserve `list_runs` error semantics and do not change `compute_audit_for`.

## Steps to Reproduce

1. Create a state store containing a run record whose `run_id` is `run-corrupt`.
2. Replace that row's `run_history` value with malformed JSON while leaving it within the newest `MAX_HISTORY_SCAN` rows. The existing `rocky-core` state test for damaged rows demonstrates this fixture shape and confirms that `list_runs(MAX_HISTORY_SCAN)` returns `Err`.
3. Run:

   ```console
   rocky audit --for run-corrupt --output json
   ```

4. The same behavior can be exercised directly by calling `compute_audit_for` against that state path.

The failure is deterministic and does not require a warehouse, credentials, timing, or concurrency.

## Expected Behavior

The run-history read failure is propagated with context, for example `failed to read run history`. The CLI should exit non-zero, and MCP should return an error rather than a successful but incomplete custody chain.

## Actual Behavior

The error is replaced with `Vec::new()`. The selector can be classified as a model, and the returned custody chain describes the run data as absent instead of unreadable.

## Proposed Fix

Propagate the existing error at the read boundary:

```rust
let r = store
    .list_runs(MAX_HISTORY_SCAN)
    .context("failed to read run history")?;
```

Add one focused regression test that injects a malformed run-history row and asserts that `compute_audit_for` returns the contextual error. No output-schema, state-schema, migration, or documentation change should be required.

## Acceptance Criteria

- `compute_audit_for` fails when `list_runs(MAX_HISTORY_SCAN)` fails.
- The returned error identifies the run-history read.
- A focused regression test fails with the current `unwrap_or_default` implementation and passes with the fix.
- Existing audit tests, engine tests, Clippy, and rustfmt pass.

## Rocky Version

`engine-v1.70.1` and current `main` at `d67951df44392ec9a7b45b72e6f994d69361bdfc`.

## OS / Architecture

Platform-independent. Initially inspected on macOS/Darwin arm64.

## Configuration

No warehouse configuration is required; the failure occurs while reading the local redb state store.

mattfaltyn:
## Summary

Make `rocky audit --for` fail closed when its run-history page cannot be read, instead of silently replacing the error with an empty history.

Closes #1447.

## Subproject(s) touched

## Changes

- Propagate `StateStore::list_runs` failures from the shared `compute_audit_for` core with `failed to read run history` context. This covers both the CLI command and MCP `audit_query`.
- Extend the existing corrupt-run test seam to `rocky-core`'s test-support feature and add a focused CLI regression test.

The successful output shape, state schema, and persistence format are unchanged. Missing state still produces the existing empty/no-data result; only a present-but-unreadable run history now returns an error.

## Failure Mode

Before this change, a malformed run row caused `list_runs` to return `Err`, which `unwrap_or_default` converted to `Vec::new()`. A real run selector could then be classified as a model and returned as a successful but unresolved custody chain. That contradicts the command's fail-closed contract.

## Test Plan

- [x] `cargo test -p rocky-cli audit_for_propagates_run_history_read_failure`
- [x] `cargo test --all-targets` (all non-ignored workspace tests and benchmark targets passed; credentialed live tests remained ignored)
- [x] `cargo fmt -- --check`

## Checklist

- [x] Commit messages follow [conventional commits](https://www.conventionalcommits.org/), scoped by subproject when relevant (e.g. `feat(engine/rocky-databricks): ...`, `fix(dagster): handle partial-success exit codes`)
- [x] If this is a CLI JSON schema or DSL syntax change, all consumers (`integrations/dagster/`, `editors/vscode/`) are updated in this same PR — not applicable; neither schema nor DSL changes

### Engine (`engine/`)

- [x] `cargo clippy --all-targets -- -D warnings` is clean (also checked with `--all-features`)

- [ ] `uv run pytest` passes — not applicable
- [ ] `uv run ruff check` is clean — not applicable
- [ ] `uv run ruff format --check` passes — not applicable

- [ ] `npm run compile` succeeds — not applicable
- [ ] `npm run test:unit` passes — not applicable
- [ ] `npm run lint` is clean — not applicable


hugocorreia90 (approved):
Approving. Correct fix, and the test genuinely pins it.

## The strongest argument for this change is one the PR does not make

The two reads are **adjacent lines in the same block**, and only one failed closed:

```rust
let d = store.list_policy_decisions()
    .context("failed to read the policy-decision ledger")?;   // fails closed
let r = store.list_runs(MAX_HISTORY_SCAN).unwrap_or_default(); // failed OPEN
```

Both feed the same custody chain. A corrupt ledger refused; a corrupt run history was silently replaced with `Vec::new()`, and the selector then fell through to model classification and returned a successful-looking but unresolved chain. That asymmetry is not a defensible tolerance, it is an oversight — which is what makes this a fix rather than a policy change.

## Verified rather than taken on trust

**The test is fix-sensitive.** Mutation-checked in an isolated worktree — reverted the fix to `unwrap_or_default()` and re-ran:

| | result |
|---|---|
| with fix | `1 passed` |
| fix reverted | `1 failed` |

So it fails for the intended reason, not incidentally.

**The seam widening does not reach the binary.** `insert_corrupt_run_history_row` moving to `pub` under `#[cfg(any(test, feature = "test-support"))]` is safe here: `rocky-cli` takes plain `rocky-core` in `[dependencies]` and only enables `test-support` in `[dev-dependencies]`, and the workspace is on `resolver = "3"`, so the feature does not unify into a normal build. Worth stating explicitly since the method writes deliberately corrupt rows.

**The no-state path is unchanged.** The `if state_path.exists()` guard still short-circuits to empty, so a project with no state store is unaffected. Only present-but-unreadable now errors — which matches the PR description.

## One thing to note, not to fix here

Two other production `list_runs` readers still swallow the error:

- `rocky-server/src/dashboard.rs:120` — `.and_then(|store| store.list_runs(1).ok())`
- `rocky-cli/src/commands/backfill.rs:634` — `let Ok(runs) = … else { return out; }`

Both are lower-stakes than a custody chain (a UI panel and advisory cost history), and both are arguably tolerable, so I would not widen this PR. But they share the shape, and in each case an unreadable history is indistinguishable from an empty one. Worth a follow-up issue deciding whether they are deliberate tolerance or the same oversight.

For contrast, `backfill.rs:351` already does the right thing with `.context(…)?`, so this change moves the codebase toward the majority convention rather than away from it.

Nothing blocking.

Issue #1397, opened 2026-08-07
bug(dagster): _dag_payload lacks the is_dir guard its siblings have, so a missing models dir loads clean with zero assets


hugocorreia90:
Found in a wide sweep of gated issues (two audits reached it independently). **Structure verified from source; the end-to-end consequence is reported, not re-run by me.**

## The asymmetry

Two of the three payload builders guard on the models directory existing; the third does not.

```python
# component.py:949  _compile_payload
if not Path(self.models_dir).is_dir():
    return None

# component.py:983  _optimize_payload   ("mirrors _compile_payload")
if not Path(self.models_dir).is_dir():
    return None

# component.py:1006-1016  _dag_payload  — no guard
return json.loads(rocky.dag(column_lineage=True).model_dump_json(by_alias=True))
```

`_optimize_payload`'s own comment says it mirrors `_compile_payload`. `_dag_payload` was left out.

## Reported consequence

For a multi-transformation project with flat roots and no `models/` directory, `rocky dag --models models` exits **0 with zero nodes**. The code location then loads clean, produces **zero assets**, and records `dag_status: "success"`.

That is the failure mode #1380 was meant to close — but #1380 makes a *failed* `rocky dag` fail the load, and here `dag` does not fail. It succeeds and reports nothing. An empty asset graph is indistinguishable from a project that genuinely has no models.

I have verified the guard asymmetry directly. I have **not** executed the zero-node scenario, so treat the exit-0/zero-assets chain as needing one reproduction before fixing.

## The fix is NOT decision-free — this is why it is an issue and not a PR

Adding the missing `is_dir()` guard makes `_dag_payload` return `None`, which makes the `"dag"` key **absent** from the payload. Absent routes to the legacy discover flow — and that flow sends `--models <dir> --all` into the *same missing directory*. Nobody measured what happens next.

So the naive one-line fix may just move the silence rather than remove it. Sequence:

1. Reproduce the zero-node load and confirm `dag_status: "success"`.
2. Determine what the legacy discover flow does with a missing models dir.
3. Only then decide whether the guard returns `None`, or whether this case needs to fail loudly rather than fall back.

Related: #1380 (fail loud when `rocky dag` fails) established that a silent fallback here is the wrong default; this is the same class one level up, where the command *succeeds* emptily.

[...]

PR #1451, opened 2026-08-17
fix(sdk,dagster): resolve the models root per pipeline for discovery and execution

### LINKED ISSUE #1348: bug(dagster/sdk): DAG discovery forces a whole-project --models, so a multi-transformation project is refused before any asset exists

hugocorreia90:
Closes #1348. Relates to #1292 and #1397.

Option A from the disposition triage: a per-call models root, with discovery and execution moving together.

## The bug, reproduced

`RockyClient.dag()` always sent `--models <models_dir>`. The engine reads that as an explicit **whole-project override** and assigns that one directory's models to *every* transformation pipeline. Against a two-pipeline fixture on 1.70.1:

```
$ rocky dag --models models_a          # what the SDK sent
Error: failed to build unified DAG
Caused by:
    model 'one' is claimed by transformation pipelines alpha and beta.

$ rocky dag                            # omitting it
nodes: 2
  transformation:one | pipeline=alpha
  transformation:two | pipeline=beta
```

Since `RockyComponent` caches `rocky.dag(...)`, definitions failed to build at all — no asset existed to be mis-routed.

## Why execution had to move in the same change

#1348 says explicitly: *do not fix this half alone.* A node discovered in one root but built from another is silently different SQL for the same asset. So `run_model()` takes `pipeline=`, and when set it sends `--pipeline` and **not** `--models`:

```
$ rocky run --model one --pipeline alpha     # resolves ./models_a, executes
$ rocky run --model one --pipeline beta      # "model 'one' not found"  ← roots agree
```

That second line is the point: execution refuses to build `one` under `beta`, because it resolved only beta's root — the same root discovery attributed nodes by.

**`--models` was never the fix here.** On a multi-transformation project the engine refuses a bare `--model` whether or not `--models` is passed:

```
$ rocky run --model one --models models_a
Error: this project has multiple transformation pipelines, so a bare --model run
cannot know which pipeline (and adapter) the model belongs to; pass --pipeline <name>
```

## Deliberately gated, not defaulted

`run_model()` keeps `--models self.models_dir` when `pipeline` is `None`, so single-pipeline callers get byte-identical argv. I considered dropping `--models` unconditionally and rejected it: I had verified that only on a freshly-built fixture where the client field and the config agreed. The case that breaks is `RockyClient(models_dir="Y")` against a config declaring `X` — today execution uses `Y`, and an unconditional change would silently switch it to `X` on the *execution* path. The gated form cannot diverge, because whenever discovery omits the override, execution is pipeline-scoped too.

`node.pipeline` is `Optional` on the engine's node; `None` takes that same fallback rather than silently dropping `--models`.

## Acceptance

| #1348 criterion | How |
|---|---|
| Two transformation pipelines in separate roots produce an asset graph | `dag()` omits the override; verified against the fixture above |
| Discovery and execution resolve the same root, **asserted by a test** | `test_discovery_and_execution_resolve_the_same_root_for_a_discovered_node` drives execution from the node `dag()` returned, not a literal |
| A caller can still ask for a whole-project override | `dag(models_dir="one-root")` |

The acceptance test is deliberately not two independent argv assertions — those would both pass if one side started forcing the override again. It fails under **both** relevant mutations.

## Mutation-checked

| Mutation | Test | Result |
|---|---|---|
| stop threading `node.pipeline` | `test_transformation_execution_names_the_nodes_own_pipeline` | **FAILED** |
| revert `dag()` to forcing `--models` | consistency test | **FAILED** |
| send *both* `--pipeline` and `--models` | consistency test | **FAILED** |

Tree verified clean before and after each.

## Gates

| Gate | Result |
|---|---|
| `uv run pytest` (sdk) | 219 passed |
| `uv run pytest` (dagster) | 713 passed |
| `ruff check` / `ruff format --check` (both) | 0 |

No engine change, no schema change, no codegen cascade, no fixture movement.

## Behaviour change worth knowing

If you construct `RockyClient(models_dir=…)` with a value that disagrees with the root your `rocky.toml` pipelines declare, `dag()` now follows the config rather than the client field. That is the intended direction — the config is the authority — but it is a change, and it is in the changelog.

## Not in scope

#1397 (`_dag_payload`'s missing `is_dir` guard) is adjacent but separate; its own issue records that the proposed guard would break replication-only projects. #1292's engine half already landed in #1358; this supplies its SDK/Dagster half, so that issue's symptom should be retested before closing.

## Red team

Reviewed by **Codex**, `needs-attention`, with all three findings confirmed by an independent **`gpt-5.6-terra`** pass. All three were real and all three are fixed. Each was verified against the code before I accepted it.

**[high] The default change introduced the divergence it was meant to remove.** This is the important one, and it means my first version made things worse for one config. With `dag()` omitting `--models` but `run_model()` (no `pipeline`) still forcing `self.models_dir`, a project whose client field and config disagree would discover a model from one root and materialize it from another. **Before the change, both calls sent the same flag and agreed.**

My advisor had judged this residual acceptable, on the grounds it "exists today in the opposite direction". Reading the code, that is not so — today they agree, and my change is what split them. Codex was right, and its third suggested option dissolves the disagreement entirely: keep `dag()`'s default and make per-pipeline resolution **opt-in**. `models_dir` is now three-state — omitted is byte-identical to before, `None` opts in, a string forces the override — and the Dagster resource opts in explicitly, which is the caller that actually needs it. Nothing moves for existing callers, and the two halves can only move together.

**[high] Supported engines do not all implement the semantics assumed.** `MIN_ROCKY_VERSION` is `1.34.0`, but per-pipeline DAG resolution shipped in engine **1.68.0** (#1261) and `--model X --pipeline P` loading from P's root in **1.69.0** (#1292) — both confirmed from the changelog. An accepted 1.68 binary would discover a pipeline's configured root and then execute from `models/` relative to the process CWD, which can build a **same-named decoy**. Mocked argv tests are structurally blind to this. Each opt-in now gates on its own floor via `_require_engine`, following the existing fail-open policy when the version is undeterminable — and logging the skip, since a silently skipped gate is how an incompatible binary surfaces later as a confusing failure.

**[high] Dagster permitted an SDK that rejects both new keywords.** The floor was `rocky-sdk>=0.9.1` while the resource passes `models_dir=` and `pipeline=` unconditionally, so a legally resolved 0.11.0 raises `TypeError` during DAG discovery and per-model execution. The monorepo path source masks this in dev and CI, so the floor is the only protection for the published wheel. Now `>=0.12.0`, with the release ordering written next to it.

Codex refuted its own earlier concern about `node.pipeline` being Optional, and found no execution defect in the other whole-project methods.

## Release ordering (new, and load-bearing)

1. **rocky-sdk 0.12.0** — contains `dag(models_dir=)` and `run_model(pipeline=)`. Minor, not patch.
2. **dagster-rocky** — only after that is on PyPI, since its floor now resolves from PyPI, not the path source.

## Gates after the follow-up

| Gate | Result |
|---|---|
| `uv run pytest` (sdk) | 222 passed |
| `uv run pytest` (dagster) | 713 passed |
| `ruff check` / `format --check` (both) | 0 |

## Still open, and honestly stated

Codex's remaining point stands: the acceptance test proves argument forwarding, not end-to-end routing. I validated routing by hand against a two-pipeline fixture — including the cross case where `--pipeline beta` correctly refuses model `one` — but that is not automated. A real-binary test over a two-root project is the right follow-up and is not in this PR.

### DIFF
--- a/integrations/dagster/CHANGELOG.md
+++ b/integrations/dagster/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **A project with two or more transformation pipelines can now produce an asset graph.** `RockyComponent` caches `rocky.dag(...)`, and that call forced a whole-project `--models`, which the engine refuses on a multi-transformation project — so definitions failed to build at all. The DAG call now lets each pipeline resolve its own models root. (#1348)
+
+- **Transformation assets execute under their own pipeline.** `run_model` is called with `pipeline=node.pipeline`, so the engine resolves the pipeline the node was discovered in rather than a project-wide root. A node whose `pipeline` is unset falls back to the previous behaviour. (#1348, #1292)
+
 ## [1.62.0] — 2026-08-11
 
 ### Changed
--- a/integrations/dagster/pyproject.toml
+++ b/integrations/dagster/pyproject.toml
@@ -27,7 +27,14 @@ dependencies = [
     # freeze-marker and schedule/tick output types; an older SDK would be missing
     # them (and also the 0.8.3 `--models` forwarding + retention `env` guard fixes
     # the model-aware resource methods depend on).
-    "rocky-sdk>=0.9.1",
+    # 0.12.0 is the first release with `dag(models_dir=...)` and
+    # `run_model(pipeline=...)`. This resource passes both keywords
+    # unconditionally, so an older SDK — legal under the previous `>=0.9.1` —
+    # raises TypeError during DAG discovery and per-model execution. The
+    # monorepo path source masks that in dev and CI, so this floor is the only
+    # thing protecting the published wheel. Release rocky-sdk 0.12.0 BEFORE any
+    # dagster-rocky release carrying this floor. (#1348)
+    "rocky-sdk>=0.12.0",
     # Floor matches the version CI actually resolves and tests (uv.lock). The
     # RockyComponent path uses dagster.components.* APIs from preview namespaces,
     # so we advertise a tested floor rather than an older, never-exercised one.
--- a/integrations/dagster/src/dagster_rocky/dag_assets.py
+++ b/integrations/dagster/src/dagster_rocky/dag_assets.py
@@ -552,8 +552,13 @@ def _make_dag_group_asset(
 
             if node.kind == "transformation":
                 context.log.info(f"Executing rocky run --model {node.label}")
+                # Scope execution to the node's own pipeline so it resolves the
+                # same models root discovery used. `pipeline` is Optional on the
+                # engine's node, and None falls back to the previous argv rather
+                # than silently dropping `--models`.
                 result = rocky.run_model(
                     node.label,
+                    pipeline=node.pipeline,
                     **partition_kwargs,  # type: ignore[arg-type]
                 )
                 context.log.info(
--- a/integrations/dagster/src/dagster_rocky/resource.py
+++ b/integrations/dagster/src/dagster_rocky/resource.py
@@ -1239,15 +1239,25 @@ class RockyResource(dg.ConfigurableResource):
         with _translating():
             return self._get_client().catalog(out=out)
 
-    def dag(self, *, column_lineage: bool = False) -> DagResult:
-        """Run ``rocky dag`` and return the full unified DAG."""
+    def dag(
+        self, *, column_lineage: bool = False, models_dir: str | None = None
+    ) -> DagResult:
+        """Run ``rocky dag`` and return the full unified DAG.
+
+        ``models_dir`` defaults to ``None`` so each pipeline resolves its own
+        configured root; pass a string only to force a whole-project override
+        (see :meth:`RockyClient.dag`).
+        """
         with _translating():
-            return self._get_client().dag(column_lineage=column_lineage)
+            return self._get_client().dag(
+                column_lineage=column_lineage, models_dir=models_dir
+            )
 
     def run_model(
         self,
         model_name: str,
         *,
+        pipeline: str | None = None,
         filter: str | None = None,
         partition: str | None = None,
         partition_from: str | None = None,
@@ -1261,6 +1271,7 @@ class RockyResource(dg.ConfigurableResource):
         with _translating():
             return self._get_client().run_model(
                 model_name,
+                pipeline=pipeline,
                 filter=filter,
                 partition=partition,
                 partition_from=partition_from,
--- a/integrations/dagster/src/dagster_rocky/resource.py
+++ b/integrations/dagster/src/dagster_rocky/resource.py
@@ -1239,9 +1239,7 @@ class RockyResource(dg.ConfigurableResource):
         with _translating():
             return self._get_client().catalog(out=out)
 
-    def dag(
-        self, *, column_lineage: bool = False, models_dir: str | None = None
-    ) -> DagResult:
+    def dag(self, *, column_lineage: bool = False, models_dir: str | None = None) -> DagResult:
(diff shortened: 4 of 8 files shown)


hugocorreia90:
Merging with an admin override. Recording why, since `BLOCKED` should not be waved past casually.

`Protect main` requires seven checks, four of which are `engine-ci` jobs — `Adapter boundary`, `Check codegen is up to date`, `Clippy`, `Format`. This PR touches only `sdk/python/` and `integrations/dagster/`, so `engine-ci`'s path filter never fires and those four are **absent, not failing**. No push can make them run.

Everything that can run, did, and passed:

```
success  Credential containment policy
success  Credential-containment regression tests
success  Test          (sdk)
success  Test          (dagster)
success  Example smoke (real binary)
success  CodeQL, Analyze (rust/python/js/actions)
```

0 failures, 0 pending on `b017bbcc`. Plus locally: sdk 222 passed, dagster 713 passed, ruff check and format clean on both, and three mutation checks confirming the new tests fail when the fix is reverted.

The override is limited to this shape — a non-engine PR blocked by absent engine checks. An engine PR with a red or missing engine check does not get one.

PR #1463, opened 2026-08-17
chore: refresh the hatchling comments now that the ceiling is lifted

hugocorreia90:
#1452 and #1457 raised the hatchling bound to `<1.33`. Both left the comment above it untouched, so both files still said:

```
Lift `<1.32` once that action carries twine >= 7.0.0; see #1395.
```

That job is done — #1395 bumped the publish action to a build with twine 7.0.0, which is what let the bound move. So the comment now reads as a pending task for finished work. Dependabot cannot know that, and no gate checks comments.

The comments now say what happened, and why the bound is still a bound: hatchling 1.30.0 and 1.32.0 both shipped a metadata-version change. A ceiling turns the next one into a caught failure instead of a broken publish.

Checked:

| check | result |
|---|---|
| both `pyproject.toml` parse | yes |
| sdk rebuilds under the new bound | 222 tests pass |
| dagster rebuilds under the new bound | 713 tests pass |

Comments only. No version, dependency, or behaviour change.

### DIFF
--- a/integrations/dagster/pyproject.toml
+++ b/integrations/dagster/pyproject.toml
@@ -72,7 +72,10 @@ dagster_rocky = "dagster_rocky"
 # 1.32.0 (2026-08-11) emits `Metadata-Version: 2.5`, which the twine 6.1.0 in
 # our allow-listed `gh-action-pypi-publish` pin rejects at the publish step.
 # `dagster-release` uses the same action, so it would have failed identically.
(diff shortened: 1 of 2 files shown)


hugocorreia90:
Merging with an admin override. Non-engine PR (two `pyproject.toml` comment blocks), so the four `engine-ci` required checks never run — absent, not failing. Everything applicable passed.

PR #1475, opened 2026-08-18
fix(engine/rocky-cli): render sidecar [columns] descriptions in rocky docs

### LINKED ISSUE #1444: `rocky docs` silently discards `[columns]` descriptions
Found while verifying the starter examples by running them (PR #1442).

A model sidecar can carry a `[columns]` table with per-column descriptions. Rocky parses it. `rocky docs` never renders it.

`run_docs` (`engine/crates/rocky-cli/src/commands/docs.rs:46`) calls:

```rust
build_doc_index(&models, &rocky_cfg, None, Some(&column_docs))
```

The `column_map` argument is hardcoded `None` at the **only** production call site — every other caller of `build_doc_index` is `#[cfg(test)]`. The descriptions render only inside the `else` branch of `if model.columns.is_empty()`, which that `None` makes unreachable.

Verified by probe: a unique marker placed in a `[columns]` description appears **0 times** in the generated HTML. The command exits **0** with no warning, so nothing tells the author their descriptions went nowhere.

## Why it matters

This is the quiet-failure shape: the input is accepted, the command succeeds, and the work is discarded. Someone documenting a wide table can write dozens of descriptions and never learn they are inert.

## Options

1. Wire it — pass the real column map at the production call site.
2. Warn — if a model has `[columns]` descriptions and they cannot be rendered, say so on stderr.

PR #1442 documents the current behaviour so readers stop expecting the output, but that is a stopgap: the field is still parsed and still ignored.

hugocorreia90:
Closes #1444.

## The bug

`rocky docs` parsed `[columns]` descriptions, then rendered none of them. Exit 0, no warning.

```
sidecar [columns] --> parsed --> build_doc_index(column_map: None)
                                          |
                          descriptions attach to columns FROM THE MAP
                                          |
                                 no map -> no columns -> dropped
```

## The fix

The column map now comes from the offline compile step's type inference — the source `build_doc_index`'s own doc comment names. No warehouse connection: source schemas come from the TTL-filtered schema cache when `[cache.schemas]` is enabled; a cold cache renders leaf models as `UNKNOWN` rather than failing.

Three decisions, each pinned by a test:

1. **Docs never fail because compile failed.** Project load is all-or-nothing (same as `rocky compile`), so one broken model costs every model its column table. That degradation is now named on stderr, not silent.
2. **A description whose column the compile cannot see warns on stderr**, naming the model and the columns, and does not drag down matching descriptions.
3. **Unknown-typed columns render as the compile reports them** — docs stay consistent with `rocky compile` output.

## Evidence

- New `engine/rocky/tests/docs_column_descriptions.rs` (3 tests). Mutation-checked: reverting the wiring to `None` fails exactly the two wiring tests; the degrade test (which does not depend on the wiring) still passes.
- `cargo test -p rocky-cli --lib`: 1453 passed. Clippy (`--all-targets --all-features`) and `fmt --check` clean.
- No `output.rs` change — `DocsOutput` is untouched, so no codegen cascade.

- **Least confident:** the per-model all-or-nothing trade-off (decision 1). A WIP project with one broken model gets no column docs at all. Per-model tolerance needs a tolerant load in the compiler — out of scope here; the degradation is at least loud now.
- **Possibly missing:** checked `rocky catalog`, the other consumer of `load_column_docs_from_tree` — it already compiles and reads `typed_models` (`catalog.rs:186-208`), so it does not have this bug. No follow-up needed.


hugocorreia90:
## Red-team disposition

Independent reviewer: **Codex (GPT-5)** on the first cut (`ead5afa6..965d5a67`). Verdict: BLOCK, 4 blocking + 2 non-blocking findings. All confirmed by trace. Disposition, now in `dce2e974`:

| finding | disposition |
|---|---|
| 1. `.rocky` sidecar descriptions never scanned | **Fixed.** Loader scans `.rocky` companion sidecars too. Test: `dsl_model_column_descriptions_render_too` (mutation-checked). |
| 2. malformed `.rocky` bypasses the promised degrade | **Claim narrowed, boundary pinned.** Load stays strict, deliberately — a docs page silently missing a model would misrepresent the project. Only compile failures degrade. Test: `an_unparseable_dsl_model_fails_docs_at_load`. Changelog rewritten. |
| 3. missing `@var` publishes sentinel-derived metadata | **Fixed.** `Ok(has_errors=true)` now degrades loudly like `Err`, and `rocky docs` gains repeatable `--var NAME=VALUE`. Test: `a_missing_run_var_degrades_and_passing_it_renders` (mutation-checked). |
| 4. `--cache-ttl` ignored | **Fixed.** Threads `cli.cache_ttl` through `with_ttl_override`, same as `rocky catalog`. No integration test — TTL semantics need a warm-cache fixture; the wiring matches the catalog precedent line for line. |
| 5. exact-case matching (non-blocking) | **Fixed anyway.** Attach + orphan check fold ASCII case, matching `column_map`'s convention. Case-colliding sidecar keys resolve deterministically (sorted, later wins). Test: `description_matching_folds_ascii_case` (mutation-checked). |
| 6. docs vs catalog column inventories differ (non-blocking, pre-existing) | **Not fixed here** — predates this PR. Tracked as a separate issue. |

Also taken: the review's compile-cost note — docs now compiles the already-loaded models (`compile_preloaded_models`), removing the double directory read.

7 integration tests total; the three new guards each mutation-checked (each mutation kills exactly its test). `rocky-core` docs (21) and models (86) unit tests pass. Clippy `--all-targets --all-features` and `fmt --check` clean.

hugocorreia90:
## Fix-delta red-team disposition

Second independent pass (Codex, **GPT-5.6-terra**) over the blocker-fix delta `dce2e974`. Verdict: BLOCK, one finding — accepted and fixed in the follow-up commit.

**Finding:** `compile_preloaded_models` has no `NoModels` rejection, so the optimization reintroduced silent success for a missing/empty models path — a typo'd `--models` rendered a blank catalog at exit 0.

**Fix:** `rocky docs` now refuses zero models outright with the sibling commands' wording (`no models found in <dir>`), before any HTML is written. Test: `docs_refuse_a_missing_or_empty_models_dir` (mutation-checked; also asserts no catalog file is left behind). This is the explicit zero-model contract the review asked for.

The pass's other four probes came back clean on the record: same-stem `.sql`/`.rocky` sidecar reads are idempotent and duplicate names are rejected upstream; the case-fold collision rule is locale-independent and deterministic; `has_errors` is `Severity::Error`-only, so warning-only projects keep their column tables.

Issue #1341, opened 2026-08-01
bug(dagster): a failing `rocky dag` silently falls back to the legacy graph instead of failing the code location


hugocorreia90:
## Description

In `dag_mode`, `_dag_payload` catches **every** exception from `rocky dag` and omits the DAG slot:

```python
# integrations/dagster/src/dagster_rocky/component.py
except Exception:
    ...  # warn, omit the slot
```

`_build_defs_from_dag` then interprets a missing slot as an *old cache* and falls back to the legacy discover-based graph.

So a project whose DAG genuinely cannot be built loads a **different, disconnected graph** with a log warning, instead of failing the code location. The two situations — "this payload predates DAG support" and "this project's DAG is invalid" — are not distinguishable at that point, and only one of them should be recoverable.

Even `strict_build=True` can fall through when discovery has sources.

## Why this surfaced now

#1340 makes two transformation pipelines writing one physical table a hard error at unified-DAG build. That is the intended behaviour — the DAG's claim that those nodes are independent is false — but it means `rocky dag` now exits non-zero for such a project, and Dagster answers by quietly loading a graph that does not reflect the project.

The failure is pre-existing and not specific to that error: any `rocky dag` failure takes this path.

## Why it matters

The fallback is not a degraded version of the real graph — it is built from discovery, so it has different assets and different dependencies. A user gets a code location that loads, looks plausible, and materializes a different plan than their project describes. A warning in the logs is not proportionate to that.

## Suggested direction

Distinguish the two cases rather than widening or narrowing the catch:

1. **A payload that predates DAG support** — no DAG slot was ever produced — is legitimately a fallback case.
2. **A `rocky dag` invocation that failed** is not. It should propagate, at minimum under `strict_build=True`, and arguably always in `dag_mode=True` since the caller explicitly asked for the DAG.

That likely means `_dag_payload` recording *why* the slot is missing rather than returning the same "absent" for both, and `_build_defs_from_dag` branching on it.

## Acceptance

- A project whose `rocky dag` fails does not silently load the legacy graph in `dag_mode=True`.
- A payload genuinely predating DAG support still falls back, with a test distinguishing the two.
- `strict_build=True` cannot fall through on a `rocky dag` failure, discovery sources or not.

## Provenance

Found by independent red-team review (Codex/GPT-5) of #1340, while checking that PR's blast radius on `rocky dag` consumers. Filed separately because the defect is in the Dagster integration's error handling, not in the engine change that exposed it.

PR #1476, opened 2026-08-18
fix(engine/rocky-cli): refuse a zero-node DAG when a transformation pipeline is declared

### LINKED ISSUE #1397: bug(dagster): _dag_payload lacks the is_dir guard its siblings have, so a missing models dir loads clean with zero assets

hugocorreia90:
Closes #1397.

## The bug

```
[pipeline.t]                 rocky dag        dagster component
models = "models/**"   -->   exit 0     -->   dag_status: success
(directory missing)          0 nodes          0 assets, load clean
```

An empty asset graph was indistinguishable from a project that genuinely has no models. #1380 made a *failed* `rocky dag` fail the load — but this dag did not fail. It succeeded and reported nothing.

## The fix — one guard, at the output level

`rocky dag` refuses when the final graph has zero nodes AND the config declares a transformation pipeline. Keying on the final node set (not on any one root) is what keeps every deliberate tolerance intact:

| shape | before | after |
|---|---|---|
| models root missing, nothing else | exit 0, 0 nodes | **refuses** |
| `--models` override pointing nowhere | exit 0, 0 nodes | **refuses** |
| seed-only pipeline (no models key, seeds/) | works | works |
| replication-only, ± `--models` (SDK default) | works | works |
| empty sibling root next to a populated one | works | works |

The last row is the strictness this area already litigated and backed out (`dag_column_lineage_ignores_roots_that_contribute_nothing` pins it) — an earlier draft of this fix refused per-root and broke three existing tests plus the seed-only shape; the output-level guard was chosen because it breaks none of them.

Dagster needs **no change**: with the engine refusing, `_dag_payload` raises and the existing #1341 machinery records `dag_status: "failed"` (or fails the deploy under `strict_build`). No fallback into the legacy discover flow, whose missing-dir behaviour was never measured. `rocky serve`'s `GET /api/v1/dag` shares the core and inherits the refusal.

## Evidence

- New `engine/rocky/tests/dag_missing_models_refusal.rs` (5 tests: 2 refusals, 3 tolerance pins). Mutation-checked: reverting the guard fails exactly the two refusal tests.
- `cargo test -p rocky-cli --lib`: 1453 passed — including the three tests the per-root draft broke.
- Clippy `--all-targets --all-features`: clean. `fmt --check`: clean.
- Swept all 123 tracked `rocky.toml` files: no in-repo project would now refuse (one apparent hit is the `include_str!` scaffold template for `rocky playground`, never dagged in place).
- No `output.rs` change — no codegen cascade.

- **Least confident:** an explicit `models = "typo/**"` root *alongside* seeds or a populated sibling still silently contributes nothing — the guard only fires when the whole graph is empty. Distinguishing an explicit-but-wrong glob from the serde default needs config provenance (`models` is `String` with a default), which is a schema change out of scope here.
- **Possibly missing:** `rocky run --dag` on the same empty-graph config still executes nothing and reports success — same family, different command. Not covered by the #1397 report; worth its own decision rather than a silent scope widening here.

The adjacent #1397 note (`rocky dag` runs neither inference pass, so the exported graph can differ from `run --dag`'s) is untouched — tracked on the issue.


hugocorreia90:
## Red-team disposition

Independent reviewers: **Codex (GPT-5)** with a second independent confirmation by **gpt-5.6-terra**. Verdict on the first cut: one HIGH finding, accepted and fixed in `732103cf`.

**Finding:** the zero-node guard also refused an EXISTING-but-empty models directory with no seeds — a supported no-op (fresh scaffold, intentionally empty pipeline), contradicting the runtime's own empty-transformation semantics.

**Fix:** the refusal now fires only when the graph is empty AND a declared transformation root **does not exist**, naming the pipeline and the path. Absence is recorded in the SAME resolution pass that loads the other pipelines (`load_transformation_models` gains `missing_roots`) — not a second resolver walk that could disagree with the first (the reviewer's TOCTOU probe, which came back clean, stays clean by construction). New test pins the no-op: `an_existing_empty_root_is_a_no_op_not_a_refusal`.

The pass's other probes, on the record:
- Consumers (serve → 500, SDK → `RockyCommandError`, dagster → `dag_status: "failed"` / strict_build deploy failure, VS Code → stderr) propagate a REAL refusal correctly; the false positive they'd have amplified is the thing this commit removes.
- Config TOCTOU: one `cfg` load drives loading, graph build, and the guard. Confirmed clean.
- The bail reaches stderr and a nonzero exit in both output modes. Confirmed clean.

Mutation-checked: disabling the refined guard fails exactly the two refusal tests; the four tolerance pins (empty root, seed-only, replication ±override, healthy) pass either way.

Issue #1445, opened 2026-08-16
Four starter examples do not work as shipped


hugocorreia90:
Found by running every example while rewriting their READMEs (PR #1442). These are the examples a new user meets first, so a failure here reads as a broken product.

| Example | What happens | Cause |
|---|---|---|
| `seed-demo` | all three documented commands exit 1 | the pipeline targets a `warehouse` catalog the in-memory DuckDB does not have |
| `compare-demo` | `rocky compare` reports nothing about the two models the README introduces | replication pipelines take compare targets from source discovery, so models never enter the comparison — `--models` does not change this |
| `watch-demo` | `rocky run --watch` never reaches a successful pass | same missing `warehouse` catalog, then a missing `raw_orders` table |
| `docs-demo` | `[columns]` descriptions never appear | tracked separately as #1444 |

PR #1442 makes each README state plainly what the reader will see and what they must change, which stops the surprise. It does not make the examples work — that is a change to the example fixtures, not the docs.

## Suggested fix

Give each affected example a catalog and tables that exist, so `./run.sh` succeeds end to end on a clean clone with no credentials. `quickstart`, `multi-layer`, `window-functions` and `test-declarative` already behave this way, so the pattern is established in-repo.

For `compare-demo` specifically the fix is larger: either ship a replication source whose discovery produces a real pair, or convert the example to the pipeline type `rocky compare` actually serves.

Related: #1443 (importer prints an invalid command), #1444 (`rocky docs` discards `[columns]`).

Issue #1435, opened 2026-08-15
Two rocky.toml sections parse but nothing consumes them: [schema_evolution] drop/grace and project [freshness] inheritance


hugocorreia90:
Found by an independent review of the documentation sweep (PR #1434). Both are config surfaces a user can set, and set successfully, with no runtime effect. Because every config struct uses `deny_unknown_fields`, these keys look validated — they load without complaint, which is exactly what makes the silence dangerous.

## 1. `[schema_evolution]` drop and grace never run

The field parses (`rocky-core/src/config.rs:2502`), but the live drift function detects only column **additions** and **type changes**, and always returns empty drop and grace collections (`rocky-core/src/drift.rs:28`, `:68`).

`detect_column_drops` exists at `rocky-core/src/drift.rs:253`, but a repo-wide search finds only **test** callers. So a user who configures a grace period for disappearing columns gets no warning, no grace window, and no eventual drop.

## 2. Project-level `[freshness]` does not inherit field by field

The conversion helper at `rocky-core/src/models.rs:327` has **no caller**.

What actually happens today:
- compile reduces the project block to a single boolean that suppresses `W005` (`rocky-cli/src/commands/compile.rs:121`, `rocky-compiler/src/typecheck.rs:1043`)
- scheduling reads only `expected_lag_seconds` (`rocky-core/src/schedule/reconcile.rs:785`)

So a project-level `time_column` or `severity` is accepted and then ignored. A model that omits those fields does not inherit them.

## Why it matters

This is the fail-open shape that keeps recurring: the configuration is accepted, so the operator believes a control is active. A grace period that silently does nothing is worse than no grace period, because it suppresses the operator's own vigilance.

## Options

1. **Wire them.** Call `detect_column_drops` from the live drift path and honour the grace window; call the freshness conversion helper so per-model fields fall back to the project block.
2. **Reject them.** Make the unwired keys a load-time error or a startup warning, so an operator learns immediately instead of after an incident.

PR #1434 takes the third, documentation-only path in the meantime: the reference page now states plainly that the drop/grace path is not active and that project `time_column` and `severity` are not inherited. That is a stopgap, not a fix — the config still accepts input it ignores.

hugocorreia90:
Verified both claims against 1.70.1. **Both hold**, and the first has a detail worth adding before the wire-vs-reject call is made.

## 1. `detect_column_drops` — confirmed unwired, and heavily tested

Repo-wide, every call site is a test:

[...]

PR #1479, opened 2026-08-18
fix(engine/examples): make seed-demo and watch-demo work as shipped

hugocorreia90:
Partial fix for #1445 (2 of 3 remaining examples; do not auto-close).

## One cause, two examples

```
[adapter.local] type=duckdb, no path -> in-memory, catalog = memory
targets name catalog 'warehouse'    -> Catalog does not exist, exit 1
```

Fix: `path = "warehouse.duckdb"` — a DuckDB file's catalog is its file stem, so `warehouse` exists on first run. `*.duckdb` is already gitignored.

## Verified from a clean clone

| command | before | after |
|---|---|---|
| seed-demo: all 3 documented `rocky seed` forms | exit 1 | exit 0, 2 seeds load |
| watch-demo: `rocky seed` + `rocky run --models` | exit 1 | exit 0 |
| watch-demo: `rocky run --watch` (live, SIGINT) | every pass failed | `run completed` then `stopped` |

watch-demo gains `seeds/raw_orders.csv`; its sidecar targets schema `main`, where the model's bare `FROM raw_orders` resolves, and the model target moves `analytics` → `main` — the pattern the playground's runnable POC models use. READMEs rewritten from failure-walkthrough (the #1442 stopgap) to the working flow.

## Not in this PR

- **compare-demo** — needs a pipeline-type decision (replication discovery pair vs converting to the type `rocky compare` serves), stays open on #1445.
- **Observed, noted on #1435:** `[target.governance] auto_create_schemas = true` did not create the model's target schema on the `run --models` path (probe: run failed with `Schema analytics does not exist` with the flag set; retargeting to `main` was the fix). Same accepted-but-inert family #1435 tracks.

- **Least confident:** `schema = "main"` sidesteps rather than answers the schema question — if the governance gap above gets wired someday, examples could go back to a named schema. The playground precedent says `main` is today's honest pattern.
- **Possibly missing:** the parse-validate POC harness does not cover `engine/examples/` (it covers `examples/playground/`), so nothing in CI runs these examples end to end; they can regress again silently. Worth a follow-up smoke check if #1445 recurs.


hugocorreia90:
## Red-team disposition

Independent reviewer: **GPT-5 Codex** (live-traced against the built binary where the read-only environment allowed). Two findings, both confirmed against the file and fixed in the follow-up commit:

1. **File tree misplaced the model sidecar** — my tree insertion landed above the pre-existing `orders_summary.toml` line, showing it under `seeds/`, and omitted the real seed sidecar. Tree corrected: `models/` holds the model pair, `seeds/` holds `raw_orders.csv` + `raw_orders.toml`.
2. **"never opens rocky.toml" was false** (pre-existing, in-scope since this PR rewrites the file) — the CLI reads it at startup to namespace the state store. Rephrased to what is true: `rocky watch` never uses the adapter and never connects to a warehouse.

Also on the record from the review: every generated artifact (`warehouse.duckdb`, `.rocky/`, state files) confirmed gitignored by source-trace; ModelIr/contract semantics untouched; and the reviewer echoes this PR's own gap note — no CI smoke test executes `engine/examples/` README flows, so drift here is not pinned. That stays tracked on #1445.

PR #1480, opened 2026-08-18
fix(engine/examples): rebuild compare-demo on the pipeline type rocky compare serves

### LINKED ISSUE #1445: Four starter examples do not work as shipped

hugocorreia90:
Closes the remaining item on #1445 (after #1479 fixed seed-demo and watch-demo). Closes #1445.

## Why a rebuild, not a patch

`rocky compare` pairs tables from **replication source discovery** — `compare.rs` bails outright without a discovery adapter. The example shipped transformation models that compare never reads, so it reported nothing about the models its README introduced. The issue offered two fixes; this takes "convert the example to the pipeline type `rocky compare` actually serves."

## The shape

Bootstraps entirely with rocky itself — no duckdb CLI, no credentials:

```
rocky seed      seeds/orders.csv -> raw__shopify.orders   (sidecar targets the schema)
rocky run       raw__shopify -> staging__shopify.orders   (discovery + schema_pattern)
rocky run --shadow --shadow-suffix _shadow                (orders_shadow)
rocky compare --shadow-suffix _shadow                     (1 pair, pass)
```

## Every README transcript is from a live run

| claim | verified |
|---|---|
| documented flow | exit 0 at each step; compare: 1 compared, 1 passed, 6=6 |
| drift arc (append source row, shadow-only re-run) | exit 1, prod 6 vs shadow 7, verdict fail |
| recovery (re-run production) | compare passes again, exit 0 |
| `--filter source=shopify` | exit 0 |
| `--output table` (global flag, before the subcommand) | prints the summary |

The transformation models are deleted — compare never served them, and the example now says plainly it is a replication example.

- **Least confident:** deleting the shipped models is a content decision; if someone wanted a models-AND-compare example, that is a different example. The issue's own framing blessed the conversion.
- **Possibly missing:** same gap #1479 recorded — no CI executes `engine/examples/` README flows end to end, so these can regress silently. Tracked as the follow-up note on #1445's closure.


hugocorreia90:
## Red-team disposition

Independent reviewer: **OpenAI GPT-5 Codex**. Verdict: BLOCK, one finding — accepted and fixed in the follow-up commit.

**Finding:** the README's compare transcripts were not what the documented commands print. At a terminal, plain `rocky compare` renders the TABLE view (TTY detection); my JSON block came from a piped run and was trimmed besides. Both blocks are now verbatim captures: the table view for the plain command (`[  OK] … (prod=6, shadow=6, diff=0.00%)`), the complete JSON under an explicit `rocky --output json compare …`, and the drift arc's real FAIL table (`prod=6, shadow=7, diff=16.67%`). The flags note now states the TTY/pipe default split.

On the record from the review, all clear: seed creates `raw__shopify` via `CREATE SCHEMA IF NOT EXISTS` and re-runs drop/recreate; no dangling references to the deleted models (the examples index line still describes the rebuilt example accurately); the drift append/restore instructions are correctly ordered; every generated artifact is gitignored (confirmed via `git status --ignored`).

PR #1488, opened 2026-08-19
fix(engine): record every partition in a batch

### LINKED ISSUE #1486: bug(engine): batched time-interval runs record only the leading partition as Computed
## Description

A successful `time_interval` execution with `batch_size > 1` writes the full
batched window to the warehouse, but records only the batch's leading
partition as `Computed` in the `PARTITIONS` state table.

The remaining keys are therefore reported as gaps by a later `--missing` run
and are needlessly recomputed, even though the original materialization
already covered them.

This contradicts the `PartitionPlan::batch_with` contract, which says the
additional keys are retained so state can mark every partition in the batch
as `Computed`.

## Version and environment

- Rocky: `main@9aa24ff47dd54730ddfa921ea4b77e8c4ced7a1c`
- Also reproduced against the code shipped in `engine-v1.71.0`
- Platform: macOS 26.6.1, ARM64
- Adapter: DuckDB

## Minimal reproduction

Configure a daily `time_interval` model with:

```toml
[materialization]
strategy = "time_interval"
time_column = "order_date"
granularity = "day"
batch_size = 3
first_partition = "2026-08-17"
```

Run one three-day batch:

```bash
rocky -c rocky.toml -o json run \
  --filter source=orders \
  --models models \
  --from 2026-08-17 \
  --to 2026-08-19
```

The successful materialization reports:

```text
key:          2026-08-17
window:       2026-08-17 through 2026-08-20
batched_with: 2026-08-18, 2026-08-19
status:       Success
```

Then run:

```bash
rocky -c rocky.toml -o json run \
  --filter source=orders \
  --models models \
  --missing
```

## Expected behavior

The first `--missing` invocation plans zero partitions because all three
logical partitions were committed by the successful batch.

## Actual behavior

1. The first `--missing` invocation recomputes `2026-08-18`, batched with
   `2026-08-19`.
2. A second identical invocation recomputes `2026-08-19` again.
3. Only the third invocation plans zero partitions.

## Impact

- The per-partition state ledger disagrees with warehouse reality.
- Recovery and audit output temporarily report completed partitions as gaps.
- `--missing` performs redundant warehouse reads and writes.
- A backfill can require up to `batch_size` invocations to converge.

This does not appear to cause data loss: `time_interval` replacement remains
idempotent. The defect is incorrect durable state and unnecessary warehouse
work/cost.

## Root cause

`run_one_partition` creates one `PartitionRecord` for
`partition_plan.partition_key` and persists that record as `InProgress`,
`Failed`, or `Computed`. It never records the keys in
`partition_plan.batch_with`; those keys are used only in output metadata.

`--missing` later trusts `PARTITIONS` as its source of truth and subtracts only
keys with a `Computed` row.

Relevant code:

- `engine/crates/rocky-core/src/plan_partition.rs`: `PartitionPlan::batch_with`
- `engine/crates/rocky-cli/src/commands/run.rs`: `run_one_partition`
- `engine/crates/rocky-core/src/state.rs`: `StateStore::record_partition`

## Proposed fix

After a batch commits successfully, persist `Computed` records for the leading
key and every key in `batch_with`, ideally in one redb transaction so the
logical batch cannot be only partly reflected in state.

## Acceptance criteria

- A successful three-partition batch creates three `Computed` partition rows.
- A subsequent `--missing` selection returns no keys for that batch.
- The default `batch_size = 1` behavior is unchanged.
- The state update for a successful batch is atomic.
- Targeted tests plus the engine's nextest, clippy, and formatting gates pass.

## Duplicate search

Searched open, closed, and merged issues/PRs for `batch_with`, `batched_with`,
`batch_size`, `time_interval`, `--missing`, `PARTITIONS`, `record_partition`,
and combinations of partition/state/computed/batch terms. The then-open #1483
(now merged) touched both state and run execution, but changed product-state
tables and typed run termination rather than partition lifecycle recording.

mattfaltyn:
## Summary

Persist every logical partition covered by a successful `time_interval` batch
as `Computed`.

Previously, a batch wrote its full window to the warehouse but stored only the
leading key in `PARTITIONS`. A later `--missing` run therefore treated the
remaining keys as gaps and recomputed already-committed data.

Closes #1486.

## Subproject(s) touched

## Changes

- Add `StateStore::record_partitions` to write multiple partition records in
  one redb transaction through the existing write-epoch path.
- Keep `record_partition` as the single-record wrapper, preserving existing
  callers.
- On successful batched execution, record the leading key and every
  `batch_with` key atomically as `Computed`.
- Add state-store and DuckDB executor regressions covering atomic persistence
  and subsequent `--missing` selection.

## Test Plan

- `cargo test -p rocky-core --lib test_record_partitions_commits_batch_atomically`
- `cargo test -p rocky-cli --lib batched_time_interval_records_every_partition_as_computed --features duckdb`
- `cargo nextest run --all-features`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo fmt -- --check`
- `bash ../scripts/lint-adapter-boundary.sh`
- `cargo build --release --bin rocky` plus the CI-equivalent tiered-cache
  `rocky doctor --check auth` smoke
- Code generation and fixture regeneration checked for drift.

## Checklist

- [x] Commit messages follow conventional commits
- [x] No CLI JSON schema or DSL syntax change
- [x] Targeted regression tests pass
- [x] Full all-feature nextest passes (6,344 passed; 112 skipped)
- [x] Full all-target/all-feature clippy is clean

## Behavioral boundary

This changes only durable partition lifecycle state after a successful
warehouse commit. It does not change planning, SQL generation, `ModelIr`,
warehouse transaction behavior, CLI JSON shape, or the default
`batch_size = 1` path.

## Standing questions

**What am I least confident about?** Whether any external consumer relied on
the incomplete ledger and reconstructed batched completion from JSON output.
No in-repository consumer does, and the documented contract makes
`PARTITIONS` authoritative per logical partition.

**What's the most important thing I may be missing?** A maintainer preference
to record every batched key as `InProgress` and `Failed` as well. This PR stays
on the demonstrated bug boundary: all keys are atomically recorded only after
the warehouse batch commits successfully. Missing selection remains correct
on failure because every non-computed key remains eligible.


hugocorreia90:
Thank you for the fix. It is correct. We will merge it.

**What I checked**

- I removed your `run.rs` change and kept your new test. The test failed. So the test really catches the bug.
- I added a check on the warehouse table itself. All three days were there. So the state records match the real data.
- Both new tests pass on my machine.

**One follow-up — filed as #1496. No change needed in this PR.**

On success, the code now records every partition in the batch. Good.
On failure, the code still records only the first partition.

This is safe on Databricks, BigQuery, Snowflake, and DuckDB. Their overwrite is all-or-nothing.
Trino is different. It runs `DELETE` and `INSERT` as two separate steps. If the `INSERT` fails after the `DELETE`, the data for the other partitions is gone. But their records still say `Computed`, so `--missing` will not rebuild them.

The fix is small: on failure, write a `Failed` record for every key in the batch, with the `record_partitions` you added. Details are in #1496.

**About the red checks**

They are not caused by your change. A new Rust release added a stricter lint, and one test broke on `main` before your branch was updated. We will fix `main`, update this branch, and merge.

hugocorreia90:
Merged as 27d9471d. Thank you for the clear report and the clean fix. The follow-up for the failure path is tracked in #1496.

PR #1489, opened 2026-08-19
fix(engine/rocky-compiler): match DECIMAL precision and scale in contract type checking

### LINKED ISSUE #1466: contracts: DECIMAL precision/scale ignored by contract type matching
`type_name_matches` in `engine/crates/rocky-compiler/src/contracts.rs` (~line 221) matches a model's inferred type against the contract's type string with:

```rust
RockyType::Decimal { .. } => type_name == "Decimal" || type_name.starts_with("Decimal("),
```

The pattern discards the model's actual precision/scale and accepts any `Decimal(...)` string. A contract that pins `type = "Decimal(18,2)"` passes compile against a model column that infers to `Decimal(38,10)` — or any other precision/scale. The contract reads as a digit-level guarantee but does not enforce the digits.

Expected: when the contract string carries `(p,s)`, parse it and compare against the inferred type's precision/scale; a mismatch should surface as the same diagnostic class as other contract type mismatches. A bare `Decimal` (no parentheses) can keep matching any precision/scale for backward compatibility.

Observed while auditing the contract-matching arms. Related but out of scope here: `RockyType::Unknown => true` in the same function ("Unknown matches anything") is a broader fail-open posture question that deserves its own look; this issue is only about the Decimal arm.

hugocorreia90:
Closes #1466.

A `.contract.toml` pinning `type = "Decimal(18,2)"` passed against a column inferred as `Decimal(10,0)`. The match tested the prefix only, so every parameter block was accepted:

```
before   type_name == "Decimal" || type_name.starts_with("Decimal(")
after    parse (p,s) and compare both
```

## The widening decision, stated

The match is **exact**, not "the inferred type fits inside the declared one". Three reasons:

1. A contract states the model's **declared output type**. This same matcher already rejects an inferred `Int32` against a contract saying `Int64` — a safe widening. Treating decimals differently would make the rule inconsistent.
2. `drift.rs::is_safe_type_widening` answers a different question: whether a **live warehouse column** can be altered in place. It is a `SqlDialect` method, and the default, Databricks and Trino allowlists all differ — so a compile-time diagnostic cannot inherit it without becoming dialect-dependent.
3. Every decimal rule in those allowlists requires **equal scale** anyway, so this issue's case (`Decimal(18,2)` over `Decimal(10,0)`) is a mismatch under them too.

Backward compatibility is kept where it costs nothing: bare `Decimal` still matches any precision and scale, so contracts written before the digits were checked keep passing. `Decimal(p)` means scale 0, matching the type checker's reading of SQL's `DECIMAL(p)`. An unparseable parameter block never matches, so a malformed contract cannot pass on the prefix alone.

## Evidence

Mutation check — revert to the prefix-only match, keep the tests:

```
test_decimal_precision_widening_is_e011   FAILED
test_decimal_scale_mismatch_is_e011       FAILED
test_bare_decimal_contract_matches_any_precision   ok   <- correctly insensitive
test_decimal_exact_match_passes                    ok   <- correctly insensitive
```

`cargo test -p rocky-compiler --lib`: 438 passed. Clippy `--all-targets --all-features` and `fmt --check`: clean. No `output.rs`/`config.rs` touch, so no codegen cascade.

## Provenance

Written by an agent whose run was interrupted; recovered from its worktree, then re-verified end to end here — full suite, both clippy and fmt gates, and the mutation check above run against this branch's committed state.

- **Least confident:** the exactness choice is a real behaviour change. A project whose contract says `Decimal(18,2)` over an inferred `Decimal(10,0)` starts failing E011 where it passed. That is the point of the issue, but it will surface as new red builds on upgrade.
- **Not swept:** I did not grep every caller of the contract gate for a tolerant consumer that downgrades E011 to a warning. If one exists, this fix is weaker than it looks there.


hugocorreia90:
Closing the gap I flagged in the PR body — I swept the consumers rather than leaving it as a caveat.

**No tolerant consumer.** E011 is emitted via `Diagnostic::error` (contracts.rs:146), so it sets `Severity::Error`; `has_errors` is computed from `is_error()` and `rocky compile` bails on it (compile.rs:62). Nothing downgrades it to a warning.

**One real weakening, already tracked.** The emission site is guarded:

```rust
if !matches && col.data_type != RockyType::Unknown {
```

So a column whose type infers as `Unknown` — cold schema cache, and every contract-typed column under `rocky test` / `rocky ci` — skips the check entirely, this fix included. That is exactly **#1240**, which is open and decision-blocked on whether an unresolved type at the E011 gate should stay silent, warn, or error.

Worth stating plainly: this PR makes the check correct **when the type is known**. It does not widen when the check runs. Anyone reading #1466 as "decimal contracts are now enforced" should read #1240 next.

Issue #1405, opened 2026-08-08
bug(engine): rocky run --watch intermittently does not exit within 30s of SIGINT — reproduced on main, and one flake aborts the rest of the suite


hugocorreia90:
## Description

`rocky run --watch` intermittently does not exit within 30 seconds of `SIGINT`, on a machine that is demonstrably not loaded. It has now failed twice in three days, on two different branches, and because `Test` is a required check with no retry policy it takes the whole gate down with it — and aborts the remaining ~4900 tests.

**This is not the #1315 load artifact, and the test says so itself.** #1315 replaced a flat 30 s deadline with `scaled_budget(one_run, 6)` precisely so a busy machine and a genuine shutdown hang would stop producing the same message. The diagnostic it added is now doing its job and pointing at the second:

| | run | branch | `one_run` calibration | budget | outcome |
|---|---|---|---|---|---|
| 1 | [31217289297](https://github.com/rocky-data/rocky/actions/runs/31217289297) | `main` | **518.875 ms** | 30 s | still running after 30 s |
| 2 | [31284672523](https://github.com/rocky-data/rocky/actions/runs/31284672523) | `feat/1281-live-identifier-case` | **506.087 ms** | 30 s | still running after 30 s |

`MIN_BUDGET` is 30 s (`run_watch.rs:42`) and `scaled_budget` clamps (`run_watch.rs:55`), so `6 × ~510 ms = ~3 s` floors to 30 s. The process was given **~60× a full run** to exit and did not. A calibration of ~510 ms in both instances is the direct evidence the runner was healthy: #1315's reported artifact had `one_run` inflated 5.6× (5.8 s isolated → 32.8 s in-suite), and neither of these is inflated at all.

The instance on `main` matters most: that run carries no PR changes, so nothing about either branch is implicated.

## Steps to reproduce

Not reproduced locally. Both observations are from `engine-ci.yml` / `Test` on `ubuntu-latest`, failing at `engine/rocky/tests/run_watch.rs:344`:

```
thread 'run_watch_reruns_on_file_change_and_exits_clean_on_sigint' panicked at rocky/tests/run_watch.rs:344:13:
rocky was still running 30s after SIGINT. One run measured 506.086542ms on this machine and
the budget is 6x that, so if `one_run` is far above a few seconds this machine is loaded
rather than rocky being stuck.
```

[...]

PR #1490, opened 2026-08-19
fix(engine/rocky-cli): keep one signal registration alive across watch iterations

hugocorreia90:
Refs #1405 — **does not close it**, see the note at the end.

## Two bugs, one line apart

Both watch loops built a fresh `tokio::signal::ctrl_c()` future on **every iteration**:

```
select! { ctrl_c() | rx.recv() }
      |
 rx.recv() wins  ->  the ctrl_c future is DROPPED
      |
      v
 [ debounce 200ms -> config load -> run start ]
 <- no listener registered in here ->
      |
      v
 the inner run registers its own handler
```

tokio discards a signal that arrives while zero listeners are registered — and because tokio had already replaced the **default disposition**, that Ctrl-C was both swallowed *and* non-fatal. The watcher kept running; a second Ctrl-C was needed. That is the intermittent hang in the issue, and the flake that aborted the rest of the engine suite.

The registration is now built once, before the first iteration, and never dropped. `pin!` alone would not do it — `pin!` does not poll, so nothing registers until the first poll.

**Second bug, fully deterministic:** neither loop had a SIGTERM arm at all. After the first run, the watcher could not be stopped by `timeout`, a CI job cancellation, or a container eviction — only SIGKILL. Both loops now handle either signal, matching the pattern `run.rs` already used.

Fixed in `run_watch.rs` (`rocky run --watch`) and `watch.rs` (`rocky watch`).

## The mutation check caught my own test being worthless

First version signalled as soon as the watcher started. It passed with the SIGTERM arm deleted — because **before** the first run, SIGTERM still has its default disposition and kills the process regardless of what the watch loop does.

The bug only exists *after* a run has installed and dropped its own handler. The test now waits for `[watch] run completed` before signalling:

```
signal at startup           arm present: ok    arm deleted: ok    <- proves nothing
signal after a completed run  arm present: ok    arm deleted: FAILED
                                                 "ignored SIGTERM for 30s"
```

Full `run_watch` suite: 3 passed. Clippy `--all-targets --all-features` and `fmt --check`: clean. No codegen cascade.

- **Least confident:** the SIGINT race itself is still not covered by a deterministic test. I fixed it and reasoned about it, but the test that pins this PR is the SIGTERM one — that failure mode is 100% reproducible, the SIGINT one is a timing window. If you want the race pinned too, it needs a fault-injection hook in the debounce path.
- **Not verified:** Windows. The SIGTERM arms are `#[cfg(unix)]` with a `pending()` fallback, and the crate compiles, but I did not run a Windows build.


hugocorreia90:
CI failed my own test, and the fix is in the test, not the product code.

```
local (idle mac)   passes
CI (linux)         FAILED: "ignored SIGTERM for 30s"
```

I asserted a **flat 30-second deadline**. This file's own helper exists to prevent exactly that, and says so:

> "finishes within 30 seconds" was a property of the machine

The test now calibrates one run — clock started before the spawn, stopped at the first completion line — and allows `SHUTDOWN_BUDGET_RUNS` run-times, clamped to [30s, 180s], matching its sibling test. The failure message reports the measured run time so a loaded runner is distinguishable from a stuck watcher.

**Stated as a hypothesis, not a conclusion:** I believe CI's shutdown simply took longer than 30s under load. I cannot prove that from here — the alternative is that SIGTERM is genuinely ignored on Linux, in which case this push just makes the test take up to 180s before failing the same way. CI is the discriminating experiment; I am not claiming it is fixed until it goes green there.

What does hold regardless: the mutation check. With the SIGTERM arm deleted the test fails, with it present it passes — so the arm is load-bearing.

Also ran the gates at CI's scope this time (`--workspace`, not `-p rocky-cli`), which is how the earlier `clippy::zombie_processes` failure slipped through.

hugocorreia90:
## Correcting myself: this PR does not close #1405

I read the CI failure wrong twice. The evidence now:

```
attempt 1  flat 30s budget    MY test failed          -> budget was wrong (my bug)
attempt 2  scaled budget      MY test PASSED on Linux -> budget fix worked
                              the OLD test failed instead
```

So the timing hypothesis was **not** dead — I misattributed the second failure. My new `run_watch_exits_on_sigterm` passes on Linux.

**But the old test still fails, and not from load:**

```
rocky was still running 30s after SIGINT.
One run measured 494ms on this machine and the budget is 6x that
```

A 494ms run means the runner was idle. rocky was genuinely stuck for 30 seconds after SIGINT — **with this fix applied**.

## What this PR does and does not do

| defect | status |
|---|---|
| `ctrl_c()` future dropped each iteration, signal discarded | fixed, reasoned + code-traced |
| no SIGTERM arm at all — only SIGKILL could stop the watcher | fixed, mutation-checked (delete the arm → 30s hang) |
| SIGINT hang after a file-change re-run | **still reproduces on Linux** |

The two fixed defects are real and independently valuable — the SIGTERM one is a hard bug on any machine, and the new test proves it. But the headline symptom in #1405 is not closed, so I have changed `Closes` to `Refs`.

## What I am not doing

Guessing at a third fix from a machine that cannot reproduce it. The hang needs a Linux reproduction to locate — my remaining hypotheses (a slow in-flight drain inside `run()`, or something keeping the process alive after the loop returns) are untested speculation, and I have already been wrong twice on this issue tonight.

Suggest landing this for the two proven fixes, keeping #1405 open with this evidence attached.

PR #1498, opened 2026-08-20
fix(engine/rocky-core): no-follow both ends of the .ff-prev backup copy

### LINKED ISSUE #1482: rocky product commit: open the .ff-prev backup source with O_NOFOLLOW to close the copy-source symlink race
## Residual (known, low severity, v0-conceded local racer)

The `rocky product` commit protocol backs up each final it will replace by copying it to a `.ff-prev` sibling, and there are TWO symlink-follow endpoints on that copy, both TOCTOU against a swap made after `contained_target` validates and before the syscall:

```
std::fs::copy(&final_path, &prev)   // engine/crates/rocky-core/src/product/commit.rs (~line 545)
                    │         └── DESTINATION: a symlink swapped in at <final>.ff-prev is followed on write
                    └── SOURCE: final_path swapped to a symlink is followed on read
```

The static ancestor/leaf symlink cases (plant once, no race) are already closed by the `contained_target` containment check + O_EXCL staging writes. What remains is only the copy pair above, and only under an active local racer running concurrently with `rocky product compile` — the hostile-local boundary v0 already concedes.

## Fix

- SOURCE: open `final_path` with `O_NOFOLLOW` (Unix: `OpenOptions::new().read(true).custom_flags(libc::O_NOFOLLOW).open(&final_path)`) and stream, instead of `std::fs::copy`.
- DESTINATION: write the `.ff-prev` backup through the same `write_new_no_follow` (O_EXCL) helper the staging leaf already uses, rather than letting `copy` create the destination.
- Gate on `#[cfg(unix)]`; Windows has no `O_NOFOLLOW` — document the platform gap.

## Also correct in the same change

- The `commit_generation` doc comment currently says "one race residual" — it understates the boundary; there are two `.ff-prev` copy endpoints (source read + destination write). Reword to name both.
- Windows junction / reparse-point behavior is untested: the symlink exploit tests are `#[cfg(unix)]` only. Add a Windows junction test (or document the coverage gap explicitly) so the containment claim is not silently Unix-only.

## Scope

Fast-follow to FF-WP-E1B (the `rocky product` verbs). NOT a merge blocker for that PR — the static exploit classes are closed and tested; this is the conceded dynamic-racer residual plus a doc-accuracy and Windows-coverage cleanup.

hugocorreia90:
Closes #1482.

`rocky product commit` backed up each final with `std::fs::copy`, which **follows symlinks on both ends**. The pre-check refuses a symlinked final, so the exposure is a TOCTOU race: a local process that can write into the models directory swaps the source between the check and the copy.

```
pre-check: final is a regular file   OK
   <-- racer swaps final -> symlink to an out-of-project secret -->
fs::copy reads THROUGH the link, stamps the secret into <final>.ff-prev
a rollback then renames .ff-prev over the final
   => the secret becomes the committed artifact
```

Both ends are now opened explicitly: the source with `O_NOFOLLOW` (via `OpenOptionsExt::custom_flags`), the destination keeping `O_EXCL`, with the mode carried across descriptors. Unix-gated; the non-unix path is unchanged and, as this module's own header says, untested — every symlink exploit test here is `#[cfg(unix)]`.

## The mutation check found a hole in the tests, not the fix

Reverting the call site to `std::fs::copy` left **all 38 tests green**. The two symlink tests call `copy_no_follow` directly — they prove the helper, never the wiring. They have to: the race needs a swap landing between the pre-check and the copy, and staging that deterministically needs a fault-injection hook this module does not have.

So the call site is now pinned by its own test:

```
call site reverted to fs::copy   before: 38 passed, 0 failed
                                 after:  38 passed, 1 FAILED
```

Its needle is assembled from fragments so the assertion cannot match its own source text — a literal would have matched itself and passed regardless.

## Verification

- `cargo test -p rocky-core --lib product::commit`: 39 passed.
- Clippy `--all-targets --all-features --workspace` and `fmt --check`: clean.
- No `output.rs`/`config.rs` touch, so no codegen cascade.

## Provenance

The `commit.rs` change was written by an agent working against the then-unmerged `feat/e1b-product-verbs` branch; it could not be proposed standalone until #1483 landed. Cherry-picked onto main here, re-verified end to end, and the wiring guard added on top after the mutation check exposed the gap.

- **Least confident:** the race itself is still not reproduced end to end — only the helper's refusal and the call site's wiring are proven. A fault-injection hook between pre-check and copy would close that, and would benefit the other symlink vectors in this module too.
- **Untested:** Windows. The no-follow path is unix-only by construction and the module documents that limitation.


hugocorreia90:
## Red-team disposition — verdict was DO NOT MERGE, and it was right

Independent reviewer: **Codex (GPT-5)**. It found that this fix, while closing #1482's exact vector, **introduced two regressions of its own** by replacing `std::fs::copy` with a hand-rolled read+write. Both are now fixed here.

### Fixed in this PR

| # | finding | fix |
|---|---|---|
| 1 | **Privilege escalation.** Source mode carried verbatim onto a Rocky-created file. Race in a `04755` source → a **Rocky-owned setuid backup**, which a rollback renames over the final. | mode masked to `0o777`. New test asserts the **unmasked** mode — the existing one masked `& 0o777` and could never have seen those bits. |
| 5 | **Denial of service.** `read_to_end` on an unchecked descriptor. A raced FIFO blocks the open until a writer appears — forever, no timeout above it. `fs::copy` refused non-regular sources and streamed. | `O_NONBLOCK`, a regular-file check on the **descriptor** (not the path — that can be swapped), and a bounded read that also catches a file growing mid-read. |
| 6 | **Orphan backup.** A failed write or chmod left `.ff-prev` behind, and recovery restores a backup it finds there. | every failure after the destination exists removes it. |

Also: the backup is now created `0600` before its final mode is applied, so a `0600` source is never briefly world-readable — and stays `0600` if the chmod fails.

Mutation-checked, both new guards:

```
remove the 0o777 mask        -> setuid test FAILED
disable the is_file() check  -> FIFO test FAILED
```

41 tests pass; clippy `--all-targets --all-features --workspace` and `fmt --check` clean.

### Filed, not fixed here

- **#1500** — containment is pathname-based: hardlinks pass the symlink-only checks, Windows keeps the original race (`O_NOFOLLOW` is `#[cfg(unix)]`), and a parent directory swapped after validation still races. Needs dirfd-relative operations and an explicit hardlink policy.
- **#1501** — Phase B reads the model sidecar with `read_to_string` **before** the pre-check and the merge preserves every unowned key, so a raced sidecar symlink exfiltrates external TOML keys into the committed artifact. Reachable per the review; I have not reproduced it end to end.

### Acknowledged, not actioned

Finding 7: the wiring test is textual and could pass from dead code or a comment. True. The reviewer confirms there is exactly one production `.ff-prev` copy call and no remaining production `std::fs::copy`, so there is no current bypass. Injecting the copy through `CommitOps` is the right answer and is a refactor beyond this fix.

hugocorreia90:
## Delta review verdict: BLOCK — fixed

Independent reviewer: **GPT-5 Codex**. It found the cleanup I added in the previous round was itself unsafe.

**The finding:** `O_EXCL` guarantees exclusivity *at open time and nothing after it*. My failure path called `remove_file(dst)` by **path**:

```
open dst with O_EXCL        this file is mine
   <-- racer unlinks it, moves another file into that name -->
write fails
remove_file(dst)            deletes a file this copy never created
```

In a directory the attacker can write, that file could be one of the user's. Deleting a bystander's file is worse than the orphan the cleanup existed to prevent.

**Fix:** capture `(dev, ino)` from the descriptor at creation, and unlink only if `symlink_metadata` of the path still matches. `symlink_metadata` rather than `metadata` so a swapped-in symlink compares as itself instead of resolving to its target.

Mutation-checked: bypass the inode comparison and `the_failure_cleanup_spares_a_file_it_did_not_create` fails. The test also asserts the matching case still cleans up, so the guard cannot pass by disabling cleanup entirely.

A window remains between the check and the unlink — closing it needs dirfd-relative removal, which is #1500 along with the other pathname-based races. Between the two failure modes, leaving an orphan is the safer one: a stale backup restored by recovery is a bug; deleting a bystander's file is data loss. That trade-off is written into the code comment rather than left implicit.

### The other four answers, on the record

- **setuid mask on non-unix:** correct that the `#[cfg(not(unix))]` branch clones unchanged. Windows `Permissions` carries only the read-only flag — there are no set-user-ID bits to strip, so there is nothing to mask there.
- **Privileged bits after rollback:** the reviewer could not confirm from the five functions in scope. Worth a look, but rollback is outside this diff.
- **`/proc` objects that report as regular files:** they pass `is_file()`, but the `take(MAX_BACKUP_BYTES + 1)` read bounds them — confirmed in the same answer.
- **Memory:** the 16 MiB bound is correct; a fixed allocation can still fail under system-wide pressure, which is true of any read.

hugocorreia90:
## Round 4: BLOCK accepted — the cleanup is gone entirely

Reviewer: **gpt-5.4-codex**. The finding: my inode-checked unlink is safer than an unconditional one but **not safer than not deleting at all**. Check-then-unlink is non-atomic, so a deletion window remains; doing nothing has none.

That is correct, and it changes the answer rather than refining it. `O_EXCL` grants exclusivity at open time and nothing after it, so *any* path-based unlink in a directory an attacker can write risks removing a file this copy never created.

```
unconditional remove_file   deletes a bystander's file outright
inode-checked remove_file   narrows the window, cannot close it
do nothing                  no deletion window            <- chosen
```

Between a stale backup and destroying a bystander's file, the stale backup is the recoverable failure. So the partial `.ff-prev` now stays, and the orphan is handled where it is consumed — **#1502**: restore must consult the journal's `has_prev` instead of trusting a backup it finds on disk. That is the reviewer's own original suggestion from round 2, and it is the right home for it.

This is not a regression against main: `std::fs::copy` could leave a partial destination on failure too.

### Where this PR now stands

```
round 1   fs::copy follows symlinks both ends      the original #1482 bug
round 2   setuid carried / FIFO blocks / orphan    regressions I introduced
round 3   cleanup deleted by path                  fixed with an inode check
round 4   inode check still has a delete window    cleanup removed entirely
```

Four independent rounds, every finding real, every one dispositioned on the record. What ships is narrower than what I first wrote, which is the correct direction for a security change.

The reviewer's third point stands unaddressed by design: the tests exercise the helpers sequentially and cannot prove behaviour under a real concurrent race. Nothing in this module can today — that needs the fault-injection seam noted in #1500.

Issue #475, opened 2026-05-11
Wire batch_row_counts trait call into rocky-adapter-sdk conformance runner


hugocorreia90:
`engine/crates/rocky-adapter-sdk/src/conformance.rs::run_conformance` currently iterates every `TestSpec` and marks all supported tests as `Passed` with zero duration. #383 wires one real trait call (`SqlDialect::format_table_ref`). This issue is the next deliberately narrow slice: wire **one more** trait call so the runner exercises another adapter method against a real fixture.

> Depends on #383 — read its merged PR as your template; this issue follows the same shape with one different trait.

## Scope

When the spec name is `"batch_row_counts"`, invoke `BatchCheckAdapter::row_counts(...)` against a stub `TableRef` (or stub adapter) and assert the returned counts vector is non-empty. Leave the other ~25 specs on the current `Passed`-stub path — follow-up issues will wire them incrementally.

Stronger assertions are adapter-specific — keep this first wiring simple.

## Code refs

- `engine/crates/rocky-adapter-sdk/src/conformance.rs:282-286` — the `batch_row_counts` spec entry
- `engine/crates/rocky-adapter-sdk/src/conformance.rs:309` — `run_conformance` signature
- `engine/crates/rocky-adapter-sdk/src/conformance.rs:337-348` — the "all supported tests marked Passed" placeholder block
- `engine/crates/rocky-adapter-sdk/src/traits.rs` — `BatchCheckAdapter::row_counts`

## Acceptance criteria

- `run_conformance` invokes `BatchCheckAdapter::row_counts` for the `batch_row_counts` spec only
- A new in-crate test exercises the wiring with a stub adapter that returns at least one row count
- All other specs continue to land on the current `Passed`-stub path
- `cargo test -p rocky-adapter-sdk` is green
- `docs/src/content/docs/guides/adapter-sdk.md` gains a one-line note that `batch_row_counts` is now wired, matching the style of the note added in #383

## Effort

2–3 hours. The bigger trait-unification work is deliberately out of scope.

hugocorreia90:
Picked this up and stopped before writing code: the premise doesn't hold in this tree, and following the issue as written would buy a breaking SDK change for a code path that never executes.

## What's actually there

There are **two different `BatchCheckAdapter` traits**, and the issue points at the one nobody implements:

- `rocky_core::traits::BatchCheckAdapter` — implemented by `rocky-snowflake`, `rocky-databricks`, `rocky-bigquery`.
[...]

PR #1506, opened 2026-08-21
fix(engine/rocky-adapter-sdk): stop reporting unimplemented conformance specs as passed

hugocorreia90:
Refs #475 — fixes the real defect, deliberately **not** the remedy in that issue's title (reasoning below).

## The false green

`run_test_spec` implements **one** arm:

```rust
match spec.name {
    "format_table_ref" => { ...actually checks... }
    _ => TestOutcome::Pass,          // <- everything else
}
```

Twenty-six specs are declared. Connect, execute_statement, schema and table lifecycle, grants, batch checks and discovery all fell through that catch-all. So `rocky test-adapter` printed **26 passing conformance tests having executed one** — for any adapter, including one that cannot connect.

They now report `Skipped` with the reason, and `tests_run` (passed + failed) counts only work that happened.

## The existing test asserted the bug

```rust
assert_eq!(result.tests_skipped, 0);
assert_eq!(result.tests_run, 26);   // "All 26 tests should run"
```

It pinned the false green in place. It now asserts the honest counts, plus the property directly — **no spec reports Passed unless its check ran**:

```rust
assert!(r.status != TestStatus::Passed || r.name == "format_table_ref", ...)
```

That form survives someone implementing a second arm: they add its name and the invariant still holds.

## Why not wire `batch_row_counts`, as #475's title asks

That remedy rests on a false premise. There are **two traits named `BatchCheckAdapter`**:

```
rocky-core/src/traits.rs:1819         implemented by snowflake, databricks, bigquery
rocky-adapter-sdk/src/traits.rs:715   implemented by NOBODY  <- what conformance calls
```

Wiring it would report `Skipped` on every adapter anyway, while forcing a breaking async signature on a sync `pub fn`. The false green is the real defect, it is independent of the trait split, and it is the part that matches the issue's `good first issue` label.

Leaving #475 open for the trait-convergence decision.

## Evidence

Mutation check — restore `_ => Pass`:

```
test_conformance_full_capabilities   FAILED
```

62 tests pass in `rocky-adapter-sdk`. Clippy `--all-targets --all-features` and `fmt --check` clean.

## Behaviour change worth knowing

Exit status is unchanged — it keys on `tests_failed`, so no pipeline breaks. But a run that printed "26 passed" now prints 1 accounted-for and 25 skipped. That is the honest number, and it will look like a regression to anyone who trusted the old one.

## Least confident

Whether any downstream consumer greps that summary line. I checked the CLI's own exit path (`test_adapter.rs:85`, keys on failures only) but did not sweep docs or external pipelines for a hardcoded "26".


hugocorreia90:
## Both non-blocking findings fixed

Reviewer: **Claude Opus 5** — verdict was `nothing blocks merge`, with two findings I have taken rather than deferred, because its suggested fix removes the class rather than the instances.

**1. Latent double-count.** The `Failed` arm's comment said *"already counted as a failure; do not double-count"* while `tests_failed += 1` ran unconditionally below it. Dead today (the first spec always skips), so it was a defect whose comment claimed the opposite of its code — one `run_test_spec` arm away from live.

**2. The assertion could not see its own bug.** `tests_run <= results.len()` is satisfied in **release** by a wrapped `usize::MAX + 1 == 0`. The mutation check that proved the underflow was therefore debug-profile-specific.

**The fix, as suggested: recount from `results` instead of patching deltas.** Delta-patching produced the original underflow *and* the double-count. Recounting makes every prior status correct by construction.

The test now asserts the accounting identity rather than a bound:

```rust
assert_eq!(tests_passed + tests_failed + tests_skipped, results.len());
assert_eq!(tests_run, tests_passed + tests_failed);
```

Plus a new test pinning the previously-dead `Failed` case. Mutation-checked — reintroducing `tests_failed += 1` after the recount fails it.

**Also from the review, now done:** `docs/.../development.md`'s `19/19 core tests passed` example was fabricated in a format the command never printed; replaced with the real `Result: N passed, N failed, N skipped` shape. And an all-skipped run now prints an explicit warning, since the exit status keys on failures and would otherwise read as success for a run that verified nothing.

**Not done, deliberately:** the rustdoc on `run_conformance` and two concepts pages still describe the old placeholder-pass behaviour. They are text-only drift from the already-merged base, and I would rather fix them in one docs pass than scatter them across this PR.

PR #1514, opened 2026-08-21
tools: guards for mutation checks and branch hygiene

hugocorreia90:
A tool, because the discipline kept failing.

## The problem

`AGENT_REVIEW.md` requires evidence that a test fails when the fix it guards is reverted. Done by hand that is: edit the source, run the test, undo the edit. The undo — `git restore` / `git checkout --` — discards **everything uncommitted in that file**, not just the mutation.

That has destroyed real work in this repo repeatedly, four times in a single session. There is no reflog for content that was never committed.

## The fix

`scripts/mutation-check.sh` refuses to run unless `git status --porcelain` is empty, so its restore can only ever take back the mutation it made.

```
scripts/mutation-check.sh <file> <mutation.py> <test-command...>

exit 0   test FAILED under mutation   -> fix-sensitive (good)
exit 1   test PASSED under mutation   -> the test proves nothing
exit 2   refused (dirty tree / target text not found / bad args)
```

Refusing on a dirty tree also enforces the right order: commit the fix **and** its test first, because the committed state is what the evidence is about.

## Verified on all three paths

```
dirty tree       REFUSED, and the uncommitted scratch line survived
real mutation    exit 0 — caught a tests_passed underflow (PR #1506)
comment-only     exit 1 — "that test does not guard this change"
```

The mutation script must assert its target appears exactly once (`assert s.count(old) == 1`), so a mutation that silently did not apply cannot read as "the test is insensitive". The script also fails if the file is unchanged after the mutation runs.

## Two failure shapes it catches, both seen this week

1. **A test that reimplements the logic instead of calling it.** In #1506 my first regression test copied the override into the test body; mutating the production code changed nothing and the test passed. Extracting the function and calling it made the check meaningful.
2. **A guard that was never wired.** In #1498 the symlink tests called the helper directly, so reverting the production call site left all 38 tests green.

`AGENT_REVIEW.md`'s test-evidence bar now names the tool and both shapes.

## Least confident

Whether `python3` for the mutation script is the right interface — it is what is already used ad hoc in this repo, and it makes the exactly-once assertion natural, but a `sed` form would have fewer moving parts for simple replacements.

---

### The review found this script had the bug it exists to catch

```
$ ./mc.sh src/x.rs mut.py this-binary-does-not-exist
mutation-check: OK - the test failed under mutation, so it is fix-sensitive.
exit=0
```

`if "$@"` treated every nonzero exit as evidence about the mutation. A missing binary, a typo'd crate name, or — the common case for Rust — a mutation that breaks compilation, all printed OK. And `AGENT_REVIEW.md` had just told reviewers to trust that signal, which is what made it blocking rather than merely wrong.

Three fixes, each verified against the reviewer's executed cases:

```
missing binary          was OK/0   now REFUSING at baseline, exit 2
mutation breaks build   was OK/0   now INCONCLUSIVE, exit 3
genuine sensitivity                still OK, exit 0
corrupt git index       guard failed OPEN (git status exits 128 printing
                        nothing, which read as clean)  now refuses
```

The baseline run — the test command must pass on the clean tree before anything is mutated — is what closes the first two. The `git status` exit check closes the third, which mattered most: the guard failing open in the one situation it exists for.

Also folded the log cleanup into `restore()`. A second `trap ... EXIT` would have silently replaced the restore trap.

### Two guards for work landing on the wrong branch

Same session, three incidents: unrelated tooling committed inside a bugfix PR, a docs edit inside an engine PR, and a branch cut from another PR's branch that then carried that PR's payload through a squash-merge — firing its `Closes #NNN` trailers under the wrong number.

`.git-hooks/post-checkout` — `git checkout <branch>` does not stop for a dirty worktree; it carries non-conflicting changes across silently. git has no pre-checkout hook, so this cannot refuse; it makes the carry impossible to miss, which is the actual failure since it happens quietly. It never blocks — the switch has already happened by the time a post-checkout hook runs.

`scripts/new-branch.sh` — refuses on a dirty tree, and defaults the base to a freshly-fetched `origin/main`. Deliberate stacking still works by passing the base explicitly, and warns about the squash-merge hazard.

Verified: the hook fires on a carried change and stays quiet on a clean switch; the script refuses dirty (exit 2), warns on an explicit stacked base, and creates cleanly from `origin/main`.

### Least confident, still

Whether a repo-level `post-checkout` hook is welcome — it is advisory and skippable with `ROCKY_SKIP_HOOKS=1`, but it prints a banner, and anyone who routinely moves work across branches on purpose will find it noisy.

### DIFF
--- a/.git-hooks/post-checkout
+++ b/.git-hooks/post-checkout
@@ -1,0 +1,69 @@
+#!/usr/bin/env bash
+#
+# Rocky monorepo post-checkout hook.
+#
+# Shouts when a branch switch CARRIED uncommitted work across with it.
+#
+# `git checkout <branch>` does not stop for a dirty worktree: when the change
+# does not conflict, git brings it along silently. The work then belongs to
+# whichever branch you happen to land on, and the next `git add -A` commits it
+# there. That has repeatedly put changes on the wrong branch in this repo —
+# unrelated tooling inside a bugfix PR, a docs edit inside an engine PR, and a
+# branch accidentally stacked on another PR's commits.
+#
+# git has no pre-checkout hook, so this cannot refuse. It can make the carry
+# impossible to miss, which is the whole failure: it happens quietly.
+#
+# Skip with ROCKY_SKIP_HOOKS=1.
+set -uo pipefail
+
+# Args: $1 = previous HEAD, $2 = new HEAD, $3 = 1 for a branch checkout,
+# 0 for a file checkout. Only a branch switch can carry work.
+prev_head="${1:-}"
+new_head="${2:-}"
+branch_switch="${3:-0}"
+
+[ "$branch_switch" = "1" ] || exit 0
+[ "${ROCKY_SKIP_HOOKS:-}" = "1" ] && exit 0
+
+# A no-op checkout of the branch you are already on carries nothing.
+[ "$prev_head" = "$new_head" ] && exit 0
+
+carried="$(git status --porcelain 2>/dev/null)"
+status_rc=$?
+if [ "$status_rc" -ne 0 ]; then
+    echo "post-checkout: WARNING — could not read git status; cannot tell whether" >&2
+    echo "               uncommitted work came across with this switch." >&2
+    exit 0
+fi
+[ -z "$carried" ] && exit 0
+
+branch="$(git rev-parse --abbrev-ref HEAD 2>/dev/null || echo '?')"
+
+cat >&2 <<WARNING
+
+╔════════════════════════════════════════════════════════════════════╗
+║  UNCOMMITTED WORK CAME WITH YOU                                    ║
+╚════════════════════════════════════════════════════════════════════╝
+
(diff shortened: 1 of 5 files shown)


hugocorreia90:
Found while using this script heavily on another branch — it reports INCONCLUSIVE on a mutation that **correctly kills a test**.

Cargo prints `error: test failed, to rerun pass …` when a test fails. The build-error grep at the end of the script matches that as `^error:`, so a working mutation looks like a broken build. It hit 3 of 9 mutations in one round.

Narrow fix: filter exactly that one line out of the mutated log before the grep — `grep -v '^error: test failed, to rerun pass'`.

Keep `error[E…]:` and `error: could not compile` tripping INCONCLUSIVE. Those are the real "the mutation broke the build so the test never ran" signals, and that branch is the false-green guard the script exists for — a mutation that fails to compile must never read as killed.

Issue #1515, opened 2026-08-21
security(H1): trusted custody of the repair driver's output bytes through Phase B


hugocorreia90:
## What's open

The fulfillment loop's repair round needs the drafting agent to rewrite the model sidecar. So before dispatching a repair worker, the loop reopens the drafting window: it byte-verifies every hash the committed manifest records, then demotes that manifest to its contract-only phase.

That fixed #1493 — a repair is no longer mis-reported as tampering — and leaves two properties the engine does not authenticate.

**D1 — the window is not hash-pinned.** The contract-only phase does not cover the sidecar the repair driver writes, so while the window is open that file is unhashed. The merge that closes it keeps every key the lowering does not own and every `[[tests]]` entry it did not generate (`lowering.rs:222`, `:238-242`, `:262-271`, `:298-304`, `:347-359`). Whatever sits in the file when the window closes is merged, hashed, and committed.

**D2 — a forged manifest skips the check.** Setting the phase back to contract-only and dropping the sidecar's entry makes the reopen find nothing to verify. Not new: the engine already says a manifest is data, not a credential (`commit.rs:823-826`).

## Why this is H1, not urgent

Both need a process that can write your models directory while the loop runs. **That process can already write an approval marker and get any plan applied** — a strictly larger capability, and one v0 concedes. Neither widens the conceded adversary. They are two more places where an artifact is trusted because it is on disk, which is exactly the H1 thesis: signed approvals and worker sandboxing are both about making a local artifact authenticatable.

## What "done" looks like

The engine can tell, at merge time, that the sidecar it is about to commit came from the worker it dispatched. Cheapest first:

1. **Custody through the outbox.** The worker already writes to a per-task outbox. If the repair's sidecar contribution arrived there and the runner merged it — instead of the worker editing in place under `models/` — the window closes entirely, and Phase B merges bytes the runner took custody of. No new trust primitive needed.
2. **Hash the window's start state and diff it.** Detects rather than prevents, and means little without an attributable record of writes.
3. **Sign the manifest.** Answers D2, not D1. Only worth doing on the same key material as signed approvals.

Whatever lands must keep #1493's property: hashes are written only by the staged commit protocol, and drift in a window where no write was authorized still fails closed.

## Already shipped — don't redo

- The reopen verifies the whole committed manifest before demoting; drift is refused with nothing mutated.
[...]

Issue #1493, opened 2026-08-20
fulfill: a repair round rewrites the merged sidecar and is then mis-reported as tamper


hugocorreia90:
## Summary

In the `rocky fulfill` loop, a repair round (triggered by a red verify) re-runs a `draft_model` call that legitimately rewrites the Phase-B-merged sidecar. The subsequent artifact-integrity check compares the current sidecar against the byte hash Phase B recorded **before** the repair, so the legitimate rewrite is classified as `tamper` and the product is wrongly blocked.

This is a correctness bug in the merged loop (present at least as of `82d10a86`). It was surfaced by the E3 walking-skeleton POC, which currently avoids it by setting `repair.calls=[]` in its replay session — i.e. the POC cannot exercise a real repair round.

## Trace (from an independent review)

```
 red verify ─► TaskKind::Repair                          machine.rs:1025
            ─► step.rs dispatches repair via the           step.rs:376
               ordinary drafting path
            ─► repair draft_model parses, reserializes,     tools.rs:2297 / :2358
               and OVERWRITES the merged sidecar
 Phase B had recorded that sidecar's exact byte hash        lowering.rs:711
            ─► DraftingFinished runs the artifact check,    step.rs:432 / machine.rs:944
               compares current merged manifest to the
               pre-repair hash ─► reports TAMPER
```

## Impact

Any real product whose first verify is red — the common case that repair exists to handle — cannot complete: the repair that fixes it is mis-read as tampering. Repair is a core loop capability, so this blocks the loop on exactly the inputs it was built to recover from.

## Fix direction (for discussion)

The artifact-integrity check must distinguish a *legitimate repair rewrite* from *external tampering*. Options: re-record the merged sidecar's hash after a repair `draft_model` writes it (repair is an authorized in-loop actor); or scope the tamper check to the pre-drafting window only; or have repair go through a path that updates the recorded hash transactionally. The invariant to preserve: an out-of-band edit between gates is still caught, but the loop's own authorized repair write is not.

## Scope

Pre-existing engine defect, not introduced by any open PR. Confidence note from the review: deterministic for repair rounds that use `draft_model`; an SQL-only repair path may avoid the sidecar rewrite. A red-verify integration test with a repairing `draft_model` call would pin the full envelope.

hugocorreia90:
Fixed in #1516 (`234f032d`).

[...]

PR #1516, opened 2026-08-21
fix(engine): let a repair round recover instead of failing as tampering (#1493)

hugocorreia90:
A product whose first verify came back red could never recover. The repair round rewrites the model sidecar — that is its job — but the integrity check compared the rewritten file against the hash recorded *before* the repair, and reported the loop's own authorized write as tampering. Repair is the loop's recovery path, so the loop could not recover.

```
 before   red verify ─► repair rewrites the sidecar ─► "tampered" ─► blocked
 after    red verify ─► reopen the drafting window ─► repair ─► re-merge ─► converge
```

## How it works

Before dispatching any drafting worker over a committed merged manifest, the loop **reopens the drafting window**: byte-verify every hash that manifest records — drift there happened in a window with no authorized writer, so it still fails closed as tampering, with nothing mutated — then re-lower the contract from the approved spec and commit it through the existing staged protocol. The next merge re-records the sidecar's hash transactionally.

Hashes are still written **only** by the commit protocol. No verification path was weakened, and there are no repair-conditionals in any check. The reopen keys on the manifest phase on disk, never on in-memory state, so a crashed run resumes into the same verify-first decision.

## Three things fixed beyond the bug itself

An independent review found these; all three are about our own surface area, not an attacker:

- **The reopen was callable by anyone.** It took a caller-supplied record — a type with public fields — so any caller could mint one and demote a merged generation without the loop ever deciding. The raw demotion is now crate-private, and the public entry reads the loop's compare-and-set record from the state store itself. A compile-fail test pins that out-of-crate.
- **A crash lost the repair's identity.** Only "drafting" was persisted, so a resume after the repair decision dispatched a *draft* — wrong brief, wrong budget. The round is now persisted and read back on both the crash-resume and driver-retry paths.
- **A concurrent `rocky product compile` could re-merge mid-window**, making the legitimate repair look like tampering again. It now refuses while a fulfillment record carries an in-flight stamp, and names the recovery. The first version of that guard compared process ids and was fail-open against pid reuse; it no longer compares pids at all.

## What is disclosed, not fixed

Two properties remain, and both are now stated publicly rather than left implicit:

- While the repair window is open the sidecar is not hash-pinned, and the merge that closes it preserves content the lowering does not own.
- A committed manifest is data, not a credential — editing it changes what gets checked.

Both need a process that can write your models directory while the loop runs. **That process can already write an approval marker and get any plan applied** — a strictly larger capability that v0 concedes. So neither widens the conceded adversary. Trusted custody of the repair agent's output bytes is tracked as **#1515**, scoped with the hardening work where it belongs.

The docs that called repair a known-broken issue now describe it working, with the residual named. The walking-skeleton POC exercises a repair round on its replay lane and pins the rounds **by kind**, so a repair dispatched as a plain draft cannot pass.

## Evidence

`cargo test --workspace` 147 suites exit 0 · `fulfill_e2e` 15 · both compile-fail crates · clippy `-D warnings` · `fmt --check` · POC `run.sh` and `mutation-pass.sh` 13/13 · docs build. Every new guard is mutation-checked, including that the derive shortcut for the persisted round would be wrong and that a recycled pid buys no ownership. No codegen surface touched (checked, not assumed).


hugocorreia90:
Updated after review. Three rounds of independent review ran on this branch; the last two found real defects **in the fixes themselves**, which is why the shape changed:

- **The compile guard no longer takes a caller's word for who it is.** The `CompileCaller` argument is deleted. Identity is now derived from the running process — pid **and** start time, via a probe shared in `rocky_core::process` — and fails closed when it cannot confirm. That removes the bypass where the public façade handed out the loop's exemption, because there is no longer an argument to hand out.
- **The reopen reads the loop's decision instead of accepting one.** The raw demotion is crate-private; the public entry takes the state store and reads the record itself.
- **The self-lockout seam is drilled.** The owner gate is the one change that could brick a working loop, so there is now a fault-injection test that crashes right after ownership is stamped and proves the product still converges — including through a repair round.

One claim was narrowed rather than defended. The reopen takes a caller-supplied store, so code inside this process could construct one that satisfies the gate. That is not a hole: an in-process caller already holds every capability the process holds, including simply rewriting the files the function would have written. The code doc now says exactly that, and the compile-fail proof states what it proves — the raw demotion is not nameable out-of-crate, and the entry reads its evidence — and explicitly not provenance.

**Deliberate route-inventory addition:** the start-time probe moved down from `rocky-fulfill` to `rocky_core::process`, because `rocky-cli` needs the same answer and cannot depend on a crate above it. The inventory gate caught it as consumed-but-not-pinned and it is now pinned with that justification.

Evidence at `274149ed`: workspace 147 suites exit 0, `fulfill_e2e` 16, rocky-core 1993, both compile-fail crates, clippy `-D warnings`, fmt, POC `run.sh` and `mutation-pass.sh` 13/13. Every new guard is mutation-checked; one mutation that did **not** kill its test is reported as such in the record rather than dropped.

Issue #1537, opened 2026-08-25
A malformed surrogate_key on an unrelated model fails a narrowed offline preview


hugocorreia90:
A malformed `[[surrogate_key]]` on **any** model makes the offline preview fail, even when the preview is narrowed to a different model.

```
rocky emit-sql                        → refuses
plan_preview --model orders           → refuses, because model `customers`
                                        has a malformed surrogate_key spec
```

The run path does not behave this way: it resolves surrogate keys only for the models actually selected.

## Where it comes from

`plan_preview_output` and `emit_sql` both load the surrogate-key tree for the whole project before rendering. `run` filters to the selected models first.

The preview recently gained surrogate-key application deliberately, so that a preview shows the same SQL a run produces — previously it silently omitted declared keys, which meant an agent could approve preview SQL that lacked a materialized key. Matching `emit-sql` was the right parity target for that change. Inheriting its unfiltered load was not a separate decision.

## Suggested fix

Filter the surrogate-key load to the selected models in both `emit_sql` and `plan_preview_output`, so a spec on an unrelated model cannot fail a narrowed preview.

Worth noting the counter-argument before changing it: an unfiltered load means a malformed spec anywhere is reported early rather than lying in wait until someone happens to select that model. If that is the intent, it should be stated rather than inherited — and the message should then say which model is at fault, which it currently does.

## Provenance

Named by the author of the parity change as a known consequence rather than found later. Filed so it is tracked rather than living only in a review thread.

hugocorreia90:
## Decision: filter the surrogate-key load to the selected models.

`emit_sql` and `plan_preview_output` both load the surrogate-key tree for the whole project before rendering. They move to what `run` already does — resolve keys only for the models actually selected.

```
 before   plan_preview --model orders   refused, because `customers`
                                        has a malformed surrogate_key

 after    plan_preview --model orders   renders orders
          plan_preview (unnarrowed)     still reports the malformed spec,
                                        still names `customers`
```

The counter-argument in the description is real but loses: an unfiltered load reports a bad spec earlier, yet it does so by failing work that has nothing to do with it. Reporting early is worth having, and an unnarrowed preview or `rocky validate` still delivers it. Blocking a narrowed preview is not the way to get it.

This also restores the parity the surrogate-key change was aiming for. Matching `run`'s SQL was the point; inheriting `emit-sql`'s unfiltered load was never a decision, and the issue says so.

## What done looks like

[...]

Issue #1536, opened 2026-08-25
rocky compile refuses an unset credential placeholder, while the offline preview now tolerates it


hugocorreia90:
`rocky compile` and the MCP `compile` tool refuse to run when a credential placeholder in `rocky.toml` is unset, even though compilation never touches a warehouse.

An initialized Databricks project with `${DATABRICKS_HOST}` unset cannot be compiled at all — the config load fails before any work starts.

## Why it is inconsistent

The offline SQL preview was in the same state and has been corrected: `rocky emit-sql` and the MCP `plan_preview` tool now load the config **credential-tolerantly**, keeping an unset `${VAR}` as literal text while still refusing a malformed or semantically invalid config. `rocky compile` still uses the strict loader.

So today, on the same project with the same unset variable:

```
rocky emit-sql     → renders SQL offline, exit 0
rocky compile      → refuses, exit 1
```

Both are offline operations. Only one of them requires credentials to be present.

## The decision to make

Either is defensible and it should be chosen deliberately rather than by which loader a call site happened to reach:

- **Tolerant** — compilation is an offline operation and should work without credentials, matching the preview. `rocky validate` remains the command that insists every variable resolves.
- **Strict** — compilation is where a project is checked end to end, and a config that cannot fully resolve is not a valid project. In that case the preview's tolerance is the anomaly and should be revisited.

## Provenance

Surfaced while correcting a regression on the preview path, where sharing a strict loader accidentally made `emit-sql` require credentials and broke its documented no-credentials contract. That regression is fixed. The `compile` behaviour predates it — it is main's shipped behaviour — so it was left alone rather than changed as a side effect. Filed so the inconsistency is a decision rather than an accident.

hugocorreia90:
## Decision: tolerant. Compile matches the preview.

Compiling never touches a warehouse, so it must not require warehouse credentials. `rocky compile` and the MCP `compile` tool move to the credential-tolerant loader that `rocky emit-sql` and `plan_preview` already use.

```
 before   emit-sql   renders offline, exit 0
          compile    refuses,        exit 1     <- same project, same unset ${VAR}

 after    emit-sql   renders offline, exit 0
          compile    compiles,        exit 0
          validate   still insists every variable resolves
```

An unset `${VAR}` stays literal text. A malformed or semantically invalid config is still refused — tolerance is about unresolved placeholders only, not about accepting a broken config.

`rocky validate` remains the command that requires every variable to resolve. That is where a project is checked end to end.

[...]

Issue #1535, opened 2026-08-25
mutation-check.sh reports INCONCLUSIVE when a mutation correctly kills a test


hugocorreia90:
`scripts/mutation-check.sh` reports **INCONCLUSIVE** for a mutation that correctly kills a test. That is the one outcome the script exists to distinguish, so it inverts the signal it is for.

## What happens

When a test fails, cargo prints a trailing summary line:

```
error: test failed, to rerun pass `-p rocky-mcp --lib`
```

The build-break detector matches it:

```sh
if grep -qE '^error(\[E[0-9]+\])?:|error: could not compile' "$mutated_log"; then
```

`^error:` matches `error: test failed…`, so a mutation that *did* what it was supposed to do — break the code and make a named test fail — is reported as a broken build instead of a killed mutant.

Hit on 3 of 9 mutations in one session, and again in a later one. Every builder using this script will hit it.

## Why it matters more than a false label

The INCONCLUSIVE branch is the script's false-green guard: it exists so a mutation that fails to *compile* can never be read as "the test caught it". That is the right instinct. But because the pattern is too broad, the guard now fires on the success case as well, and the two become indistinguishable — which is exactly the ambiguity the guard was added to remove.

The practical effect is that a builder either learns to ignore INCONCLUSIVE (defeating the guard) or hand-writes a wrapper each time (what happened here, twice).

## Fix

Filter that one line before the grep:

```sh
grep -v '^error: test failed, to rerun pass' "$mutated_log" | grep -qE '^error(\[E[0-9]+\])?:|error: could not compile'
```

Keep `error[E…]:` and `error: could not compile` tripping INCONCLUSIVE — those are the real "the mutation broke the build so the test never ran" signals.

## Provenance

Reported on #1514 before it merged; the merged `scripts/mutation-check.sh:139` still carries the original pattern, so the comment arrived too late to be picked up. Filing so it is not lost. Found while mutation-checking guards on the fulfillment observation branch, where the workaround had to be rebuilt twice.

hugocorreia90:
Closing as a duplicate of #1547, which reports the same defect with a fuller reproduction and a sharper fix.

Same line, same cause:

```
scripts/mutation-check.sh:139
  grep -qE '^error(\[E[0-9]+\])?:|error: could not compile'
                ^^^^^^^ also matches "error: test failed, to rerun pass ..."
```

[...]

Issue #1533, opened 2026-08-25
inspect_schema returns an empty source list when discovery fails, indistinguishable from an empty warehouse


hugocorreia90:
`inspect_schema` appends physical warehouse tables on a best-effort basis. When that discovery cannot run, the tool still returns success and simply omits them, so a caller cannot tell "the warehouse has no other tables" from "discovery did not happen".

## Scope — narrower than it first looks

Only the **appended physical tables** degrade. `models` and the compile-derived `sources` are exact: a compile failure other than "no models found" already returns `ToolError::compile_failed`.

So with models present, the result is partly trustworthy and partly silent. At cold start — before any model exists, which is exactly when an agent leans on discovery — the whole `sources` list can come back empty.

## The two ways it goes quiet

In `engine/crates/rocky-mcp/src/tools.rs`:

```rust
// Best-effort across warehouses:
// the discovery query degrades to an empty list on any error.
if let Ok(Some(adapter)) = self.warehouse_adapter() {
    for entry in discover_source_tables(adapter.as_ref()).await {
        ...
    }
}
```

1. **`if let Ok(Some(adapter))` discards the `Err` arm.** `warehouse_adapter` returns `Ok(Some(..))` or propagates `Err` — it never returns `Ok(None)`, so `Err` is the only way this can miss, and the `None` arm is dead. Its doc justifies the `Result<Option<..>>` shape as what enables the degradation, which has it backwards.
2. **`discover_source_tables` returns an empty `Vec` on a failed query.**

`InspectSchemaResult` carries no availability or reason field, unlike `SampleRowsResult`.

## One case where the empty return is the ordinary outcome, not an edge

The discovery query selects `FROM information_schema.columns` with **no catalog qualification**, while `discover_source_tables` is called with whatever adapter the pipeline target resolved — any of five.

- **BigQuery: documented in this repo as not resolving.** `engine/crates/rocky-bigquery/src/dialect.rs`: "The view lives per-dataset (a bare `INFORMATION_SCHEMA.COLUMNS` doesn't resolve)". On BigQuery, discovery returning nothing is therefore the normal path.
- **Snowflake and Databricks: unverified.** Rocky qualifies the view for both in its own batch paths (Snowflake to `<database>.`, Databricks to `<catalog>.`), which shows the house convention but does not prove the unqualified form fails. Snowflake in particular resolves an unqualified `information_schema` against the session database, so it may work. Someone should check against a live warehouse before claiming otherwise.

## Why it matters

[...]

PR #1539, opened 2026-08-25
feat(engine/rocky-fulfill): run the declared data checks at observation and route a failure into a repair round

hugocorreia90:
Observation now runs the data checks a product declared about itself, and a failure routes back into a repair round that produces a NEW proposal a human must approve.

Before this, only compile- and contract-shaped defects could reach repair. The declared data checks cannot run at verify — they need a materialized table — so a wrong number, a violated grain, or a broken freshness budget surfaced after apply and stopped there.

```
 before
   draft ─► verify ─► propose ─► review ─► apply ─► observe ─► (halt)
              │                                        │
        compile/contract red                    data-red seen here
              │                                        │
              └──► repair                          nothing routes back

 after
   observe ─► data-red ─► repair ─► new proposal ─► HUMAN review ─► apply
```

The bad data is already applied when this fires. A data-red grants the loop no authority it does not otherwise have: every repaired output goes through a fresh propose → human review → apply cycle, with the same gates and the same digest binding as any other change.

## What is in here

- A new state — applied, and the applied output is failing a check the product declared about itself. Distinct from a healthy `observing` and from a pre-apply `blocked`. Schema version 24 (1.73.0 released v23; see the changelog "On upgrade" note).
- The deferred checks execute at observation through the **same** typed core as `rocky test --declarative`, not a second check engine.
- A data-red spends one repair round, bounded by its own counter. Exhausting the ceiling lands `blocked` naming the check, so a live table is never thrashed.
- A custody gate: the checks that run must be the checks this generation verified.
- Freshness stays report-only. Staleness is usually a scheduling fact, not a model defect.

## The custody gate, and what it cost to get right

Read this part first. It was wrong four times, and each version looked finished.

The sidecar holding the declared checks is byte-verified at Phase B and then not looked at again until observation — which is after an apply, and arbitrarily later.

1. **No gate.** Removing the checks read as passing them: an empty set yields a clean verdict, moves a known-red product to a healthy `observing`, clears the evidence, and refunds the repair budget while the bad table sits untouched. Deleting a check beat fixing it.
2. **Hashing the sidecar.** A no-op for any product using `[[use_test]]`: the sidecar entry carries only a name, while the check's type and SQL come from `models/test_definitions.toml`, which is not a lowering artifact and is hashed nowhere.
3. **Digesting the loader's expanded set — but never reading it.** The digest was computed at verify, persisted, and consulted by nothing. The whole suite passed with the feature inert.
4. **Comparing it, but re-reading the world afterwards.** Two loads meant the digest covered snapshot A while snapshot B executed.

What is here now: one owned snapshot, loaded once, digested over `{tests, target}`, executed without a second read. The adapter is **bound** rather than digested — digesting `rocky.toml` would make every credential rotation a custody hold whose printed remedy (restore the file) would be wrong. The digest carries a scheme tag outside the hash, so a future preimage change is re-pinnable rather than permanently stranding a generation.

Two properties stated precisely rather than assumed:

- **Verified, not approved.** No human approves check bytes; a plan payload carries operational metadata, not SQL. The strongest true statement is what the verify bundle went green on, and that is what is compared.
- **Absence is a failure.** A missing digest is a custody failure, not a pass. "Every claim matched" and "the claim I needed was made" are different questions.

## Known limitations, stated plainly

- **Observation IS now bound to the warehouse the apply used.** This was the one open decision, and it is closed — Hugo chose Option A. Apply refuses unless the config's routing identity equals the `config_identity` its plan authorised; observation now asks the same question of the config that chose its adapter, comparing the same `config_policy_identity` value rather than a second derivation that could disagree with the apply gate silently. A mismatch is a hold (`RoutingDiverged`), never a reading. Absence is a hold too. See **The routing gate** below.
- **A custody divergence holds until you restore the file you changed.** The loop cannot adopt an edit: nothing after an apply can re-verify a new set of checks. To keep a change, restore first, re-run until the loop leaves `applied`, and only then put it in the spec and approve again — approving is refused while `applied`, so the order is not optional. `rocky product compile` is **not** a remedy for either drift class; a test asserts the loop does not offer it.
- **The spec cannot express every check shape.** `output.checks` is a list of SQL boolean strings lowered to error severity, and `output.grain` lowers to unique. A warning severity, a `row_count_range`, a filter, or a `[[use_test]]` has no spec spelling, so for those the restore is the only faithful route.
- **A worker-written check is pinned, not rejected.** Removing `draft_check` from the worker profile closes the MCP route; it does not confine the worker's filesystem. A check already on disk when the plan was verified is inside the verified set and runs. The custody digest closes a window, not the content question — #1491 (OS sandbox) and #1515 (trusted custody).
- 🔴 **What the guards over served text buy, stated exactly.** The substring pins are **negative regression rules**. Each bites on a *restored* violation; **none bites on a reworded one — every one would pass a paraphrase.** Independent review confirmed there is no lexical technique that catches a reworded semantic over-claim without pretending to understand semantics, and that substring pins are the practical ceiling for automatic policing.
  So the branch adds the one non-semantic technique that does help: a **golden over the whole serialized served surface** — hashed rows, both profiles, one row per item so a diff names which surface moved, bidirectional (CHANGED / NEW / GONE). It does **not** detect false meaning. It makes any wording change impossible to land *silently*, converting an unbounded semantic problem into a bounded review one. Re-blessing is the review record; an empty or unrecognised bless value does not bless.
  It earned that twice. On its first real change it caught a default-profile description still claiming `test` runs declarative checks — a claim an earlier round had fixed on the worker half only. And the split was measured, not argued: a mutation rewording "loop" to "cycle" in two served payloads **passed both lexical guards and was caught only by the golden**, with a paired negative control against the pre-widening row confirming the widening is a real delta.
  Two deliberate scope limits, written into the code: surfaces carrying run-dependent payloads are excluded, because a digest that drifts every run gets blessed reflexively; and the golden covers what the **server** serves, not the out-of-band task brief.
- **`rocky compile` still refuses an unset credential placeholder** while the offline preview now tolerates it — **#1536**. Main's shipped behaviour, deliberately left alone rather than changed as a side effect of the preview fix. Needs a decision.
- **A malformed `surrogate_key` on an unrelated model fails a narrowed preview** — **#1537**. Inherited from `emit-sql`, the parity target the preview was matched to; the run path filters to selected models instead.
- **Declarative `expression` checks are raw-interpolated SQL** with a sandbox contract nothing implements — **#1524**. Engine-wide, deliberately out of scope.
- **Plan-id reuse after local state loss** — **#1525**. Pre-existing; the schema note that had called that state loss harmless is corrected.

## Review record

Independent adversarial review by **Codex (a different model)**, which ran its own independent red team (`gpt-5.6-terra`) in most rounds. **Eighteen rounds, every one returning BLOCK**, and each found the previous fix weaker than claimed.

Round 18 reviewed the routing gate itself and found five defects, all fixed: the gate ran **after** the freshness query, so a re-route still sent one real `SELECT MAX(...)` to the wrong warehouse and journaled the answer (the gate is now the first thing observation does); missing plan evidence printed a restore remedy that could not work at a state where approving is refused (now typed — a genuinely-legacy plan is exempt, mirroring apply's own rule, and missing-or-unreadable evidence routes to `blocked`, whose `--retry` genuinely exits); the hold's message claimed "a different warehouse" when the compared value covers every adapter and pipeline (it now states that scope); the new tests leaked artifacts into the test cwd; and no test drove the gate end to end (POC assert 12 now does — it re-routes to a **copy** of the applied warehouse, where every check would pass, so without the gate the failure mode is a green reading).

The through-line worth knowing, because it predicts where the next defect will be: **every derived rule in this crate is lexical, and lexical derivation closes the *name* axis completely and the *meaning* axis not at all.** Every meaning-axis defect on this branch was found by reading the served text, never by a test. The sharpest instance: the identifier matcher was corrected so `propose_only` would *not* match, and the authoring skill then quotes `propose_only` while handing the worker the runner's posture verbs — the lexical fix created the blind spot only reading caught.

A related pattern, five times over: **fixing an over-claim produced a new false statement pointing the other way.** The most recent was a fix that made `emit-sql` require credentials, breaking its documented no-credentials contract; and a projection built to strip false claims silently stripped a true safety warning.

Everything is fixed, filed (#1524, #1525, #1526, #1532, #1533, #1535, #1536, #1537), or listed above. One finding is deliberately **not** fixed — the first limitation.

## Evidence

All on one commit, on a tree verified quiescent before and after each run:

```
base              rebased on main (engine-v1.72.0)
toolchain         1.98.0 — the version engine-ci.yml pins, not the local default
workspace tests   PASS   6604 passed, 0 failed (fresh target dir)
clippy            PASS   --all-targets --all-features -D warnings
fmt               PASS
codegen drift     0      `just codegen` reproduces the committed bytes
fixture drift     0      `regen_fixtures.sh` reproduces the committed bytes
```

The POC's replay lane, including the new post-apply data-red:

```
[11] data-red: a violated declared check routes into a repair round behind a NEW human gate
     OK  red -> observed_failing (exit 4) -> repair round -> NEW plan (!= the prior plan id),
         bare apply refused, fresh review -> observing

Mutation pass: 16 / 16 gate mutations caught (12 asserts)
```

Two of those mutations are the ones that matter: skipping the injected duplicate proves the data-red branch is reached, and skipping the fresh review of the repaired plan proves a data-red cannot inherit the failing plan's approval.

That lane needs the POC seed fix (**#1526**) to run at all — before it, a pre-existing failure aborted the script several asserts earlier, so nothing in this lane had ever executed.

## Standing questions

**Least confident about:** whether a custody hold that can only be cleared by restoring a file is the right ergonomics, or whether it will read as the product being stuck. The behaviour is honest — the loop genuinely does not know whether the output is right when the checks changed under it — but honesty and usability are not the same thing, and this has not met a real user.

**Most important thing I am missing:** an ownership rule for post-apply model sidecars. The engine treats manifest hashes as authoritative before merge, then treats mutable disk as authoritative afterwards. This PR makes observation refuse to act on that ambiguity rather than resolving it. That is the same question the warehouse-identity gap belongs to: what, exactly, is an applied generation bound to?


hugocorreia90:
Rebased onto `main` (merge `e1e7a7f6`) and moved the F3 vocabulary to **state schema v24** — 1.73.0 released v23, and `observed_failing` is a new tagged-enum variant, which a v23 reader cannot parse.

**Merge resolution**
- `emit_sql.rs`: kept main's shipped #1624 form (`load_project_config` + local `resolve_dialect`); dropped this branch's rewire to `preview_dialect`. The branch's added test stays.
- `state.rs`: main's released v23 text kept; the F3 material is a new v24 entry; `EXPECTED_VERSION` pinned to 24; dagster fixtures recaptured (version fields only).
- `test.rs`: main's loader now takes the project `[freshness]` block. Both custody sites load with `None`; `declarative_run` keeps the inheritance through `resolve_warehouse_adapter`. A new test pins that the digest ignores the inherited block (it hashes `tests` and `target`; inheritance writes `freshness` only).
- `plan.rs`: `apply_surrogate_keys` now returns the validate-before-interpolation error; the preview propagates it like `emit_sql` and `run` do.

**Red team: Codex (independent model), round 1 — BLOCK, 2 findings, both verified and fixed in `2f7d7dc8`.**
1. The changelog said `recreate` "does a full refresh". False: it bootstraps a fresh local store and suppresses the upload (`state.rs:810-815`); the run keeps each model's strategy, so a merge model merges. Reworded.
2. "Every other command stops with a schema-mismatch error" was too strong: `rocky state show` and `rocky doctor --check state_schema` use `peek_schema_version`, which reports a newer version instead of refusing. Reworded.
Sound per the same round: the freshness-independence of the digest (A), the `?` in the preview (C), and the two dialect derivations agreeing today (D — duplication noted as maintenance risk, not a defect).

**Carried forward, not in this PR:** the released 1.73.0 entry for v22 → v23 makes the same two overclaims. Separate docs fix.

hugocorreia90:
**Codex round 2 on the corrected paragraph — BLOCK, 2 findings. Disposition (fixed in `031a49df`):**
1. "never uploads it back over the newer remote state" — *rebutted on its trigger, sentence narrowed anyway.* Codex said the transformation-pipeline arm never reads the recreate flag and so would upload a recreated store. The arm has no suppression, true — but it never recreates: `run_local.rs:153/660/1196` open with plain `StateStore::open`, not `open_with_policy`, so a v24 store makes it refuse. Its `Err` path does finalize the session, but the local file is then the v24 file it just downloaded and no write happened, so the same bytes go back. Only the replication and model-only arms (`run.rs:2901`, `:2006`, `:7805`) and `rocky load` (`load.rs:224`) follow the policy and suppress the upload. The paragraph now says exactly that.
2. `rocky doctor --check state` records the mismatch as a Warning (`doctor.rs:184/198`) and exits 0 — *confirmed.* Now: "`rocky state show` and `rocky doctor` report the mismatch instead of refusing."

Sound per the same round: recreate starts a fresh local store; execution keeps `model.config.strategy`; `state show` and `doctor --check state_schema` peek. Filed the docs-vs-code gap this exposed (which commands honour `on_schema_mismatch`) as a separate issue.

hugocorreia90:
**Codex round 3 — BLOCK, 5 findings, all on the rollback paragraph. Disposition (in `07ba61f0`):**
1. Inventory wrong (`run.rs:7805` is backfill execution via `rocky apply`, not a `rocky run` arm; load is reachable two ways) — *accepted.* The per-command inventory is out of the changelog; it lives in #1679, where it can be made exact.
2. `rocky run --pipeline <transformation> --idempotency-key` recreates through the claim's `open_with_policy` before dispatch, while the same run without the key refuses — *verified, added to #1679.* That asymmetry is a code question, not a release-note sentence.
3. `rocky backfill --model` discards the read-open error and continues — *verified, added to #1679.*
4. "Mixed-version fleet sharing one state file" — *accepted, the important one.* Remote keys are `v<schema>/…` (`state_sync.rs:141-148`), so versions never share a remote object; the hazard is a physically shared local file, which `recreate` deletes (`state.rs:1143`). The paragraph now says exactly that and nothing about fleets.
5. `rocky doctor --check state` warns and exits 0; `--check state_schema` is Critical and exits 3 — *accepted.* The note names only the `state_schema` check and its exit code.

**Carried forward:** the released 1.73.0 entry for v22 → v23 makes the same fleet claim. Separate docs fix, on `main`.

hugocorreia90:
**Codex round 4 — BLOCK, 2 precision findings; both applied in `863c995e`. This closes the loop.**
1. "The hazard is a local file" overstated: a rolled-back binary also cannot see progress recorded under the v24 key (it reads its own, possibly stale, v23 object). Now: "It reads whatever is under its own v23 key, which may be stale, and it cannot see progress recorded under v24. The incompatible-blob hazard is a local `state.redb` …".
2. `v23/…` is the object-store shape only; Valkey is `<prefix>v23:<key>` (`state_sync.rs:154`), tiered CAS `<prefix>cas:v23:<key>` (`:1952`). Now: "`v23/state.redb` on an object store, `…v23:…` on Valkey".

Confirmed by the same round: the isolation claim holds for S3/GCS (`state_sync.rs:2904`), Valkey (`:3182`) and tiered; the periodic uploader never reads remote state and writes through the same qualified legs (`:1184`, disabled under CAS at `:1124`); `recreate` deletes and reopens the local file (`state.rs:1143`); `doctor --check state_schema` is Critical and exits 3 (`doctor.rs:1003`, `:114`); deferring the command inventory to #1679 makes no unsupported claim.

Four rounds total (R1 code+prose, R2–R4 prose). Independent reviewer throughout: Codex.

PR #1555, opened 2026-08-30
fix(tools): stop reading a killed mutant as a broken build

### LINKED ISSUE #1535: mutation-check.sh reports INCONCLUSIVE when a mutation correctly kills a test

### LINKED ISSUE #1547: bug(tools): mutation-check.sh reports INCONCLUSIVE for every failing cargo test
## Description

`scripts/mutation-check.sh` reports `INCONCLUSIVE` (exit 3) for every `cargo test` invocation, including the one in its own usage docstring. A test that genuinely ran and genuinely failed under mutation — the exact evidence the script exists to produce — is misreported as "the mutation broke the build".

The build-break detector at `scripts/mutation-check.sh:139` is:

```bash
if grep -qE '^error(\[E[0-9]+\])?:|error: could not compile' "$mutated_log"; then
```

`cargo test` prints `error: test failed, to rerun pass ...` at the start of a line when a test fails. That matches `^error:`, so a real test failure is classified as a compile failure.

## Version

- `scripts/mutation-check.sh` at `54619294`

## Steps to reproduce

Any mutation that compiles and makes a `cargo test` target fail. Two hit in #1546:

```
$ scripts/mutation-check.sh engine/rocky/src/main.rs /tmp/mut.py \
    cargo test --manifest-path engine/Cargo.toml --bin rocky dag_conflicts_with_the_resume_flags

mutation-check: applied mutation to engine/rocky/src/main.rs
thread 'tests::dag_conflicts_with_the_resume_flags' panicked at rocky/src/main.rs:5033:13:
["rocky", "run", "--dag", "--resume-latest"] must be a clap conflict
error: test failed, to rerun pass `-p rocky --bin rocky`
---
mutation-check: INCONCLUSIVE — the mutation broke the build, so the test
never ran.

$ echo $?
3
```

The test ran. The test failed. The mutation compiled. The verdict should be 0.

## Expected behavior

A test that compiled, ran, and failed exits 0 (fix-sensitive). Only a genuine compile failure exits 3.

## Impact

The script is the repo's evidence gate for `AGENT_REVIEW.md`'s fix-sensitivity requirement, and it was written as a tool precisely because a written rule was not enough. Right now it cannot deliver a passing verdict for `cargo test`, which is the form its own docs recommend. The failure direction is safe — it under-reports evidence rather than inventing it — but in practice it means the verdict has to be overridden by reading the log by hand, which is the habit the tool replaced.

## Suspected cause

The pattern is too broad. `error: test failed, to rerun pass ...` and `error: test run failed` (nextest) are runner summaries, not compiler diagnostics.

## Suggested fix

Match the compiler's own shapes and exclude the runner summaries, for example:

- keep `^error\[E[0-9]+\]:` and `error: could not compile`
- drop the bare `^error:` arm, or negate `^error: test (failed|run failed)` before it

Then pin it: a mutation-check run whose command is a failing `cargo test` must exit 0, and one whose mutation does not compile must exit 3. Both are cheap to fake with a stub command.

## Existing issue / PR review

Searched open and closed issues and PRs for `mutation-check`, `INCONCLUSIVE`, and `fix-sensitive`. #1514 added the script and the branch-hygiene guards; nothing reports this false verdict.

hugocorreia90:
`scripts/mutation-check.sh` reported `INCONCLUSIVE` for every failing `cargo test` — the exact evidence it exists to produce.

## The defect

The build-break detector matched a bare `^error:`. Cargo prints its own failure summary at the start of a line:

```
error: test failed, to rerun pass `-p rocky --bin rocky`
```

So a mutation that **compiled, ran the test, and made it fail** — a killed mutant, the passing verdict — was reported as a mutation that broke the build.

```
                        wanted        got
 test ran and failed      0 (OK)      3 (INCONCLUSIVE)   <- the bug
 build broke              3           3
 test still passed        1           1
```

nextest's `error: test run failed`, and the bench and doctest summaries, all have the same shape.

## The fix

Strip the runner's own summaries, then look for compiler diagnostics in what is left. `error[E…]:`, `error: could not compile`, and uncoded compiler errors such as ``error: linking with `cc` failed`` still report `INCONCLUSIVE`. The guard against reading a broken build as evidence is unchanged.

Filtered into a file rather than piped. `set -o pipefail` is on, and a `grep -v` that filters out every line exits 1 — so a piped conditional would misread a log that was nothing but the summary.

## Evidence

New `scripts/tests/mutation-check.test.sh` — seven cases using stub commands in scratch git repos. No Rust toolchain, runs in about a second.

**Fix-sensitive, demonstrated.** The suite takes a `MUTATION_CHECK_SCRIPT` override so the check runs against a modified copy without dirtying the worktree. With the fix reverted:

```
  FAIL  cargo test failure is a killed mutant   — wanted 0, got 3
  FAIL  nextest failure is a killed mutant      — wanted 0, got 3
  FAIL  doctest failure is a killed mutant      — wanted 0, got 3
  ok    coded compiler error is inconclusive
  ok    could not compile is inconclusive
  ok    uncoded compiler error is inconclusive
  ok    surviving mutant fails the check
```

Exactly the three regression cases fail and none of the four guard cases do — so the tests catch this bug specifically, and the fix demonstrably does not weaken the build-error guard.

## Wiring

- `scripts-ci.yml` gains a job. That workflow already runs on `scripts/**`, and its shellcheck sweep now covers the new file — which is clean at **full** severity, not just the workflow's current error-severity gate.
- `just mutation-check-selftest`, next to the existing `soak-verdict-selftest` precedent.

Closes #1547
Closes #1535


hugocorreia90:
## The runner strings, measured rather than assumed

The regex was written from memory of what cargo prints. Generated the real output instead, in scratch crates, one per failure mode:

| command | line it actually prints | this change |
|---|---|---|
| `cargo test --lib` | ``error: test failed, to rerun pass `--lib` `` | stripped → verdict 0 |
| `cargo test --doc` | ``error: doctest failed, to rerun pass `--doc` `` | stripped → verdict 0 |
| `cargo nextest run` | `error: test run failed` | stripped → verdict 0 |
| a type error | `error[E0308]: mismatched types`<br>`error: could not compile \`probe\` (lib) due to 1 previous error` | **kept** → verdict 3 |

Two checks on the dangerous direction — reporting verdict 0 when the build actually broke:

- A real compile failure emits `error[E0308]:` **and** `error: could not compile`. The filter strips neither, so a broken build still reports INCONCLUSIVE.
- When several targets fail, cargo stops at the first and prints only `error: test failed, to rerun pass ...`. No separate multi-target summary line exists to miss.

**One residual, stated rather than papered over:** `error: bench failed` is in the pattern but is **not** verified — a bench harness needs nightly and I could not produce the line on stable. If the real string differs, the effect is that a failing `cargo bench` still reports INCONCLUSIVE, which is the safe direction and the behaviour on `main` today. It is not a new risk.

## Red team

Dispatched to Codex (a different model, per `AGENTS.md`). It confirmed the suite passes and that reverting the fix fails exactly the three asserted cases, then **timed out at 10 minutes without delivering a verdict** — it was part-way through probing genuine cargo/nextest output, which is the table above. Recording that rather than claiming a review that did not complete. Re-dispatched with a narrower scope.

hugocorreia90:
## Red team: BLOCK, confirmed and fixed in `00074b34`

Reviewer: **Codex** (`codex-rescue`, a different model; its finding independently confirmed by `gpt-5.6-terra`). It found a **fail-open** path — verdict 0, the *passing* verdict, for a mutation that broke the build. That is the one direction that matters, and I introduced it.

### The defect

`set -u` does not catch a failed `mktemp`: the variable ends up **set and empty**, not unset.

```
mktemp fails  ->  filtered_log=""
              ->  > ""   ambiguous redirect, writes nothing
              ->  || true swallows the failure
              ->  grep finds nothing in a file never written
              ->  "OK — the test failed under mutation"   exit 0
```

The EXIT trap is installed *after* those calls, so nothing cleaned up behind it either.

### Reproduced, not just reasoned about

A stub `mktemp` that succeeds twice and fails on the third call, with the command reporting a real `error[E0308]` plus `error: could not compile`:

```
against d43f27e7    a failed mktemp — wanted exit 2 or 3, got 0     <- the bug
against 00074b34    a failed mktemp refuses rather than reporting
                    success (exit 2)                                 <- fixed
```

### The fix — both halves fail closed

- Every `mktemp` is checked for **failure and for an empty path**, and refuses with exit 2.
- The filter keeps grep's exit 1 (every line was a runner summary — normal) but treats **2 or above** as `INCONCLUSIVE`. A log that could not be searched yields no verdict rather than a passing one.

The suite is now 8 cases. The eighth pins this exact trace.

### Codex's other finding, recorded

It found **no** ordinary cargo build-break log where the new `grep -v` strips every surviving error marker — the fail-open was via the `mktemp`/redirect route, not the pattern being too broad. That matches the runner strings I measured in the comment above, so two independent methods agree the pattern itself is sound.

### One note on process

The first Codex dispatch **timed out at 10 minutes with no verdict**, part-way through probing real cargo output. I said so rather than treating silence as approval, measured that part myself, and re-dispatched with a narrower scope. The narrow run returned this BLOCK in four minutes.

PR #1558, opened 2026-08-30
fix(engine): compile offline without warehouse credentials

hugocorreia90:
`rocky compile` refused to run when a warehouse credential was unset, even though compiling never connects to a warehouse.

```
 before   emit-sql   renders SQL offline, exit 0
          compile    refuses,             exit 1   <- same project, same unset ${VAR}
 after    both compile offline; validate still insists every variable resolves
```

## Scope — tolerance is field-scoped, not global

This is the part the first draft got wrong (see the review below). An unresolved `${VAR}` is tolerated **only inside `[adapters.<name>]`, and not on its `type` or `kind` keys**:

```
 [adapters.wh] host / token / password / ...   unset ${VAR} tolerated
 [adapters.wh] type / kind                     REFUSED — selects behaviour
 [imports] path / snapshot / pin               REFUSED
 [state], [policy], [cache], everything else   REFUSED
```

The check walks the normalized TOML tree so it knows the key path, and matches against the variable names that **actually went unresolved** rather than the `${` shape — so a config whose value legitimately contains that text is not refused.

Beyond that one rule nothing changes: the TOML must parse, the whole `CONFIG_VALIDATORS` chain still runs, and `rocky validate` keeps the strict loader.

## Verified end to end

Against a binary built from this branch, with the sha checked identical before and after the build:

```
credential field unset                TOLERATE   exit 0
adapter type unset                    REFUSE     exit 1
[imports] path unset                  REFUSE     exit 1
[adapter] shorthand credential        TOLERATE   exit 0
malformed TOML                        REFUSE     exit 1
validator failure                     REFUSE     exit 1
rocky validate, credential unset      REFUSE     exit 1
```

248 config tests and 15 env-var tests pass.

## Review history — the first draft was wrong

**Codex** (a different model) blocked the first version. Its central point:

> no allowlist exists distinguishing "credential" env vars from structural config fields — so "credential tolerant" is currently **unrestricted placeholder tolerance**

Correct. Two confirmed consequences, both fixed by the scoping above:

- **`[imports]` enforcement was silently skipped.** An unresolved `path`/`snapshot` makes the snapshot read fail, and `load_failure_diagnostic` (`imports_check.rs:329`) classifies that as a **W012 warning**. Compile reported `has_errors = false` while E030–E033 pin and compatibility checks never ran.
- **An unresolved adapter `type` was accepted**, because the validator chain skips adapter types it does not recognise (`config.rs:5615`).

Codex cleared the strict path: same first-missing name, same byte span, unchanged across no-vars, all-set, `${VAR:-default}`, comments, first/last-missing and missing-at-EOF.

## Still blocked on #1559

Codex found a third defect that scoping does **not** fix, because it triggers on a legitimately-tolerated field. `draft_model` writes files, compiles, then gates — and the gate does `load_rocky_config(...).ok()` (`apply.rs:1610`), so a config that fails to load reads as `PolicyGate::NotConfigured`, which `tools.rs:2474` matches in the **same arm as `Allow`**. A configured `deny` stops denying.

That is **pre-existing on `main`** — both the `.ok()` and the shared arm predate this PR — so it is filed as **#1559**. But this change makes it easier to reach, so this PR stays a draft until #1559 lands.

## One assumption the tests refuted

The `[adapter]` shorthand test failed on the first run. `normalize_toml_shorthands` only wraps a bare `[adapter]` into `[adapter.default]` — it does **not** rename the key to `adapters`; serde does that later, on a tree this check never sees. Both spellings are now matched.

Refs #1536
Blocked by #1559


hugocorreia90:
## Blast radius, traced

The risk with tolerance is a literal `"${DATABRICKS_HOST}"` escaping the offline boundary and reaching something that connects or writes. Traced every use of the tolerantly-loaded config.

**Exactly one tolerant call site exists in the tree:**

```
engine/crates/rocky-cli/src/commands/compile.rs:121
```

`run` and `apply` have zero. Every executing path still uses the strict loader, so a config that never resolved its variables cannot reach a warehouse.

**What `project_config` feeds inside `compile.rs`** — all five uses:

| line | use | can it connect or write? |
|---|---|---|
| 141 | `config.cache.schemas` → `load_cached_source_schemas(&schema_cfg, state_path)` | no — only `enabled` + `ttl_seconds` |
| 163 | `[mask]`, `[classifications]`, `[freshness]` | no — diagnostic data |
| 195 | `[portability]` | no — lint config |
| 287 | `[imports]` → `imports_diagnostics` | no — reads, produces diagnostics |

The one that mattered: **`state_path` is a separate parameter to `compile_output`, not derived from the config.** A literal placeholder in `[state]` therefore cannot become the state DB path. That was the defect worth worrying about, and it is not reachable.

## One consequence, stated rather than left to be found

`compile_output` has six callers, and one is the fulfillment loop's verify leg (`rocky-fulfill/src/step.rs:466`). So a project with unset credentials now compiles **green** in verify where it used to go red.

That is intended, not an oversight. Compile was never the right gate for credentials — `apply` is, and it still loads strictly and still refuses. The effect is that such a project fails later, at the step that actually needs the credential, instead of failing at a step that does not.

Worth knowing if you are reading loop behaviour, so it is on the record here.

hugocorreia90:
## Red team: BLOCK. Converted to draft — this must not merge as written.

Reviewer: **Codex** (`codex-rescue`, a different model). Its central finding reframes the change, and it is right:

> no allowlist exists distinguishing "credential" env vars from structural config fields (paths, adapter types, import pins) — so "credential tolerant" is currently **unrestricted placeholder tolerance**

I named it credential tolerance and implemented placeholder tolerance. Everything below follows from that gap. All three defects verified directly in the code before accepting them.

### 1. Import enforcement silently skipped — CONFIRMED

```toml
[imports.orders]
path = "${IMPORT_DIR}"
snapshot = "${SNAPSHOT}"
pin = "deadbeef"
```

Unset, those stay literal, the snapshot read fails, and `load_failure_diagnostic` (`imports_check.rs:329`) classifies a missing or unreadable file as **W012, a warning**. Compile reports `has_errors = false` and E030–E033 pin/compatibility enforcement never runs. A pin that looks enforced and checks nothing.

### 2. A deny policy stops denying — CONFIRMED, and filed separately as #1559

`draft_model` writes files, then compiles, then gates. The gate does `load_rocky_config(config_path).ok()` (`apply.rs:1610`), so a config that fails to load yields `None` → `PolicyGate::NotConfigured` — and `tools.rs:2474` matches `NotConfigured | Allow` in the same arm and defuses the rollback.

```
 configured deny -> config load fails -> .ok() -> None
                                            -> NotConfigured
                                            -> same arm as Allow -> write persists
```

**Pre-existing on `main`** — both the `.ok()` and the shared arm predate this PR — so it is #1559 rather than a fix here. But this PR makes it reachable, because strict loading used to stop first. **#1558 should not merge before #1559 is settled.**

### 3. Validators are not placeholder-aware — CONFIRMED

`type = "${ADAPTER_TYPE}"` unset is accepted: the validator chain skips unknown adapter types (`config.rs:5615`). Typed numbers and enums still fail at deserialization; string fields survive.

### 4. "Compile is offline" is not strictly true

`--with-seed` executes `data/seed.sql` through DuckDB `execute_batch`, and the MCP compile **auto-enables it** when a seed exists (`tools.rs:817`). `COPY ... TO` can write files. Strict config loading used to stop that first for a project with unset variables. This undercuts my justification as written, and the PR description needs correcting whatever the fix turns out to be.

### What Codex cleared

Existing strict callers are unaffected: same first-missing name, same byte span, same behaviour across no-vars, all-set, `${VAR:-default}`, a var in a comment, first-missing, last-missing and missing-at-EOF. The unconditional tail copy is unobservable on the strict error path. `rocky validate` stays strict.

That matches my own trace, which found exactly one tolerant call site in the tree and `run`/`apply` untouched.

## The fix

Tolerance has to be **field-scoped, not global**: unresolved placeholders permitted only inside `[adapters.<name>]`, and not on its `type` or `kind` keys. `[imports]`, `[state]`, `[policy]`, `[cache]` and every pin go back to refusing. That kills 1 and 3 at the root rather than patching each downstream consumer.

2 is not fixed by narrowing — `host` is a field that legitimately stays tolerant — so #1559 has to land as well.

Reworking. Clippy and Format also failed here and will be fixed in the same pass.

PR #1560, opened 2026-08-30
fix(engine): refuse when the policy cannot be read, instead of allowing

### LINKED ISSUE #1559: security: a policy that fails to load is treated as no policy, so a deny policy stops denying
The apply policy gate swallows config load errors. A `[policy]` block that cannot be loaded becomes `NotConfigured`, and `NotConfigured` is handled the same as `Allow`. A configured **deny** therefore stops denying.

## The chain

```rust
// engine/crates/rocky-cli/src/commands/apply.rs:1610
let policy = rocky_core::config::load_rocky_config(config_path)
    .ok()                       // <- ANY load error becomes None
    .and_then(|cfg| cfg.policy);
```

`None` yields `PolicyGate::NotConfigured`. Every consumer then treats that as permission:

```rust
// engine/crates/rocky-mcp/src/tools.rs:2474
match gate {
    PolicyGate::NotConfigured | PolicyGate::Allow => {
        rollback.defuse();      // <- the agent's writes persist
```

So "the policy said no" and "we could not read the policy" are indistinguishable, and both mean yes.

```
 configured deny  ->  config load fails  ->  .ok() -> None
                                                |
                                                v
                                        NotConfigured
                                                |
                                                v
                                     same arm as Allow -> write persists
```

## Reaching it

Any config that parses at the file level but fails `load_rocky_config` will do. An unset `${VAR}` is the easy one:

```toml
[adapter.wh]
type = "databricks"
host = "${DATABRICKS_HOST}"
token = "unused"

[policy]
version = 1
default_agent_effect = "deny"
```

Start the MCP server with `DATABRICKS_HOST` unset and call `draft_model({"name":"blocked","sql":"SELECT 1","intent":"proof"})`. The deny is read as unconfigured and `models/blocked.sql` plus its sidecar persist.

`draft_contract`, `draft_check` and `draft_metadata` share the write-then-compile-then-gate shape (`tools.rs:2570-2596`, `2694-2732`, `2910-2932`).

## Why it matters

This is the failure direction a policy plane exists to prevent. An operator who configures `default_agent_effect = "deny"` has said the agent may not write. A typo in an unrelated env var silently revokes that, with no diagnostic — the gate reports "not configured" for a policy that is plainly configured, sitting in the same file it just failed to read.

`Unknown at a gate is fail-open` is a shape this repo has hit before.

## Suggested fix

Two distinct changes, both needed:

1. **Do not swallow the error.** `apply.rs:1610` must distinguish "no `[policy]` block" from "the config could not be read". The second is not an answer and must not be reported as one.
2. **Give the gate a third outcome.** `NotConfigured` and `Unreadable` must not share an arm with `Allow`. An unreadable policy should refuse, and say why.

Worth checking every `PolicyGate` match arm at the same time: any that groups `NotConfigured` with `Allow` inherits this.

## Provenance

Found by independent adversarial review (Codex, a different model) while reviewing #1558, and verified directly against both call sites before filing. **Pre-existing on `main`** — both the `.ok()` and the shared match arm predate that PR. It is filed separately because it is not that change's defect, but #1558 makes it easier to reach and should not merge before this is settled.

hugocorreia90:
A `[policy]` block that could not be **loaded** was reported as "no policy configured", and every consumer treated that the same as `Allow`. A configured **deny** therefore stopped denying.

```rust
// engine/crates/rocky-cli/src/commands/apply.rs
let policy = load_rocky_config(config_path)
    .ok()                      // <- ANY load error becomes None
    .and_then(|cfg| cfg.policy);
```

`None` yields `PolicyGate::NotConfigured`, and the MCP draft tools match `NotConfigured | Allow` in one arm and defuse the rollback — so an agent write a deny rule forbids was **kept on disk**.

```
 configured deny -> config fails to load -> .ok() -> None
                                         -> NotConfigured
                                         -> same arm as Allow -> write persists
```

Reaching it needs only a config that parses but does not load. An unset `${VAR}` will do.

## The fix already existed, at one seam

`rocky gc` had the right shape (`gc.rs:1661`) with a comment saying exactly why:

> a config-load ERROR would otherwise silently unenforce a possibly-configured `[policy]` block. Bail instead. A genuinely-missing config file keeps the NotConfigured posture.

That reasoning was never applied to the shared gate. Both `.ok()` sites now go through one helper with gc's semantics:

| config state | before | after |
|---|---|---|
| no file at all | `NotConfigured` → proceed | **unchanged** — `NotConfigured` → proceed |
| exists, fails to load | `NotConfigured` → **proceed** | `Unloadable` → **refuse** |
| loads, no `[policy]` | `NotConfigured` → proceed | unchanged |

The standalone `rocky compile --models …` case with no `rocky.toml` is deliberately untouched.

## Why a new variant, not a bool

`PolicyGate::Unloadable` makes the compiler name every consumer, because the enum has no `_` arms. It found seven: two apply seams, the fulfill propose seam, and all four MCP draft tools. Every one now refuses.

The draft tools deliberately do **not** defuse the rollback, so the artifact is removed — matching the existing `Deny` arm.

`Unloadable` also ranks above `Deny` in `gate_rank`. An unreadable config might have denied, and the aggregate keeps the most restrictive outcome, so a readable `allow` on another model must not out-vote it.

## Two comments that documented the bug as a feature

`rocky-mcp` carried these, both now corrected:

> an unloadable config resolves to `NotConfigured` in the gate, so it reads no markers either

## Evidence

Three tests: an unreadable config is `Unloadable` and its apply refuses; a genuinely absent config is still `NotConfigured` and still proceeds; `Unloadable` outranks every other gate.

**Fix-sensitive**, checked with `scripts/mutation-check.sh`. Reverting both loads to `.ok()` fails exactly `an_unreadable_config_is_unloadable_not_notconfigured`.

111 apply tests and 32 MCP tests pass. Clippy and fmt clean across the workspace.

## Provenance

Found by independent adversarial review (Codex, a different model) while reviewing #1558, and verified against both call sites before filing. Pre-existing on `main` — the `.ok()` and the shared match arm both predate that PR. **#1558 is blocked on this.**

Closes #1559

### DIFF
--- a/engine/crates/rocky-cli/src/commands/apply.rs
+++ b/engine/crates/rocky-cli/src/commands/apply.rs
@@ -1148,6 +1148,18 @@ pub enum PolicyGate {
     /// (AiAuthored → require a review marker; Run/Promote → ungated), so
     /// absent-`[policy]` behaviour is byte-identical to today.
     NotConfigured,
+    /// The config could not be READ, so whether a `[policy]` block exists is
+    /// unknown. Distinct from [`PolicyGate::NotConfigured`] on purpose: "there
+    /// is no policy" and "we could not find out" are different answers, and
+    /// only the first is permission.
+    ///
+    /// Every consumer must refuse on this. Collapsing it into the
+    /// `NotConfigured | Allow` arm is what let a configured `deny` stop denying
+    /// when an unrelated `${VAR}` was unset (#1559).
+    Unloadable {
+        /// Why the config could not be read, for the refusal message.
+        reason: String,
+    },
     /// Every touched model resolved to `allow`. Proceed without a marker.
     Allow,
     /// The most-restrictive effect is `require_review`. A human review marker
@@ -1508,9 +1520,10 @@ pub fn evaluate_apply_policy(
     // the pre-gate sync decision, so the sync-guard and this gate can never
     // disagree about whether `[policy]` is configured (finding A — config-snapshot
     // TOCTOU).
-    let policy = rocky_core::config::load_rocky_config(config_path)
-        .ok()
-        .and_then(|cfg| cfg.policy);
+    let policy = match load_policy_for_gate(config_path) {
+        Ok(policy) => policy,
+        Err(reason) => return PolicyGate::Unloadable { reason },
+    };
     evaluate_apply_policy_with_policy(
         policy.as_ref(),
         plan_id,
@@ -1581,6 +1594,28 @@ pub(crate) fn evaluate_apply_policy_with_policy_matching(
     )
 }
 
+/// Load the `[policy]` block for a gate decision, distinguishing "no config
+/// file" from "the config could not be read".
+///
+/// `Ok(None)` means there is genuinely no project config — the standalone case,
+/// which keeps the pre-policy-plane posture. `Err` means the file exists and
+/// did not load, so any `[policy]` block in it is unknown and unenforceable.
+///
+/// This is the shape `rocky gc` already uses at its own seam (`gc.rs:1661`),
+/// with the comment that says why: a config-load error "would otherwise
+/// silently unenforce a possibly-configured `[policy]` block". That reasoning
+/// was never applied to the shared gate, which swallowed the error with `.ok()`
+/// and reported `NotConfigured` — indistinguishable from permission (#1559).
+fn load_policy_for_gate(
+    config_path: &Path,
+) -> Result<Option<rocky_core::config::PolicyConfig>, String> {
+    match rocky_core::config::load_rocky_config(config_path) {
+        Ok(cfg) => Ok(cfg.policy),
+        Err(rocky_core::config::ConfigError::FileNotFound { .. }) => Ok(None),
+        Err(e) => Err(format!("{e}")),
+    }
+}
+
 /// [`evaluate_apply_policy`] evaluated over BOTH the pre-image and the
 /// post-image classification state, returning the most restrictive verdict.
 ///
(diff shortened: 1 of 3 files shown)


hugocorreia90:
## Red team: BLOCK, and the finding was worse than a missed case — fixed in `a1f3c8d0`

Reviewer: **Codex** (a different model; a second reviewer, `gpt-5.6-terra`, confirmed the first two findings).

### 1. The fulfill arm I added was UNREACHABLE — confirmed

A **third** `.ok()` survived at `fulfill_api.rs:381`:

```rust
let cfg = rocky_core::config::load_rocky_config(config_path).ok();
```

So an unreadable config still became `None`, the gate saw no `[policy]`, and it returned `NotConfigured` — never `Unloadable`. The refusing arm beside it could not fire.

That is worse than not adding the arm: the path *looked* defended. And the arm cannot catch it by construction — `evaluate_apply_policy_with_policy` is **handed** the policy and never loads a config, so it can never report `Unloadable` itself. The refusal has to happen at the read.

The comment above that line already promised to "load the config ONCE and thread that snapshot" into both gates. It did thread one snapshot; it just threw the load error away.

**Fixed.** `load_policy_for_gate` generalized to `load_config_for_gate` so the whole config gets one fail-closed read. The new `ProposeError::PolicyUnreadable` made the compiler route the refusal through the MCP propose tool, where it maps to `policy_denied` — a refusal, not a transient fault to retry.

I also swept the other three config loads in that file rather than fixing only the one named: `:204` degrades to a cold schema cache (not security-relevant), `:576` and `:676` already propagate with `?`. Only `:381` was fail-open.

**Test, and it is fix-sensitive.** A propose against a config that parses but does not load refuses with `PolicyUnreadable` and persists no plan. Restoring the `.ok()` fails it. Its doc comment records *why* it cannot go through the gate arm, so the inert shape is not reintroduced.

### 2. `FileNotFound` covers a dangling symlink and a deletion race — accepted, not fixed here

Every `ErrorKind::NotFound` maps to `ConfigError::FileNotFound` (`config.rs:5875`), so deleting a symlink target between compile and gate yields `NotConfigured`. Real, and a genuine TOCTOU.

Not fixed in this PR, deliberately: distinguishing "absent" from "removed underneath us" needs one config snapshot carried from the first security-relevant read through enforcement — which is the reviewer's own "most important missing" point and a larger change than this fix. Recording it rather than pretending the window is closed. The exploit requires the ability to mutate the project directory mid-call, which is already a larger capability than the gate defends against.

### 3. The rollback discards its filesystem errors — filed as #1561

`restore_or_remove` does `let _ = std::fs::remove_file(path)` (`tools.rs:4748`), so "a denied draft leaves NO file" is not guaranteed. **Pre-existing** — it affects the existing `Deny` arm too, so it is #1561 rather than this PR.

What I did change here: my four new messages said "The draft was not kept." They now say "The draft was rolled back" — the action Rocky takes, not a filesystem end state it cannot promise.

---

Clippy and fmt clean across the workspace. 111 apply, 5 fulfill_api, and 32 MCP tests pass.

PR #1562, opened 2026-08-30
chore(ci): bump taiki-e/install-action and its allow-list entry together

hugocorreia90:
Supersedes #1529, which could not pass its own security gate.

## Why Dependabot cannot land this alone

Dependabot bumps the `uses:` pin but not the **allow-list entry** the credential-containment policy checks against, so #1529 failed with:

```
.github/workflows/engine-release.yml: release step action source is not allow-listed
credential-containment policy: 1 violation(s)
```

The allow-list in `check_credential_containment.py` duplicates the SHA **on purpose** — it is the trust anchor. Deriving it from the workflow would make it constrain nothing. So the two have to move together, and Dependabot only knows about one of them.

All three sites move here:

```
 .github/workflows/engine-release.yml       uses: taiki-e/install-action@<sha>
 .github/workflows/engine-ci.yml            uses: taiki-e/install-action@<sha>
 .github/scripts/check_credential_containment.py   the allow-list entry
```

Verified the old SHA `82cd3e76` appears nowhere under `.github/` afterwards, and `check_credential_containment.py` passes locally.

## The pin was verified, not taken on trust

Rather than accepting Dependabot's `# v2.86.8` comment, checked it upstream:

```
$ gh api repos/taiki-e/install-action/git/ref/tags/v2.86.8
commit fcf5432d9f50d67e37ee6e29bdb7a224ff67b4a7
```

A lightweight tag pointing straight at that commit — which is the SHA being pinned.

## Swept for the same drift elsewhere

Compared all 15 allow-listed pins against the 16 distinct pins used across the workflows:

- allow-listed but unused by any workflow: **none**
- any action pinned at two different SHAs across the allow-list and the workflows: **none**

So this was the only drift.

Closes #1529

### DIFF
--- a/.github/scripts/check_credential_containment.py
+++ b/.github/scripts/check_credential_containment.py
@@ -61,7 +61,7 @@ GH_RELEASE_SOURCE = (
     "softprops/action-gh-release@3d0d9888cb7fd7b750713d6e236d1fcb99157228"
 )
 TAIKI_INSTALL_SOURCE = (
-    "taiki-e/install-action@82cd3e7658a6f96c86c0234aeeda1748937cb0a1"
+    "taiki-e/install-action@fcf5432d9f50d67e37ee6e29bdb7a224ff67b4a7"
 )
 # Actions a job may run when it is not reachable from a pull request. Release,
 # publish and deployment workflows never execute candidate code, so the
--- a/.github/workflows/engine-ci.yml
+++ b/.github/workflows/engine-ci.yml
@@ -83,7 +83,7 @@ jobs:
           sudo fallocate -l 4G /mnt/swapfile && sudo chmod 600 /mnt/swapfile && sudo mkswap /mnt/swapfile && sudo swapon /mnt/swapfile || true
       - run: sudo apt-get update && sudo apt-get install -y cmake
       - name: Install cargo-nextest
-        uses: taiki-e/install-action@82cd3e7658a6f96c86c0234aeeda1748937cb0a1 # v2.85.13
+        uses: taiki-e/install-action@fcf5432d9f50d67e37ee6e29bdb7a224ff67b4a7 # v2.86.8
         with:
           tool: nextest
       # nextest runs all test binaries in one global scheduler instead of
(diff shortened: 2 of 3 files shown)


hugocorreia90:
## The red `Credential containment policy` is EXPECTED here, and the procedure says so

This is a **frozen trust-root update**. `.github/SECURITY_ENVIRONMENTS.md:164` documents the bootstrap, and step 2 is literally "expect the credential-containment check to reject the changed frozen file."

Two violations are reported, and both are the design working:

```
check_credential_containment.py: frozen trust root differs from trusted base
engine-release.yml: release step action source is not allow-listed
```

The first is the freeze itself — the checker is in `FROZEN_TRUST_ROOTS` (`:164`), so any PR changing it fails by construction. That is correct: a security checker must not be editable by a pull request that the checker gates.

The second follows from the first. The gate runs **`main`'s** checker (`trusted/...`) against this candidate, and main's copy still carries the old allow-list. So an allow-listed action can never be bumped by an ordinary PR — the pin and its trust anchor have to cross the boundary together, under human authority.

## Step 2's verification, run against this exact SHA (`c5594f6a`)

```
policy-test suite   110 tests, OK
checker vs tree     credential-containment policy: passed
actionlint          clean
diff                3 files, 3 insertions, 3 deletions
```

Working tree clean; sha unchanged before and after.

## Step 3 — no candidate code executes before credentials

The change adds no step, job, trigger or import. It swaps one 40-hex SHA in three places:

```
 check_credential_containment.py   TAIKI_INSTALL_SOURCE constant
 engine-ci.yml                     uses:
 engine-release.yml                uses:
```

Nothing else in the diff. `test_credential_containment.py` — also a frozen trust root — is untouched.

## What is left, and it is not mine to do

Steps 4 and 5 need an authorized maintainer:

4. Merge this reviewed SHA with the narrow branch-protection override. **Do not** disable the policy or broaden the allow-list to get around it — the allow-list entry here is a one-for-one replacement, not an addition.
5. The confirmation run is **the next PR opened after this merges**, and it must not itself touch a trust root. `ci-security-policy.yml` only triggers on `pull_request_target`, so main cannot be re-checked on demand. #1553 or #1556 would serve — neither touches a frozen file.

The pin was verified against upstream rather than taken from Dependabot's comment: `v2.86.8` is a lightweight tag pointing directly at `fcf5432d9f50d67e37ee6e29bdb7a224ff67b4a7`.

hugocorreia90:
Superseded by #1588, which carries the newer taiki-e pin (v2.87.1), the action-gh-release bump from #1585, and both allow-list entries in one trust-root update.

PR #1564, opened 2026-08-30
fix(engine): refuse a default-plan flag when a plan subcommand follows

### LINKED ISSUE #1550: bug(engine): plan's default-subcommand args are accepted and silently dropped when a subcommand follows
## Description

`rocky plan` carries a large set of default-subcommand arguments — `--filter`, `--pipeline`, `--model`, `--models`, `--all`, `--resume`, `--resume-latest`, `--shadow`, `--shadow-suffix`, `--shadow-schema`, `--branch`, the partition selectors, `--dag`, `--idempotency-key`, `--env` and more. Each is declared `global = false` with a doc comment saying "Applies to the default plan subcommand only."

Clap accepts every one of them when a **subcommand** follows, and the subcommand dispatch ignores them all. Nothing warns, and nothing refuses.

This is not resume-specific. `--resume-latest` is just how it was noticed.

## Version

- Reproduced on `54619294` (main) with a debug build.

## Steps to reproduce

```shell
$ rocky plan --resume-latest promote branch-x
Error: branch 'branch-x' not found — see 'rocky branch list'
```

The parse succeeded and the promote dispatch ran. `--resume-latest` was accepted and discarded. Any of the other default-plan flags behaves the same way — `rocky plan --shadow promote <branch>` and `rocky plan --filter tenant=acme promote <branch>` both parse and both ignore the flag.

## Expected behavior

Supplying a default-plan argument together with a `plan` subcommand is operator error and should be refused at parse time, naming the flag. The doc comments already state the contract; the parser does not enforce it.

## Actual behavior

`engine/rocky/src/main.rs` (the `Command::Plan` variant, roughly lines 760-880) declares these args on the parent. The `Some(PlanSubcommand)` dispatch arm reads none of them.

## Impact

Low blast radius per invocation, but it is a whole class of silently-dropped flags. `rocky plan --shadow promote <branch>` reads as "promote in shadow mode" and does the production thing. That is the #1272 shape — a routing flag accepted and ignored — which the repo already treats as a defect worth refusing rather than tolerating.

## Suspected cause

The default-plan args and the subcommands share one clap variant. `global = false` stops them being *inherited* by subcommands; it does not stop them being *accepted* before one.

## Acceptance criteria

- A default-plan argument supplied alongside any `plan` subcommand is rejected at parse time, naming the argument.
- The fix is generic over the whole argument set, not per-flag.
- An order-sensitive parser test covers at least one routing flag and one selection flag.

## Existing issue / PR review

Searched issues and PRs for `plan subcommand`, `global = false`, `PlanSubcommand`, and the individual flag names. #1272 covers `--shadow` accepted-and-ignored on pipeline kinds that do not route, and #1173 covers `--model` + `--dag` on a persisted plan — the same class, different surface. Nothing covers the parent-flag-before-subcommand shape.

Found during an adversarial review of #1546 (Codex / GPT-5.6), then reproduced against a built binary.

hugocorreia90:
`rocky plan --resume-latest promote <branch>` parsed, ran the promote, and threw `--resume-latest` away without a word.

```
 before   $ rocky plan --resume-latest promote branch-x
          Error: branch 'branch-x' not found        <- parse OK, flag discarded

 after    $ rocky plan --resume-latest promote branch-x
          error: the subcommand 'promote' cannot be used with '--resume-latest'
```

The same held for every default-plan flag: `--filter`, `--pipeline`, `--model`, `--models`, `--all`, `--resume`, `--shadow` and its pair, `--branch`, the partition selectors, `--dag`, `--idempotency-key`, `--env`.

Each is declared `global = false` with a doc comment saying *"Applies to the default plan subcommand only"*, and the `Some(PlanSubcommand)` dispatch reads none of them. The contract was written down and never enforced — so supplying one read as "the flag applied."

## The fix is one attribute, and it is the in-repo precedent

`args_conflicts_with_subcommands` makes clap refuse the combination at parse time and name the offending flag. `rocky fulfill` already carried it.

## Complete, not a spot fix

Swept every command in the CLI for the same shape — an optional subcommand plus its own args. There are exactly two:

```
 Fulfill    own_args=  1   args_conflicts_with_subcommands = YES
 Plan       own_args= 24   args_conflicts_with_subcommands = no
```

So `plan` was the only gap.

## Evidence

Six flag/subcommand pairs must fail to parse, plus **both** over-restriction directions:

- the same flags still parse on the bare default path, where they are the whole point — verified `rocky plan --shadow` yields `subcommand: None, shadow: true`
- `rocky plan promote <branch>` with no default-plan flag is untouched

Checked against a built binary too: the three reported spellings refuse with exit 2, while `plan --shadow`, `plan --filter …`, `plan --resume-latest` and bare `plan` all still parse and reach execution.

**Fix-sensitive**, checked with `scripts/mutation-check.sh` — removing the attribute fails the test.

28 binary tests pass; clippy and fmt clean.

Closes #1550


hugocorreia90:
## Over-restriction check, against a built binary

The real risk with `args_conflicts_with_subcommands` is breaking a legitimate invocation. Tested rather than assumed.

**Global flags still work alongside a plan subcommand:**

```
plan promote b --output json     parsed
-o json plan promote b           parsed
-c rocky.toml plan promote b     parsed
--state-path X plan promote b    parsed
```

**The flag is not lost — it moves to where it is actually read.** `promote` has its *own* `--filter`, `--pipeline`, `--models` and `--cache-ttl`:

```
 rocky plan --filter t=acme promote b     now REFUSED (parent's flag, never read)
 rocky plan promote b --filter t=acme     parses    (promote's own flag, read)
```

That is the point of the change. Before, the first spelling was accepted and discarded, so an operator got the *promote* behaviour with their filter silently ignored. Now the parser says where the flag belongs.

**One apparent failure that is not mine:** `-v plan promote b` is refused with `unexpected argument '-v' found`. Checked against a binary built from `main` — it is refused there too. `-v` is not a flag this CLI has.

**Positive controls, all still parsing:** `plan --shadow`, `plan --filter …`, `plan --resume-latest`, bare `plan`, and `plan promote b`.

hugocorreia90:
## Red team: BLOCK, confirmed and fixed in `827ef28c`

Reviewer: **Codex** (a different model; `gpt-5.6-terra` confirmed). It found that the fix rejected legitimate invocations:

```
rocky plan --output json promote b
rocky plan --principal agent promote b
```

Both flags are global and both are consumed by the promote path. `args_conflicts_with_subcommands` conflicts **inherited globals** as well when they appear before the subcommand — not only the command's own args.

Verified against a binary built from `main`:

```
argv                                 MAIN       previous fix
plan --output json promote b         parsed     REFUSED     <- regression
plan --principal agent promote b     parsed     REFUSED     <- regression
```

### My probe missed the one position that breaks

I tested the flag **after** the subcommand (`plan promote b --output json`) and **before** `plan` (`-o json plan promote b`). Both still work. I never tested it **between** `plan` and `promote`, which is the only ordering the attribute breaks. The reviewer's "most important missing" was exactly that: positive parser coverage for a global before the subcommand. It is now a test.

### The fix

The attribute is reverted. In its place, a guard at the head of the plan dispatch that checks **only the flags `plan` itself declares**:

```
argv                                 MAIN       now
plan --output json promote b         parsed     parsed      <- restored
plan --principal agent promote b     parsed     parsed      <- restored
plan --resume-latest promote b       parsed     REFUSED     <- the bug
plan --shadow promote b              parsed     REFUSED
plan --filter t=a promote b          parsed     REFUSED
plan promote b --filter t=a          parsed     parsed
plan --shadow  (default path)        parsed     parsed
plan promote b                       parsed     parsed
```

It sits before the `match` because the `Promote` pattern rebinds `base`, shadowing the outer value inside that arm.

Detection is by value rather than clap's `ValueSource`. The three flags with a `default_value` are compared against that default, so passing one explicitly *at* its default is not reported — an invocation indistinguishable from omitting it and identical in effect. Stated in the helper's doc rather than left implicit.

### Tests

The old parse-conflict assertions are gone; they no longer describe the behaviour. Replaced with:

- **every `global = false` flag on `Plan` appears in the guard array**, read out of this file's own source. A flag added later cannot be silently unguarded — which is the recurrence risk, since the guard is an explicit list. It refuses to pass vacuously if the scan finds too few flags.
- **a global before a plan subcommand still parses** — the coverage whose absence let this through.
- the helper reports the first supplied flag and nothing otherwise.

**Fix-sensitive:** deleting one flag from the guard array fails the coverage test.

30 binary tests pass; clippy and fmt clean.

PR #1571, opened 2026-08-30
fix(ci): make required engine and codegen contexts always report

### LINKED ISSUE #1563: ci: four required checks are path-filtered on engine/**, so every non-engine PR is permanently unmergeable
A pull request that touches no `engine/**` path can never merge. Four required status checks never run, and a check that never runs is never satisfied.

## Measured, not inferred

Required contexts on the `Protect main` ruleset (from `gh api repos/rocky-data/rocky/rulesets`):

```
Adapter boundary
Check codegen is up to date
Clippy
Credential containment policy
Credential-containment regression tests
Format
Test
```

What #1555 — a `scripts/` + `justfile` + one workflow change — actually reported:

```
required but ABSENT:
  Adapter boundary
  Clippy
  Format
  Test
```

Result: `mergeStateStatus: BLOCKED` with **zero failures and zero pending**. The same holds for #1552 (docs) and #1554 (vscode deps).

`engine-ci.yml` filters on `engine/**`, `schemas/**` and itself. `Adapter boundary` has its own filter. Neither matches a docs, scripts, SDK, or dagster change.

## Why this is worth fixing rather than routing around

Today the workaround is an admin merge for anything outside `engine/`. That means the merge protection is routinely bypassed for the exact class of change least likely to have been reviewed by CI — and bypass becomes ordinary rather than exceptional, which is the thing branch protection is trying to prevent.

It also reads exactly backwards in the UI. A maintainer sees a PR with every visible check green, blocked for no stated reason.

`.github/SECURITY_ENVIRONMENTS.md:70-74` already flags the same root cause from the security side:

> Every other required context is path-filtered on `engine/**`, so a pull request that bundles any engine source edit with a hostile `.github` edit reports all of them green.

So the path filters cause a merge problem *and* a detection problem. One change fixes both.

## Suggested fix — make the required jobs always report

The standard shape: drop the workflow-level `paths:` filter, add a `changes` job that computes what was touched, and give each real job an `if:` on it. The job still **runs and reports**, it just short-circuits to success when its paths were untouched.

```
 before   engine-ci.yml has paths: engine/**
          non-engine PR -> workflow never runs -> 4 contexts ABSENT -> blocked

 after    engine-ci.yml always runs
          changes job -> engine touched?  no -> jobs report success immediately
                                          yes -> jobs do the real work
```

Cost is a handful of near-instant runner minutes per PR, which is small next to the current admin-merge tax. The cost-saving intent of the filters (`engine-ci.yml` comment) is preserved — the expensive steps still only run when engine paths change.

Worth doing at the same time: a check that every context named in the ruleset is produced by a workflow that runs unconditionally. That is the durable form — this class of bug is invisible until someone opens a PR that happens not to touch `engine/`.

## Related

- #1557 fixed one instance narrowly: `.claude/skills/rocky-ai-workflow/SKILL.md` is compiled into the binary but was outside the filter, so it both skipped engine CI and could not merge. This issue is the general form.
- #1507 (merge queue) is adjacent but separate; it is about paying the suite twice, not about contexts that never report.

## Currently blocked by this

#1555, #1552, #1554 — all `ready`, all zero failures, none mergeable.

hugocorreia90:
## Summary

Every non-engine PR is permanently unmergeable today (#1563). Four required contexts (`Test`, `Clippy`, `Format`, `Adapter boundary`) live in `engine-ci.yml` and one (`Check codegen is up to date`) in `codegen-drift.yml`, and both workflows path-filter their `pull_request` trigger. A filtered-out workflow never reports. An absent required check blocks the merge forever.

This PR makes both workflows always run on PRs. A cheap `changes` job diffs the merge commit against its base parent. Each real job skips itself when no relevant path was touched. A job skipped by an `if:` conditional still reports, and satisfies the required check (``[GitHub docs: handling skipped but required checks](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks)).``

```
 before   non-engine PR -> workflow never runs -> 5 contexts ABSENT -> blocked forever
 after    non-engine PR -> changes job says "false" -> jobs skip -> contexts report -> mergeable
          engine PR     -> changes job says "true"  -> jobs run exactly as before
```

Unblocks #1555, #1552, #1554.

Closes #1563

## Subproject(s) touched

## Changes

- `engine-ci.yml`: drop the `pull_request` `paths:` filter; add a `changes` job (merge-commit diff, `fetch-depth: 2`); gate `test`, `clippy`, `fmt`, `adapter-boundary`, `release-build-smoke` on its output.
- `codegen-drift.yml`: same shape for the `drift` job, with its own path list.
- `SECURITY_ENVIRONMENTS.md`: the rationale paragraph described the old pending-context behavior; reworded for always-report.

Design points:

- **Fails closed twice.** A diff error keeps the answer `true`. The gate is `engine != 'false'` with `!cancelled()`, so a failed detection job (empty output) runs everything instead of skipping it.
- **The detection regexes mirror the old filters exactly** — verified against sample paths, including anchored single files (`justfile` matches, `justfile.bak` does not).
- **`push` triggers keep their filters.** Main-branch cost is unchanged. Non-engine PRs pay one near-instant detection job per workflow.
- **No new action sources.** The detection job reuses the pinned `actions/checkout` already in the file and plain `git diff`, so the credential-containment allow-list is untouched.

## Test Plan

- `python3 .github/scripts/check_credential_containment.py` → passed on the modified tree.
- `.github/tests/test_credential_containment.py` → 110 tests, all pass.
- `actionlint` v1.7.7 on both workflows → clean.
- Regex simulation for both path lists against engine, scripts, docs, vscode, schemas, and near-miss paths (`.bak` suffixes) → matches the old filter semantics exactly.
- The live proof is this PR itself once merged: #1555 (scripts-only) should flip from `BLOCKED` to mergeable with the four contexts reporting as skipped.

## Open questions, on the record (per AGENTS.md)

**Least confident about:** the ruleset treating a `skipped` required context as satisfied. GitHub's own docs say a job skipped by a conditional reports and does not block, and this is the documented pattern for path-filtered required checks — but the definitive test is the first non-engine PR after this merges. If it stays blocked, the fallback is letting each gated job run and short-circuit its steps to a real `success` conclusion.

**Most important thing possibly missing:** the issue's suggested durable form — a CI check that every ruleset-named context is produced by an unconditionally-running workflow. Reading rulesets needs API access the unprivileged CI token may not have, so it is left as follow-up rather than widening this PR.

**Red team:** the repo's designated independent red team (Codex) is not available in this session. Flagging that here per AGENTS.md instead of skipping silently.

### DIFF
--- a/.github/SECURITY_ENVIRONMENTS.md
+++ b/.github/SECURITY_ENVIRONMENTS.md
@@ -68,14 +68,15 @@ choice. Require branches to be up to date before merging, do not grant ordinary
 bypass access, and keep any emergency override limited to named maintainers.
 
 Requiring these two checks is not optional hardening, and the repository's other
-required checks are not a substitute. Every other required context is
-path-filtered on `engine/**`, so a pull request that bundles any engine source
-edit with a hostile `.github` edit reports all of them green. If the containment
-check is merely present-and-red rather than required, GitHub returns a mergeable
-state and the change lands with no override and no administrator involvement. The
-absence of an engine change is likewise not a barrier: a `.github`-only pull
-request leaves the engine contexts pending, which is the same state that ordinary
-docs-only work produces, so it is routinely cleared rather than investigated.
+required checks are not a substitute. The engine contexts run real work only for
+changes that touch engine paths (since #1563 their workflows always report — a
+change-detection job skips the expensive jobs on other changes, so the contexts
+read green either way). A pull request that bundles any engine source edit with
+a hostile `.github` edit therefore reports all of them green, and a `.github`-only
+pull request reports them green as skipped. Only the containment check examines
+the `.github` tree itself. If it is merely present-and-red rather than required,
+GitHub returns a mergeable state and the change lands with no override and no
+administrator involvement.
 
 Verify the rule with a disposable pull request that makes a policy regression:
 the first check must fail from trusted `main` tooling, the candidate regression
--- a/.github/workflows/codegen-drift.yml
+++ b/.github/workflows/codegen-drift.yml
@@ -21,22 +21,15 @@ on:
       - 'scripts/copy_project_schema.py'
       - 'justfile'
       - '.github/workflows/codegen-drift.yml'
+  # No `paths:` filter on pull_request — deliberately. "Check codegen is up
+  # to date" is a required merge context. A workflow-level path filter leaves
+  # it ABSENT on a PR that does not match, and an absent required check
+  # blocks the merge forever (#1563). Instead the `changes` job below detects
+  # whether codegen inputs were touched, and the drift job skips itself — a
+  # job skipped by an `if:` conditional still reports, and satisfies the
+  # required check.
   pull_request:
     branches: [main]
-    paths:
-      - 'engine/**'
-      - 'schemas/**'
-      - 'docs/public/openapi.json'
-      - 'sdk/python/src/rocky_sdk/types_generated/**'
-      - 'integrations/dagster/tests/fixtures_generated/**'
-      - 'editors/vscode/src/types/generated/**'
-      - 'editors/vscode/schemas/rocky-project.schema.json'
-      - 'examples/playground/pocs/**'
-      - 'scripts/regen_fixtures.sh'
-      - 'scripts/_normalize_fixture.py'
-      - 'scripts/copy_project_schema.py'
-      - 'justfile'
-      - '.github/workflows/codegen-drift.yml'
 
 # CI only reads the repo; restrict the token to the minimum (CodeQL ql-for-actions).
 permissions:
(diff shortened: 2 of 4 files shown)


hugocorreia90:
Post-merge red team: Codex, two verification lenses per finding.

Confirmed, fixed in #1580: `git diff --name-only` hides rename SOURCES, so a file renamed out of `engine/` skipped validation; and `printf | grep -q` under `pipefail` dies of SIGPIPE past ~64 KB of paths, read as "no match" — both fail-open, both reproduced. Also corrected: the detector does run on push (~10 s), so "main-branch cost is unchanged" was false.

Refuted: "cancellation makes the heavy jobs report skipped and satisfy required checks." A never-started job in a cancelled run reports `cancelled`, which does not satisfy a required check — probed on real runs.

PR #1572, opened 2026-08-31
fix(engine/rocky-mcp): report a failed draft rollback instead of claiming a clean refusal

### LINKED ISSUE #1561: the draft rollback discards its own filesystem errors, so "a denied draft leaves no file" is not guaranteed
`DraftRollback` states a contract it cannot keep. Both filesystem calls discard their result.

```rust
// engine/crates/rocky-mcp/src/tools.rs:4748
fn restore_or_remove(path: &Path, prior: Option<&[u8]>) {
    match prior {
        Some(bytes) => { let _ = std::fs::write(path, bytes); }   // <- discarded
        None        => { let _ = std::fs::remove_file(path); }    // <- discarded
    }
}
```

The guard's own doc comment (`tools.rs:4759-4766`) states the contract:

> unwinding past it would leave a denied/broken draft on disk, violating the "a denied draft leaves NO file" contract

If `remove_file` fails — a read-only mount, a permissions change, a locked file on Windows — the artifact stays and Rocky reports the opposite.

## Why it matters

The affected arms are exactly the refusal paths:

```
 draft_model / draft_contract / draft_check / draft_metadata
   PolicyGate::Deny        -> "the draft was not kept"
   PolicyGate::Unloadable  -> "the draft was rolled back"
```

A `deny` rule exists to stop an agent's write landing. When cleanup fails silently, the tool reports a denial while the file the denial was about is on disk — and the agent, reading a structured `policy_denied`, has no reason to check.

## Reproduce

Have any draft tool write a NEW artifact (no prior bytes, so the rollback path is `remove_file`), then make `models/` unwritable before the gate returns, with a policy that denies. The tool returns `policy_denied`; the file remains.

## Suggested fix

Two parts:

1. **Do not discard the result.** `restore_or_remove` should report failure to its caller.
2. **Say what actually happened.** A refusal whose cleanup failed must not claim the draft was removed — it should name the path and say the artifact could not be cleaned up, so the operator knows there is something to delete.

Worth deciding at the same time whether a failed rollback should change the error class. An agent that reads "denied, nothing written" behaves differently from one told "denied, but a file is still there."

## Scope

**Pre-existing.** The `Deny` arm and the guard have carried this since the draft tools landed. It is filed separately from #1559 because it is not that fix's defect — though #1559 adds a fourth arm with the same dependency, and its wording was chosen to describe the action taken rather than a guaranteed end state, precisely because of this.

## Provenance

Found by independent adversarial review (Codex, a different model, with a second reviewer confirming) during review of #1560. Verified directly at `tools.rs:4748-4757` before filing. The reviewer's own stated uncertainty was whether post-write filesystem mutation is inside the accepted threat model — that is the question to settle first, and an explicit rollback contract plus a fault-injection test would settle it.

hugocorreia90:
## Summary

`DraftRollback` stated a contract it could not keep: both filesystem calls in `restore_or_remove` discarded their result, so a `PolicyGate::Deny`/`Unloadable` refusal could report "the draft was not kept" while the denied file stayed on disk (#1561).

```
 before   deny -> remove_file fails silently -> "the draft was not kept"   file remains
 after    deny -> remove_file fails          -> refusal names the path,
                                                says it could not be removed
          deny -> rollback succeeds          -> message byte-identical to before
```

Closes #1561

## Subproject(s) touched

## Changes

- `restore_or_remove` returns `std::io::Result<()>`. Removing an already-absent file counts as success.
- New `DraftRollback::rollback()` consumes the guard, rolls back now, and returns the paths it could not clean (`RollbackFailure`: path + action + I/O cause). All 8 deliberate refusal arms (Deny + Unloadable across `draft_model` / `draft_contract` / `draft_check` / `draft_metadata`) call it. The `Drop` impl stays as the unwind-path net and logs failures via `tracing::warn!`.
- The error class stays `policy_denied` — agents key off it. One optional envelope field `rollback_failed_paths` (`skip_serializing_if`) rides only on failure, so the wire is unchanged on the clean path; the three `propose` wire-parity goldens still pass byte-for-byte.
- Verified nothing in rocky-mcp feeds `schemas/` codegen (that registry is `rocky-cli`'s `export_schemas.rs` only) — no codegen cascade.

## Test Plan

- `cargo test -p rocky-mcp` — lib 40 passed, roundtrip 64 passed, wire_parity 3 passed; re-run after rebasing onto current main (`21639f0`).
- `cargo clippy -p rocky-mcp --all-targets --all-features -- -D warnings` — clean. `cargo fmt -- --check` — clean.
- New tests: `draft_rollback_failure_is_reported_and_names_the_path` (root-proof fault: the path occupied by a non-empty directory), `draft_rollback_reports_a_read_only_directory` (`#[cfg(unix)]` 0o555; probes and stands down under root, exercises EACCES on non-root runners), `rollback_disposition_clean_keeps_the_claim`, `rollback_failed_paths_ride_the_envelope_only_on_failure`.
- **Fix-sensitive, demonstrated:** with the discarding body restored (test kept), `draft_rollback_failure_is_reported_and_names_the_path` fails with `left: None, right: Some(["models/shadow.sql"])`; re-applying the fix, all rollback tests pass.

## Open questions, on the record (per AGENTS.md)

**Least confident about:** the wording contract of refusal messages. Clean-path messages are byte-identical and the goldens pin them; no test pins the new failure-path prose, so a downstream scraper of message text (against the repo's own advice) sees new prose only on the failure path.

**Most important thing possibly missing:** a true MCP round-trip test of deny-with-failed-rollback. Root defeats permission-based fault injection in this environment, so coverage sits at the guard/disposition seam; the arm-to-envelope threading is plain interpolation covered by compile plus the envelope test.

**Red team:** the repo's designated independent red team (Codex) is not available in this session — flagged here per AGENTS.md rather than skipped silently. The issue itself carries prior independent review (found by Codex during #1560's review), and this fix follows its two-part suggested shape.

## Checklist

- [x] Commit messages follow conventional commits, scoped by crate
- [x] Not a CLI JSON schema or DSL syntax change — no consumer cascade

### Engine (`engine/`)
- [x] `cargo test -p rocky-mcp` passes (full `--all-targets` runs in CI)
- [x] `cargo clippy -p rocky-mcp --all-targets -- -D warnings` is clean

### DIFF
--- a/.github/SECURITY_ENVIRONMENTS.md
+++ b/.github/SECURITY_ENVIRONMENTS.md
@@ -68,14 +68,15 @@ choice. Require branches to be up to date before merging, do not grant ordinary
 bypass access, and keep any emergency override limited to named maintainers.
 
 Requiring these two checks is not optional hardening, and the repository's other
-required checks are not a substitute. Every other required context is
-path-filtered on `engine/**`, so a pull request that bundles any engine source
-edit with a hostile `.github` edit reports all of them green. If the containment
-check is merely present-and-red rather than required, GitHub returns a mergeable
-state and the change lands with no override and no administrator involvement. The
-absence of an engine change is likewise not a barrier: a `.github`-only pull
-request leaves the engine contexts pending, which is the same state that ordinary
-docs-only work produces, so it is routinely cleared rather than investigated.
+required checks are not a substitute. The engine contexts run real work only for
+changes that touch engine paths (since #1563 their workflows always report — a
+change-detection job skips the expensive jobs on other changes, so the contexts
+read green either way). A pull request that bundles any engine source edit with
+a hostile `.github` edit therefore reports all of them green, and a `.github`-only
+pull request reports them green as skipped. Only the containment check examines
+the `.github` tree itself. If it is merely present-and-red rather than required,
+GitHub returns a mergeable state and the change lands with no override and no
+administrator involvement.
 
 Verify the rule with a disposable pull request that makes a policy regression:
 the first check must fail from trusted `main` tooling, the candidate regression
--- a/.github/workflows/codegen-drift.yml
+++ b/.github/workflows/codegen-drift.yml
@@ -21,22 +21,15 @@ on:
       - 'scripts/copy_project_schema.py'
       - 'justfile'
       - '.github/workflows/codegen-drift.yml'
+  # No `paths:` filter on pull_request — deliberately. "Check codegen is up
+  # to date" is a required merge context. A workflow-level path filter leaves
+  # it ABSENT on a PR that does not match, and an absent required check
+  # blocks the merge forever (#1563). Instead the `changes` job below detects
+  # whether codegen inputs were touched, and the drift job skips itself — a
+  # job skipped by an `if:` conditional still reports, and satisfies the
+  # required check.
   pull_request:
     branches: [main]
-    paths:
-      - 'engine/**'
-      - 'schemas/**'
-      - 'docs/public/openapi.json'
-      - 'sdk/python/src/rocky_sdk/types_generated/**'
-      - 'integrations/dagster/tests/fixtures_generated/**'
-      - 'editors/vscode/src/types/generated/**'
-      - 'editors/vscode/schemas/rocky-project.schema.json'
-      - 'examples/playground/pocs/**'
-      - 'scripts/regen_fixtures.sh'
-      - 'scripts/_normalize_fixture.py'
-      - 'scripts/copy_project_schema.py'
-      - 'justfile'
-      - '.github/workflows/codegen-drift.yml'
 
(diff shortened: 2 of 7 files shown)


hugocorreia90:
Post-merge red team: Codex, two verification lenses per finding, all reproduced end-to-end over the rmcp harness.

Confirmed, fixed in #1581: a dangling symlink at the draft path leaked the denied bytes OUTSIDE the project while the refusal read clean; an existing-but-unreadable `.sql` snapshotted as absent, so a refusal deleted the user's file; "STILL ON DISK" was asserted without checking. #1581 also closes the deeper check-then-write race with real no-follow writes and refuses a redirected `models/` directory.

Refuted: the "no schema change" claim was scoped to CLI codegen and holds — nothing in rocky-mcp feeds `schemas/`, and the new envelope field is optional and absent on the clean path.

PR #1579, opened 2026-08-31
fix(engine/rocky-compiler): propagate outer-join nullability

### LINKED ISSUE #1578: bug(engine/rocky-compiler): outer joins preserve NOT NULL and bypass nullable contracts
## Description

The compiler copies a source column's non-nullability through an outer join. A `NOT NULL` source column becomes nullable on the unmatched side, so this can incorrectly accept a `nullable = false` output contract without E012.

This is a compile-time correctness issue. It does not change the generated SQL or independently corrupt stored data.

## Steps to Reproduce

1. Supply these source schemas to `rocky_compiler::compile::CompilerConfig.source_schemas` (equivalently, compile with their discovered source schemas):

   | Source | Column | Rocky type | Nullable |
   |---|---|---|---|
   | `raw.orders` | `id` | `Int64` | false |
   | `raw.orders` | `customer_id` | `Int64` | false |
   | `raw.customers` | `id` | `Int64` | false |
   | `raw.customers` | `name` | `String` | false |

2. Compile this `joined.sql`:

   ```sql
   SELECT o.id, c.name AS customer_name
   FROM raw.orders AS o
   LEFT JOIN raw.customers AS c ON o.customer_id = c.id
   ```

   `joined.toml`:

   ```toml
   name = "joined"
   [target]
   catalog = "warehouse"
   schema = "silver"
   table = "joined"
   ```

   `joined.contract.toml`:

   ```toml
   [[columns]]
   name = "customer_name"
   type = "String"
   nullable = false
   ```

3. Inspect `result.type_check.typed_models["joined"]` and the compiler diagnostics. The right-hand `customer_name` must be nullable and the contract must emit E012. The preserved left-hand `id` should remain non-nullable.

The underlying SQL behavior is reproducible without Rocky:

```sql
CREATE SCHEMA raw;
CREATE TABLE raw.orders (id BIGINT NOT NULL, customer_id BIGINT NOT NULL);
CREATE TABLE raw.customers (id BIGINT NOT NULL, name VARCHAR NOT NULL);
INSERT INTO raw.orders VALUES (1, 10), (2, 99);
INSERT INTO raw.customers VALUES (10, 'Alice');
SELECT o.id, c.name AS customer_name
FROM raw.orders AS o
LEFT JOIN raw.customers AS c ON o.customer_id = c.id
ORDER BY o.id;
-- (1, 'Alice'), (2, NULL)
```

## Expected Behavior

Infer `(String, nullable = true)` for `customer_name` and reject its non-null contract with E012. Infer `(Int64, nullable = false)` for `o.id`. An otherwise identical INNER JOIN should preserve source nullability.

Apply the same rule to the left input of RIGHT JOIN and both inputs of FULL JOIN. Keep aliases distinct in self-joins and carry the result through downstream model inference.

## Actual Behavior

Current code treats the selected column as a direct lineage edge and copies `(String, false)` from `raw.customers.name`. Contract validation therefore has no nullable-column mismatch to report.

Evidence at [`540f2f0`](https://github.com/rocky-data/rocky/commit/540f2f0e0d4815bd10d961de7173628e66f833e3):

- [`extract_tables`](https://github.com/rocky-data/rocky/blob/540f2f0e0d4815bd10d961de7173628e66f833e3/engine/crates/rocky-sql/src/lineage.rs#L193-L200) does not retain the join operator or which relation is null-extended.
- [`compute_model_typecheck`](https://github.com/rocky-data/rocky/blob/540f2f0e0d4815bd10d961de7173628e66f833e3/engine/crates/rocky-compiler/src/typecheck.rs#L449-L468) copies upstream nullability for direct references and ordinary casts.
- [`validate_contract`](https://github.com/rocky-data/rocky/blob/540f2f0e0d4815bd10d961de7173628e66f833e3/engine/crates/rocky-compiler/src/contracts.rs#L163-L181) relies on that inferred nullable bit for E012.

Executed reproduction on Linux/arm64 with Rust 1.98.0: added the public-compile regression to unchanged upstream compiler code at `540f2f0`, then ran:

```text
cargo test --locked -p rocky-compiler --lib outer_join_rejects_non_null_contract -- --nocapture

assertion `left == right` failed: LEFT JOIN
  left: []
 right: [("E012", "joined")]
```

The test reached contract validation and failed because the compiler emitted no E012. [PR #1579](https://github.com/rocky-data/rocky/pull/1579) includes this regression and its INNER JOIN control, plus ordinary-cast and grouped-aggregate compatibility coverage.

The SQL control was also executed twice with DuckDB 1.5.2: LEFT JOIN returned two rows with one null name; INNER JOIN returned one row with no null name. Twelve additional DuckDB controls confirm the merged-key nullability rules for LEFT/RIGHT/FULL JOIN USING, with nullable and non-null inputs.

## Rocky Version

Canonical `main` at `540f2f0e0d4815bd10d961de7173628e66f833e3`, rechecked on 2026-08-31. The affected lineage and typecheck files are also unchanged in release `engine-v1.72.0`.

## OS / Architecture

Compiler regression: Linux / arm64 (local Colima container), Rust 1.98.0. SQL controls: macOS / arm64, DuckDB 1.5.2.

## Related Work / Duplicate Check

Searched canonical issues and PRs in all states, including drafts and merged PRs, with `outer join`, `LEFT JOIN`, `nullable join`, `nullability`, `null extension`, `E012`, and `infer_select_types`; inspected the matching discussions and relevant recent tracker updates on 2026-08-31. No existing item addressing this root cause was found.

#1148 / #1155 cover fallible casts, #1034 covers consumed-column completeness, and #1467 / #1513 cover the `no_new_nullable` rule. Those do not account for nulls introduced by outer joins.

mattfaltyn:
## Summary

Outer joins can introduce nulls into columns declared `NOT NULL` at the source. The compiler previously copied source nullability through direct lineage edges and could silently accept a false non-null contract.

This PR carries join-side nullability into the compiled schema, so an affected `nullable = false` contract emits E012 before warehouse execution.

Fixes #1578.

## Subproject(s) touched

- [x] `engine/` (Rust compiler)
- [x] Compiler documentation (`docs/`)

## Changes

- Build expression scope from the query's actual relations, keeping aliases separate through LEFT, RIGHT, FULL, and chained or nested outer joins.
- Keep merged `USING` / `NATURAL` keys separate from qualified input columns. Preserve the appropriate merged-key nullability.
- Widen direct-reference and ordinary column-reference cast nullability without inventing concrete types for unresolved expressions. Preserve computed-cast contract behavior; expression-level ordinary casts still preserve input nullability, and TRY_CAST / SAFE_CAST remain nullable.
- Retain exact compiler schema lookup and existing wildcard duplicate-name handling. Avoid cloning the whole project schema map.
- Add a public compile/E012 regression plus coverage for preserved sides, self-joins, stars, casts, merged keys, alias column lists, and incremental/full equivalence. A grouped SUM cast control preserves existing valid contracts.

No public IR fields, CLI JSON schemas, dependencies, or generated SQL change. Existing contracts that relied on incorrectly non-null outer-join outputs now fail intentionally. Unknown types remain conservative. WHERE filters do not narrow nullability; broader pre-existing qualified-star schema and semi/anti/asof inference limits are outside this fix.

## Test Plan

- Upstream regression: fails on `540f2f0` because LEFT JOIN emits no E012; the same test passes with the fix.
- `cargo test --locked -p rocky-compiler --lib outer_join -- --nocapture`: **9 passed** on the final head, including the grouped-cast compatibility control.
- `cargo test --locked --all-targets --all-features --no-fail-fast`: **6,618 passed / 112 repository-defined ignored / 0 failed** locally.
- `cargo clippy --locked --all-targets --all-features -- -D warnings`: passed on the final GitHub head. Focused compiler Clippy also passed locally.
- `cargo fmt --all -- --check`: passed (Rust 1.98.0).
- Repository mutation check: passed. Disabling the compiler overlay makes `outer_join_rejects_non_null_contract` fail on the missing E012 diagnostic; restoring the source returns all nine regressions to green.
- Final commit `7c5d552f`: **all 11 GitHub checks passed**, including **6,618 engine tests passed / 112 skipped** and all nine new regressions.
- Python SDK: **241 passed**. Dagster: **721 passed**. Both documented Ruff lint/format checks passed.
- Editor: TypeScript compilation, **208 tests passed**, ESLint and bundle passed; **14 Electron integration tests passed**. Used one test worker after worker-startup timeouts in the parallel run.
- Documentation site: final build passed (**105 pages**).
- DuckDB controls: LEFT/INNER reproduction repeated twice, plus **12** merged-key controls for nullable and non-null inputs.

Local validation used Linux/arm64 in Colima, Rust 1.98.0, Rust's bundled LLD, Node 22.23.2, and Python 3.11. The focused compiler doctest command passed; its single example is repository-defined ignored. Source hashes and the final commit are recorded per run.

## Independent review

GPT-5.5 reviewed the complete patch. Its review led to exact compiler lookup, duplicate-star handling, and explicit USING/NATURAL merged-key handling. The alias-column concern was rebutted against sqlparser's actual type and covered by tests. CI exposed overly broad widening of CAST(SUM(...)); the correction limits the compile overlay to column-reference casts and adds a regression. Independent review approved the final head with no blocking findings and confirmed ModelIr / contract semantics are preserved for the stated correction.

Least confident: the semantic graph's existing qualified-star schema limitations, which this PR does not claim to fix. Most important missing evidence: none for the stated compiler correction; credential-dependent repository tests remain at their existing ignored status.

## Checklist

- [x] Conventional commit scoped to the engine/compiler; no `Co-Authored-By` trailers.
- [x] `cargo test --all-targets` passes (validation also enables all features).
- [x] `cargo clippy --all-targets --all-features -- -D warnings` is clean.
- [x] `cargo fmt --check` passes.
- [x] Relevant compiler documentation is updated.
- [x] CLI JSON schema and DSL syntax are unchanged; no generated consumer updates are required.

**ModelIr / contract semantics preserved: yes** — SQL and IR shape are unchanged; non-null contracts now reject outer-join outputs that can contain nulls.

### DIFF
--- a/docs/src/content/docs/concepts/compiler.md
+++ b/docs/src/content/docs/concepts/compiler.md
@@ -110,6 +110,16 @@ It infers types from:
 Every model comes out with a typed schema: a list of `TypedColumn` entries,
 each with a name, a `RockyType`, and a nullability flag.
 
+Outer joins can introduce nulls even when source columns are non-nullable.
+Rocky marks direct references and ordinary casts of those references from the right side of `LEFT JOIN` as nullable.
+`RIGHT JOIN` marks the accumulated left side; `FULL JOIN` marks both sides.
+Aliases remain distinct in self-joins, and downstream models inherit the resulting nullability.
+A `nullable = false` contract on an affected column with a known type raises `E012`.
+This analysis is conservative: a later `WHERE` filter does not narrow nullability.
+It does not change the existing nullability inference for casts of computed expressions.
+
+For `USING` and `NATURAL` joins, Rocky distinguishes merged join keys from qualified references to either input.
+
 ### 5. Validate contracts
 
 If a contracts directory exists, Rocky loads the `.contract.toml` files and
--- a/engine/crates/rocky-compiler/src/compile.rs
+++ b/engine/crates/rocky-compiler/src/compile.rs
@@ -791,6 +791,137 @@ mod tests {
     use super::*;
 
     #[test]
+    fn outer_join_rejects_non_null_contract() {
+        let dir = tempfile::tempdir().unwrap();
+        std::fs::write(
+            dir.path().join("joined.toml"),
+            "name = \"joined\"\n[target]\ncatalog = \"warehouse\"\nschema = \"silver\"\ntable = \"joined\"\n",
+        )
+        .unwrap();
+        std::fs::write(
+            dir.path().join("joined.contract.toml"),
+            "[[columns]]\nname = \"customer_name\"\ntype = \"String\"\nnullable = false\n",
+        )
+        .unwrap();
+        let column = |name: &str, data_type| TypedColumn {
+            name: name.to_string(),
+            data_type,
+            nullable: false,
+        };
+        let config = CompilerConfig {
+            models_dir: dir.path().to_path_buf(),
+            source_schemas: HashMap::from([
+                (
+                    "raw.orders".to_string(),
+                    vec![
+                        column("id", RockyType::Int64),
+                        column("customer_id", RockyType::Int64),
+                    ],
+                ),
+                (
+                    "raw.customers".to_string(),
+                    vec![
+                        column("id", RockyType::Int64),
+                        column("name", RockyType::String),
+                    ],
+                ),
+            ]),
+            ..Default::default()
+        };
+        for (expression, join, nullable) in [
+            ("c.name", "LEFT JOIN", true),
+            ("c.name", "INNER JOIN", false),
+            ("CAST(c.name AS STRING)", "LEFT JOIN", true),
+            ("CAST(c.name AS STRING)", "INNER JOIN", false),
+            ("c.name::STRING", "LEFT JOIN", true),
+            ("c.name::STRING", "INNER JOIN", false),
+            ("CAST(CAST(c.name AS STRING) AS STRING)", "LEFT JOIN", true),
+            (
+                "CAST(CAST(c.name AS STRING) AS STRING)",
+                "INNER JOIN",
+                false,
+            ),
+        ] {
+            std::fs::write(
+                dir.path().join("joined.sql"),
+                format!(
+                    "SELECT o.id, {expression} AS customer_name FROM raw.orders o \
+                     {join} raw.customers c ON o.customer_id = c.id"
+                ),
+            )
+            .unwrap();
+            let result = compile(&config).unwrap();
+            let diagnostics: Vec<_> = result
+                .diagnostics
+                .iter()
+                .filter(|d| d.is_error())
+                .map(|d| (&*d.code, d.model.as_str()))
+                .collect();
+            assert_eq!(
+                diagnostics,
+                if nullable {
+                    vec![("E012", "joined")]
+                } else {
+                    vec![]
+                },
+                "{expression}: {join}"
+            );
+            assert_eq!(result.has_errors, nullable, "{expression}: {join}");
+            let columns = &result.type_check.typed_models["joined"];
+            assert_eq!(columns[0].data_type, RockyType::Int64);
+            assert!(!columns[0].nullable, "preserved order id: {join}");
+            assert_eq!(
+                columns[1].data_type,
+                RockyType::String,
+                "{expression}: {join}"
+            );
+            assert_eq!(columns[1].nullable, nullable, "{expression}: {join}");
+        }
+    }
+
+    #[test]
+    fn outer_join_propagation_preserves_grouped_cast_contracts() {
+        let dir = tempfile::tempdir().unwrap();
+        std::fs::write(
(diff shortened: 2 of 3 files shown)


hugocorreia90 (approved):
Verified the outer-join nullability rule against SQL semantics for LEFT, RIGHT, FULL, chained, nested and self-join shapes, with DuckDB as ground truth. Confirmed no IR, schema, dependency or generated-SQL drift.

One regression needed a fix before merge, pushed to this branch: the merged USING/NATURAL key shared the FULL OUTER rule with plain JOIN and INNER JOIN via a catch-all arm, which widened a key that cannot be null and rejected valid contracts with E012. FULL OUTER keeps its own arm unchanged. Added a regression over all four input combinations.

Nice catch on the underlying bug, and the reproduction in #1578 made it quick to confirm.

PR #1588, opened 2026-09-02
chore(ci): bump taiki-e/install-action and action-gh-release with their allow-list entries

hugocorreia90:
Supersedes #1562 and #1585. One PR moves the workflow pins and the allow-list entries together, so `main` never holds a pin its own checker rejects.

## Why one PR

The policy job runs the checker from the **trusted base**, so a PR that bumps a pinned release action is red by design. That is the bootstrap in `.github/SECURITY_ENVIRONMENTS.md`, steps 4–5, and it needs the narrow override.

The trap is a half-merge. Dependabot bumps the `uses:` pins but not `RELEASE_ACTION_SOURCES`. Overriding #1585 as-is would leave `main` with pins its own checker rejects, and every later PR would go red.

```
#1562   taiki-e v2.86.8  + allow-list     older
#1585   taiki-e v2.87.1  + gh-release     newer, no allow-list
this    taiki-e v2.87.1  + gh-release + both allow-list entries
```

## What moves

| Site | Before | After |
|---|---|---|
| `engine-ci.yml`, `engine-release.yml` | `taiki-e/install-action@82cd3e7` v2.85.13 | `@742a331` v2.87.1 |
| `engine-wasm-release.yml`, `vscode-release.yml` | `softprops/action-gh-release@3d0d988` v3 | `@efb3536` v3 |
| `check_credential_containment.py` | both old SHAs | both new SHAs |

The old SHAs appear nowhere under `.github/` afterwards.

## Pins verified upstream, not taken from the comment

```
$ gh api repos/taiki-e/install-action/git/ref/tags/v2.87.1
742a3317eac7bd62f91cd888b4eead5e784ba833   (lightweight tag -> commit)

$ gh api repos/softprops/action-gh-release/git/ref/tags/v3
5113cdc9... (annotated tag)
$ gh api repos/softprops/action-gh-release/git/tags/5113cdc9...
efb35369e0ad2afab669f228072c1b0d510eae64   (dereferences to the pinned commit)
```

## Bootstrap step 2, run against this exact tree

- `check_credential_containment.py "$PWD"` with the candidate as its own trusted base: `passed`
- `python3 -m unittest discover -s .github/tests`: 110 tests, OK
- `actionlint` on the four touched workflows: clean

## After the override

Step 5: the **next** PR that does not touch a trust root is the confirmation run. Require its `Credential containment policy` to be green before merging anything else.

## Review record

Independent review by **Codex** (`gpt-5.6-sol`, read-only sandbox). Verdict: **approve**.

What it checked on its own, not from this description:
- Re-resolved both pins from the upstream tags with `gh api`; both exact. The `action-gh-release` commit carries a GitHub-verified signature; the `taiki-e` tag is lightweight (a provenance limit, not a finding).
- All 15 `RELEASE_ACTION_SOURCES` entries match every workflow use of their repositories; no other drift.
- Walked the upstream ranges: 89 commits in `taiki-e/install-action` (one runtime-script change, `action.yml` unchanged; the bundled `cargo-zigbuild` moves 0.23.0 -> 0.23.3, still a hash-pinned download from the same host), 16 in `action-gh-release` (error-classification fix, deps, docs, rebuilt bundle; no new API call or network destination). Zero published advisories in either.
- Confirmed the policy run fails with exactly the four expected bootstrap violations and that no `pull_request_target` / `workflow_run` execution path changed.

Its two standing answers: public history cannot exclude a non-public compromise (signed provenance would); and the step-5 trusted-base run is the one thing the candidate self-check cannot prove.


hugocorreia90:
**This got smaller. Half of it is now unnecessary, and that half is the reason it conflicts.**

#1647 replaced `softprops/action-gh-release` with `gh release upload` in both workflows that used it, because the action finalizes an existing draft after it uploads — which would have published each release before the registry step the new draft gate exists to protect. On `main` today:

```
uses: softprops/action-gh-release   ->  no workflow, anywhere
uses: taiki-e/install-action        ->  engine-ci.yml:172, engine-release.yml:132  (both v2.85.13)
```

The only remaining mentions of the release action are two explanatory comments. So this PR's `engine-wasm-release.yml` and `vscode-release.yml` hunks now apply to lines that no longer exist, and bumping a pin nothing uses buys nothing.

**What is still worth landing:** the `taiki-e/install-action` bump from `82cd3e7` (v2.85.13) to `742a331` (v2.87.1) at those two call sites, plus its `RELEASE_ACTION_SOURCES` entry. That is still a real bump on a real pin, and it still needs the narrow override, because the policy job runs the checker from the trusted base and a PR that moves a pinned release action is red by design.

**The `softprops` allow-list entry:** I left it in place in #1647 rather than deleting it, so this PR would not have to re-derive it. If the action stays unused, removing the entry is tidy-up for a separate change; leaving it costs nothing and the containment check passes either way (verified against the merged tree).

**Suggested reshape:** rebase onto `main`, drop both release-action hunks and the `softprops` allow-list line, and keep this as a two-line `taiki-e` bump plus its allow-list entry. That is a much smaller thing to grant an override for.

Still owner-gated either way — I have not touched this branch.

hugocorreia90:
This cannot pass its own gate. Filed the mechanism as #1769.

Short version: the PR is correct — it bumps the workflow pins *and* the allow-list entries together, which is exactly right. But the allow-list constants live inside `.github/scripts/check_credential_containment.py`, and that file is a frozen trust root compared byte-for-byte against the base branch (`:1770-1772`, no exemption branch). So editing the allow-list is itself a violation, while *not* editing it leaves the new SHAs un-allow-listed. Both errors appear in run 33703776062.

**This needs an admin merge.** Not a rebase, not a re-run — there is no version of this PR that goes green.

The general problem is worse than this one bump: none of the ~15 pinned action SHAs can ever be rotated by a pull request, including a security bump. #1769 has the mechanism and the shape of a fix; the design call is yours.

hugocorreia90:
## Closed: `main` has moved past both halves of this PR

This was queued as "rebase, then admin-merge". A rebase is the wrong operation now — both changes are stale in ways a conflict resolution would quietly get wrong.

### The `taiki-e/install-action` half is superseded, and merging it would go backwards

| | This PR | `main` today |
|---|---|---|
| `engine-ci.yml`, `engine-release.yml` | `742a3317` (v2.87.1) | `5bf6ce01` (**v2.87.5**) |
| `TAIKI_INSTALL_SOURCE` | `742a3317` (v2.87.1) | `5bf6ce01` (**v2.87.5**) |

#1774 landed v2.87.5 in the workflows, and #1809 caught the allow-list up to it. Merging this PR would **downgrade** the pin and point the allow-list at a SHA nothing uses.

### The `softprops/action-gh-release` half no longer has a call site

On current `main`, that action is not used by any workflow. `vscode-release.yml` and `engine-wasm-release.yml` finalize their drafts with `gh release upload` / `gh release edit` instead. The name survives only in explanatory comments:

```
# softprops/action-gh-release finalizes an existing draft after it
```

This PR still edits `uses:` lines that no longer exist on `main`, so a rebase would conflict there — and a naive resolution would **re-introduce** the action into workflows that deliberately stopped using it.

### What this leaves behind

`GH_RELEASE_SOURCE` in `.github/scripts/check_credential_containment.py` is now a **dead constant** — it allow-lists an action no workflow runs. Harmless (an unused allow-list entry grants nothing that is invoked), but it should be removed, or the comment above it corrected, so the next reader does not take it as evidence the action is still in the release path. Not doing that here; it is unrelated to what this PR set out to do.

### The ordering lesson, recorded because it cost a red `main`

An allow-list extension has to land **before** the pin bump it covers, never after. This PR was the extension for the earlier bumps and sat open while #1774 went in first — which put `main` in violation of its own required `Credential containment policy` check until #1809 cleared it.

Nothing here was wrong when it was written. Closed as overtaken by #1774 and #1809, not as rejected. Refs #1769.

PR #1599, opened 2026-09-03
OWNER MERGE — fix(engine/rocky-cli): fail a replication run whose error-severity checks failed

### LINKED ISSUE #1598: rocky run exits 0 on a replication pipeline whose declared checks failed
A `rocky run` on a replication pipeline can finish with exit 0 while a declared check in `check_results` failed. The exit path looks at table and model failures and never at the check results it just assembled.

Found by a red-team review of #1595, which added a `not_evaluated` state for a check the engine refused to run. That state is recorded correctly and still cannot change the exit code, so the honest reporting it added does not reach an orchestrator that only reads the status.

## Where

`engine/crates/rocky-cli/src/commands/run.rs:5922` — the replication exit path emits `pipeline_complete` and returns success after checking table and model failures. Failed entries in `check_results` are not consulted.

```
  table failed   -> non-zero
  model failed   -> non-zero
  check failed   -> exit 0        <- this
```

## Why this is a decision, not a bug fix

Changing it changes the contract for every replication user and every orchestrator reading the exit code. A project whose checks have been failing quietly would start failing its pipeline on upgrade. That may well be right — a check exists to be acted on — but it is not a change to make as a side effect of a security PR.

The options:

- **(A) Fail the run on any failed check.** Simple and honest. Breaks any project relying on today's behaviour.
- **(B) Fail only on `severity = "error"` checks.** Matches how severity already reads elsewhere, and leaves warnings advisory. Still a behaviour change, but a narrower one.
- **(C) Add a config switch, defaulting to today's behaviour.** No surprise on upgrade, and the default stays wrong.
- **(D) Leave it, and document that replication check results are advisory.** Cheapest, and it makes the check surface much weaker than it appears.

There is a related precedent worth checking before deciding: `fail_on_error` already exists under `[pipeline.<name>.checks]`, so part of this may be a matter of wiring an existing flag rather than inventing one.

## A second consumer, same shape

The Dagster component excludes every `candidate` cross-source check from its declared specs (`integrations/dagster/src/dagster_rocky/component.py:1452`) and then drops undeclared results (line 3258). So even a correctly recorded failure is absent there. Whichever option is chosen, that path needs the matching change or the result stays invisible to the main consumer.

Related: #1595, #1524.

hugocorreia90:
A replication run whose error-severity checks **failed** exited 0 and persisted as `Success`. It now fails, using the flag that already promised exactly that.

**This changes behaviour for every replication project on upgrade**, so it is marked for owner merge.

## What was wrong

`RunOutput::derive_run_status` keys only on `tables_copied` / `tables_failed` / `interrupted`. It ignores `check_results`. Nothing on the replication path folded a failed check into `tables_failed`.

```
  table copy failed    ->  tables_failed  ->  exit 1 / 2
  model compile failed ->  tables_failed  ->  exit 1 / 2
  error check failed   ->  check_results  ->  exit 0        <- this
```

## This is not a new policy

`[pipeline.<name>.checks] fail_on_error` already exists, defaults to `true`, and is documented as: *"the quality run exits non-zero if any error-severity check fails"*. The quality path honours it — `run_local.rs`, `error_failures > 0 && pipeline.checks.fail_on_error` folded into `tables_failed`. The replication path never read it.

So this wires a documented control that was not configured. That is the same class of gap several reviews found this week: a promise in the config reference that no code kept.

## Placement, and why it differs from the quality path

The quality path folds *after* its JSON emit, deliberately, so the payload is unchanged. Replication is the other way round: it calls `persist_run_record` **before** `print_json`. Folding after the emit would record `Success` and then exit non-zero — the record and the exit code would disagree, and the schedule reconciler reads the record.

So the fold sits immediately before `persist_run_record`. The JSON payload is unaffected either way: `tables_failed` already carries compile failures, and `check_results` is untouched.

## Two rules worth stating

- **A `not_evaluated` check counts as failed.** It carries `passed: false` and its configured severity. "The engine could not run this check" is not a pass — that is the whole reason #1595 gave it a state distinct from `overlap_count: 0`.
- **A warning-severity failure never folds**, with or without the flag. Severity is the author's statement about what a failure means; the flag governs only whether an *error* is fatal.

## Evidence

| Gate | Result |
|---|---|
| `cargo fmt --all -- --check` | pass |
| `cargo test --workspace --no-fail-fast` | **6919 passed, 0 failed** |
| `just codegen` + drift | clean |
| `./scripts/regen_fixtures.sh` + drift | clean |

`HEAD` and `git status --porcelain` were captured before and after the test run and were identical, so the run is not void.

Mutation check — the fold is neutered, the test kept:

```
assertion `left == right` failed: an error failure must fold
  left: 0
 right: 1
```

The test covers five cases: a failed error check folds and exits non-zero; `fail_on_error = false` opts out; a warning failure does not fold; a passing check folds nothing; and a `not_evaluated` check counts as failed.

While writing it I disabled a neighbouring test by inserting between its doc comment and its `#[test]`. Clippy caught it as "function is never used". Both attributes are correct now and the module runs 240 tests.

## Upgrade impact, stated plainly

A project whose error-severity checks have been failing quietly will start failing its pipeline. That is the intent, and it is still a change someone can be surprised by. The opt-out is one line:

```toml
[pipeline.<name>.checks]
fail_on_error = false
```

Worth a CHANGELOG entry under a clear heading before the next `engine-v*` tag.

## Not in scope, and filed

The Dagster component excludes `candidate` cross-source checks from its declared specs and then drops undeclared results, so a correctly recorded failure is still invisible there. That needs the matching change on the Dagster side; it is noted in #1598 rather than bundled here.

## Standing questions

**Least confident about:** whether folding into `tables_failed` is the right representation, as opposed to a separate `checks_failed` tally. Folding means `tables_failed` no longer means only "tables". I chose it because it is exactly what the quality path does, so both paths derive status the same way — but a reviewer who dislikes the overloading has a fair point, and the alternative is a wider change to `derive_run_status`.

**Most important thing I might be missing:** whether any consumer reads `tables_failed` as a count of tables rather than as a failure tally. I checked `derive_run_status`, `run_status_exit_result` and the persisted `RunRecord`; I did not audit every Dagster or SDK reader.

Closes #1598

### DIFF
--- a/engine/crates/rocky-cli/src/commands/run.rs
+++ b/engine/crates/rocky-cli/src/commands/run.rs
@@ -189,6 +189,31 @@ pub struct RunFailed {
 /// JSON consumer reads regardless of exit code. This is what makes a
 /// compile-failed model a first-class run failure instead of a silently
 /// skipped no-op that still reported `status: "Success"`, exit 0.
+/// Fold failed error-severity checks into `tables_failed` so the replication
+/// path's exit code, recorded status and check results agree (#1598).
+///
+/// [`RunOutput::derive_run_status`] keys only on `tables_copied` /
+/// `tables_failed` / `interrupted` and ignores `check_results`, so a run whose
+/// assertions failed persisted as `Success` and exited 0. `[pipeline.X.checks]
+/// fail_on_error` already promised the opposite and defaults to `true`; it was
+/// never read on this path. The quality path folds the same condition in
+/// [`super::run_local::run_quality`].
+///
+/// A `not_evaluated` check counts as failed. It carries `passed: false` and its
+/// configured severity, and "the engine could not run this check" is not a
+/// pass — the reason #1595 gave it a state distinct from `overlap_count: 0`.
+///
+/// A warning-severity failure never folds, with or without the flag: severity
+/// is the author's statement about what a failure means, and the flag governs
+/// only whether an *error* is fatal.
+pub(crate) fn fold_failed_checks_into_tables_failed(output: &mut RunOutput, fail_on_error: bool) {
+    if !fail_on_error {
+        return;
+    }
+    let (error_failures, _) = super::run_local::count_failures_by_severity(output);
+    output.tables_failed += error_failures;
+}
+
 pub(crate) fn run_status_exit_result(output: &RunOutput, run_id: &str) -> Result<()> {
     match output.derive_run_status() {
         rocky_core::state::RunStatus::PartialFailure => Err(PartialFailure {
@@ -5743,6 +5768,30 @@ pub async fn run(
     if model_phase_clean {
         write_recipe_manifests(&output, governance_adapter.as_ref(), &run_id).await;
     }
+
+    // Replication check trap (#1598), the twin of the quality-path fold in
+    // `run_local::run_quality`. `RunOutput::derive_run_status` keys only on
+    // `tables_copied` / `tables_failed` / `interrupted` and IGNORES
+    // `check_results`, and nothing on this path folded a failed check into
+    // `tables_failed` — so a replication run whose error-severity assertions
+    // FAILED persisted as `Success` and exited 0. `[pipeline.X.checks]
+    // fail_on_error` already promised the opposite ("the run exits non-zero if
+    // any error-severity check fails") and defaults to true; it was simply
+    // never read here. This wires the documented control rather than inventing
+    // a policy.
+    //
+    // Placed BEFORE `persist_run_record` — unlike the quality path, which folds
+    // after its JSON emit — because this path persists first (below) and prints
+    // later (~`print_json`). Folding here keeps the recorded status, the
+    // reconciler's view and the exit code agreeing; the JSON payload is
+    // unaffected either way, since `tables_failed` already carries compile
+    // failures and the `check_results` array is unchanged.
+    //
+    // A `not_evaluated` check counts as failed: it carries `passed: false` and
+    // its configured severity, and "the engine could not run this check" is not
+    // a pass. That is the whole reason #1595 gave it a state distinct from
+    // `overlap_count: 0`.
+    fold_failed_checks_into_tables_failed(&mut output, pipeline.checks.fail_on_error);
 
     // Persist the RunRecord so `rocky history`, `rocky replay`,
     // `rocky trace`, and `rocky cost` have real data to read.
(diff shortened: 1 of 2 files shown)


hugocorreia90:
**Converted to draft. Do not merge this shape.** An independent Codex review blocked it, and the central finding is that this change would create a worse failure than the one it closes.

## The blocking finding: resume turns a check failure into a false green

`--resume-latest` accepts a `PartialFailure` record — `run.rs`, `RunStatus::Failure | RunStatus::PartialFailure => Ok(())`. Verified on `main`.

Today a check-only failure never records `PartialFailure`, so resume never sees it. **This change makes it.** And a resumed run excludes every table that already copied, while check inputs are built only from tables materialized in the current invocation. So:

```
run 1   copies 1 table, an error check fails
        -> tables_failed = 1 -> PartialFailure -> exit 2
        -> the CLI message says: use --resume <id> to retry

run 2   resume excludes the copied table -> copies nothing
        -> no check inputs -> no checks run
        -> Success, exit 0, and the violating data is unchanged
```

The command's own error text points the user straight into it. That is strictly worse than exiting 0 the first time.

## Three more findings, all real

**JSON status goes stale.** `merge_replication_compile_and_copy_errors` stamps `output.status` from the *old* `tables_failed`, before my fold. So one copied table plus one failed assertion yields `status: "Success"` with `tables_failed: 1` and `errors: []` in the JSON, while the run record says `PartialFailure`, idempotency says `Failed`, and the process exits 2. Four observations, three different answers. `output.rs` documents `status` as derived from the counts.

**`tables_failed` is a documented count of tables, not of checks.** `count_failures_by_severity` counts individual checks, so one table with three failed assertions reports `tables_failed: 3` with `errors: []`. The schema and docs both call it a table/model count. This was my own stated "least confident" question in the PR body — the review confirmed it as a public contract break, with citations.

**The check-only failure lands in the compile-only branch**, which fires `pipeline_error` saying "N model(s) failed to compile" when no model failed, and returns a sentinel about failure "during parallel execution".

**The condition is also incomplete** — several check-evaluation failures never reach `check_results` and so still exit 0: freshness query errors are only warned; null-rate generation and result-parsing failures are skipped; assertion table-reference formatting errors are skipped. Worse, a row-count query failure defaults both source and target to zero and produces a **false pass**. That last one is pre-existing on `main`, independent of this change, and is filed separately.

## What the fix actually needs

Not a fold. The review's findings say the shape is:

1. A signal distinct from `tables_failed`, so the documented count keeps its meaning and the JSON stays consistent — which means an `output.rs` change and the full `just codegen` cascade into the Pydantic and TypeScript bindings.
2. Status derivation that includes it, so JSON, the run record, idempotency and the exit code agree.
3. Resume semantics for a check-only failure: it must not be resumable as a copy checkpoint, or the resumed run must re-run checks against the existing tables. This is the load-bearing one.
4. A decision on the incomplete-condition class above — which of those silent skips should also fail a run.

That is a design change, not a patch, and it belongs back on #1598 with these constraints recorded. None of the four options originally sketched there anticipated the resume interaction.

The commit stays on the branch as the starting point for whoever takes it, but the fold alone must not ship.

Reviewer: Codex (`gpt-5.6-sol`), read-only, 23 tool calls, against frozen commit `6ab965ec`. Its first attempt was killed by a full disk mid-run and produced no verdict; this is the completed re-run.

hugocorreia90:
Superseded by #1671, which merged today with the design this PR's own review round 3 showed was needed: an honest `PartialFailure` status plus a `check_gate_failed` field, rather than `Success` with a side field (that shape produced false greens across five consumers). Closing; reopen if any part of this diff is still wanted.

Issue #1597, opened 2026-09-02
SQL string literals are encoded without a dialect, so quote-doubling alone is unsound


hugocorreia90:
Rocky builds SQL string literals by doubling single quotes, in about a dozen places, with no dialect involved. That is only correct where `''` is the *sole* escape. Snowflake, DuckDB and BigQuery also honour a backslash escape, so a value ending in a backslash can move the literal's boundary.

#1595 closes the cheap half of this by refusing what it cannot encode. This issue is the real fix: **make literal encoding dialect-owned, or bind parameters.**

## Why refusal cannot finish the job

`reject_unquotable_literal` refuses any backslash in a config-supplied constant. That is sound, and it is also blunt:

- It **rejects values that are legal today.** A backslash in `accepted_values` is fine on Trino and on ordinary DuckDB strings. Those projects now get a refusal.
- It **cannot be applied to `regex_match.pattern`**, where backslashes are ordinary regex syntax. `validate_regex_pattern` rejects quotes, backticks and semicolons but deliberately permits backslashes, and the adapters then wrap the pattern in single quotes — `rocky-databricks/src/dialect.rs:197`, `rocky-snowflake/src/dialect.rs:270`. Databricks interprets backslash escapes in ordinary string literals.

So the same defect has two incompatible patches, which is the signal that the patch is at the wrong layer.

## Sinks found so far

Each takes a user-controlled string and wraps it in `'…'` with quote-doubling only.

```
rocky-core/src/tests.rs          accepted_values[].values      guarded in #1595
rocky-core/src/quarantine.rs     accepted_values[].values      guarded in #1595
rocky-duckdb/src/discovery.rs    schema_pattern.prefix         guarded in #1595
rocky-bigquery/src/discovery.rs  schema_pattern.prefix         guarded in #1595
rocky-databricks/src/dialect.rs  regex_match.pattern           NOT guarded
rocky-snowflake/src/dialect.rs   regex_match.pattern           NOT guarded
rocky-core/src/lakehouse.rs:165  format_options.table_properties[].value   NOT guarded
rocky-core/src/lakehouse.rs:178  format_options.comment        NOT guarded
rocky-core/src/catalog.rs:305    governance tag keys + values  NOT guarded
```

`validate_tag_value` and the `lakehouse.rs:120` validation both reject several delimiters but not a backslash.

[...]

Issue #1596, opened 2026-09-02
DSL string-literal lowering still doubles quotes without a dialect (last site of #1596)


hugocorreia90:
Rocky builds a `'...'` SQL literal in several places by doubling single quotes:

```rust
format!("'{}'", value.replace('\'', "''"))
```

That is correct only where `''` is the **only** escape. Snowflake, DuckDB and BigQuery also honour a backslash escape, so a value ending in `\` escapes the first quote of the pair and the second quote closes the literal. `rocky-snowflake/src/governance.rs` already documents this for tag values and escapes backslashes first.

Escaping backslashes as well is not a general fix: on Trino and standard-SQL DuckDB a doubled backslash is two literal backslashes, so it changes the value.

#1595 added `rocky_sql::validation::reject_unquotable_literal`, which refuses a backslash, and wired it into the sites that surface in the declarative-check path.

**Closed by #1595**

- `rocky-core/src/tests.rs` — `accepted_values` values
- `rocky-core/src/quarantine.rs` — `accepted_values` values
- `rocky-duckdb/src/discovery.rs` — `schema_pattern.prefix`
- `rocky-bigquery/src/discovery.rs` — `schema_pattern.prefix`

**Still open — this issue**

- `rocky-lang/src/lower.rs` — DSL string-literal lowering
- `rocky-core/src/seeds.rs` — seed cell values
- `rocky-core/src/arrow_loader.rs` — loader cell values
- `rocky-core/src/unit_test.rs` — unit-test fixture strings
- `rocky-duckdb/src/discovery.rs` — the per-schema name read back from the warehouse

`rocky-core/src/checks.rs` also doubles quotes for the cross-source `_src` tag, but its input is a table reference built from `validate_identifier`-checked parts, so no backslash can reach it. Not a seam.

Refusing a backslash is the cheap guard. The real fix is dialect-owned literal encoding — a `SqlDialect::string_literal(&str) -> String` that each adapter implements for its own escape rules — or parameter binding where the execution path supports it. Several of the open sites have a dialect in hand already, so that route is available.

hugocorreia90:
**Status after #1648.** Ten sinks now encode through `SqlDialect::string_literal`. Two remain, and this issue tracks those.

[...]

PR #1605, opened 2026-09-03
feat(engine/rocky-sql): make SQL literal escaping a dialect-owned fact

hugocorreia90:
Rocky splices user text into `'…'` in about twenty places, each with its own hand-written rule. No single rule is right on all five warehouses. This adds one encoder and makes the choice a fact each dialect states.

**No sink is routed here.** Every existing call site is unchanged and no generated SQL differs. This is step 1 of 3; the follow-ups route the sinks and delete the interim guard from #1595.

**One user-visible change, and it is the judgment call in this PR.** `rocky init-adapter` now scaffolds a crate that *deliberately does not compile*: the new method is emitted as a `compile_error!` naming both rules, so the adapter author has to state the fact instead of inheriting a guess. It came from a blocking red-team finding — a scaffold that answers for you is the same silent wrong fact the required method exists to remove. Veto that one piece and the rest of the PR still stands.

## The two rules

```
                 quote      backslash   line break
Standard         ''         literal     raw          Trino, DuckDB
Backslash        \'         \\          \n  \r       Snowflake, Databricks, BigQuery
```

Doubling the quote on a backslash-reading lexer is bypassable: a leading `\` escapes the first quote of the `''` pair and the second one closes the literal. That is the bypass `rocky-snowflake/src/governance.rs` already documents. A raw line break is worse than bypassable on BigQuery — its quoted string cannot hold one at all, so a raw newline is a parse error there whatever else is right.

Order inside `Backslash` is load-bearing, and only the first pass matters: it is the only one that rewrites an existing backslash, so the backslashes the later passes add can never be doubled. Quote-first would turn `a\'b` into `a` + four backslashes + a bare `'` — an even run that escapes itself and frees the quote.

## Evidence per dialect

| Dialect | Rule | Evidence | Where |
|---|---|---|---|
| DuckDB | `Standard` | **executed**, in-process, every CI run | `rocky-duckdb/tests/literal_escape_round_trip.rs` |
| Trino 483 | `Standard` | **executed**, Docker (`trino-conformance`) | `rocky-trino/tests/conformance.rs` |
| Snowflake | `Backslash` | **executed**, live account (PAT) | `rocky-snowflake/tests/dialect_sweep_live.rs` |
| Databricks | `Backslash` | **executed**, live workspace, default parser | `rocky-databricks/tests/lakehouse_initial_ddl_live.rs` |
| BigQuery | `Backslash` | **doc-derived** — sandbox expired | `rocky-bigquery/src/dialect.rs` |

Every executed test asserts the value comes **back byte-identical**, not that a statement parsed. Same corpus on all four: `it's`, `C:\tmp`, `a\'b`, `trailing\`, the empty string, `line1\nline2`, `crlf\r\nvalue`, `'; DROP TABLE t; --`, `\'; DROP TABLE t; --`, `plain`. DuckDB additionally stores each value in a `VARCHAR` column and compares it with `=`, so the engine reads the literal, not just the projection.

**BigQuery, both halves, labelled doc-derived in the code.** The GoogleSQL / ZetaSQL lexical specification lists the supported escapes and then says "Any sequence not in this table produces an error". `\'` is listed, so `Backslash` is valid. `''` is not listed, so `Standard` would be a **syntax error** there, not a wrong value. The same page says "Quoted strings can't contain newlines, even when preceded by a backslash", which is why `Backslash` encodes the line break instead of passing it through. BigQuery is the reason this is an enum rather than one hand-written encoder.

## What is required, and what guards that

`SqlDialect::literal_escape` has **no default body**. A default is the silent wrong fact this lane exists to remove. `cargo check --workspace --all-targets --all-features` enumerated the sites: **17 in-tree implementations** (5 shipping adapters + 12 test stubs), plus one in the `string_literal` doctest, plus the `rocky init-adapter` template — which is text, not compiled.

Because every implementation already answers, *adding* a default later would leave the whole suite green. So the requirement is guarded like a claim: `rocky-core-compiletest/tests/compile_fail/missing_literal_escape.rs` implements every other required method and omits only this one, pinned to `E0046`. Give the trait a default and that case compiles and the test fails.

`rocky init-adapter` now emits `compile_error!` for the method rather than a variant. A scaffolded rule that compiles is a guess about somebody else's lexer, and the wrong guess is silent. The generated crate refuses to build until a human states the fact. Verified with `rustc`: `compile_error!` fires during expansion and produces exactly one error, no confusing return-type mismatch. `todo!()` and `unimplemented!()` would *not* work — their type is `!`, so they compile.

Three render tests guard the template, because nothing in CI compiles the scaffold — its `mod tests` are text. While there, two pre-existing scaffold defects are fixed: the emitted `Cargo.toml` never declared the `rocky-ir` and `chrono` it uses, and `merge_into` still took `&[String]` against the trait's `&[Arc<str>]`. So `rocky init-adapter` emitted a non-compiling crate on `main` already, for reasons unrelated to this change.

## Nothing added to the adapter SDK, deliberately

`rocky_adapter_sdk::traits::SqlDialect` is a separate trait and gets nothing. A `''`-doubling default there would be dormant fail-open: nothing consumes an SDK dialect's literals today, and there is no RPC round-trip to reuse. The honest shape, if the SDK ever needs it, is a required fallible method sourced from a new `AdapterManifest` field — a manifest can carry a fact, not an algorithm, which is another argument for the enum. **Stated as a limit:** an out-of-tree process adapter implementing the SDK trait still compiles without answering anything. The `rocky-new-adapter` skill and the adapters doc now say so.

## Gates

| Gate | Result |
|---|---|
| `cargo fmt --all -- --check` | exit 0 |
| `cargo clippy --all-targets --all-features -- -D warnings` | exit 0 |
| `cargo test --workspace --no-fail-fast` | exit 0 — **6571 passed, 0 failed, 114 ignored** across 149 suites |
| `just codegen` then `git status --porcelain` | clean after reverting a `package-lock.json` diff of only `libc` lines (known local npm artifact; verified the diff contains nothing else) |
| `./scripts/regen_fixtures.sh` then drift | clean, 40 fixtures, zero diff |
| `git rev-parse HEAD` before / after the test run | `c5b556ab` / `c5b556ab`, tree clean both times |

`cargo test` here omits `--all-features` (CI runs `nextest --all-features`). The feature-gated code is still compiled and linted: the `clippy --all-targets --all-features` pass above builds the `trino-conformance` test and both live-test files — that is how an unused import in the Trino test surfaced. The `#[ignore]` live tests and the Docker Trino test were run by hand; their output is above.

## Mutation checks

Both via `scripts/mutation-check.sh` (refuses on a dirty tree), re-run against `c5b556ab` after the final suite, with the tree clean before and after.

**1. Flip DuckDB's stated rule to `Backslash`** — the executed round trip must fail, not only the pin:

```
test the_duckdb_dialect_states_the_standard_rule ... FAILED
test the_duckdb_dialect_literal_round_trips_byte_identical ... FAILED
test the_duckdb_dialect_literal_matches_the_same_value_stored_in_a_column ... FAILED
test the_backslash_rule_is_wrong_on_duckdb ... ok

DuckDB rejected `SELECT 'it\'s' AS v`: DuckDB error: Parser Error: unterminated quoted string at or near "' AS v"
DuckDB rejected `INSERT INTO probe VALUES (0, 'it\'s')`: DuckDB error: Parser Error: syntax error at or near "s"

test result: FAILED. 1 passed; 3 failed; 0 ignored; 0 measured; 0 filtered out
```

**2. Move the backslash pass from first to last in the `Backslash` chain:**

```
literal::tests::backslash_encodes_a_backslash_quote_pair_without_reopening_the_bypass
  left: "'a\\\\\\\\'b'"   right: "'a\\\\\\'b'"
literal::tests::backslash_escapes_a_quote_with_a_backslash
  left: "'it\\\\'s'"      right: "'it\\'s'"
literal::tests::backslash_encodes_a_line_break
  left: "'a\\\\nb'"       right: "'a\\nb'"
literal::tests::backslash_keeps_a_literal_backslash_n_distinct_from_a_line_break
  assertion `left != right` failed

test result: FAILED. 21 passed; 4 failed; 0 ignored; 0 measured; 249 filtered out
```

The mutant emits four backslashes before a bare quote for input `a\'b`, which is exactly the even-run bypass the comment describes.

## Red team

Reviewer: **Codex `gpt-5.6-sol`** (`codex exec --sandbox read-only`, reasoning effort high), which itself consulted an independent second model (`gpt-5.6-terra`). Two rounds. Round 1: 103 tool calls. Round 2: 41 tool calls. Round 2 verdict: **no blocking finding**.

| # | Round | Finding | Disposition |
|---|---|---|---|
| 1 | 1 | **BLOCKING** — `Backslash` passed a raw line break through, which BigQuery rejects outright. `\n` / `\r` are documented on all three `Backslash` dialects, so this needs no third rule. | **Fixed.** `Backslash` now encodes LF and CR. Re-executed all four live dialects afterwards. A new unit test pins that a literal `\`+`n` in the value stays distinct from a real line break. |
| 2 | 1 | **BLOCKING** — the scaffold emitted `LiteralEscape::Standard`, silently supplying the default the trait forbids. A TODO is not a forced choice. | **Fixed.** The scaffold emits `compile_error!`. Verified with `rustc` that it produces one error and no type-mismatch noise. |
| 3 | 1 | The live Snowflake / Databricks oracle coerced a returned SQL `NULL` to `""`, so the empty-value case could pass on a wrong result. | **Fixed.** Both accept only `Value::String`. Both re-run live. |
| 4 | 1 | Requiredness had no mutation guard: adding a default would leave every test green. | **Fixed.** Added the `trybuild` `E0046` case described above. |
| 5 | 1 | Pre-existing: the render tests do not establish a *compilable* scaffold — the template declared neither `rocky-ir` nor `chrono` and `merge_into` had a stale signature. | **Fixed** both. Actually compiling a generated fixture is still not done — stated as a limit, not claimed. |
| 6 | 1 | The replace-order comment said three backslashes; quote-first actually yields four. | **Fixed.** The mutation output above confirms four. |
| 7 | 2 | "no error path … any `&str`" overstates: a value can still exceed BigQuery's 1 MB statement limit. Fails loudly, not silently. | **Fixed.** The claim is now scoped to the lexer. |
| 8 | 2 | The guarantee is core-trait-only; an out-of-tree SDK adapter still compiles without answering. | **Fixed** in the skill and the adapters doc. The SDK omission itself stays deliberate — see above. |
| 9 | 2 | "the only thing stopping the crate from compiling" is not literally true — workspace membership also stops it. | **Fixed** in the changelog and the doc. |
| 10 | 2 | The Trino corpus was missing `\'; DROP TABLE t; --`, which the other three carry. | **Fixed** and Trino re-run live. |

Round 2 confirmed independently: the encoder is total for the challenged characters (U+2028/U+2029 and other surrogate-free characters pass through unchanged; BigQuery rejects raw LF/CR, which the encoder no longer emits); the regex two-stage decode is the identity in both directions, so PR2's BigQuery `r'…'` → `'…'` conversion hands re2 the same bytes; the three render tests reject `Standard`, `todo!()` and `unimplemented!()`; and no production SQL path was edited.

## The two standing questions

**What am I least confident about?** That BigQuery accepts `\'` and rejects `''`. It is the one rule nobody can execute — the sandbox is expired — and it is the rule that decided the whole design. It is derived from the GoogleSQL lexical page, which lists `\'` and says an unlisted escape errors. It is labelled doc-derived in the code, in the changelog and in this table. What settles it: one `SELECT` against a live BigQuery project. Until then the direction of harm is the safe one — if the spec is wrong about `''`, `Backslash` is still valid, because `\'` is documented either way. Second, smaller: a raw NUL or U+2028 as literal SQL source is not in the executed corpus on any warehouse. No sink can produce one today (TOML forbids raw control characters), but that is an argument, not a test.

**What is the most important thing I'm missing?** The sink diff. This PR has no production caller, so nothing here proves that the twenty sinks PR2 routes will use the wrapper correctly — an encoder nobody calls is guarded only by its own tests. The related gap that PR2 must decide, not this one: `spark.sql.parser.escapedStringLiterals=true` turns Databricks' backslash reading off, and it breaks both ways — `\'` becomes a parse error and `\\` becomes silent double-backslash corruption. The live test ran under the default. There is no policy for that session flag yet, and PR2 routes Databricks sinks.

## Not in this PR

`rocky-adapter-sdk` (see above). Every sink. The DSL lowering path in `rocky-lang`, which has no dialect by construction. `#1597` is already closed as a duplicate of `#1596`; this closes neither — `#1596` closes when A1-PR3 lands.

Refs #1596


hugocorreia90:
Follow-up filed: #1607 — a test that renders the scaffold and actually compiles it. The three render tests added here check text, and their required-method list is hand-maintained, so trait drift can still ship a broken `rocky init-adapter` crate. Stated as a limit in the PR body; #1607 is where it gets closed.

Issue #1609, opened 2026-09-03
rocky run: config refusals land after the destination adapter and the state store are already open


hugocorreia90:
`rocky run` refuses some configurations before it touches the warehouse — the target-collision preflight (#1461), and now the `metadata_columns` value guard (#1594). Both land before any catalog, schema, tag, binding, grant or table copy.

Two things still happen earlier, on every run whatever the outcome:

```
config → REGISTRY (opens the destination adapter) → state store → ... → preflight refusal
             │                                          │
             └─ DuckDB: creates the database file       └─ commits version + tables
                                                           on a fresh store
                            ... and on the way out: the end-of-run retention sweep
                                runs on the Err path too, deleting expired history
                                rows and stamping last_retention_sweep_at
```

Neither is a governance or data mutation, and both predate the refusals:

- Adapter construction creating the DuckDB file is shared with `rocky discover`, a read-only command.
- The retention sweep on `Err` is deliberate — its comment says a failed run should not postpone state cleanup.

So this is not a bug in either guard. It is the gap between what a refusal *ought* to promise ("nothing happened") and what it can promise today.

## What would close it

1. Split registry construction so the destination adapter is not opened until after the preflight, or make the DuckDB adapter open lazily.
2. Decide whether a preflight refusal should skip the retention sweep. The current answer is deliberate; the question is whether a refusal that never started the run is the same thing as a run that failed.
3. Give the tree a recording adapter, so a test can assert "this run issued zero catalog / tag / binding / grant calls" instead of inferring it from one DuckDB observation.

Point 3 is the load-bearing one. Three review rounds on #1594 turned on *when* something happens, and every time the only evidence available was reading control flow plus a single DuckDB schema check.

Raised by the independent review of #1594; scoped out of that PR deliberately, because changing run's startup ordering touches every command and every failure path.

hugocorreia90:
## Proceeding on the recommended default: build the recording adapter first

Taken under the standing "proceed on the recommended default" authority — the substance here is test infrastructure, so it did not warrant a blocking question. Say so and I will change it.

Point 3 in the issue is the load-bearing one and goes first: **a recording adapter**, so a test can assert "this run issued zero catalog / tag / binding / grant calls" instead of inferring it from one DuckDB schema observation. Three review rounds on #1594 turned on *when* something happens, and every one of them had only control-flow reading as evidence.

With that in place, points 1 and 2 become measurable rather than arguable:
[...]

Issue #1607, opened 2026-09-03
rocky init-adapter: nothing compiles the scaffold, so the template can drift from the traits


hugocorreia90:
`rocky init-adapter <name>` writes a crate from template strings in `engine/crates/rocky-cli/src/commands/init_adapter.rs`. Nothing in CI ever compiles what it writes. The `#[cfg(test)] mod tests` inside the template is text, not code.

That let two defects sit unnoticed on `main` until #1605:

- the emitted `Cargo.toml` used `rocky_ir` and `chrono` without declaring either;
- the emitted `merge_into` took `&[String]` while `SqlDialect::merge_into` takes `&[std::sync::Arc<str>]`.

#1605 fixed both, and added three render tests over a new `render_dialect_rs`. Those tests check **text**, not compilation, and the required-method list in `the_scaffolded_dialect_emits_the_methods_this_list_names` is hand-maintained. A required method added to `SqlDialect` and forgotten in both places still ships a broken scaffold.

**What would close this:** a test that renders the scaffold into a temp directory and compiles it, so trait drift fails in CI rather than in a new contributor's terminal.

Two things it has to handle:

1. The scaffold's `Cargo.toml` uses `edition.workspace = true` and `{ workspace = true }` dependencies, so it cannot build standalone — the fixture needs a generated workspace root, or the template needs pinned values under test.
2. `literal_escape` is emitted as a deliberate `compile_error!` (#1605). The fixture must substitute a variant before compiling, and should assert that the unsubstituted form *fails* — that refusal is the feature.

Cheap and worth doing at the same time: `run_init_adapter` resolves its output against a relative `Path::new("crates")`, so it can only be run from the engine directory and cannot be pointed at a temp dir. Taking a root path would make it testable end to end.

Issue #1461, opened 2026-08-17
bug(engine): --shadow-schema collapses every replication connector into one schema, silently overwriting same-named tables


hugocorreia90:
`rocky run --shadow --shadow-schema <X>` on a multi-connector replication pipeline collapses every connector into one schema, so same-named tables overwrite each other. The run reports success for both.

Found while probing #1273 fix shapes; reproduced from scratch on 1.70.1.

## Reproduction

Config — the `schema_template` is what keeps connectors apart:

```toml
[pipeline.poc]
strategy = "full_refresh"
source.schema_pattern = { prefix = "raw__", separator = "__", components = ["source"] }
target = { catalog_template = "poc", schema_template = "staging__{source}" }
```

Two sources, each with an `orders` table:

```
raw__shopify.orders  ->  3 rows
raw__stripe.orders   ->  1 row
```

**Baseline — normal run.** Correct: the template separates them and both survive.

```
copied: 2  failed: 0  status: Success
  staging__shopify.orders  ->  3 rows
  staging__stripe.orders   ->  1 row
```

**With shadow isolation:**

```
$ rocky run --shadow --shadow-schema shadow_x
copied: 2  failed: 0  status: Success
  materializations: duckdb.shopify.orders, duckdb.stripe.orders
```

What actually exists afterwards:

```
shadow_x.orders  ->  1 row
```

**One table, one row.** Two tables were reported copied; one connector's data overwrote the other's. No error, no warning, exit 0.

## Cause

`run.rs:2698-2704` takes `schema_override` verbatim and discards the template:

```rust
let target_schema = if let Some(shadow_cfg) = shadow_config {
    shadow_cfg
        .schema_override
        .clone()
        .unwrap_or_else(|| parsed.resolve_template(target_schema_template, target_sep))
} else {
    parsed.resolve_template(target_schema_template, target_sep)
};
```

This sits inside the per-connector loop, so every connector resolves to the same literal schema. `{source}` — the only component distinguishing them — is dropped precisely when isolation is requested. The replication path has no collision check, so two writes to one identity are indistinguishable from two writes to two.

## Why this is worse than a wrong table name

`--shadow` exists to give `rocky compare` a trustworthy baseline. Here the shadow output is silently *wrong*, and `tables_copied: 2` asserts otherwise — so a comparison run validates against corrupted data and reports clean. The failure is invisible at every layer that would normally catch it: exit code, counts, materialization list (which still names both connectors), and the error array.

[...]

PR #1610, opened 2026-09-03
fix(engine): validate metadata_columns[].value at the IR boundary — OWNER MERGE

### LINKED ISSUE #1594: metadata_columns value is spliced into the replication SELECT unchecked
`[pipeline.<name>] metadata_columns` takes `{ name, type, value }`. The `name` is validated as an identifier and the `type` goes through `validate_sql_type`. The `value` is spliced into the generated SELECT as written:

```
, CAST({value} AS {data_type}) AS {name}
```

Every warehouse dialect does this (`rocky-databricks/src/dialect.rs`, `rocky-snowflake/src/dialect.rs`, `rocky-trino/src/dialect.rs` and the others). The Trino comment says so out loud: "the trusted `value` expression is intentionally left as-is".

So a `value` that closes the `CAST(` can add SQL to the replication copy statement, which runs with the project's warehouse credentials. This is the same class as #1524, on a different field.

#1524 added `rocky_sql::validation::reject_statement_terminator` for exactly this shape. Applying it to `value` would bound it to a single expression. It was left out of that PR on purpose: the fix would touch every adapter crate's `select_clause`, which would make a focused security change much harder to review.

Two things to decide:

- Where the check goes. Each dialect's `select_clause` already calls `validate_sql_type` on `data_type`, so a call beside it is the consistent spot — but that is five call sites and a default trait method, and a new adapter would have to remember. Validating once where `MetadataColumn` is built from `MetadataColumnConfig` is one place, but a direct constructor then bypasses it.
- Whether `value` should be bounded further than "one expression". `value = "NULL"` and `value = "current_timestamp()"` are the real uses. An allowlist would be tighter, and would also close the read primitive that a terminator check does not.

Same limitation applies as in #1524: refusing a terminator stops statement injection. It does not stop a subquery or a DuckDB `read_csv` call inside the expression.

hugocorreia90:
`[[pipeline.X.metadata_columns]] value` was spliced raw into `CAST({value} AS {type}) AS {name}` by every adapter. `name` and `data_type` were validated; `value` was not.

It is now validated at one boundary, and the two kinds of text in it get two different rules. The refusal lands before any catalog, schema, tag, binding or grant is written.

Closes #1594

## The path

```
rocky.toml                                warehouse
value = "'{tenant}'"                      schema name "src__acme__shopify"
     │                                          │
     │                                          ├─ SchemaPattern::parse
     │                                          │    splits on the separator,
     │                                          │    checks NOTHING
     │                                          ▼
     │                                    component "acme"
     │                                          │
     │                                          ├─ validate_identifier   ◄── NEW
     │                                          │   warehouse text was never
     │                                          │   an expression
     ▼                                          ▼
  rocky_core::schema::resolve_metadata_columns   ── the ONE producer
     │                                              (rocky plan + rocky run)
     ▼
  rocky_ir::MetadataColumn::new
     ├─ validate_identifier(name)
     ├─ validate_sql_type(data_type)
     └─ reject_statement_terminator(value)   ◄── NEW, the author's own text
     │
     ▼
  <dialect>::select_clause  ── repeats all three scans (defence in depth)
     │
     ▼
  CAST({value} AS {type}) AS {name}
```

Ordering inside `rocky run`:

```
config → registry → list_tables (read) → PREFLIGHT ← the refusal lands here
                                             │
                                             ▼
                            catalogs, schemas, tags, bindings, grants
                                             │
                                             ▼
                                     table collection → copy
```

## Why `value` was not trusted

The Trino comment called it "the trusted `value` expression". It is not.
`resolve_template` fills its `{placeholder}`s from **source schema names read
back from the warehouse**, and `SchemaPattern::parse` splits those on the
configured separator with no identifier check.

A statement-terminator scan alone does not bound that text. The scan tracks no
parentheses, so a component free of `;` with balanced quotes can still reshape
the expression inside the `CAST`. Hence the split: the author's own template
keeps Rocky's fragment rule (one statement, reads unbounded), and every
substituted component must be a plain `[A-Za-z0-9_]+` identifier.

**Severity, stated honestly.** Round 3 of the review could find no in-tree
executed path where such a component reaches SQL: `format_table_ref` refuses
the same non-identifier schema name one frame later, when it builds the `FROM`
clause of the very statement the value lands in. So this is defence in depth
plus an ordering improvement, not the closing of a live end-to-end hole. The
end-to-end test's control run demonstrates exactly that.

## The six plan corrections, and where each was wrong

| # | Correction | Disposition |
|---|---|---|
| 1 | `#[non_exhaustive]`, do not privatise fields | **Partly wrong.** `#[non_exhaustive]` does block a cross-crate struct literal — the compiler proved it — but `pub` fields still allow `col.value = hostile;` on a benignly-built one, so "every construction path crosses validation" was false. Fields are now **private** with `name()` / `data_type()` / `value()` readers. `#[non_exhaustive]` kept. |
| 2 | 22 construction sites outside `rocky-ir` | **Confirmed**, exactly 22 (some line numbers had drifted; located by grep). Benign fixtures go through `new(…).unwrap()`; hostile ones through `new_unchecked`. |
| 3 | Add `new_unchecked` for tests, and therefore also scan `value` inside each `select_clause` | **Confirmed and taken.** One correction: there is **one** `select_clause_rejects_injection_in_metadata_type_and_name`, in Trino — not four. BigQuery, Databricks and Snowflake had benign metadata tests only, and DuckDB had no `select_clause` test at all. All five now have a hostile-value test. `new_unchecked` sits behind a `test-support` feature, not `#[doc(hidden)]` — a `pub fn` with a hidden doc is documentation, not access control. |
| 4 | `validate_sql_type` must move to `rocky_sql::validation`; `rocky-ir` cannot import `rocky-core` | **Confirmed** (`rocky-ir/Cargo.toml` depends on `rocky-sql` only). One deviation: `sql_gen::validate_sql_type` **delegates** rather than re-exports, because `drift.rs` propagates its `SqlGenError` with `?` and a re-export would change the error type there. One rule, one body, zero caller churn. |
| 5 | `Deserialize` is a second constructor; no JSON-schema cascade | **Confirmed.** `#[serde(try_from)]` added with tests. `just codegen` produced zero drift (the only diff was the known `libc`-lines-only `package-lock.json` artifact, reverted); `./scripts/regen_fixtures.sh` produced zero drift. Also found a fourth deserialization site the plan did not list: `replay.rs:259`. |
| 6 | Do not turn the `debug_assert!` into the constructor call | **Confirmed.** The constructor scans the *rendered* value (`md5(cast(coalesce(cast(<col> …`), which carries no terminator even for a hostile input column. `apply_surrogate_keys` is now fallible and runs `validate_surrogate_key_spec` in **every** build; propagated at `emit_sql.rs`, and both `run.rs` call sites. |

Also done, as asked: the "trusted `value`" comments in Trino and BigQuery are
replaced with the true statement, and the `init-adapter` scaffold emits all
three guards with a render test.

`ProcessDialect::select_clause` in `rocky-adapter-sdk` is different from what
the plan expected. The SDK **cannot** call the validators: it has no
`rocky-sql` dependency, deliberately (`traits.rs` lists that as the reason
`clone_table_for_branch` is not in the SDK either). Re-implementing the
fragment scanner there would give one rule two bodies. So it **refuses**
metadata columns instead — the same fail-closed shape `merge_into` beside it
already uses. Nothing in-tree reaches it: `rocky_adapter_sdk::MetadataColumn`
is a separate struct from `rocky_ir::MetadataColumn` and no code converts one
to the other. It is a behaviour change for an out-of-tree process adapter that
passes metadata columns: from silent unvalidated splicing to a loud refusal.

## Cost — please read this part

**1. What an author may no longer write in a `value`.** The #1595 scanner
refuses, in these lexical positions:

| Refused | Where |
|---|---|
| `;` | outside a quoted region or comment |
| a backtick | outside a quoted region or comment |
| a `//` or `#` line comment | outside a quoted region or comment |
| a scanner-recognized dollar-quote opener (`$$`, `$tag$`) | outside a quoted region or comment. A bare `$` inside a word (`col$1`) is still fine |
| a triple quote | at the start of a quoted region |
| a backslash | **only inside** a single- or double-quoted region |
| a nested `/*` | inside a block comment |
| an unterminated `'…'`, `"…"` or `/* … */` | anywhere |

So `'C:\x'` and a backtick-quoted BigQuery identifier are refused where they
used to work. Loud, named, and fixable in one line.

**2. What a source schema may no longer contain.** Any component that is not
`[A-Za-z0-9_]+` is refused **when, and only when, it is substituted into a
metadata value**. That includes an *empty* component from adjacent, leading or
trailing separators. Target catalog/schema templates and every other
`resolve_template` consumer are untouched.

**3. Persisted IR.** A stored `ModelIr` whose metadata `name`, `data_type` or
`value` is now refused stops deserializing — the whole record, not just the
field. Three commands read one; all three are fail-closed and were checked:
`rocky replay` refuses that record, `rocky restore` likewise, and `rocky gc`'s
`storage_prefix_from_provenance` returns `None`, which **holds** the eviction.
No deletion path results.

**4. Rust API.** `rocky_ir::MetadataColumn` is no longer constructible,
mutable, destructurable or field-readable by struct syntax from another crate —
use `new`, `name()`, `data_type()`, `value()`.
`rocky_core::models::surrogate_key_metadata_columns` and `apply_surrogate_keys`
changed signature (both fallible now; the former also takes a model name).
`rocky_sql::validation::ValidationError` and `rocky_core::schema::SchemaError`
each gained variants, which breaks an exhaustive downstream `match`.

**5. Adapter SDK.** `ProcessDialect::select_clause` now refuses any non-empty
metadata list. Not reachable through the engine registry, but externally
observable through the public SDK.

## Where the refusal lands — and what it does not promise

The refusal runs in `run.rs`'s existing pre-mutation preflight, lazily for the
first table that survives the same filter / missing-from-source / `enabled =
false` skips the collection loop uses. So it lands before every **warehouse
governance** operation — catalog creation, schema creation, tags, workspace
bindings, grants — and before any table copy.

It does **not** precede two things that happen on every `rocky run` whatever
the outcome, neither of which this guard causes:

- **The destination database file.** Adapter construction opens it, which
  creates it when absent on DuckDB. Probed: `rocky discover` — a read-only
  command — on a fresh directory creates the file, while `rocky run --pipeline
  <unknown>` creates none, because config resolution refuses before the
  registry is built.
- **The state store, and the end-of-run retention sweep.** The store opens
  before all of `run`'s command logic, and the sweep runs on the `Err` path by
  explicit design (its own comment: a failed run should not postpone state
  cleanup). An idempotency claim is taken before the preflight too; on failure
  `dedup_on = "success"` deletes it and `dedup_on = "any"` keeps a terminal
  `Failed`. Under a `[policy]` block a `PolicyDecisionRecord` may also be
  written — a record of an attempted evaluation, not of a run.

There is **no** `RunRecord`, watermark advance, resume checkpoint or
remote-state upload: those sites are all later, and the error path abandons the
remote-state session.

The decisive point: the target-collision refusal already living in this same
preflight block (#1461) has exactly these side effects too. If this guard leaks
state, so does that one, and so does every other `?` in `rocky run`. Changing
that is a run-lifecycle refactor touching every command — filed as #1609, not
attempted here.

## Merged with #1605

`#1605` ("make SQL literal escaping a dialect-owned fact") landed on `main`
while this branch was in review, and it edits the same scaffold template. One
file conflicted: `commands/init_adapter.rs`. The resolution keeps main's shape
and layers this branch's guards on top:

- main's `render_dialect_rs` name and doc, plus this branch's sentence on why
  the template needs a render test;
- main's `LiteralEscape` import and its `literal_escape` block with the
  `compile_error!` that forces the author to choose;
- main's `merge_into` signature fix (`&[Arc<str>]`) — which closes the scaffold
  defect round 3 flagged here as pre-existing;
- this branch's `select_clause`: identifier check on explicit columns, and
  name / type / value validation on every metadata column, through the
  accessors the private fields now require.

Both render tests are kept; all five `init-adapter` tests pass.

**This branch routes no sink through `encode_string_literal`.** That is lane
A1-PR2's job. Nothing here builds a SQL literal — the change is validation and
ordering only.

**#1607 applies to the scaffold change.** Nothing in CI compiles the generated
crate, so a scaffold edit can drift from the traits with nothing noticing. The
render test here asserts on the emitted *text*, which catches a dropped guard
but cannot catch a signature drift. A human reading it is the only other check.

Every gate below was re-run on the merged tree; the pre-merge run does not
cover what is shipping.

## Gates

All re-run on the **merged** tree at `5114b640`; `git status` empty before and
after each run, `HEAD` unmoved.

| Gate | Result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --all-targets --all-features -- -D warnings` | exit 0, zero findings |
| `cargo test --workspace --no-fail-fast` | **6596 passed, 0 failed, 114 ignored** across 150 suites, exit 0 |
| `just codegen` + `git status --porcelain` | zero drift (`package-lock.json` `libc`-lines-only artifact reverted) |
| `./scripts/regen_fixtures.sh` + drift check | zero drift, 40 fixtures |

The IR goldens that would catch a serialization change all pass:
`ir_json_matches_snapshot`, `ir_round_trip_byte_stable`, `ir_recipe_hash_pinned`,
`frozen_provenance_records_stay_replayable`. `Serialize` is untouched.

One note on evidence hygiene: an earlier full-suite run reported one failure in
`rocky-core-compiletest` with `error[E0514]`. That was a poisoned target dir —
a review process had built in it with the default toolchain (1.96.0) instead of
the pinned 1.98.0. The suite was re-run after `cargo clean`; the numbers above
are from that clean run. Two other earlier failures
(`fires_due_cron_each_tick_without_drift`, `upload_state_succeeds_when_endpoint_is_prompt`,
the latter "took 1.02510625s (expected <1s)") were load flakes at load average
20-39 with three concurrent cargo jobs; both pass in isolation and both pass in
the clean run.

## Mutations

`scripts/mutation-check.sh`, eight mutations, every one
`OK — the test failed under mutation, so it is fix-sensitive`. All eight were
re-run on the merged tree.

| Mutation | Tests that failed |
|---|---|
| drop the constructor's `reject_statement_terminator` | 5 `rocky-ir` tests, including both serde ones |
| drop the per-component `validate_identifier` | 3 `rocky-core` schema tests |
| drop Trino's inline value scan | `select_clause_rejects_a_hostile_metadata_value` |
| drop BigQuery's | same |
| drop Databricks's | same |
| drop DuckDB's | same |
| drop Snowflake's | same |
| drop the `run.rs` preflight resolution | `a_refused_metadata_value_lands_before_governance_setup`, on the assertion "the refusal must land BEFORE governance setup" |

The component mutation is the informative one. With the check removed, the
mutated run prints the value it built —
`MetadataColumn { name: "_tenant", data_type: "VARCHAR", value: "'a' || 'b'" }`
— which is the escape shape: semicolon-free and quote-balanced.

## Red team

Independent reviewer: **Codex (`gpt-5.6-codex`)**, five rounds, each running
its own second-model pass (Boole, Darwin, Carver — all `gpt-5.6-terra`).
Rollout audit: 66, 74, 28 and 31 `custom_tool_call` entries on the substantive
rounds — real work, not receipts. Round 5 returned **no blocking finding** and
`ModelIr / contract semantics preserved: yes`. Round 6 reviewed the merge
resolution above.

Log-integrity note: this session's scratchpad is shared across lanes, and
round 1's transcript was written to a bare path and picked up interleaved
output from a sibling lane. Its `workdir:` header names this worktree and its
verdict block cites only this worktree's paths, so the round-1 findings are
this diff's. Rounds 2-6 were written under a lane-unique path and carry zero
sibling citations.

| Round | Finding | Disposition |
|---|---|---|
| 1 | BLOCKING: the value guard does not enforce a single-expression boundary | **Fixed.** Per-component identifier rule + the single producer. |
| 1 | BLOCKING: the construction boundary is publicly bypassable (`pub` fields, `pub` `new_unchecked`) | **Fixed.** Private fields + accessors; `new_unchecked` behind `test-support`. |
| 1 | `ProcessDialect` is a public-contract change | **Disclosed**, kept. Rounds 2 and 3 both agreed refusing is the right direction. |
| 2 | BLOCKING: the refusal landed after catalogs, schemas, tags and grants | **Fixed.** Moved into the pre-mutation preflight; end-to-end test added. |
| 2 | BLOCKING: the `init-adapter` scaffold no longer compiled (field access) | **Fixed.** Accessors emitted; render test updated; `rocky-ir` + `chrono` declared. |
| 2 | plan/run parity for disabled tables | **Fixed.** `plan` resolves after the exclusion branches. |
| 3 | BLOCKING: "nothing mutates the warehouse before preflight" is false — adapter construction creates a missing DuckDB file | **Rebutted; accepted in round 4.** Probes above; claim narrowed to warehouse governance; residual disclosed; #1609 filed. |
| 3 | TEST-GAP: the test's boundary is weaker than its name | **Accepted as a limit.** Test renamed and also asserts the target table is absent; the module header states what DuckDB cannot observe (no catalogs) and that a recording adapter would settle it. |
| 3 | the generated scaffold still does not compile (`merge_into` signature) | **Accepted, not fixed.** Pre-existing, more than one mismatch, and unverifiable without a compile test — a blind signature edit would move the false claim. Stated plainly; part of #1609. |
| 3 | the cost disclosure is incomplete | **Fixed.** See the five-part cost section. |
| 4 | BLOCKING: state changes outlive the refused run (store open, retention sweep on `Err`) | **Rebutted; accepted in round 5.** Every one of those side effects is shared byte-for-byte with the #1461 collision refusal in the same block, and the sweep-on-`Err` is deliberate. Claim narrowed to warehouse governance; test renamed and `--state-path` pinned; #1609 filed. |
| 4 | test claim too wide; incomplete reset; non-discriminating table assertion | **Fixed.** All three. |
| 4 | disclosure still incomplete (restore, `name`/`data_type`, field reads, signatures, `ValidationError`) | **Fixed.** |
| 5 | LOW: say "warehouse-governance"; name the idempotency claim and `PolicyDecisionRecord`; add `SchemaError` variants, triple-quote openers, the SDK refusal; lexical framing | **Fixed** in this body. |
| 5 | NIT: the SQL-type doc states an absolute safety property | **Fixed.** The doc now says it is an allowlist, not a type grammar. |

## Out of scope, on the record

Same posture as #1595: refusing a terminator bounds the fragment to one
statement; it does not bound reads. A `value` the project author writes can
still be a correlated subquery or a DuckDB `read_csv`. That is the documented
posture for every SQL fragment Rocky takes from config, not a gap specific to
this field — and it is why warehouse-derived substitutions get the stricter
identifier rule instead.

## The two standing questions

**What am I least confident about?** That the per-component identifier rule
costs nothing real. Round 3 found one false-refusal shape I had not: an *empty*
component, produced by adjacent, leading or trailing separators in a source
schema name. That schema shape is already broken for the copy path
(`format_table_ref` refuses it), so I believe the cost is zero in practice —
but that is an inference from reading, not from a project that hits it. A user
report of a refused `metadata_columns` value on a schema that used to work
would settle it.

**What is the most important thing I am missing?** A recording adapter. Three
of the five review rounds turned on *when* something happens rather than
whether it happens, and every time the only available evidence was reading
control flow plus one DuckDB observation. The tree has no way to assert "this
run issued zero catalog, tag, binding or grant calls". That is the test I could
not write, it is why the ordering claim here is scoped to what the DuckDB
fixture can actually see, and it is the load-bearing item in #1609.


hugocorreia90:
**OWNER MERGE — why I stopped instead of merging.**

My dispatch said this lane was not owner-gated. I am not merging it, and the reason is on the record rather than an oversight.

The plan's authority section reserves for you anything that changes a public contract, and names *"refusing a config that works today"* as the shape. This PR does exactly that: a `metadata_columns` `value` holding a backslash inside quotes (`'C:\x'`) or a backtick runs today on every dialect and is refused after.

The plan disclosed that cost inside A2 and still placed A2 in the ungated part — so the plan is internally inconsistent here. The tiebreaker is #1595: the identical trade-off, on other config fields, was owner-gated and owner-merged one PR ago. Same judgement call, same field class, so it is yours.

Everything else is done: CI green, six red-team rounds with the last two returning no blocking finding, eight mutations re-confirmed on the merged tree, `just codegen` and `regen_fixtures.sh` both zero-drift.

Two things to read before merging: the **Cost** section (five parts), and the residual in **Where the refusal lands** — the refusal precedes every warehouse governance operation but not adapter construction or the state store, which is #1609.

PR #1624, opened 2026-09-03
OWNER MERGE fix(engine): refuse the inert [schema_evolution] block and wire project freshness inheritance

### LINKED ISSUE #1435: Two rocky.toml sections parse but nothing consumes them: [schema_evolution] drop/grace and project [freshness] inheritance

hugocorreia90:
Two `rocky.toml` sections parsed, validated, and did nothing. A user could configure either and get silence. One is now refused with a remedy; the other is wired.

**This PR was recovered, not authored end to end.** The agent that wrote it died on an API error before opening a PR, leaving six commits on one machine with no trace in the tracker. I pushed the branch, merged `main`, and re-ran every gate myself. The red-team disposition below is from its own rounds; where I could not confirm something, I say so.

## `[schema_evolution]` drop / grace — refused, not wired

`detect_drift` hard-returned `columns_to_drop: Vec::new()` and `grace_period_columns: Vec::new()`, so the block could never take effect.

Wiring it is a **data-destructive default change with no opt-in**, so this PR removes the dead surface instead — and makes the removal **loud**. A config that still sets those keys is now rejected with a message naming the remedy, rather than accepted and ignored. Replacing one silence with another would have been the same defect.

**The finding worth carrying forward:** the persistence layer for wiring it *already exists and is entirely unwired* — a `grace_periods` redb table with `GracePeriodRecord`, `set_` / `get_` / `list_` / `delete_grace_period`, listed in `EXPECTED_TABLES` and eagerly created. `detect_column_drops` already takes those records as a parameter. Neither the issue nor its four comments mention it. That makes the wiring far cheaper than recorded, and it deserves its own issue rather than being lost here.

## Project `[freshness]` inheritance — wired

A model with no `[freshness]` sidecar now inherits the project block. `ModelFreshness::from_project_default` existed with **zero callers** and a doc comment asserting behaviour that did not happen; it is wired now rather than left as a claim.

## Two fixes beyond the brief

The lane found and fixed two things nobody asked for, both real:

- **`rocky emit-sql` could emit another warehouse's SQL when `rocky.toml` did not load.** It swallowed the config error and continued against a default.
- **The LSP lost `[mask]` and the project `[freshness]` block after the first keystroke.** Opening a project read them; the first edit dropped them.

It also closed two paths that bypassed the new refusal and the new inheritance — the kind of gap that makes a guard look present while a caller routes around it.

## Evidence

| Gate | Result |
|---|---|
| `cargo fmt --all -- --check` | pass |
| `cargo test --workspace --no-fail-fast` | **6626 passed, 0 failed** |
| build after the `main` merge | clean |

Gates ran on the post-merge tree. The only merge conflict was `engine/CHANGELOG.md`, where both sides appended under `[Unreleased]`; both sets of bullets are kept, none dropped.

Removing config fields regenerates the project JSON schema and the Pydantic and TypeScript bindings — those are committed.

## Upgrade impact

A project that sets `[schema_evolution]` drop or grace keys will now fail to load, with the remedy in the message. Those keys never did anything, so no behaviour is lost — but the failure is new and someone can be surprised by it.

## Standing questions

**Least confident:** whether the refusal message is the right shape for someone who set those keys years ago expecting drops to work. It names the remedy, but "this never did anything" is an awkward thing to learn from an error.

**Most important thing missing:** the `grace_periods` table is left in place, unwired. Keeping a built, migrated table for work that is not scheduled is a bet that the work happens. If the wiring issue is closed as won't-do, the table should go with it.

Closes #1435


hugocorreia90:
## Red team: Codex (GPT-5), read-only, 45 tool calls. Verdict: **BLOCK**, and it was right.

This PR had no completed review — the agent that wrote it died on an API error mid-round — so this was the first full pass over the diff.

### The blocking finding

The PR said it "closed two paths that bypassed the new refusal". **Five more exist.**

```
  invalid rocky.toml
    run / plan / compile / validate / emit-sql   ->  refuses, shows the remedy
    rocky lineage                                ->  .ok() -> unwrap_or_default() -> compiles
    MCP plan_preview                             ->  .ok() -> DuckDB dialect -> emits SQL
    MCP compile_full                             ->  compiles with defaults
    LSP (initial + didChange)                    ->  defaults, no diagnostic
    serve POST /api/v1/compile                   ->  degrades, returns success
```

I verified the sharpest one rather than taking it on trust: `lineage.rs` does `.ok()` then `.unwrap_or_default()` — the identical fail-open this PR fixed in `emit-sql`.

The review's concrete trigger: a Snowflake project with an invalid section, called through MCP `plan_preview`, returns **DuckDB SQL** and no word that anything was wrong. And `POST /api/v1/compile` returns success *unconditionally* after degrading, which makes the SDK internally inconsistent — its `GET` route is strict, its `POST` route is not.

### What changed here

The changelog claimed the refusal fires "on every command". That was false, and it is the exact defect this PR exists to fix: a guarantee stated more strongly than the code keeps it. Corrected in `0a6fe040` to name which commands surface the refusal and which swallow it.

**No code change.** Everything the review verified as sound is untouched.

### What the review confirmed as correct

- The sentinel deserializer always refuses the section with an actionable deletion message; absence stays valid.
- Freshness precedence is whole-block and correct — sidecar, then directory defaults, then project — and sidecar fields do not fall through.
- Full and incremental compilation both receive project freshness; W005 keeps its existing semantics.
- `tick` deliberately collects only declared budgets; the reconciler still takes the minimum declared and otherwise falls back to the project value.
- `emit-sql` tolerates only a missing config and uses credential-tolerant loading otherwise.
- The generated JSON, Python and TypeScript surfaces no longer advertise `schema_evolution`, and the changelog merge lost and duplicated nothing.

### Filed, not fixed here

**#1625** — one shared fallible project loader where only `FileNotFound` becomes `None`, used by every model-resolution and preview entry point.

Fixing five callers inside this PR would repeat the mistake it was blocked for: this is the fifth through seventh instance of the same defect, precisely because each caller decides independently what a `ConfigError` means. A per-caller patch leaves the next entry point inheriting nothing.

One question left open rather than decided: whether degradation is *intentional* for the LSP and `serve`. An editor staying usable with a broken config is defensible — but it should be a stated contract, not an accident of `.ok()`, and the refusal should surface as a diagnostic instead of vanishing. That was the reviewer's own least-confident point and it is not answerable from the code.

Note: the review was static only. Cargo tests and `just codegen` ran separately — 6626 passed, 0 failed, fmt and clippy clean.

Issue #1500, opened 2026-08-20
security: product commit containment is pathname-based — hardlinks, Windows, and parent-dir races bypass the symlink guards


hugocorreia90:
Found by an independent red team on PR #1498 (which closes #1482's Unix source-leaf case). These are the vectors that fix does **not** cover. Filed together because they share one root: the guards are pathname-based and symlink-specific.

## The gaps

```
symlink at the leaf, unix   -> covered (O_NOFOLLOW, #1498)
hardlink                    -> NOT covered: passes the is_symlink checks
                               (commit.rs:223) and is copied into .ff-prev
Windows                     -> NOT covered: O_NOFOLLOW is #[cfg(unix)]
                               (commit.rs:362); the original race stands
parent directory swapped    -> NOT covered: static escaping parents are
  after validation             rejected, but mutations are pathname-based
                               (commit.rs:599), so a post-check swap races
```

The parent-directory vector needs write access to the parent of `models` or the product state directory — not merely their contents — so it is a narrower threat than the leaf cases.

## What full containment needs

- **dirfd-relative operations** (`openat`/`renameat` against a directory descriptor captured once) so a path component swapped after validation cannot be reached.
- **an explicit hardlink policy** — either refuse a source with `nlink > 1`, or accept and document why.
- **Windows reparse-point protection**, or a documented statement that the guarantee is unix-only. Today the module header says every symlink exploit test is `#[cfg(unix)]`, which is honest about test coverage but not about the guarantee.

## Also reported, same review

`rocky-fulfill`'s candidate writer (`step.rs:922`) validates only the final path, derives a second temporary path, then `std::fs::write`s it at line 931 — a planted `products/<name>.toml.ff-candidate-tmp` symlink is written through during elicitation. That one is independent of this module and could be fixed on its own with O_EXCL/no-follow.

hugocorreia90:
**Chain link: this issue carries the fix for #1501 too.** Same root — validate a path, then re-resolve it by name.

```
this issue  hardlink / Windows / parent-dir race   integrity failures
#1501       Phase B sidecar re-read                data exfiltration
shared fix  dirfd-relative operations (openat / renameat against a
            directory descriptor captured once)
```

[...]

Issue #1625, opened 2026-09-03
Six entry points each decide what an invalid rocky.toml means, and five treat it as absent


hugocorreia90:
Six entry points load `rocky.toml` and each decides for itself what an invalid config means. Five of them treat "present but broken" the same as "absent" and carry on against defaults. So a refusal that one command reports, another silently discards.

Found by the red-team review of #1624, which fixed exactly one of them and claimed the class was closed. It was not — that claim is corrected in the changelog.

## The shape

```
  config invalid
    rocky run / plan / compile / validate / emit-sql   ->  refuses, shows the remedy
    rocky lineage                                      ->  .ok() -> unwrap_or_default() -> compiles
    MCP plan_preview                                   ->  .ok() -> DuckDB dialect -> emits SQL
    MCP compile_full (lineage, dependents, AI helpers) ->  compiles with defaults
    LSP (initial + didChange)                          ->  defaults, no diagnostic
    serve POST /api/v1/compile  (SDK lineage)          ->  degrades, returns success
```

Concrete trigger from the review: configure Snowflake plus an invalid section, then call MCP `plan_preview`. Loading fails, `.ok()` erases the refusal, and the tool returns **DuckDB SQL** — for a Snowflake project — with no indication anything was wrong.

`POST /api/v1/compile` is the sharpest one: it returns success *unconditionally* after degrading. And the SDK is internally inconsistent about it — `GET /api/v1/compile` is strict because it calls `compile_output`, while the POST route and the CLI lineage route are not.

## Why a fix per caller will not hold

This is the fifth, sixth and seventh instance of the same defect. #1624 fixed `emit-sql` and closed two more; the review then found five it had missed. Every one is a caller deciding independently what a `ConfigError` means, so the property is caller-by-caller rather than structural — and a new entry point inherits nothing.

## Suggested shape

One shared fallible project-context loader, used by every model-resolution and preview entry point, where **only `FileNotFound` becomes `None`** and every other `ConfigError` propagates. Thread mask, classifications and project freshness from that single snapshot rather than re-deriving them per caller.

For the LSP and `serve` the answer may legitimately differ — staying usable with a broken config is a defensible product choice for an editor. But it should be a stated contract, not an accident of `.ok()`, and the refusal should surface as a diagnostic rather than vanish.

## Deciding question

[...]

PR #1627, opened 2026-09-03
fix(engine/rocky-core): refuse a hardlinked write or backup target

hugocorreia90:
Two helpers in the product commit protocol now refuse a hardlinked target. `write_no_follow` and `read_no_follow` `fstat` the descriptor they already hold and refuse `nlink > 1`.
A second name for one inode **is** an ordinary regular file, so every pathname check passed it and `O_NOFOLLOW` had no link to refuse.

**This closes two helpers, not the hardlink class.** The independent review found two more hardlink paths with no timing to win, both outside these helpers. Recovery publishes a hardlinked `.ff-prev` through a pathname `rename` after an `exists()` check. The fulfillment imports read outside bytes through `read_to_string` and `std::fs::read`, and the elicitation digest there is derived from the same bytes it later checks, so it verifies nothing. Both are filed as #1633, which carries the structural fix: one descriptor-bound rule at every import and publication sink, instead of a third per-caller check here. The review also settled the false-refusal worry below in this PR's favour: git checkout creates working-tree files with `O_EXCL`, a local clone's hardlinks stay under `.git/objects`, and a macOS `cp -c` clone leaves `nlink == 1`.

## What is and is not covered

```
  vector                        unix                        windows
  --------------------------------------------------------------------------
  symlink at the leaf           refused at the syscall      pre-check only
                                (O_EXCL on a create,        (O_NOFOLLOW is a
                                O_NOFOLLOW on an            unix flag)
                                in-place write or read)
  hardlink at the leaf          refused on the opened       not checked
                                DESCRIPTOR (nlink > 1)      (see below)
  parent dir swapped after      NOT closed — only dirfd-relative APIs close it,
    validation                  and v0 does not use them (unchanged)
```

Two guarded call sites, both after the existing regular-file check:

| helper | harm closed | callers |
|---|---|---|
| `write_no_follow` (in-place truncate) | writing **through** to a name the project does not own | MCP `draft_model` / `draft_contract` / `draft_check` / `draft_metadata`, and the rollback restore |
| `read_no_follow` (behind `copy_no_follow`) | copying a file the project does not own **into** it (`.ff-prev`), and the MCP draft snapshot | `commit_generation`'s backup, Phase B's sidecar read, `DraftRollback::snapshot` |

Not guarded, on purpose:

- **The O_EXCL create helpers** (`create_new_no_follow`, `write_new_no_follow`). The descriptor they return names a file this process just created, so the link count is 1 by construction; a check there could only fire on a link added afterwards, which no open-time check can see. Pinned by a test (`the_o_excl_create_unlinks_a_hardlinked_squatter_instead_of_writing_through_it`), not asserted in prose.
- **`commit_generation`'s final replacement.** It renames the staged file over the final, which swaps a directory entry — it never writes through. So the commit path's exposure was the backup read, not a write, and the docs now say that.
- **Windows.** `O_NOFOLLOW` is a unix flag and `std::os::windows::fs::MetadataExt::number_of_links` is unstable (`windows_by_handle`, rust-lang/rust#63010). There is no Windows `cargo test` lane in this repo, so a Windows guard would be untested code claiming a protection nobody had run. The module header states the gap instead.
- **The dirfd/`openat` conversion.** It is the complete fix for the parent-directory race and needs either a new workspace dependency (only `libc` is present today) or a hand-rolled `openat` across ~20 helpers. Gated on evidence of a hostile-parent-directory threat. The existing TOCTOU note still reads true and was updated to enumerate the new guard.

Also read (not written) through a plain `std::fs::read_to_string`: the `briefs_dir` override load in `rocky-fulfill/src/briefs.rs`. It uses neither changed helper, so it is out of scope here, but it is a genuine read-through gap.

## Gates

Toolchain: `cargo +1.98.0` (CI pin) throughout, macOS arm64.

| gate | result |
|---|---|
| `cargo fmt --all -- --check` | exit 0 |
| `cargo clippy --all-targets --all-features -- -D warnings` | exit 0 |
| `cargo test --workspace --no-fail-fast` | exit 0 — 150 suites, **6633 passed, 0 failed, 114 ignored** |
| `just codegen` + `git status --porcelain` | no drift (a `package-lock.json` diff of only `libc` lines is a known local artifact; reverted) |
| `./scripts/regen_fixtures.sh` + drift | no drift — 40 fixtures captured, byte-identical |

`git rev-parse HEAD` and `git status --porcelain` were captured before and after the test run; the tree did not move.

## Mutation check

`scripts/mutation-check.sh`, one mutation per guarded call site, both on the committed tree.

**1. Delete the `nlink` check from `write_no_follow`:**

```
test product::commit::tests::the_hardlink_guard_does_not_fire_on_an_ordinary_single_link_file ... FAILED
test product::commit::tests::write_no_follow_refuses_a_hardlinked_leaf_and_leaves_both_names_untouched ... FAILED
test result: FAILED. 4 passed; 2 failed; 0 ignored; 0 measured; 2070 filtered out
test tools::tests::draft_paths_refuse_a_hardlinked_leaf_out_of_the_project ... FAILED
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 50 filtered out
```

```
panicked at crates/rocky-core/src/product/commit.rs:3174:59:
a hard link is refused: ()
panicked at crates/rocky-mcp/src/tools.rs:6165:14:
a hard-linked draft path refuses the write: ()
```

**2. Delete the `nlink` check from `read_no_follow`:**

```
test product::commit::tests::read_no_follow_bytes_refuses_a_hardlinked_leaf ... FAILED
test product::commit::tests::the_backup_copy_refuses_a_hardlinked_source_and_produces_no_backup ... FAILED
test product::commit::tests::fresh_commit_refuses_a_hardlinked_final_and_never_copies_it_into_the_project ... FAILED
test result: FAILED. 3 passed; 3 failed; 0 ignored; 0 measured; 2070 filtered out
test tools::tests::draft_paths_refuse_a_hardlinked_leaf_out_of_the_project ... FAILED
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 50 filtered out
```

```
panicked at crates/rocky-core/src/product/commit.rs:3233:48:
a hard-linked source is refused: ()
panicked at crates/rocky-mcp/src/tools.rs:6154:14:
a hard-linked leaf refuses the snapshot instead of being read through
```

Every new test is fix-sensitive, and no test passes under either mutation that should not.

**Caveat on the script's exit code, not on the evidence.** The first run (without `--no-fail-fast`) exited 0. Every run *with* `--no-fail-fast` exited **3 — "INCONCLUSIVE: the mutation broke the build"**, which is false: the log above shows the tests compiled, ran, and failed. That is the known #1547 misread of a failing `cargo test`. The logs are the evidence, not the exit code.

## Two standing questions

**What am I least confident about?** Whether the *read*-side refusal can break a setup that works today. A hardlinked model file is not something Rocky or git ever creates — a git checkout never hardlinks working-tree files, and macOS `cp -c` makes an APFS clone, which leaves `nlink == 1`. But an external tool that populates a models tree with `cp -l`, or a content-addressed cache, would now be refused at `rocky product compile`. The refusal is loud and prints the remedy, which is the direction of harm this repo prefers — but one real report of a hardlinked models tree should demote the read-side guard to a warning while the write-side guard stays a refusal.

**What is the most important thing I am missing?** That this closes the *leaf* class only, and the attacker it assumes is the same attacker the still-open parent-directory race serves better. Anyone who can plant a hardlink inside `models/` can usually also swap a parent directory — and that vector is untouched here. So the threat model that justifies this guard also justifies the dirfd conversion; deferring it is a cost decision (a new workspace dependency or a hand-rolled `openat` across ~20 helpers), not a claim that the residual is small. #1500 stays the record for it.

Refs #1500 (option B: refuse hardlinks, state the Windows limit, defer dirfd). #1500 stays open until #1633 closes the two remaining hardlink paths.


hugocorreia90:
## Red team: Codex (GPT-5), read-only, 50 tool calls. Verdict: **BLOCK**.

The agent that wrote this PR died on a server error before recording any review, so this was the first independent pass over the diff. Running it was worth it.

### What it blocked on

**The completeness claim is false.** Two paths import or publish a file Rocky did not create, and neither goes through the guarded helpers. Neither needs to win a race.

1. **Recovery publishes a hardlinked `.ff-prev`.** Residue is validated with `is_symlink` only, then `if entry.has_prev && prev.exists() { std::fs::rename(&prev, final_path) }` — an existence check and a pathname rename. I verified this in the source. A hardlink at `<allowed-final>.ff-prev`, plus a journal entry with `has_prev: true` and a non-matching manifest digest, selects rollback; recovery sees a regular file and renames it over the final. The committed file then names an outside inode.

2. **Fulfillment imports through following reads** — `read_to_string` for brief overrides, `std::fs::read` for the elicitation outbox. And the digest that appears to protect the second is **tautological**: the driver derives `expected_digest` from the same bytes it just read, so the later comparison checks an input against its own hash.

Filed as **#1633**, with the structural fix Codex recommends: one descriptor-bound containment rule at every import and publication sink, rather than a check per helper. Same shape as #1625 earlier today.

### What it confirmed as correct

- The `O_EXCL` reasoning holds: `create_new(true)` atomically refuses an existing file or symlink, so a successfully returned descriptor names a file with `nlink == 1` by construction. A check there would add nothing durable.
- The `fstat` is genuinely descriptor-based, `nlink > 1` is the right predicate after the regular-file check, and both helpers refuse before truncating or reading.
- **The false-refusal worry is unfounded**, which resolves the author's own least-confident point. Git checkout creates working-tree files with `O_EXCL`; local-clone hardlinks apply to `.git/objects`, not the working tree. macOS `cp -c` makes copy-on-write clones, not hardlinks. No fixture, POC or documented workflow in the repo materializes project files with `cp -l`. And a benign multi-name hardlink is deliberately indistinguishable from a hostile one — downgrading the read side to a warning would restore the vulnerability.
- The Windows disclosure is accurate: handle-based `number_of_links` is still nightly-only and the test lane is Ubuntu-only.

One correction it made to the module header: `dirfd + renameat` prevents ancestor re-resolution but still resolves the source **leaf** name. Closing leaf substitution needs publication bound to the staged inode. That overstatement is noted in #1633.

### Disposition

The PR body's completeness claim is corrected to say what this actually closes: two helpers, not the class. The code is unchanged — everything the review verified as sound stays as it is, and the two bypasses are filed rather than patched here, because a third per-caller check would repeat the mistake this was blocked for.

One question left open rather than decided: whether project-configured brief overrides are a **trusted** input. If they are, half of finding 2 is by design and should be documented. There is no written trust-boundary specification to consult, and that absence is worth closing on its own.

PR #1642, opened 2026-09-04
fix(dagster): fail a contract check when the compile result is missing

### LINKED ISSUE #1619: dagster: contract checks pass green when the compile result is missing
A Dagster contract check reports `passed=True` when there was no compile output at all, not only when there was nothing to report.

`integrations/dagster/src/dagster_rocky/component.py`

- a compile transport failure leaves the cached compile slot unset
- the missing slot becomes an empty diagnostics list
- contract `AssetCheckSpec`s are declared independently of that, from the `.contract.toml` files

`contract_check_results_from_diagnostics` then sees an empty list and yields `passed=True` for every declared check.

So a failed compile looks identical to a clean contract. `E010`, `E011`, `E012` and `E013` would all be missing in that state, and every contract check would still go green.

This predates #1240 and is wider than it: the fix is to distinguish "compiled, nothing to report" from "did not compile", and emit an unverified or failing result for the second.

Found while reviewing #1240.

hugocorreia90:
## Contract checks no longer pass green when the compile result is missing

A `RockyComponent` contract check now reports `passed=False` ("not verified") when the cached `rocky compile` slot is missing, instead of `passed=True`. On the way, the default `streaming` mode no longer fails the step with `returned an output ... multiple times` for a source asset that has a matching `.contract.toml`.

```
                          cached state
                     ┌──────────────────────┐
   rocky compile ───►│  compile slot ?      │
                     └──────────┬───────────┘
                present         │        missing (compile failed or skipped)
          ┌─────────────────────┴─────────────────────────┐
          ▼                                               ▼
  diagnostics for the model?             BEFORE: read as []  → passed=True
    none        → passed=True            AFTER:  None        → passed=False, WARN,
    E010..E014  → passed=False                                 rocky/compile_missing=true
```

### What changed

- `_CompileState.compiled` records whether a compile result was cached. `from_result(None)` gives `compiled=False`.
- `contract_check_results_from_diagnostics(diagnostics=None, ...)` means "no compile output": every declared check fails at `WARN` with `rocky/compile_missing = true`, a `status` and a `description`. An empty list keeps its meaning ("compiled, nothing to report") and passes. `CONTRACT_COMPILE_MISSING_METADATA_KEY` is exported.
- The run logs a warning when contract rules exist and the compile slot is missing.
- Streaming mode: the contract results are built before the run and prime the placeholder pass's dedup set, the same guard the compliance results already use. Before, the placeholder pass emitted each contract check as `passed=True` ("not produced by rocky") and the contract pass emitted it again, so Dagster failed the step. That second defect is fixed here because the #1619 fix cannot deliver a result in the default mode without it, and the placeholder's green is itself a green with no compile behind it. Both halves date from the initial commit.

Severity is `WARN`, matching the integration's other "not checked" results (unmaterialized table, failure containment). `passed=False` is what removes the false green; severity is presentation.

Not affected: DAG mode and derived-model assets declare no contract check specs (`build_dag_specs` and `build_model_specs` accept `contract_rules_by_model` and ignore it), so the only emit site is `_make_rocky_asset`, in both execution modes.

### Evidence

Fail-before, on `main`, end-to-end probe with `contracts_dir` set and no compile slot, streaming mode:

```
contract_required_columns   passed=True  meta={'status': 'not produced by rocky (check type: contract_required_columns)'}
contract_protected_columns  passed=True  ...
contract_column_constraints passed=True  ...
FAILURE: DagsterInvariantViolationError: Compute for op "rocky_acme" returned an output
         "fivetran__acme__us_west__shopify__orders_contract_required_columns" multiple times
```

The same probe with the compile slot present gives the same crash: the duplicate does not depend on the slot.

Pass-after: `uv run pytest -q` in `integrations/dagster` gives 732 passed (was 724). `ruff check .` and `ruff format --check .` are clean.

Mutation checks with `scripts/mutation-check.sh`, all three mutants killed:

| Mutation | Tests that fail under it |
|---|---|
| `diagnostics=None` falls through as `[]` (`contracts.py`) | 2 unit tests + streaming and pipes end-to-end "not verified" tests |
| `from_result(None)` returns `compiled=True` (`component.py`) | `_CompileState` test + streaming and pipes end-to-end "not verified" tests |
| contract pairs not primed into the placeholder pass (`component.py`) | both streaming end-to-end tests (step fails with the duplicate output) |

Closes #1619

### DIFF
--- a/integrations/dagster/CHANGELOG.md
+++ b/integrations/dagster/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **A contract check now fails as "not verified" when there is no `rocky compile` output. Before, it passed.** `RockyComponent` caches `rocky compile` when it writes its state. That step is best-effort: if the binary is missing, the compile times out or crashes, or `models_dir` does not exist, the `compile` slot is left out of the state and the component keeps working. Contract checks read their verdict from that slot's diagnostics. A missing slot read as an empty list, and an empty list means "nothing to report", so every declared contract check went green with no compile behind it.
+
+  The two cases are now told apart. With a cached compile and no diagnostics for the model, the check passes as before. With no cached compile, each declared contract check reports `passed=False` at `WARN` severity, with `rocky/compile_missing = true` and a `status` that says why. The run logs a warning too. Refresh the state after `rocky compile` succeeds to clear it. (#1619)
+
+  For hand-rolled assets, `contract_check_results_from_diagnostics()` now accepts `diagnostics=None` to mean "no compile output". An empty list keeps its old meaning. `CONTRACT_COMPILE_MISSING_METADATA_KEY` is exported.
+
+- **In the default `streaming` mode, a source asset with a matching `.contract.toml` no longer fails its run with `returned an output ... multiple times`.** The placeholder pass that covers declared checks the engine did not produce also covered the contract checks, and reported each one as passing ("not produced by rocky"). The contract pass then reported the same check a second time, and Dagster failed the step. The contract results are now built first and the placeholder pass skips them, so each contract check is reported exactly once. Only `execution_mode="streaming"` was affected; `"pipes"` has no placeholder pass. (#1619)
+
 ## [1.65.0] — 2026-09-03
 
 Pairs with engine 1.73.0 and `rocky-sdk` 0.14.0. The `rocky-sdk>=0.13.0` floor is unchanged: nothing here reads a field that 0.13.0 lacks.
--- a/integrations/dagster/src/dagster_rocky/__init__.py
+++ b/integrations/dagster/src/dagster_rocky/__init__.py
@@ -26,6 +26,7 @@ from .component import (
 )
 from .contracts import (
     CONTRACT_COLUMN_CONSTRAINTS_CHECK,
+    CONTRACT_COMPILE_MISSING_METADATA_KEY,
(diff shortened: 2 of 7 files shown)


hugocorreia90:
**Independent red team: Codex (GPT-5), two rounds, `VERDICT: PASS` in both.**

Round 1 (full PR) and what happened to each finding:

- A missing compile slot now fails as "not verified" on every supported path (streaming, pipes, tenant-collapsed partitions, sensor-triggered runs): confirmed, no change.
- A compile that ran but did not find the contract's model emits `W011`. This integration does not map that code, so the check stays green. Real, same class as #1619, outside this PR's scope. Filed #1644.
- The cached compile slot can be stale until the state is rewritten. True, and it is how the state-backed component works: rewriting the state is the refresh. Not changed here.
- DAG mode and derived-model assets declare no contract checks, so there is nothing to fix there. The `surface_derived_models` docstring said otherwise; corrected in round 2.
- When a run ended in a quota-breach retry, the contract results were never yielded (pre-existing order; the run fails, so no false green). Round 2 yields them before the raise, in both modes, with a test that fails if the order is reverted (mutation-checked).
- `passed=False` at `WARN` is the right shape for "not verified": agreed.

Round 2 (the delta only): no findings. It traced both execution modes and confirmed each yields the contract results exactly once, that the new quota-breach test runs the real component and fails if the yield order is reverted, and that the docstring correction matches `build_model_specs` and `_make_derived_model_asset`. Its own run: 52 tests, `ruff check` and `ruff format --check` green.

Transcript audit: 31 commands run in round 1, 39 in round 2.

Found on the way, out of scope: a declared default check that the engine did not produce reports `passed=True` on a materialized table. Filed #1645.

PR #1643, opened 2026-09-04
OWNER MERGE — fix(engine/rocky-core): report a load column whose warehouse type is unrecognised instead of matching it

### LINKED ISSUE #1614: rocky load: an unrecognized warehouse type makes a contract type check pass
The `rocky load` contract gate has the same shape #1240 just closed on the compile-time gate.

`engine/crates/rocky-core/src/contracts.rs`

- `warehouse_type_to_rocky` maps any type string it does not recognise to `RockyType::Unknown` (its own doc says "which matches anything").
- `validate_contract_typed` then skips the type comparison when either side is `Unknown`, and staging is promoted.

So a landed column whose warehouse type string is outside the map satisfies whatever type the contract declares. The declared side has the same hole: a contract type string the mapper does not know also becomes `Unknown`.

This is a runtime gate on live `DESCRIBE` output, not the compiler, so it was left out of #1240. The compile-time fix there was: branch on `Unknown` before the matcher, report it at info severity naming the column and the declared type, and make the matcher answer "not a match" for `Unknown`. The same shape would work here.

Found while reviewing #1240.

hugocorreia90:
> **Owner merge.** One load that promotes today can now fail: a contract declaring the precision-only `DECIMAL(10)` was read as `DECIMAL(38,0)` and matched a landed 38-digit column. It now means precision 10, scale 0, as SQL reads it and as the compile-time gate already did, so it correctly fails there. The contract was never enforcing the digits it named. Remedy: write the digits the data really has, or drop to a bare `DECIMAL` to match any.

`rocky load` now reports a required column whose type it could not compare. Before, a warehouse type string outside Rocky's type map — on the landed side or the declared side — matched anything, and staging was promoted with nothing said. The load still promotes. The gate result now says which column was not checked, and why.

## The hole

```
validate_contract_typed  (engine/crates/rocky-core/src/contracts.rs)

  landed   'GEOMETRY' ──warehouse_type_to_rocky──► Unknown ─┐
                                                            ├─► is_assignable(Unknown, _) == true
  declared 'BIGINT'   ──warehouse_type_to_rocky──► Int64   ─┘        │
                                                                     ▼
                                                        "conforms" ─► promote, nothing said
```

Now:

```
  either side Unknown ─► warnings += "column 'geo' landed as 'GEOMETRY' and the contract
                          declares 'BIGINT'; Rocky does not recognise the landed type, so
                          the declared type was not checked. ..."       (passed unchanged)
  both sides known    ─► landed_type_conforms(..) ─► violation on mismatch   (unchanged)

  landed_type_conforms(Unknown, _) and (_, Unknown) ─► false             (was true)
```

## What changed

- `validate_contract_typed` branches on `Unknown` before the matcher. The column goes to `ContractResult.warnings` with the column name, the landed type string, the declared type string, and which side Rocky did not recognise. `passed` does not change.
- `landed_type_conforms` answers "not a match" for `Unknown` on either side. `validate_contract_typed` no longer reaches it with `Unknown`; the arm exists so a future caller that skips the branch gets a violation, not a silent pass. `rocky_ir::is_assignable` is untouched — its `Unknown` rule is right for the compiler's best-effort inference.
- Where the warning is seen: `rocky load` already logs every gate warning with `tracing::warn!` (`load.rs`; the default filter is `info`), and `--output json` already carries `files[].contract.warnings`. No new field.
- Four doc comments described the skip as designed. Corrected. Two are schema descriptions (`ContractResult.warnings` in `load.schema.json`, `RequiredColumn.type` in `rocky_project.schema.json`), so their bindings and `docs/public/openapi.json` are regenerated. Text only.
- `engine/CHANGELOG.md`, under Unreleased.

## Not changed

- No load that promotes today is refused. `passed` is computed from `violations` only, and this adds no violation.
- The compile-time gate (`rocky-compiler`, #1240 / `I003`) is untouched.
- `validate_contract` (the raw-string, source-vs-target variant) has no `Unknown` concept and is untouched.

## Evidence

Fail-before, the new tests against the old code:

```
test_typed_unknown_landed_type_is_reported_not_passed_silently   FAILED  warnings: []  (expected 1)
test_typed_unknown_declared_type_is_reported_not_passed_silently FAILED  warnings: []  (expected 1)
test_typed_both_types_unknown_reported_once                      FAILED  warnings: []  (expected 1)
test_landed_type_conforms_rejects_unknown_on_either_side         FAILED  !landed_type_conforms(&Unknown, &Int64)
```

Pass-after: `cargo +1.98.0 test -p rocky-core --lib contracts::tests` 27 passed. `cargo +1.98.0 test -p rocky-cli --lib commands::load::tests` 18 passed, including the new end-to-end `test_contract_gate_reports_a_landed_type_it_cannot_compare`: a CSV column DuckDB sniffs as `TIME` (outside the map), contract declares `TIMESTAMP`. The load promotes, and the gate result names `'at'`, `'TIME'` and `'TIMESTAMP'`.

The old `test_typed_unknown_type_never_fails` pinned the fail-open (it asserted only `passed`). The first test above replaces it. The happy-path test now also asserts zero warnings, so known matching types cannot start warning.

## Mutation checks

Three runs of `scripts/mutation-check.sh` (it refuses a dirty tree; all three on the committed state). Each exit 0: the test failed under the mutation, so it is fix-sensitive.

| Mutation | Test command | Result |
|---|---|---|
| Drop the `Unknown` branch (say nothing, as before) | `cargo test -p rocky-core --lib contracts::tests` | 3 tests fail: the landed-side, declared-side and both-sides reports |
| Same mutation | `cargo test -p rocky-cli --lib commands::load::tests::test_contract_gate_reports_a_landed_type_it_cannot_compare` | fails — the end-to-end DuckDB path is wired to the branch |
| Remove the `Unknown` guard in `landed_type_conforms` | `cargo test -p rocky-core --lib contracts::tests::test_landed_type_conforms_rejects_unknown_on_either_side` | fails on `!landed_type_conforms(&Unknown, &Int64)` |

## Gates

`cargo +1.98.0 clippy --all-targets -p rocky-core -p rocky-cli -- -D warnings` exit 0. `cargo +1.98.0 fmt --all --check`: my first local run read the exit code through a pipe and reported green falsely; CI's `Format` job caught a three-line wrap in one test, fixed in the second commit, after which the check exits 0. `export-schemas`, `codegen-sdk`, `codegen-vscode`, `codegen-vscode-project-schema`, `export-openapi` rerun; the diff is the two descriptions only.

Closes #1614


hugocorreia90:
**Independent red team: Codex (GPT-5), `VERDICT: BLOCK`, both findings applied.**

**Medium — the compatibility claim was false.** The changelog said a load that promoted before still promotes. It does not, in one case, and that case is the point of the change: a contract declaring the precision-only `DECIMAL(10)` used to be read as `DECIMAL(38,0)`, so it matched a landed 38-digit column. `DECIMAL(10)` means precision 10, scale 0 — how SQL reads it, and how the compile-time gate already read `Decimal(10)` — so it now fails against that column. The contract was never enforcing the digits it named.

The reviewer was explicit that the new reading is correct and the *claim* was the defect. The changelog now states the regression, names the remedy (write the real digits, or drop to a bare `DECIMAL` to match any), and this PR is marked for owner merge because it refuses a load that promotes today.

**Low — whitespace before the parameter list.** `DECIMAL (10,2)` is valid SQL and was being read as an unrecognised type. A `.contract.toml` is written by hand, so the space is now trimmed. Fixed, with the case added to the grammar test alongside a lowercase and an outer-whitespace spelling.

Two findings it checked and cleared: the warning path never touches `passed` or `violations`, and the matcher's new "not a match" answer for `Unknown` cannot flip a load, because the gate branches on `Unknown` before reaching it.

Its stated limit: it could not confirm that every warehouse's `DESCRIBE` canonicalises away the space before `(`, since it was held to the diff. That is why the fix trims rather than relying on the inventory.

hugocorreia90:
**Before merging: this widens a divergence between two mappers that are documented as mirrors. It is worth a deliberate yes, not an accidental one.**

`warehouse_type_to_rocky` opens with:

> Mirrors the compiler's `default_type_mapper` so the load gate compares types in the same vocabulary the rest of Rocky uses.

That claim was already partly untrue before this PR, and this PR makes the gap wider. On `main` today, for the decimal family:

| Type string | `default_type_mapper` (compiler) | `warehouse_type_to_rocky` after this PR |
|---|---|---|
| `DECIMAL(10,2)` | `Decimal(10,2)` | `Decimal(10,2)` |
| `DECIMAL` | `Decimal(38,0)` | `Decimal(38,0)` |
| `DECIMAL(10)` | `Decimal(38,0)` | **`Decimal(10,0)`** |
| `DECIMAL(a,b)`, `DECIMAL(10,2,3)` | `Decimal(38,0)` | **`Unknown`** |
| `NUMBER(38,0)`, `BIGNUMERIC(76,38)` | **`Unknown`** (the compiler's arm matches only `DECIMAL`/`NUMERIC` prefixes) | `Decimal(38,0)` / `Decimal(76,38)` |

The last row is pre-existing and is the larger of the two gaps: Snowflake's `NUMBER` and BigQuery's `BIGNUMERIC` are already read differently by the two gates today.

**I still think this PR is the right direction.** Reading `DECIMAL(10)` as precision 10 scale 0 is what SQL means and what `decimal_type_matches` already does for `Decimal(10)`, and refusing to invent `(38,0)` for a string that does not parse is the whole point. Moving the load gate first is defensible. But it means the load gate is briefly stricter and more correct than the compile gate, and the doc comment should not keep claiming they mirror each other.

**Two ways to take it:**

1. Merge as is, and land #1646 next — it already covers bringing `default_type_mapper` into step (malformed → `Unknown`) plus the bare BigQuery `NUMERIC`/`BIGNUMERIC` defaults, which is a live wrong-refusal on BigQuery. I would then correct the "Mirrors" sentence in the same change.
2. Hold this until #1646 lands so both mappers move together, and neither is briefly ahead.

I would take option 1: the gap in the table's last row already exists, this PR removes fabricated types rather than adding any, and #1646 is well specified. Happy to do #1646 straight after either way.

Flagging it because "same source as X" claims that stop being true are the kind of thing this repo has been bitten by, and a reviewer reading only the diff would not see the sentence at the top of the function.

PR #1647, opened 2026-09-04
chore(ci): create every release as a draft and publish it after the registry upload

### LINKED ISSUE #1639: Release workflows: the engine's local fallback and all three sibling workflows publish a non-draft GitHub Release before the artifacts exist
`engine-release.yml` creates its GitHub Release as a **draft** and publishes only after every target uploads, so a failed matrix build cannot publish a broken engine release. Nothing else in the release surface has that gate.

## Three fail-open paths

1. **`scripts/release.sh` (`just release-engine`), the documented local fallback.** It pushes the tag and creates the release itself, non-draft, with the locally built archives and no `checksums.txt`. `engine-release.yml`'s `ensure-release` job then sees the existing release and keeps it as is (`"Release $TAG already exists (created by scripts/release.sh)"`) — it does not convert it to a draft. `engine/install.sh` resolves "latest" from the public release list and requires `checksums.txt`, so during the 15–25 minute matrix build a user gets a public tag that is not installable, and if the matrix fails the incomplete public release is what remains.
2. **`sdk-release.yml`, `dagster-release.yml`, `vscode-release.yml`** all run `gh release create` with no `--draft` before building or publishing. A failed `uv build`, PyPI upload or VSIX package leaves an empty or partial public release.
3. **`vscode-release.yml`'s Marketplace step exits 0 when `VSCE_PAT` is unset** ("skipping marketplace publish (this is expected on forks)"). On a tag pushed to this repository that is not a fork condition; a missing or rotated secret makes a canonical release "succeed" without publishing.

## Proposed

- `scripts/release.sh`: create the release with `--draft`, and have `ensure-release` in `engine-release.yml` convert any existing non-draft release to draft (`gh release edit --draft`) before the matrix runs — or refuse to continue.
- The three sibling workflows: `--draft` on create, `gh release edit --draft=false` as the last step after the registry publish succeeds.
- `vscode-release.yml`: fail when `VSCE_PAT` is unset and `github.repository == 'rocky-data/rocky'`; keep the soft no-op for forks.

These are `.github/workflows` changes and go through the credential-containment gate as usual.

For the 1.73.0 release the mitigation is procedural: tags are pushed directly (no local fallback), and each sibling workflow is watched to completion before the next tag goes out. Found by independent review of the release PR (#1637).

hugocorreia90:
Every release path now creates the GitHub Release as a draft and publishes it only after the artifacts and the registry upload exist. A failed step leaves a draft, which the public release list omits, instead of an empty or partial public release.

Closes #1639

## What changed

**`.github/workflows/sdk-release.yml`, `dagster-release.yml`**
- `ensure-release` creates the release with `--draft --latest=false`. An existing draft (the local fallback) is used as it is. An existing published release stops the run with an error that names the fix (`gh release edit <tag> --draft`, or delete it).
- New last step `Publish the GitHub Release`: `gh release edit <tag> --draft=false --latest=false`, after the PyPI upload and the `dist/*` attach (wheel, sdist, two `.publish.attestation` files) both succeed. `--latest=false` is repeated because GitHub's API documents `make_latest` as defaulting to `true` for a newly published release, and the engine release owns the "Latest" badge.

**`.github/workflows/vscode-release.yml`**
- Same `ensure-release` gate and last-step publish.
- The VSIX is attached with `gh release upload` instead of `softprops/action-gh-release`. At the pinned commit the action finalizes an existing draft after it uploads (`src/github.ts`, `finalizeRelease`: it returns early only when `draft: true` is passed or the release is already public), which would have published the release before the Marketplace step. This removes an action pin; it adds none.
- The Marketplace step fails when `VSCE_PAT` is unset and `GITHUB_REPOSITORY` is `rocky-data/rocky`. A fork keeps the soft no-op.

**`.github/workflows/engine-release.yml`**
- `ensure-release` accepts an existing draft and refuses an existing published release. Refusing rather than converting: `gh release edit --draft` on a live release would retract it on every full re-run of the workflow, and leave it retracted if the re-run failed. Refusing changes nothing and names the fix.

**`scripts/release.sh`**
- Engine: creates the release with `--draft` (no `--latest`; CI's `publish` job sets it when it flips the draft). The final messages say so and give the manual publish command for when CI cannot run, with the caveat that `install.sh` still needs `checksums.txt` and the `rocky-lsp` archives, which this script does not build.
- sdk / dagster / vscode: `--draft` unless `--publish` was given. With `--publish` the registry already has the artifact, so the release is created published (`gh release create` with assets saves a draft, uploads, then publishes: `pkg/cmd/release/create/create.go`, `draftWhileUploading`). The tag push's CI run then stops at `ensure-release`; the script says that is expected.

**`.claude/skills/rocky-release/SKILL.md` + `.agents/skills/rocky-release/SKILL.md`** (byte-identical): the per-artifact job lists, the local-fallback paragraphs, the "already exists" sentence, a recovery line for a job that fails after its registry upload, and the post-release checklist (`--json isDraft` is `false`). Also corrects the dagster section, which listed three jobs; the workflow has two.

**`CONTRIBUTING.md`**: one sentence, the local paths create a draft.

## What was exercised

Tag workflows cannot be dry-run from a PR. What ran:

- `actionlint` with shellcheck on all four workflows: clean. `bash -n` and `shellcheck` on `release.sh`: clean (the two `SC2155` warnings are pre-existing, lines 24-25).
- `.github/scripts/check_credential_containment.py` against this tree: `passed`. The 110 policy regression tests: `OK`.
- The new `run:` bodies, extracted and executed against a stub `gh` (PATH shim, `GH_HOST` pointed at a non-existent host): `ensure-release` with an existing draft proceeds, with a published release exits 1, with no release or an API error calls `gh release create --draft`; the `VSCE_PAT` guard exits 1 unset in `rocky-data/rocky`, 0 unset on a fork, publishes when set; the VSIX glob guard errors on zero files and uploads with `--clobber` on one; the publish step calls `gh release edit --draft=false --latest=false`.
- Claims grounded in source: `gh release view` finds a draft by its pending tag (`pkg/cmd/release/shared/fetch.go`, `FetchRelease` at v2.99.0); the pinned `action-gh-release` finalizes drafts (`src/github.ts`); GitHub's OpenAPI description for `PATCH /repos/{owner}/{repo}/releases/{release_id}` gives `make_latest` a default of `true`.

What was only read: the first real run of each workflow is the next tag push. The `engine-v1.73.0` release (draft created 00:05:47Z, published 00:40:54Z) is the existing proof that `gh release upload` and `gh release edit` work on a draft by tag name under `GITHUB_TOKEN`.

**`.github/workflows/engine-wasm-release.yml`**
- Same `ensure-release` gate and same last-step publish. This one is not named in the issue, but leaving it out would have made the sentence at the top of this description false: it created a non-draft release and published to npm afterwards, so a failed build or a failed `npm publish` left an empty public release behind.
- An unset `NPM_TOKEN` now fails the run on `rocky-data/rocky` and stays a soft no-op on a fork, matching the `VSCE_PAT` guard. `vscode-release.yml`'s comment used to point at this workflow as the soft-no-op precedent; it now describes the shared rule instead.
- `actionlint` with shellcheck: clean.

- The asset moves from `softprops/action-gh-release` to `gh release upload`, for the same reason the vscode workflow did: at the pinned commit the action finalizes an existing draft after it uploads, so leaving it in front of the npm publish would have made the release public before the step the draft gate exists to protect. No workflow uses that action now. Its `RELEASE_ACTION_SOURCES` entry stays, so #1588's in-flight bump does not have to be re-derived.

## Out of scope

Nothing in the release surface is left creating a public release before its artifacts exist.


hugocorreia90:
**Independent red team: Codex (GPT-5), `VERDICT: BLOCK`, one high finding, applied.**

**High — the local fallback could still leave a partial public release.** `scripts/release.sh --publish` passed `--draft=false` to `gh release create` in the same call as the artifacts. The description argued that `gh` keeps such a release a draft until its uploads finish (`draftWhileUploading`), and that is what `gh` does today. The finding stands anyway, for a better reason than the one given: the guarantee lived inside `gh`'s create command, where a version bump could move it, and the script read as though it published up front.

The script now always creates a draft, attaches the artifacts, and flips the draft in a separate `publish_release` step. That is the same shape as the five workflows, and it states the ordering here instead of depending on a tool's internals. The three stale comments that cited the `gh` behaviour are gone, and the skill's release pages follow in both mirrors.

Four things it checked and cleared:

- **Ordering.** All five workflows publish last, and no workflow still has `softprops/action-gh-release` in front of its publish step.
- **Re-run safety.** Existing drafts are reused, uploads use `--clobber`, and refusing an already-published release strands no legitimate CI recovery.
- **Permissions.** Every job making a new `gh` call already grants `contents: write`.
- **Secret guards and the Latest badge.** Both correct. It noted the repository-identity test is a literal, so a rename would need updating — true, and the same is already true elsewhere in these workflows.

Reviewed at the tag-workflow level only; the first real exercise of any of this is the next tag push, as the description says.

Issue #1524, opened 2026-08-23
Declarative expression checks are raw-interpolated SQL with an unbacked sandbox contract


hugocorreia90:
# Declarative `expression` checks are raw-interpolated SQL with an unbacked sandbox contract

**Area:** `engine/crates/rocky-core/src/tests.rs` · **Severity:** medium · **Found by:** FF-WP-F3 adversarial review

## What is wrong

A declarative `[[tests]]` block of `type = "expression"` becomes SQL by string interpolation:

```rust
// engine/crates/rocky-core/src/tests.rs:492
Ok(format!(
    "SELECT COUNT(*) FROM {table} WHERE {}NOT ({expression})",
    ...
))
```

The only validation is that the expression is non-empty (`tests.rs:487`). The comment above it states the contract:

> Expression is user-supplied SQL — we cannot validate it as an identifier. The caller is responsible for sandboxing execution.

**No caller sandboxes execution.** The adapter executes the string with the warehouse credentials the project is configured with.

## Why it matters now

The contract was defensible while the only caller was a human typing `rocky test --declarative` — a person choosing, in the moment, to run SQL derived from their own project.

FF-WP-F3 added a second caller: the fulfillment loop evaluates a product's declared checks automatically after every apply, unattended. The check text is no longer read by a human before it runs.

F3 shipped with the untrusted half of that exposure closed — `draft_check` was removed from the worker MCP profile, so an untrusted drafting worker can no longer author an expression at all. What remains is the operator-authored case: an expression in a spec or sidecar the operator wrote is still arbitrary SQL evaluated with full credentials.

## Blast radius, as measured

- **DuckDB: statement injection appears blocked.** `execute_sql` uses `conn.prepare(sql)` (`engine/crates/rocky-duckdb/src/lib.rs:66`), which takes a single statement, so a trailing `; DROP TABLE …` should fail at prepare.
- **Expression-scoped evaluation is NOT blocked** on any adapter. A correlated subquery can read tables the check has no business reading, and DuckDB exposes filesystem-reading functions (`read_text`, `read_csv`). The result is one aggregate per run, so exfiltration is slow — but it is a real read primitive with no boundary.
- **Multi-statement behaviour on Databricks, Snowflake, BigQuery, and Trino is UNVERIFIED.** Do not assume the DuckDB result generalises; each adapter's execute path needs checking before any claim is made.

## Suggested fix

[...]

PR #1648, opened 2026-09-04
fix(engine/rocky-core): encode SQL string literals through the dialect at ten sinks

hugocorreia90:
Ten SQL string-literal sinks now encode through the dialect's own lexer rule (`rocky_core::sql_gen::string_literal`, the encoder #1605 added). Where a sink is routed, the interim backslash refusal from #1595 is deleted. A backslash in `accepted_values` or `schema_pattern.prefix` is accepted again and carried as a value.

Generated SQL for a value without a quote, backslash or line break is byte-identical to `main`. This is proven below, not assumed.

## Sink table

| Sink | Dialect source | State |
|---|---|---|
| `rocky-core/src/tests.rs` `accepted_values` (with a dialect) | `generate_test_sql_with_dialect` argument (`adapter.dialect()` in `rocky test` / `rocky run`) | **routed**, guard deleted |
| `rocky-core/src/tests.rs` `accepted_values` (dialect-less `generate_test_sql`) | none by construction | **guarded** (refuses a backslash) — no `rocky` command reaches it |
| `rocky-core/src/quarantine.rs` `accepted_values` | `lower_valid_predicate` argument | **routed**, guard deleted |
| `rocky-core/src/seeds.rs` seed cells | `generate_insert_sql` argument | **routed** |
| `rocky-core/src/arrow_loader.rs` loader cells (BigQuery local-file `INSERT` path) | `generate_batch_insert_sql` argument — was `_dialect`, unused | **routed** |
| `rocky-core/src/unit_test.rs` fixture strings | `rocky-engine/src/test_runner.rs` passes `DuckDbSqlDialect` (the only runner) | **routed** |
| `rocky-core/src/checks.rs` cross-source `_src` tag | `generate_cross_source_overlap_sql` argument | **routed** — validated identifiers, so byte-identical |
| `rocky-duckdb/src/discovery.rs` `schema_pattern.prefix` (inside `LIKE`) | `DuckDbSqlDialect` (the crate's own) | **routed**, guard deleted |
| `rocky-duckdb/src/discovery.rs` schema name read back | `DuckDbSqlDialect` | **routed** |
| `rocky-bigquery/src/discovery.rs` `schema_pattern.prefix` (inside `STARTS_WITH`) | `BigQueryDialect` | **routed**, guard deleted |
| `rocky-snowflake/src/governance.rs` tag values | `SnowflakeSqlDialect` | **routed** — was already escaped correctly by hand; now one encoder |
| `rocky-lang/src/lower.rs` DSL string literal | none — lowers to SQL before a dialect exists | **remaining** |
| `regex_match.pattern` (`regex_match_predicate` on DuckDB / Snowflake / Databricks / BigQuery) | each dialect's own | **remaining** — a backslash is regex syntax; routing it changes today's Snowflake / Databricks behaviour (see below) |
| `rocky-core/src/catalog.rs` Databricks tag values | none (Unity Catalog SQL, no dialect argument) | **guarded** (refuses a quote); a trailing backslash still breaks the statement |
| `rocky-core/src/lakehouse.rs` `COMMENT` / `TBLPROPERTIES` | none | **guarded** (refuses a quote); same backslash gap |
| Snowflake stage URI / path, Databricks `COPY INTO` path, CSV delimiters | n/a | **guarded** (refuse quote, backslash and newline) |

Not a seam, verified: `WHERE table_schema = '{schema}'` and similar in the adapters splice `validate_identifier`-checked names only.

## Evidence

Tests were written first, against the unfixed code, and failed for the real reason. Then the bodies were routed.

| Sink | Fail-before (unfixed code) | Pass-after |
|---|---|---|
| `accepted_values` (`tests.rs`, `quarantine.rs`) with a dialect | `unwrap()` on the #1595 refusal — the guard fired although a dialect was in hand | `NOT IN ('pending', 'trailing\', 'it''s')` on `Standard`; `NOT IN ('pending', 'trailing\\', 'it\'s')` on `Backslash` |
| seed cell, loader cell, fixture string | `Backslash` dialect emitted `'C:\tmp\'` — the trailing `\'` escapes the closing quote | `'C:\\tmp\\'` and `'it\'s'`; `Standard` unchanged: `'C:\tmp\'`, `'it''s'` |
| DuckDB discovery | `discover(r"raw\")` returned the guard's error | executed in-process: schemas `raw\`, `raw\x` found for prefix `raw\` (`rawx` correctly not), `it's` found for prefix `it'`, with its table |
| BigQuery discovery | emitted `STARTS_WITH(schema_name, 'raw\')` — on BigQuery `\'` escapes the closing quote | `STARTS_WITH(schema_name, 'raw\\')`, `'it\'s'` |
| Snowflake tags | emitted `'\\''; DROP TABLE x; --'` — hand-escaped and valid on Snowflake; this sink was already sound | `'\\\'; DROP TABLE x; --'`, `'Hugo\'s pipeline'` — the one encoder |

**Live DuckDB probe** (the branch's own `rocky` binary, a scratch transformation project, DuckDB 1.x file):

- `accepted_values = ["plain", "it's", "C:\\tmp\\"]` on a column holding `plain`, `it's`, `C:\tmp\`, `other`. `rocky test --declarative` generated `SELECT DISTINCT kind FROM probe.demo.items_mart WHERE kind NOT IN ('plain', 'it''s', 'C:\tmp\')`, DuckDB executed it, and the check reported exactly `1 unexpected value(s)` — the quote and the trailing backslash round-tripped. On `main` this config is refused before any SQL is built.
- `rocky seed` on a CSV whose `note` cells are `plain`, `it's`, `C:\tmp\`: 3 rows loaded into `probe.seeds.notes`; read back with `SELECT note, length(note), note = 'C:\tmp\', note = 'it''s'` — lengths 5 / 4 / 7 and both equality tests true, so both values round-tripped byte-exact.

**Byte-identical for plain values.** Unchanged, still-passing assertions on exact SQL: `tests.rs` `NOT IN ('pending', 'shipped')`, `seeds.rs` `'it''s fine'`, `arrow_loader.rs` `'hello'`, `checks.rs` `'cat.s1.orders' AS _src`, `governance.rs` `managed_by = 'rocky'` and friends, `rocky-duckdb` `test_discover_matching_schemas`. New: `accepted_values_plain_values_are_byte_identical_under_both_rules` (dialect-less, `Standard` and `Backslash` all equal the pre-existing string) and `schemata_sql_plain_prefix_is_byte_identical`. Under `Standard` the encoder *is* quote-doubling, so DuckDB and Trino output cannot differ for any value.

**Gates** (`cargo +1.98.0`, exit codes captured directly, tree hash checked before and after): `fmt --all --check` 0; `test -p rocky-sql -p rocky-core -p rocky-duckdb -p rocky-bigquery -p rocky-snowflake -p rocky-engine` 0 (45 suites, 2078 rocky-core lib tests, 0 failed); `clippy --all-targets` on those six with `-D warnings` 0; `check --workspace --all-targets` 0.

**Mutation check** (`scripts/mutation-check.sh`, clean tree enforced): the `Backslash` arm of `encode_string_literal` reverted to plain quote-doubling — the exact defect this PR closes, applied at the one point every routed sink goes through. Five tests fail, one per sink, and nothing else moves:

```
arrow_loader::tests::format_cell_encodes_a_backslash_and_a_quote_per_dialect
quarantine::unit_tests::quarantine_accepted_values_encodes_a_backslash_and_a_quote_per_dialect
seeds::tests::format_value_encodes_a_backslash_and_a_quote_per_dialect
tests::unit_tests::accepted_values_encodes_a_backslash_and_a_quote_per_dialect
unit_test::tests::json_to_sql_literal_encodes_a_backslash_and_a_quote_per_dialect

test result: FAILED. 2073 passed; 5 failed
```

The two discovery sinks carry their own dialect-specific assertions rather than routing through this arm, so they are covered by the fail-before evidence above instead.

## Not in this PR

- `rocky-lang/src/lower.rs`: the DSL lowers to a SQL string before any dialect is chosen. Routing it needs the dialect at lowering time, which is a design change, not a call-site fix.
- `regex_match.pattern`: a backslash is regex syntax. On Snowflake and Databricks the lexer consumes it today, so `\d` reaches the regex engine as `d`, and a user who worked around that by writing `\\d` would see the encoder double it again. That is a behaviour change for a config that works today on those two warehouses, so it is a separate PR for the owner to judge.
- Databricks tag values (`rocky-core/src/catalog.rs`) and lakehouse `COMMENT` / `TBLPROPERTIES` (`rocky-core/src/lakehouse.rs`): both refuse a quote, neither refuses a backslash, and neither takes a dialect. A trailing backslash breaks the statement rather than escaping the literal (a quote cannot follow it). Follow-up.

## Red team

Codex review in progress — reviewer, rounds and the disposition of each finding are posted as one comment and summarised here when done.

## The two standing questions

**Least confident:** that the dialect-less `generate_test_sql` path is unreachable from a `rocky` command with user values. I grepped every caller: the only non-test caller is `rocky-core/tests/integration.rs`, and both production callers (`commands/test.rs`, `commands/run_local.rs`) use `generate_test_sql_with_dialect`. The guard stays there, so the direction of harm if I am wrong is a refusal, not a bypass. Second: the DuckDB `LIKE` fact (a backslash is a plain character, no default `ESCAPE`) is proven by the executed test on the DuckDB version the crate links; a future DuckDB default could change it, and the test would fail loudly.

**Most important thing I might be missing:** the encoder's `Backslash` rule is one rule for three warehouses, and Databricks can turn backslash reading off (`spark.sql.parser.escapedStringLiterals=true`, noted in #1605). Under that session flag every routed Databricks sink would now emit `\'`, which is a parse error there, where `''` used to work. This PR does not add a policy for that flag; it inherits #1605's position that the default parser is the supported one.

Refs #1596 (not closed: four sinks remain, listed above). Refs #1524.

### DIFF
--- a/engine/CHANGELOG.md
+++ b/engine/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Ten SQL string-literal sinks now encode through the dialect's own lexer rule, and the interim backslash refusal from #1595 is gone where they do.** A value in `accepted_values` (declarative checks and quarantine assertions), a seed cell, a `rocky load` cell on BigQuery's local-file `INSERT` path, a `rocky test` unit-test fixture string, `schema_pattern.prefix` in DuckDB and BigQuery discovery, the schema name DuckDB discovery reads back from the warehouse, a Snowflake tag value and the cross-source `_src` tag were each spliced into `'…'` by doubling quotes. On Snowflake, Databricks and BigQuery a value ending in `\` escaped the first quote of that pair and let the second one close the literal; #1595 refused any backslash at four of those sites, on every warehouse. Each sink now calls `rocky_core::sql_gen::string_literal(dialect, value)`, so a backslash stands for itself on DuckDB and Trino and is doubled, with the quote written as `\'`, on Snowflake, Databricks and BigQuery.
+
+  **On upgrade:** a backslash in `accepted_values` or in `schema_pattern.prefix` is accepted again and carried as a value. For a value with no quote, backslash or line break the generated SQL is byte-identical to 1.73.0. For a value that has one, it differs only on the three backslash-reading warehouses. Snowflake tag values were already escaped correctly and now share the one encoder. The dialect-less `generate_test_sql` API still refuses a backslash, because it has no lexer rule to ask; every `rocky` command passes a dialect.
+
+  **Still not routed:** a string literal written in a `.rocky` model (`rocky-lang` lowers to SQL before any dialect is chosen); `regex_match.pattern`, where a backslash is regex syntax and Snowflake and Databricks read it as an escape before the regex engine sees it; a Databricks tag value and a lakehouse `COMMENT` or `TBLPROPERTIES` value, which refuse a quote but still let a trailing backslash break the statement. (Refs #1596, #1524)
+
 ## [1.73.0] — 2026-09-03
 
 ### Added
--- a/engine/crates/rocky-bigquery/src/discovery.rs
+++ b/engine/crates/rocky-bigquery/src/discovery.rs
@@ -22,10 +22,12 @@ use std::sync::Arc;
 use async_trait::async_trait;
 
 use rocky_core::source::{DiscoveredConnector, DiscoveredTable, DiscoveryResult};
+use rocky_core::sql_gen::string_literal;
 use rocky_core::traits::{AdapterError, AdapterResult, DiscoveryAdapter, WarehouseAdapter};
 use rocky_sql::validation::{validate_gcp_project_id, validate_identifier};
 
 use crate::connector::BigQueryAdapter;
+use crate::dialect::BigQueryDialect;
 
 /// BigQuery discovery adapter that lists datasets matching a prefix.
 pub struct BigQueryDiscoveryAdapter {
@@ -59,6 +61,23 @@ impl BigQueryDiscoveryAdapter {
     }
 }
 
+/// The `INFORMATION_SCHEMA.SCHEMATA` query for every dataset whose name
+/// starts with `schema_prefix`.
+///
+/// `project` and `region` are validated by the caller. The prefix is user
+/// text from `schema_pattern.prefix`, encoded by BigQuery's own lexer rule
+/// (it reads backslash escapes, and `''` is a syntax error there rather
+/// than an escaped quote). `STARTS_WITH` reads the result as plain text, so
+/// `_` and `%` need no `LIKE` treatment.
+fn schemata_sql(project: &str, region: &str, schema_prefix: &str) -> String {
+    let prefix_lit = string_literal(&BigQueryDialect, schema_prefix);
+    format!(
+        "SELECT schema_name FROM `{project}.{region}.INFORMATION_SCHEMA.SCHEMATA` \
+         WHERE STARTS_WITH(schema_name, {prefix_lit}) \
+         ORDER BY schema_name"
+    )
+}
+
 #[async_trait]
 impl DiscoveryAdapter for BigQueryDiscoveryAdapter {
     async fn discover(&self, schema_prefix: &str) -> AdapterResult<DiscoveryResult> {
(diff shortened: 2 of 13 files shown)


hugocorreia90:
**Independent red team: Codex (GPT-5), `VERDICT: PASS`, no findings.**

Two rounds. The first sprawled off the diff into unrelated files and compacted its own context without reaching a verdict, so it was discarded, not counted. The second was pinned to the extracted diff with exploration disabled.

What it was asked, and what it answered:

- **Escape order in the backslash rule.** Sound. Doubling existing backslashes first is what stops a trailing backslash from escaping the closing quote, and the `\n`, `\r` and `\'` that later passes add cannot be doubled by an earlier one. It found no input that closes the literal early.
- **Guard retirement.** Safe at all four deletions. It checked each one reaches the encoder on the same path: `rocky-bigquery/src/discovery.rs` via `schemata_sql`, `quarantine.rs` per value, `tests.rs` via `accepted_value_literal`, and both `rocky-duckdb/src/discovery.rs` sites. The dialect-less `generate_test_sql` keeps its refusal, which is the intended split.
- **Dialect assignment.** Correct everywhere. The only inline escape rule left is the guarded dialect-less path.
- **Plain-value regression.** None. Values with no quote, backslash, CR or LF emit the same bytes, and DuckDB's `%` stays inside the same literal.

Its own stated limit, which is fair and matches this PR's evidence section: it could not execute anything, so warehouse lexer behaviour was taken from the brief. The Snowflake, Databricks and BigQuery round trips remain unexecuted here; the DuckDB probe in the description is the one that actually ran against a warehouse.

The mutation check in the description was added after this review: reverting the backslash rule to plain quote-doubling fails five tests, one per routed sink.

PR #1651, opened 2026-09-04
fix(bigquery): collect all paginated query results

### LINKED ISSUE #1650: BigQuery content-addressed models silently truncate paginated query results
## Description

`BigQueryAdapter::execute_query` returns only the first REST result page. The connector sets `maxResults` to 10,000, but `BigQueryResponse` does not deserialize `pageToken`, and both the synchronous `jobs.query` completion path and the deferred `jobs.getQueryResults` completion path return immediately. `execute_query` then converts only that response's `rows`.

BigQuery documents `maxResults` as a per-page limit and a non-empty `pageToken` as the signal that more rows must be fetched:

- https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query
- https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults

This is more than a display issue. `execute_content_addressed_model` obtains the model's rows through `execute_query`, converts them to Arrow/Parquet, and commits the result as the complete table. A BigQuery content-addressed model with more than one result page can therefore report success while committing only the first page. Live hash re-derivation and restore use the same path and can reproduce or validate the same incomplete bytes.

## Steps to reproduce

1. Point `BigQueryAdapter` at the existing wiremock test seam.
2. Return this successful `jobs.query` response:

```json
{
  "jobComplete": true,
  "jobReference": {"projectId": "test-project", "jobId": "job-paged"},
  "schema": {"fields": [{"name": "id", "type": "INTEGER"}]},
  "rows": [
    {"f": [{"v": "1"}]},
    {"f": [{"v": "2"}]}
  ],
  "totalRows": "3",
  "pageToken": "next-page"
}
```

3. Call `adapter.execute_query("SELECT id FROM paged")`.
4. Assert that the result contains all three rows.

A nearby control response with `totalRows: "2"`, the same two rows, and no `pageToken` correctly returns two rows.

## Expected behavior

Rocky follows `pageToken` through `jobs.getQueryResults` until it is empty and returns every row in order. If BigQuery supplies a page token without a usable job reference, Rocky fails closed instead of returning an incomplete successful result.

## Actual behavior

The call succeeds with two rows. The token is discarded during deserialization, no second request is issued, and callers cannot tell that the result is incomplete.

```text
single-page control rows: 2
paged response rows returned: 2 (BigQuery totalRows: 3)
assertion failed: adapter must fetch every BigQuery result page
  left: 2
 right: 3
```

The reproduction is deterministic and produced the same result twice.

## Affected paths

- `engine/crates/rocky-bigquery/src/connector.rs`: `run_query_once`, `poll_query_results`, `BigQueryResponse`, and `execute_query`
- `engine/crates/rocky-cli/src/commands/run_content_addressed.rs`: content-addressed writes and live hash re-derivation
- `engine/crates/rocky-cli/src/commands/restore.rs`: artifact rebuilding
- Generic query consumers such as `rocky shell`

Warehouse-side CTAS, merge, and full-refresh statements are not affected because they do not pull the materialized rows through `execute_query`.

## Suggested fix and regression coverage

Deserialize `pageToken`, centralize `jobs.getQueryResults` page fetching, and append pages until the token is empty. Preserve schema and row order. Add wiremock coverage for synchronous and deferred first pages, a later-page HTTP/job error, a page token without a job reference, and a successful 10,001-row result.

## Rocky version

- `upstream/main`: `e03ac5599dffa0659d4f3fe377c24153ff3d9b59`
- Latest release: `engine-v1.73.0`
- The affected files are unchanged between the release and current main.

## OS / architecture

Reproduced on macOS 26.6.2 arm64 with Rust 1.97.1. The failure is API-response handling and is platform-independent.

## Configuration

No live credentials or project configuration are required for the wire-level reproduction; it uses the adapter's existing `test-support` base-URL seam and static bearer-token test authentication.

## Prior-art check

All-state searches across issue/PR titles, bodies, and comments for `pageToken`, `maxResults BigQuery`, `BigQuery pagination`, `totalRows BigQuery`, and truncation variants found no report or active fix. PRs #129, #795, and #796 cover initial polling, HTTP mocks, retries, and error mapping, but do not retrieve subsequent pages.

mattfaltyn:
## Summary

Fetch every BigQuery `jobs.getQueryResults` page before returning an `execute_query` result. This prevents successful content-addressed model runs, restores, and query consumers from silently using only the first 10,000-row page.

Fixes #1650.

## Subproject(s) touched

## Changes

- Deserialize and follow BigQuery `pageToken` values for both inline and deferred query completions.
- Preserve schema and row order while accumulating pages, then validate the accumulated count against `totalRows`.
- Fail closed on missing job references, repeated tokens, incomplete later pages, and later-page API/job errors.
- Apply the adapter's bounded transient retry, backoff, and shared retry budget to later-page requests.
- Add wire-level coverage for 10,001-row pagination, deferred pagination, transient recovery, partial-failure refusal, malformed tokens, and row-count mismatch.

## Test Plan

- `cargo nextest run --all-features` — 6,828 passed; 115 credential/measurement tests skipped by the repository.
- `cargo test` — passed, including doctests; credential-gated live tests remained ignored.
- `cargo test -p rocky-bigquery --all-features --test wiremock_tests -- --nocapture` — 36 passed.
- `scripts/mutation-check.sh ... inline_query_collects_all_result_pages` — passed; removing the page collection makes the assertion fail at 10,000 versus 10,001 rows.
- Monorepo unaffected-consumer checks: SDK 241 passed, Dagster 724 passed, VS Code 208 passed.

`cargo test --all-targets` cannot complete on current `main`: its `rocky-cli` `binary_startup` benchmark starts a nested `cargo run` against the target directory locked by the parent Cargo process. The run was stopped at that pre-existing deadlock after all preceding tests passed. The CI-equivalent nextest and all-target clippy commands above completed successfully.

## Adversarial review

Fermat (independent GPT-5 Codex) initially found that later-page requests bypassed transient retry handling. The patch now uses the bounded adapter retry policy and includes a 503-then-success regression. Fermat's second-pass disposition is **accept**, with no blocking findings and ModelIr/contract semantics preserved.

- **Least confidence:** rare live BigQuery response shapes around optional `jobReference` behavior are mocked, not exercised against a real project.
- **Most important missing evidence:** a live BigQuery content-addressed run above 10,000 rows that verifies the final committed Parquet artifact end to end. The adapter boundary and content-addressed caller were traced, and the wire contract is covered locally.

## Checklist

- [x] This does not change CLI JSON schema or DSL syntax

### Engine (`engine/`)

- [x] All engine tests pass under `cargo nextest run --all-features` and `cargo test`

### DIFF
--- a/engine/crates/rocky-bigquery/src/connector.rs
+++ b/engine/crates/rocky-bigquery/src/connector.rs
@@ -1,5 +1,6 @@
 //! BigQuery REST API connector — jobs.query + jobs.getQueryResults.
 
+use std::collections::HashSet;
 use std::time::Duration;
 
 use async_trait::async_trait;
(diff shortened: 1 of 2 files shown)


hugocorreia90:
Merged — thank you. The pagination fix is right, and the wiremock coverage of the deferred path, the retry budget and the repeated-token case is what made it easy to verify without a live BigQuery.

PR #1652, opened 2026-09-04
fix(engine/rocky-cli): report a replication check the engine could not run

### LINKED ISSUE #1602: A row-count check whose query fails reports a pass, because both sides default to zero
A row-count check whose query fails is reported as **passing**, not as an error. Both sides default to zero, and zero equals zero.

Found by a Codex review of #1599 (a different change); this one is pre-existing on `main` and independent of it.

## The shape

In the replication check path, the source and target row-count results are each unwrapped with a zero default. A query that errors, returns no rows, or returns an unreadable cell therefore contributes `0`. The check then compares `0` against `0`, finds no difference, and records a pass.

```
  source query fails  ->  0  ┐
                            ├─  0 == 0  ->  check PASSES
  target query fails  ->  0  ┘
```

So the failure mode is not "the check did not run" — it is "the check said the data is fine". That is the fail-open direction.

## Why it matters more than it looks

A row-count check exists to catch a copy that silently moved nothing. The one circumstance most likely to break the check's own query — a missing table, a permission change, a warehouse error — is also a circumstance where the copy may genuinely have moved nothing. The check is least trustworthy exactly when it is most needed.

## Siblings worth deciding at the same time

The same "skip quietly" shape appears nearby, though those produce *no* result rather than a false pass:

- freshness: a per-table query error is only warned, and a NULL or unparseable timestamp emits no check at all, because only `Some(ts)` creates one
- null-rate: SQL generation failures, malformed result cells, missing column names and query errors are all skipped
- assertions and custom checks: a table-reference formatting error is skipped

Those are less severe — an absent check is honest in a way a false pass is not — but they are the same decision: what should the engine do when it cannot evaluate a check it was asked to run?

## Suggested direction

`not_evaluated` already exists for exactly this. #1595 added it for the cross-source overlap check so a check the engine could not run would stop reading as a clean zero. The same state fits here, and it keeps "could not evaluate" distinct from both "passed" and "failed".

The row-count case should change regardless of the wider decision: a defaulted zero must never produce a pass.

## Direction of harm

Today: a broken check reports success. After: it reports that it could not run. The second is loud and recoverable; the first is silent and looks like health.

Related: #1595 (introduced `not_evaluated`), #1598 (whether a failed check should fail the run at all).

hugocorreia90:
A row-count check whose source or target query failed reported a **pass**. Both sides defaulted to zero, and zero equals zero, so a copy that moved nothing looked exactly like a copy that moved everything.

```
  source query fails  ->  0  ┐
                            ├─  0 == 0  ->  check PASSES
  target query fails  ->  0  ┘
```

Four more checks on the replication path had the same defect in the other direction: they disappeared from the tally instead of failing.

Closes #1602

## What was wrong, per check

| Check | Failure | Before | Now |
|---|---|---|---|
| row count | either side's query errors, or returns an unreadable cell | side defaults to `0`, compares equal, **passes** | fails, reason names the side |
| row count | the batch query silently omits a table | table absent from the results, no check emitted | fails, reason says it was left out |
| freshness | query fails, or the cell is not a timestamp | result dropped, no check emitted | fails, reason names the cell |
| null rate | query fails, or its SQL cannot be generated | table skipped, no column reported | every configured column fails |
| null rate | one column has no readable row | that column skipped | that column fails, the others keep their measurements |
| assertion / custom | the table reference cannot be formatted | check skipped | fails, reason names the table |

## The field

`not_evaluated` already existed on the cross-source-overlap check. It moves up to `CheckResult`, so every check kind can carry it and `--output json` consumers read it on `checks[]` directly.

While it is set, `passed` is `false` and the numeric fields (`source_count`, `target_count`, `lag_seconds`, `null_rate`, `failing_rows`, `result_value`, `overlap_count`) are placeholders, not readings. This is the schema change in this PR; `just codegen` regenerated `schemas/run.schema.json`, the Pydantic model, the TypeScript interface and the OpenAPI document. The hand-written SDK model already declared the field, so only its comment changed.

## Two things that deliberately did not change

**An empty null-rate sample is still a measured zero.** This is the defect the first draft of this change shipped, and it is worth naming. `SUM(CASE WHEN col IS NULL …)` over an empty sample returns NULL, not `0`, and `sample_percent` defaults to **10**. A small table sampled at 10% produces that row on an ordinary, healthy run. Reading the null count before the sampled count would therefore have reported a failing check for every small table — making a skipped check honest, and firing it on the common case. The sampled count is read first: zero sampled rows is an empty sample and a rate of zero. The surrounding code already said it wanted that; the branch was simply unreachable.

**An empty table still emits no freshness check.** `MAX(ts)` over an empty table is NULL because there is no row to be fresh, which is not the same as a cell that failed to parse.

**Exit codes are unchanged.** A failed replication check still does not fail the run. That is #1598, which was re-scoped after an earlier attempt (#1599) was blocked for creating a false green on resume. This PR only makes the reporting honest, which is the prerequisite that issue named.

## Evidence

Thirteen tests, written against the unfixed code. Fail-before, in one line each: a row count with both queries failing reported `passed: true`; a freshness query that failed emitted no check at all; a null-rate query that failed left the table out; an assertion on an unaddressable table was silently dropped.

- `cargo +1.98.0 test -p rocky-core -p rocky-cli --lib`: **0 failed** (1612 rocky-cli, 2073 rocky-core).
- `uv run pytest -q` in `sdk/python` (241) and `integrations/dagster` (724): pass. `ruff check` and `ruff format --check`: clean.
- **Mutation check** (`scripts/mutation-check.sh`, clean tree enforced) on the empty-sample guard — the one case that would have made this change worse than the bug. Reverting it to read the null count first fails `an_empty_null_rate_sample_is_measured_not_reported` and nothing else. Verdict: `OK — the test failed under mutation, so it is fix-sensitive.`
- No dagster fixture changed: the field is omitted when unset, and no captured fixture holds a check the engine could not run.

## Also here

The first commit lifts the batched replication checks out of `run()` into their own functions, with no behaviour change. The check bodies were unreachable from a test before that, which is why these five defects had no coverage.


hugocorreia90:
**Independent red team: Codex (GPT-5), `VERDICT: BLOCK`, three findings. One fixed here, two filed as pre-existing.**

**HIGH — batched freshness dropped a table the batch query left out. Fixed in this PR.** The freshness pass iterated the results the adapter returned, so an omitted table produced no entry and therefore no check — the same silent drop this PR closes everywhere else. Worse, the batch path never populates the per-table failure map, so nothing else would have noticed. Row counts already reconciled against the *requested* tables; freshness now does the same. An empty table still emits no freshness check, and the test pinning that still passes. New test `a_table_the_batch_freshness_left_out_is_not_evaluated`, mutation-checked: reverting the loop to read the returned entries fails it and nothing else.

This one was a real gap between what the description claimed and what the code did, on the path every batching warehouse takes.

**HIGH — cross-source overlap reports a tolerated keyless sibling as a failed check. Pre-existing → #1654.** The reviewer read this as new because the refactor commit moves the function; the identical arm is on `main` today, verified before filing. It is a real defect, and the fix needs to tell a tolerated missing key from a genuine query failure, which the error type alone cannot do. Out of scope here.

**MEDIUM — a batched query error aborts the run while the per-table path reports it. Pre-existing → #1655.** Also on `main`. This PR makes the per-table side honest, which widens the gap rather than creating it. Aborting is fail-closed, so it is a consistency and blast-radius problem, not a false green.

**Cleared:** every other ordinary-run shape it checked — empty null-rate samples, empty and no-match row-set assertions, empty freshness, a first run with no history — reaches no new failing branch. Every constructor setting `not_evaluated` also sets `passed = false` with documented placeholder numerics, and it found no duplicate emission.

Its stated limit: it could not see the adapter implementations, so how each one represents a non-NULL timestamp that fails to parse is unsettled. The reconciliation above makes the omission case safe regardless; a batch adapter that returns `None` for an unreadable cell would still read as an empty table, which is #1655's territory.

Issue #1645, opened 2026-09-04
dagster: a declared default check the engine did not produce reports passed=True


hugocorreia90:
A declared default check that the engine did not produce is reported as `passed=True` on a materialized table.

`integrations/dagster/src/dagster_rocky/component.py`, `_emit_placeholder_checks`, last branch:

```python
        materialized = cs.asset_key in materialized_keys
        if materialized:
            reason = f"not produced by rocky (check type: {cs.name})"
            severity = dg.AssetCheckSeverity.ERROR
        ...
        yield dg.AssetCheckResult(..., passed=materialized, severity=severity, metadata={"status": ...})
```

For `row_count_anomaly` and `compliance_exception` this is by design: no event means no anomaly, and the docstring on `_build_check_specs` says so. But `_build_check_specs` declares `row_count`, `column_match` and `freshness` on every asset too, and those can be switched off in `rocky.toml` (`engine/crates/rocky-core/src/config.rs`, `ChecksConfig`: `enabled`, `row_count: AggregateCheckToggle`, `column_match: AggregateCheckToggle`, `freshness: Option<FreshnessConfig>`). The engine gates that check on the toggle (`engine/crates/rocky-cli/src/commands/run.rs`, `check_row_count: pipeline.checks.row_count.enabled()` and `row_count_enabled = ...`), so with `row_count` disabled it emits no `row_count` result, and Dagster reports the check as **passed** with `status = "not produced by rocky (check type: row_count)"`. The status text is honest; the verdict is not, and the UI badge shows the verdict.

Same class as #1619 (a "not run" state that reads as "clean"), on the default checks rather than the contract checks. Options: `passed=False` at `WARN` for a declared check the engine did not produce (keeping the pass-by-absence only for the two checks that are defined that way), or declare the spec only when the check is enabled in discover's check projection.

Refs #1619, #1642.

hugocorreia90:
**Verified, and the two options are not equal in cost. Here is what the code actually allows today.**

Confirmed the reported behaviour: `DEFAULT_CHECK_NAMES` is `("row_count", "column_match", "freshness", "row_count_anomaly")` and `_build_check_specs` declares all four on every asset unconditionally. In the last branch of `_emit_placeholder_checks`, a materialized table with no result for a declared check yields `passed=materialized`, which is `True`. The `severity = ERROR` on that line never shows, because severity only reads on a failing result. So the badge is green and only the status text says the check never ran.

[...]

PR #1656, opened 2026-09-04
fix(dagster): fail a contract check when the compiler did not find the model

### LINKED ISSUE #1644: dagster: a contract for a model the compiler did not find (W011) reports green
A Dagster contract check reports `passed=True` when the compiler could not find the model the contract names.

`engine/crates/rocky-compiler/src/compile.rs` (`validate_all_contracts`): when a contract's model is not in `typed_models`, the engine emits **`W011`** ("contract exists for '<model>' but model was not found in project") instead of running `validate_contract`. No `E010`–`E014` can follow, because nothing was validated.

`integrations/dagster/src/dagster_rocky/contracts.py`: `_CONTRACT_CODE_TO_CHECK` maps `E010`, `E011`, `E012`, `E013`, `E014` and `W010`; `_CONTRACT_UNVERIFIED_CODE_TO_CHECK` maps `I003`. `W011` is in neither map, and the dispatch in `contract_check_results_from_diagnostics` drops an unmapped code silently. So every declared contract check for that model passes.

How a user reaches it: `contracts_dir` holds `orders.contract.toml`, a source table is named `orders` (so the check specs are declared), but the compiled project has no model named `orders`, for example after a rename. The engine says `W011`; Dagster says green three times.

Same class as #1619 (fixed in #1642 for the missing-compile case): a "not checked" state that reads as "clean". Suggested shape: map `W011` to every declared check kind for that model as a not-verified failure (`passed=False`, `WARN`), the way #1642 does for a missing compile.

Refs #1619, #1642.

hugocorreia90:
A Dagster contract check reported `passed=True` when the compiler never found the model the contract names.

```
  contract names `orders`  ->  model missing from the project
                           ->  compiler emits W011 INSTEAD of validating
                           ->  no E010-E014 can follow
                           ->  W011 unmapped, dropped at the dispatch
                           ->  every declared contract check GREEN
```

Closes #1644

## Why nothing else caught it

`validate_all_contracts` branches: with the model present it calls `validate_contract`, and with the model absent it pushes `W011` and returns. So an empty violation set on that path is not evidence of a clean contract, it is evidence that nothing looked. `_CONTRACT_CODE_TO_CHECK` maps the five `E0xx` codes and `W010`; `_CONTRACT_UNVERIFIED_CODE_TO_CHECK` maps `I003`. `W011` was in neither, and the dispatch drops an unmapped code silently.

The way a user reaches it is a rename: the `.contract.toml` still names `orders`, a source table is still called `orders` so the specs are still declared, and the compiled project has no model by that name.

## What changed

`W011` for this model now short-circuits the dispatch and reports every **declared** rule kind as not verified: `passed=False` at `WARN`, with `rocky/contract_model_not_found` in the metadata and a status naming the model. That is the shape #1642 already established for a missing compile result, so the two "nothing was checked" paths now report alike and share one fan-out helper.

A `W011` naming a *different* model is still filtered out by the existing `diag.model != model_name` guard, so one missing model cannot blank another model's checks. There is a test for exactly that, because the short-circuit would be easy to write in a way that does not respect it.

## Evidence

Two tests, written against the unfixed code.

- `test_results_not_verified_when_the_model_was_not_found` — fail-before: all three declared checks returned `passed=True` with no metadata. Now all three fail at `WARN` with the new key.
- `test_a_model_not_found_warning_for_another_model_does_not_leak` — guards the filter.

Gates from `integrations/dagster/`: `uv run pytest -q` **735 passed**; `ruff check .` and `ruff format --check .` clean.

**Mutation check** (`scripts/mutation-check.sh`, clean tree enforced): disabling the new branch fails `test_results_not_verified_when_the_model_was_not_found` and nothing else. Verdict `OK — the test failed under mutation, so it is fix-sensitive.`

## Not in this PR

#1645 is the sibling case on the *default* checks rather than the contract checks: a declared `row_count`, `column_match` or `freshness` check that the engine did not produce because it is switched off in `rocky.toml` also reports green. That one needs a decision, because the obvious fix would turn a currently-green check red for a config that works today.


hugocorreia90:
**Independent red team: Codex (GPT-5), `VERDICT: PASS`, one low note, no changes required.**

- **Short-circuit is safe.** `W011` and a real violation cannot coexist for the same model. I can close the gap the reviewer flagged as its own least-confident point, because it was held to the diff and could not read the producer: `validate_all_contracts` in `engine/crates/rocky-compiler/src/compile.rs` is a plain if/else — `if let Some(schema) = typed_models.get(model_name)` calls `validate_contract`, and the `else` pushes `W011` and nothing more. The two branches are mutually exclusive by construction, so the early return loses nothing that could have been produced.
- **Model filter is correctly ordered.** `diag.model != model_name` is evaluated before the new branch, so a `W011` for another model cannot blank this one's checks.
- **The extracted fan-out is behaviour-preserving.** Both paths emit required, then protected, then column-constraints, and the missing-compile builder and its metadata key are untouched.
- **No false failure on a healthy project.** A model that compiles produces no `W011` naming itself.

**Low, and accepted as written:** the companion test that a `W011` for a *different* model does not leak is not reversion-sensitive — it passes with or without the fix. That is what it is for. It guards the ordering of the model filter against the short-circuit, which is the mistake this change could easily have made, and the primary test is the one that is mutation-checked.

PR #1657, opened 2026-09-04
fix(engine/rocky-duckdb): match a discovery prefix as text, not as a LIKE pattern

### LINKED ISSUE #1649: DuckDB discovery treats an underscore in schema_pattern.prefix as a LIKE wildcard
`rocky-duckdb/src/discovery.rs` (line 47 on `main` after #1648) finds schemas with `WHERE schema_name LIKE '<prefix>%'`. In `LIKE`, `_` matches any one character. A prefix such as `src__` therefore also matches `srcab`, `src1x` and every other schema whose first five characters fit that shape.

What happens today:

```
prefix = "src__"           LIKE 'src__%'
schemas: src__orders  ──▶  matched   (wanted)
         srcab        ──▶  matched   (not wanted)
```

The over-match is caught later. `rocky discover` (`rocky-cli/src/commands/discover.rs`, line 121) tries to parse each schema name against the pattern, fails on the unwanted ones, logs `skipping source with unparseable schema`, and moves on. So the visible cost is one warning and one `information_schema.tables` query per unwanted schema, not wrong output.

The BigQuery discovery adapter uses `STARTS_WITH`, which reads the prefix as plain text. DuckDB has `starts_with(schema_name, '<prefix>')`, so the same shape works there and removes both the `_` and the `%` wildcard reading.

Found while routing the literal through the dialect encoder in #1648; left out of that PR because it changes which schemas the query returns, not how the literal is encoded.

hugocorreia90:
DuckDB discovery found schemas with `WHERE schema_name LIKE '<prefix>%'`. In a `LIKE` pattern `_` matches any single character, so a prefix of `raw__` also matched `rawXY`, `raw12`, and every other schema of that shape.

```
prefix = "raw__"        LIKE 'raw__%'       starts_with(…, 'raw__')
  raw__orders     ──▶   matched   (wanted)    matched
  raw__customers  ──▶   matched   (wanted)    matched
  rawXY           ──▶   matched   (NOT)       not matched
```

Closes #1649

## What it cost

Not wrong output. `rocky discover` parses each returned schema against the pattern, fails on the unwanted ones, logs `skipping source with unparseable schema`, and moves on. So the price was a warning per unwanted schema plus one wasted `information_schema.tables` query each — on a warehouse where many schemas share a prefix shape, once per discovery.

## The fix

`starts_with(schema_name, '<prefix>')`. It reads the prefix as plain text, which is what a prefix means, and it removes the `%` as well. The BigQuery adapter already used `STARTS_WITH` for exactly this reason, so the two adapters now agree.

A prefix with no underscore behaves exactly as before.

## Evidence

`test_discover_matching_schemas` gains a decoy schema, `rawXY`, that the old query matched and the new one does not. Fail-before: the test asserts 2 connectors and the `LIKE` query returns 3.

- `cargo +1.98.0 test -p rocky-duckdb discovery`: 3 passed, 0 failed — including the existing backslash/quote literal test from #1648, so the dialect encoding is unchanged.
- `clippy --all-targets -p rocky-duckdb -- -D warnings`: clean. `fmt --all --check`: clean.
- **Mutation check** (`scripts/mutation-check.sh`, clean tree enforced): restoring the `LIKE '<prefix>%'` query fails `test_discover_matching_schemas` and nothing else. Verdict `OK — the test failed under mutation, so it is fix-sensitive.`

Found while routing this literal through the dialect encoder in #1648, and deliberately left out of that PR because it changes which schemas the query returns rather than how the literal is encoded.


hugocorreia90:
**Independent red team: Codex (GPT-5), `VERDICT: BLOCK`, two medium findings. One applied, one rejected with evidence.**

**Applied — the changelog claim was too narrow.** It said a prefix with no underscore behaves exactly as before. `%` is the other `LIKE` wildcard and changes meaning too, so `prefix = "raw%"` now means those literal characters. Both are named now.

The reviewer also asked whether anyone could be relying on the wildcard reading deliberately. They cannot, and this is the fact that settles it — it was outside the diff so the reviewer could not see it. Every other use of this same `prefix` value already treats it literally: `SchemaPattern::parse` (`engine/crates/rocky-core/src/schema.rs`) does `strip_prefix(&self.prefix)` and errors with `schema '<name>' does not start with prefix '<prefix>'`. So a wildcard prefix discovered extra schemas that then **failed to parse** — the `skipping source with unparseable schema` warning in the issue is exactly that. Discovery was the odd one out; this makes it agree with the parser. The changelog now says so.

**Rejected — "the test would pass after reversion".** It does not. The reviewer inferred this from the changelog's note that unmatched schemas are skipped later, but that skipping happens in `rocky discover` (the CLI), not in the adapter under test. `DuckDbDiscoveryAdapter::discover` returns one connector per schema the query returned, so with `LIKE` the decoy `rawXY` becomes a third connector and the `len() == 2` assertion fails.

That is not an argument, it is a run. `scripts/mutation-check.sh` restored the `LIKE '<prefix>%'` query on a clean tree and reported:

```
failures:
    discovery::tests::test_discover_matching_schemas
```

Its own second reviewer (`gpt-5.6-terra`) said the same thing I am saying here and was overruled internally on the strength of the changelog sentence. The empirical result stands.

**Cleared:** `starts_with(p)` and `LIKE p || '%'` agree for any prefix without `_` or `%`, including the empty prefix, quotes and backslashes, under DuckDB's comparison rules; removing the appended `%` reopens no injection seam; and `starts_with` is the correct function name and long predates the DuckDB version this engine bundles (1.5.5 via `libduckdb-sys 1.10505.0`).

Issue #1548, opened 2026-08-29
bug(engine): --resume-latest resumes a successful run and copies nothing


hugocorreia90:
## Description

`rocky run --resume-latest` resumes the most recent run of any status, including one that already succeeded. The CLI help and the docs both call it "the most recent **failed** replication run". `get_latest_run_progress()` picks purely by `started_at`.

When the latest run succeeded, every table in its checkpoint is `Success`, so the resume skips all of them and exits 0 having copied nothing — even when new source rows have landed since.

This predates #1544, but #1544 raised the stakes: a resume that finds no checkpoint now hard-fails, which teaches an operator that a non-error means "my recovery worked". Here a non-error means "there was nothing to do", and the two are indistinguishable from the exit code.

## Version

- Reproduced on `54619294` (main, after #1544) with a debug build.

## Steps to reproduce

Using the replication config from #1543:

```shell
rocky -c rocky.toml --state-path state.redb -o json run
#   "tables_copied": 1

rocky -c rocky.toml --state-path state.redb -o json run --resume-latest
#   "status": "Success"
#   "tables_copied": 0
#   "tables_skipped": 1
#   "resumed_from": "run-20260829-193742-331"
```

The second command exits 0 and reports a successful resume. It copied nothing. New rows arriving in `raw__orders.one` between the two commands are not picked up.

## Expected behavior

One of these, decided deliberately:

- `--resume-latest` selects the most recent run that did **not** succeed, matching what the help text already promises; a fully successful latest run is refused with a message saying there is nothing to resume, and
- the crash case is defined explicitly — a run killed before writing a terminal status must still be resumable, so "not succeeded" cannot simply mean "status == Failure".

Or, if the current behavior is intended, the help text and `docs/src/content/docs/reference/cli.md` stop saying "failed run".

## Actual behavior

`engine/crates/rocky-core/src/state.rs:2838-2850` iterates the `RUN_PROGRESS` header table and keeps the row with the greatest `started_at`. No status is consulted, and `RunProgress` carries no run-level status field to consult.

## Impact

An orchestrator that retries with `--resume-latest` after a run it believes failed — but which actually succeeded, or which partially succeeded and was later completed by another path — silently does no work and reports success. The failure mode is a stalled pipeline that looks healthy, which is worse than an error.

## Suspected cause

`get_latest_run_progress()` has no status predicate, and `RunProgress` (`state.rs:2239-2244`) stores only `run_id`, `started_at`, `total_tables`, `tables`.

[...]

PR #1658, opened 2026-09-04
fix(engine/rocky-cli): refuse an explicit resume of a run that already succeeded

### LINKED ISSUE #1635: rocky run --resume <run-id> accepts a checkpoint whose run already succeeded
`--resume-latest` refuses to resume a run that succeeded (#1573). The explicit form does not.

## Where

`engine/crates/rocky-cli/src/commands/run.rs`, `resolve_resume_progress`:

```
--resume-latest        require_resume_progress -> ensure_resume_scope -> ensure_latest_run_is_resumable
--resume <run-id>      require_resume_progress -> ensure_resume_scope -> (returns)
```

`ensure_latest_run_is_resumable` bails on `RunStatus::Success`, `SkippedIdempotent` and `SkippedInFlight`. The `resume_run_id` arm never calls it.

## Consequence

`rocky run --resume <id of a run that succeeded>` loads that run's progress, treats every `Success` entry as done, and exits 0 having copied nothing. A script that pins a stale run id gets a green run that did no work. Default retention keeps 100 run records, so the checkpoint is there to be found.

## Reproduce

1. Run a replication pipeline to success; note its run id from `rocky -o json history`.
2. `rocky run --resume <that id>` — exits 0, `tables_skipped` = every table, `resumed_from` set.
3. `rocky run --resume-latest` on the same state — refused with `nothing to resume: run <id> already succeeded`.

## Fix shape

Call the same gate in both arms. The doc on `ensure_latest_run_is_resumable` and its name say "latest"; the check is about the run record, not about how it was selected.

Found by independent review of the checkpoint-resume POC rewrite (#1634).

hugocorreia90:
`rocky run --resume <run-id>` accepted a checkpoint whose run already succeeded. It loaded the progress, treated every entry as done, copied nothing, and exited 0.

```
--resume-latest   require_progress -> ensure_scope -> ensure_resumable -> refuses
--resume <id>     require_progress -> ensure_scope ->        (nothing)  -> exit 0
```

Closes #1635

## Why it matters

A script that pins a stale run id gets a green run that did no work. Default retention keeps 100 run records, so the id stays resolvable long after the run it names. The run reports success because, from the checkpoint's point of view, there was nothing left to copy — which is exactly what the existing gate was written to catch.

## This is not a new policy

`ensure_latest_run_is_resumable` already refused `Success`, `SkippedIdempotent` and `SkippedInFlight` for `--resume-latest` (#1548). The explicit arm never called it. A run id names the same checkpoint however it was chosen, so the gate now runs on both, with the same message:

```
nothing to resume: run <id> already succeeded
```

The function is renamed to `ensure_run_is_resumable`, because it was named for a path it no longer exclusively serves.

## Nothing recoverable stops being recoverable

The gate deliberately keeps two cases resumable, and widening a refusal is exactly where those get lost by accident:

- `Failure` and `PartialFailure` still resume.
- **A checkpoint with no run record at all still resumes.** That is the crash case: a run killed before it reached a terminal status still has post-copy work to finish, and a complete-looking checkpoint is what such a crash leaves behind.

Rather than adding a parallel pair of tests, both existing tests now run their assertions against both entry points, so the allow-side case is proven on the explicit form too.

## Evidence

- Fail-before: with a `Success` record seeded, `resolve_resume_progress(store, Some("run-1"), false, scope)` returned the progress instead of an error.
- `cargo +1.98.0 test -p rocky-cli --lib`: **1613 passed, 0 failed.** `clippy --all-targets -p rocky-cli -- -D warnings` and `fmt --all --check`: clean.
- **Mutation check** (`scripts/mutation-check.sh`, clean tree enforced): removing the new `ensure_run_is_resumable` call from the explicit arm fails `resume_refuses_runs_with_nothing_to_recover` and nothing else. Verdict `OK — the test failed under mutation, so it is fix-sensitive.`

## On the merge class

This refuses a command that exits 0 today, so it is worth being explicit: the run it refuses did no work and reported success for it. The refusal replaces a false green, it does not stop anything that was being accomplished, and it makes the explicit form agree with a rule `--resume-latest` has enforced since 1.72.0. A user who genuinely wants to re-run that pipeline runs it without `--resume`.


hugocorreia90:
**Independent red team: Codex (GPT-5), `VERDICT: PASS`, no findings. A second model (`gpt-5.6-terra`) agreed.**

The question I most wanted answered was whether widening a refusal quietly closes a recovery path, and it does not:

- **Nothing recoverable is newly refused.** `Failure`, `PartialFailure` and the no-record crash case all still resume through the explicit form. Selecting an older id still selects an older checkpoint.
- **No asymmetry is justified.** Naming an id picks a checkpoint; it does not mean "force a re-execution". Nothing in the behaviour supports reading `--resume <id>` as an override, and a user who wants to re-run that pipeline simply runs it without `--resume`.
- **Ordering is unchanged.** The scope check still runs first on both paths, so a checkpoint that is both out-of-scope and non-resumable still reports the scope error.
- **No stale naming.** The rename is consistent across the function, its call sites, the tests, the comments and the changelog.

On tests, its second reviewer raised the one thing worth raising: could the refusal test stop at scope validation and never reach the gate? It does not — the checkpoint's scope and the current scope are identical in the fixture, and the assertion requires the gate's own message. The mutation check settles it independently: removing the new call fails that test and nothing else.

Its stated limit is fair. It was held to the diff, so it could not rule out references or additional run statuses elsewhere in the tree. `RunStatus` has five variants and the gate matches all five exhaustively, so a new one would have to be added deliberately and would not compile without a decision.

PR #1659, opened 2026-09-04
fix(engine/rocky-cli): emit NDJSON for a watched transformation pipeline

### LINKED ISSUE #1604: run --watch -o json: a transformation pipeline pretty-prints instead of emitting NDJSON
`rocky run --watch --output json` promises one compact JSON object per iteration, one per line. A transformation pipeline emits a pretty-printed object spread over many lines instead, so anything piping the stream cannot read it line by line.

## Measured

Same DuckDB fixture, same flags, four seconds of watching, one iteration each:

| Pipeline type | stdout lines for one iteration |
|---|---|
| replication | 1 (compact, as documented) |
| transformation | 65 (pretty-printed) |

## Where it comes from

`run_watch.rs` sets the process-global `COMPACT_JSON` flag, and `output.rs`'s `print_json` honours it. The transformation arm does not go through `print_json` — `engine/crates/rocky-cli/src/commands/run_local.rs` calls the pretty serializer directly, so the flag never reaches it.

## Why it was never caught

The only watch test in the tree used a replication fixture, which takes the `print_json` path. The NDJSON assertion in `run_watch_reruns_on_file_change_and_exits_clean_on_sigint` therefore passes while the transformation arm is broken.

## Suggested work

Route the transformation JSON output through `print_json`, and add a transformation-pipeline watch test that asserts the same one-object-per-line contract the replication test already asserts.

Found while adding the mid-iteration signal test for #1590, which needed a transformation fixture.

hugocorreia90:
`rocky run --watch --output json` promises one compact JSON object per iteration, one per line. A replication pipeline delivered that. A transformation pipeline pretty-printed instead, spreading a single iteration across roughly 65 lines, so nothing reading the stream line by line could parse it.

```
                     COMPACT_JSON flag ──▶ output::print_json ──▶ one line
  replication   ──▶  print_json                                   1 line
  transformation ─▶  serde_json::to_string_pretty (direct)       65 lines
```

Closes #1604

## Cause

`run_watch.rs` sets the process-global `COMPACT_JSON` flag, and `output::print_json` is the only reader of it. The transformation arm called `serde_json::to_string_pretty` directly at three sites in `run_local.rs`, so the flag never reached it.

All three now go through `print_json`. No other direct pretty-serializer call remains in either run module.

**Nothing outside `--watch` changes.** The flag is set only by the watch loop, so an ordinary `rocky run --output json` is still pretty-printed for humans, which is what that path is for.

## Why it survived

The one existing watch test uses a replication fixture, and replication always took the `print_json` path. The test asserting the NDJSON contract was therefore structurally unable to see this.

The new test drives a transformation pipeline and asserts that **each stdout line parses as JSON on its own** — the assertion a pretty-printed object fails on its very first line, `{`. It reuses the existing transformation fixture with its source left as an ordinary CSV, so the run simply completes, and it needs only one iteration. It runs in about 6 s.

## Evidence

- Fail-before: with the pretty serializer restored, the new test fails on `stdout line 0 is not valid single-line JSON`.
- `cargo +1.98.0 test -p rocky --test run_watch`: passes, 5.8 s.
- `clippy --all-targets -p rocky-cli -p rocky -- -D warnings` and `fmt --all --check`: clean.
- **Mutation check** (`scripts/mutation-check.sh`, clean tree enforced): reverting all three call sites to `to_string_pretty` fails the new test and nothing else. Verdict `OK — the test failed under mutation, so it is fix-sensitive.`


hugocorreia90:
**Independent red team: Codex (GPT-5), `VERDICT: BLOCK`, one high finding. The finding's premise is wrong; its suggestion was right and is applied.**

**The claim:** the test invokes `run --watch` without `--output json`, so it "does not explicitly exercise the reported contract" and "likely fails with empty stdout".

**Why that is not so.** `rocky` resolves its output format by TTY detection when `--output` is unset — `resolve_output` in `engine/rocky/src/main.rs`: an explicit flag wins, an interactive terminal gets `table`, and **anything else gets `json`**. The comment there says the pipe fallback is load-bearing precisely because Dagster, the LSP and CI capture stdout as a pipe. The test pipes stdout, so `output_json` was already true and `run_watch.rs` set `COMPACT_JSON`. The sibling replication test has always relied on the same thing.

The decisive evidence is not that argument, though. It is the mutation check, which ran before this review: reverting all three call sites to `to_string_pretty` **failed** the new test. A test that never selected the contract could not have detected that.

**Applied anyway, because the suggestion is right.** A test *about* the JSON contract should not depend on TTY detection to select the thing under test. Run with stdout attached to a terminal, the old form would quietly stop testing anything. `--output json` is now passed explicitly. Re-verified after the change: the test still passes (5.9 s), and the mutation check still reports `OK — the test failed under mutation, so it is fix-sensitive.`

**Cleared:** no unintended scope — the three sites are compact only during watch, since nothing but the watch loop sets the flag, so a plain `rocky run --output json` stays pretty-printed. No behavioural delta beyond compactness: same stdout, same single trailing newline, errors still propagate through `?`. No other direct serializer call remains reachable. And the fixture reuse is safe: every test builds its own temp directory, database, CSV, config and model, so there is no shared state.

The reviewer named its own limit accurately: `print_json` and the flag setters were outside the diff, which is exactly what left it unable to settle points 1 and 2 independently. Both are quoted above from the code.

Issue #1617, opened 2026-09-03
several commands drop every non-error compiler diagnostic


hugocorreia90:
A compiler diagnostic that is not an error is invisible in some command outputs. The compiler reports it; the command never shows it.

- `rocky publish-ir` checks `has_errors` and writes the snapshot. It renders nothing (`engine/crates/rocky-cli/src/commands/publish_ir.rs`). Running it on `examples/playground/pocs/00-foundations/00-playground-default` exits 0 and prints only `published IR snapshot`, with three `I003` diagnostics hidden.
- `rocky test --output json` has no `diagnostics` field at all. `TestOutput` in `engine/crates/rocky-cli/src/output.rs` never carries them, even though `rocky_engine::test_runner::TestResult` holds them.

This is not specific to `I003`. `W001`, `W004`, `W005`, `P002`, `I001` and `I002` are dropped the same way. `rocky compile` and `rocky ci` do show them.

`rocky test` matters most: it is the command most likely to run in CI, and it is one of the two that compile with empty source schemas, so it is where unchecked-contract reports are most likely to exist.

Adding `diagnostics` to `TestOutput` is a codegen-cascade change (`just codegen` plus regenerated SDK and vscode bindings, then `just regen-fixtures`).

Found while reviewing #1240.

PR #1660, opened 2026-09-04
fix(engine/rocky-ai): report the declared types the contract loop did not check

### LINKED ISSUE #1618: rocky ai-contract accepts a drafted type the compiler could not verify
The AI contract-drafting loop verifies a generated `.contract.toml` by running `validate_contract` and keeping only errors:

`engine/crates/rocky-ai/src/contract.rs`

```rust
let diagnostics = validate_contract(model_name, inferred_schema, &contract);
let errors: Vec<String> = diagnostics.iter().filter(|d| d.is_error()) ...
if errors.is_empty() { Ok(contract) } else { Err(...) }
```

With a cold schema cache every leaf column infers `Unknown`, so the type check does not run and the loop accepts whatever type the model wrote. `rocky ai-contract --save` then writes that contract to disk, and the MCP tool returns it, with nothing saying the declared types were never checked.

#1240 added `I003` for exactly this condition, so the information now exists in the diagnostics list. Carrying it into the `ai-contract` output (and the MCP result) would tell the user which declared types were accepted unverified.

That is a schema change on `AiContractOutput`, so it needs the codegen cascade.

Found while reviewing #1240.

hugocorreia90:
`rocky ai-contract` drafts a `.contract.toml` and verifies it by compiling. The verify step keeps only error-severity diagnostics, so a declared type the compiler could not check passed silently and the draft read as verified.

```
  column type inferred      -> declared type COMPARED -> E011 on mismatch
  column type Unknown       -> I003, no comparison    -> dropped -> looks clean
```

Closes #1618

## Why this is not a corner case

`validate_contract` reports `I003` when a column's own type could not be worked out: the declared type is accepted without being compared to anything. **With a cold schema cache that is every leaf column.** So a contract that drafted "without errors" was not evidence its types were right, and `--save` wrote it to disk with nothing said. The MCP tool returned it the same way.

The information already existed — #1240 added `I003` for exactly this — it was simply filtered out one line after being produced.

## What changed

`verify_contract_toml` returns the contract together with the `I003` messages instead of discarding them, and `DraftedContract` carries them. Three surfaces report them:

| Surface | How it appears |
|---|---|
| `rocky ai-contract --output json` | `unverified_types`, a list of messages |
| MCP `ai_contract` result | same field, same content |
| Human output | a named list above the contract, counted |

The field is **absent when every declared type was checked** (`skip_serializing_if`), so a fully verified draft looks exactly as it did. The human path prints nothing extra in that case either.

Schema change, so `just codegen` regenerated `schemas/ai_contract.schema.json`, the Pydantic model, the TypeScript interface and the OpenAPI document. The Python SDK re-exports the generated `AiContractOutput` rather than keeping a hand-written copy, so there was nothing to edit by hand there.

## Evidence

- **Fail-before:** with a schema where `amount` infers `Unknown` and a contract declaring `type = "Decimal"` for it, the loop returned `Ok` and reported nothing. New test `verify_reports_a_declared_type_it_could_not_check` asserts the message names the column and says the type was not checked.
- The existing happy-path test now also asserts `unverified_types` is **empty** when every type resolved, so the two cases are pinned apart rather than only the new one being covered.
- `cargo +1.98.0 test -p rocky-ai -p rocky-cli -p rocky-mcp --lib`: 0 failed. `clippy --all-targets` on all three with `-D warnings`, and `fmt --all --check`: clean.
- `uv run pytest -q` in `sdk/python` and `integrations/dagster`: both pass, including the schema-parity test.
- **Mutation check** (`scripts/mutation-check.sh`, clean tree enforced): replacing the `I003` filter with an empty vector fails the new test and nothing else. Verdict `OK — the test failed under mutation, so it is fix-sensitive.`

## Not in this PR

#1617 is the wider version of the same complaint: `rocky publish-ir` renders no diagnostics at all and `rocky test --output json` has no `diagnostics` field, so `W001`, `W004`, `W005`, `P002`, `I001` and `I002` are dropped the same way. That one is a larger surface and is tracked separately.


hugocorreia90:
**Independent red team: Codex (GPT-5), two rounds. Round 1 `BLOCK` found a real gap and it is fixed. Round 2's remaining finding is an artefact of how I scoped the diff, not a gap in the branch.**

## Round 1 — accepted, and it was right

It would not accept the `I003`-only selector without evidence that it was complete and exclusive. Fair: I had asserted it. The inventory settled both halves, and the second half was **wrong in my favour**:

- **Exclusive.** `I003` has exactly one emission site in the engine, in `validate_contract`, and it is the unchecked-declared-type case. Every other occurrence is a test assertion or the constant.
- **Not complete.** `validate_contract` emits `E010`, `E011`, `E012`, `E013`, `E014`, `W010` and `I003`. All five `E` codes are error severity and have already returned from the loop. That leaves exactly **two** non-failing diagnostics, and I had handled one. `W010` — the draft declares a column the model does not produce — was being dropped by the same filter. A missing *required* column fails as `E010`, so only the optional arm was silent, which is precisely the case where an LLM invents a column name and is never contradicted.

Now reported as `unmatched_columns`, kept as its own field so each name means what it holds, with its own test.

## Round 2 — verified and rejected

The one remaining finding was that the codegen cascade was missing and `codegen-drift` would fail. It is not missing. **I had excluded generated files from the diff I handed the reviewer** (`':!*generated*' ':!schemas' ':!*openapi*'`) to keep it focused on logic, so it correctly reported what it could not see. My prompt caused that, not the branch. Both models flagged it independently, which is a fair signal my scoping was misleading.

The commit contains all four regenerated artefacts, each carrying the new field:

```
schemas/ai_contract.schema.json                                 1
sdk/python/src/rocky_sdk/types_generated/ai_contract_schema.py  1
editors/vscode/src/types/generated/ai_contract.ts               1
docs/public/openapi.json                                        1
```

No Dagster fixture regeneration is needed: the corpus holds no `ai_contract` fixture, and its 735 tests pass.

## Confirmed in round 2

Both lists separate and accurately named; the shared `collect_code` helper preserves the `I003` path's filtering, conversion and ordering; both fields omitted when empty on **both** structs so no existing consumer sees a wire change; and the `W010` test fails if its collection is removed.

**Gates:** `cargo test`/`clippy --all-targets`/`fmt --check` on `rocky-ai`, `rocky-cli`, `rocky-mcp` all clean; `sdk/python` and `integrations/dagster` pytest green. Mutation check on the `I003` filter: `OK — the test failed under mutation, so it is fix-sensitive.`

hugocorreia90:
**Also fixed here, found while resolving a changelog conflict:** two entries were sitting under the released **1.73.0** heading instead of Unreleased — the BigQuery pagination fix (#1650) and the replication-check fix (#1602). Neither shipped in 1.73.0, so the released section was claiming work it does not contain.

Cause: both were inserted by anchoring on the first `### Fixed` in the file. At the time Unreleased had no such heading, so the first one was 1.73.0's. Moved into Unreleased, which now lists all seven unreleased fixes and leaves the released section untouched.

Worth knowing for anyone else scripting a changelog insert: anchor on the heading *within* `## [Unreleased]`, not the first one in the file.

PR #1661, opened 2026-09-04
fix(engine/rocky-cli): surface the non-error diagnostics test and publish-ir dropped

### LINKED ISSUE #1617: several commands drop every non-error compiler diagnostic

hugocorreia90:
An error already fails `rocky test` and `rocky publish-ir`. What was invisible is everything that does not fail: `W001`, `W004`, `W005`, `P002`, `I001`, `I002`, `I003`.

```
  compile reports ──┬─ error ──────────▶ command fails, user sees it
                    └─ warning/lint/info ▶ DROPPED
```

Closes #1617

## The two commands

| Command | Before | Now |
|---|---|---|
| `rocky test --output json` | no `diagnostics` field at all, though `TestResult` has always carried them | `diagnostics`, omitted when the compile reported nothing |
| `rocky publish-ir` | printed only `published IR snapshot`, whatever the compile said | renders them to **stderr** via the same `render_diagnostics` `rocky compile` uses, so its own stdout stays parseable |

`rocky test` is the one that matters. It is the command most likely to run in CI, and it is one of the two that compile with **empty source schemas** — which is exactly the condition that produces `I003`, "the declared type was never compared". A silent "unchecked" reads as "clean" precisely where nobody is watching.

## The regenerated fixtures are the evidence

This is not a hypothetical. `just regen-fixtures` re-captured the two `test` fixtures from real POC runs, and both grew the same three previously-hidden reports:

```
+  "diagnostics": [
+    { "severity": "Info", "code": "I003",
+      "message": "column 'customer_id' declares type Int64 in the contract, but Rocky
+                  could not work out the column's type, so it did not check the declared type",
+      "model": "revenue_summary" },
```

The playground project's contract has been unverified this whole time and the JSON said nothing. Additions only, 52 lines across the two files; nothing else moved.

## Evidence

- **Fail-before**, measured with the built binary on a scratch project before the change: `rocky test --output json` had no `diagnostics` key.
- New integration test `test_json_carries_the_compiles_non_error_diagnostics`: asserts the `I003` reaches the caller **and** arrives at `Info` severity. It deliberately does *not* assert the command exits 0 — the fixture's model reads a source that does not exist, so the run fails for a reason unrelated to the diagnostic, and asserting success would have been testing the fixture's accident. The first draft did assert it and failed for exactly that reason.
- `cargo +1.98.0 test -p rocky --test test_model_scope`: **13 passed, 0 failed**.
- `uv run pytest -q`: `integrations/dagster` 735 passed against the regenerated fixtures, `sdk/python` 241 passed.
- `just codegen` committed alongside: schema, Pydantic, TypeScript, OpenAPI. The SDK aliases its old hand-written test model to the generated one, so nothing needed editing by hand.


hugocorreia90:
**Independent red team: Codex (GPT-5), `VERDICT: PASS`, no findings. A second model (`gpt-5.6-terra`) also found no blocker.**

Confirmed: `publish-ir` writes only to stderr so its stdout contract is untouched; the new field is omitted when empty so an existing `rocky test --output json` consumer sees no change; no error-severity diagnostic can reach this field in a way that alters exit codes or double-reports; and the test discriminates — removing the wiring makes the field absent and the test fails before its assertions.

Two things I checked rather than took on trust, because the reviewer was held to the diff:

**Wiring completeness.** Its answer covered the two sites the diff touched. `TestOutput` is constructed at **four** sites, and the other two were never shown to it. They are correct, for a reason the diff could not reveal: those are the `--declarative` paths, and that path does not compile the project at all — it loads models through `models_loader::load_project_models`. There is no compile result there, so there are no diagnostics to carry. The two paths that *do* compile are the two that now carry them.

**Runtime mutation evidence**, which it named as its one gap. Now produced (`scripts/mutation-check.sh`, clean tree enforced): removing both `with_diagnostics` calls fails `test_json_carries_the_compiles_non_error_diagnostics` and nothing else. Verdict `OK — the test failed under mutation, so it is fix-sensitive.`

For the record, this review cost one round rather than two because the prompt stated which files were excluded from the diff and that `just codegen` and `just regen-fixtures` had run and were committed. On the previous PR the same omission produced a spurious "the codegen cascade is missing" finding from both models.

Issue #1638, opened 2026-09-03
engine-weekly Coverage job: the linker dies with SIGBUS while tarpaulin builds the test binaries (red since 2026-08-24)


hugocorreia90:
The `Coverage` job in `engine-weekly.yml` has failed on every run since 2026-08-24 (last green: 2026-08-17). It is not a test failure. `cargo tarpaulin` fails to build the test binaries:

```
error: linking with `cc` failed: exit status: 1
  = note: collect2: fatal error: ld terminated with signal 7 [Bus error], core dumped
error: could not compile `rocky-cli` (test "remote_state_bypass") due to 1 previous error
error: could not compile `rocky-cli` (lib test) due to 1 previous error
error: could not compile `rocky-fulfill` (lib test) due to 1 previous error
error: could not compile `rocky` (test "examples_models_load") due to 1 previous error
```

Same signature on the scheduled runs of 2026-08-24 and 2026-08-31 and on a manual dispatch of the branch for #1634 on 2026-09-03 (run `33811298534`). The `Security Audit` and `POC Smoke Tests` jobs in the same workflow are green on that dispatch, so the runner and the code build fine outside tarpaulin.

## What changed between the last green run and the first red one

- The Rust toolchain pin moved to 1.98.0 on 2026-08-20 (the weekly job pins it like the other engine jobs).
- The link line uses `-fuse-ld=lld` with `-pie -z relro -z now`, and the `rocky-cli` test binaries link every adapter crate, DuckDB, aws-lc, ring and the OpenTelemetry stack. A `SIGBUS` from `ld` on a runner with a 4 GB swap file added by the job itself points at memory or `/tmp` exhaustion during link, not at a code defect.

## What would settle it

1. Re-run the job with `CARGO_BUILD_JOBS=1` for the tarpaulin step, or split it per crate, and see whether the bus error goes away.
2. If it does not, try `-C link-arg=-fuse-ld=bfd` for the coverage build only.
3. Check `df -h /tmp` and `free -m` right before the step.

Until then the weekly coverage number is missing; nothing else in the release gate depends on it.

Raised during the `engine-v1.73.0` pre-flight.

hugocorreia90:
**Cause established, with numbers. Three fixes tried, two disproved, and the last one is a decision rather than a tweak — so I am stopping here and handing it over.**

## What it is

[...]

PR #1663, opened 2026-09-04
chore(ci): record the machine state around the coverage job, and free more disk

hugocorreia90:
**This does not fix the Coverage job.** It establishes why the job fails, with numbers, and keeps the two things that proved useful. The job stays red until a decision is made — see the options in #1638.

Refs #1638

## What the diagnostics found

A linker writes its output through `mmap`, so a filesystem that fills mid-write faults instead of returning ENOSPC. That is why every failure since 2026-08-24 carries `signal 7 [Bus error]` and never `No space left`.

With `df` around the step ([run 33911564622](https://github.com/rocky-data/rocky/actions/runs/33911564622)):

```
before the step    /dev/root  145G   40G  106G   28% /
after the failure  /dev/root  145G  145G  149M  100% /
```

The tarpaulin-instrumented build consumes **~105 GB**. Corroborated on [run 33906729207](https://github.com/rocky-data/rocky/actions/runs/33906729207), where the runner *agent* itself crashed with `System.IO.IOException: No space left on device` writing its own diagnostic log.

## What is in the diff

- **`df -h` and `free -m` bracketing the coverage step**, permanently. The second carries a comment naming its own limit: when the disk fills hard enough the runner agent dies, so `if: always()` does not run and no logs upload at all. A Coverage job whose `Run coverage` step has no conclusion and an empty log has **lost its runner** — the disk causes that here, but so do cancellation and ordinary infrastructure loss, so the job's annotations carry the agent's own error and are what to read first. That is how the cause was identified.
- **More aggressive cleanup**: boost, swift, the bundled node modules, the CodeQL toolcache, cached Docker images. Worth ~25 GB. Not sufficient on its own, kept because it is real headroom.

## What I tried and removed

- **`--jobs 1`**, on the theory that four concurrent links of a several-hundred-rlib binary exhausted memory. Failed identically. Removed — it only made a slow job slower.
- **`debug = line-tables-only`** for the dev and test profiles. Failed the same way. Removed — a change that does not work, and that narrows what coverage can map, is worse than not having tried it.

Both theories were mine and both were wrong. The 105 GB figure is the only claim here I would rely on.

## Why it is not fixed here

The build outgrew the machine, so freeing space cannot close it. The remaining options each change what the job is — a larger runner, per-crate coverage, or swapping tarpaulin for `cargo-llvm-cov` — and that is a decision about what this repository measures, not a fix. Laid out with a recommendation in #1638.

`actionlint`: clean.

### DIFF
--- a/.github/workflows/engine-weekly.yml
+++ b/.github/workflows/engine-weekly.yml
@@ -103,21 +103,31 @@ jobs:
           # while competing for the repo's 10 GB budget against caches that
           # every PR restores from. Latency does not matter on a schedule.
           save-if: false
+      # The runner filled its root filesystem here, which is what has been
+      # failing this job since 2026-08-24. It surfaced as
+      # `collect2: fatal error: ld terminated with signal 7 [Bus error]`
+      # because a linker writes its output through mmap: a filesystem that
+      # fills mid-write faults instead of returning ENOSPC. The proof is the
+      # runner agent's own crash on the next run —
+      # `System.IO.IOException: No space left on device` while writing
+      # `_diag/Worker_*.log` — which is also why that run uploaded no logs at
+      # all. Removing dotnet, android and ghc was not enough for a
+      # tarpaulin-instrumented debug build of this workspace.
       - name: Free disk space and add swap
         run: |
-          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc
+          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc \
+                      /usr/local/share/boost /usr/share/swift \
+                      /usr/local/lib/node_modules /opt/hostedtoolcache/CodeQL
+          sudo docker image prune --all --force || true
           sudo fallocate -l 4G /mnt/swapfile && sudo chmod 600 /mnt/swapfile && sudo mkswap /mnt/swapfile && sudo swapon /mnt/swapfile || true
       - run: sudo apt-get update && sudo apt-get install -y cmake
       - name: Install cargo-tarpaulin
         # Pinned + --locked for the same reason as cargo-audit above.
         run: cargo install cargo-tarpaulin --version 0.31.2 --locked
-      # Kept permanently, not just while #1638 is open. The failure this job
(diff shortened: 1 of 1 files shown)


hugocorreia90:
**Independent red team: Codex (GPT-5), `VERDICT: BLOCK`, three findings. All three were my comments being wrong. All three applied.**

It cleared the mechanics — the cleanup is safe (JS actions use the runner's bundled Node, not `/usr/local/lib/node_modules`, corroborated by the checkout post-action succeeding), it is confined to this job's runner, the readings are placed correctly, and nothing masks the failure: no `continue-on-error`, coverage still decides the outcome.

What it blocked on was prose:

1. **A comment pointing at a setting I had removed.** The cleanup comment said "see the debug-info setting on the coverage step" and claimed the two together keep the job under the disk. I reverted that setting one commit earlier for not working, so the comment referenced nothing *and* asserted a fix that does not exist. Now states the truth: ~25 GB reclaimed, ~105 GB needed, job stays red.
2. **An overclaim about the diagnostic signature.** I wrote that an empty log with no step conclusion **is** the disk case. It is not — that shape means the runner was lost, and cancellation or infrastructure loss produce it too. Corrected to say the annotations carry the agent's own error and are what to read first, which is in fact how I identified the cause.
3. **"Six weeks of failures."** Arithmetic I carried forward without checking. The failures began 2026-08-24; that is 11 days. Fixed here, in the description, and in the write-up on #1638.

Its second reviewer (`gpt-5.6-terra`) reached BLOCK independently on the same three.

Worth noting the shape: the code was fine and every defect was in what I said about it — including one claim invalidated by my own revert two commits earlier.

hugocorreia90:
Correction to the description and to my earlier comment: "`debug = line-tables-only` failed the same way" overstates it. **It was never applied.** Tarpaulin 0.31.2 forces `-Cdebuginfo=2` via `RUSTFLAGS` unconditionally and rustc takes the last `-C debuginfo`, so that run built with full debuginfo like every other. Reverting it was still right — it did nothing — but the reason is "inert", not "disproved". Detail and the corrected next step on #1638.

PR #1664, opened 2026-09-04
OWNER MERGE — fix(dagster): declare the freshness check only when the pipeline configures freshness

hugocorreia90:
**A project whose `rocky discover` state reports no `[checks.freshness]` no longer shows a `freshness` check on its assets. It used to show one, and it was green.**

The engine emits a `freshness` result only when the pipeline declares `[checks.freshness]`. `RockyComponent` declared the spec on every asset anyway, so the placeholder pass reported the check as `passed=True` on every materialized table. The status text said "not produced by rocky". The badge said pass.

## Before / after

```
rocky.toml has no [checks.freshness]
  ├─ engine  : emits no `freshness` CheckResult   (run.rs gates the whole
  │                                                block on
  │                                                pipeline.checks.freshness)
  ├─ before  : spec declared → placeholder → passed=True, severity=ERROR
  │            → GREEN badge for a check that never ran
  └─ after   : spec not declared → no `freshness` row on the asset

rocky.toml has [checks.freshness]
  └─ before == after : spec declared, engine result drives the verdict
```

## What changed

- `freshness.py` gains `freshness_is_configured(checks)`. It is the one predicate for "the engine will emit a `freshness` result".
- `_build_check_specs` gains a **required** keyword-only `declare_freshness`. No default: a default of `True` is the fail-open shape that lets a future call site re-introduce the bug.
- `RockyComponent.build_defs` passes `freshness_is_configured(discover.checks)` — the same predicate that already decides the **pipeline-level** `FreshnessPolicy`, so the badge and that policy cannot disagree. `test_freshness.py` pins that equivalence across all three `checks` shapes.

A per-model `[freshness]` frontmatter is separate: `_build_group_contexts` does `model_policies.get(table.name, default_policy)`, so that one asset still gets a `FreshnessPolicy` with no `[checks.freshness]`. It never produced a `freshness` CheckResult either — per-model `max_lag_seconds` is read by `rocky tick` and `rocky validate`, not by the check runner — so removing the spec loses nothing there.

`discover.checks` is `None` both when the pipeline declares no `[checks]` block (the engine's `ChecksConfigOutput::from_engine` returns `None` when there is neither a freshness config nor a configured check) and when the `rocky` binary predates the projection. Both read as "no freshness".

## Why this is owner-merge

A declared check disappears from every asset on a project that never configured freshness. Anything selecting `freshness` by name — an asset-check selection, a job, an alert, a sensor — stops matching. Smaller blast radius than flipping a verdict red, but not zero.

## Scope: `row_count` and `column_match` are NOT fixed

Both can be switched off in `rocky.toml` and both have the same false green. Dagster cannot fix them today. `ChecksConfigOutput` (`schemas/discover.schema.json`) has exactly two fields:

```
configured_checks   per-model NON-default check names (the four defaults are excluded by design)
freshness           the FreshnessConfig, or null
```

There is no `enabled`, no `row_count` and no `column_match` toggle. Fixing those needs `engine/crates/rocky-cli/src/output.rs` to add them to `ChecksConfigOutput`, the full `just codegen` cascade, then the same gating here. `configured_checks_by_model` cannot stand in: it is populated only when `surface_configured_checks` is on, and it lists only non-default names.

`row_count_anomaly` and `compliance_exception` are pass-by-absence by design — no event genuinely means no anomaly. Unchanged.

## Surfaces checked

`_build_check_specs` is the only place that declares a `freshness` check.

- `check_specs` reaches exactly one `@dg.multi_asset`, in `_make_rocky_asset`.
- `_make_derived_model_asset` passes no `check_specs` at all.
- `assets.py::load_rocky_assets` returns only `AssetSpec` — it already gates the `FreshnessPolicy` on the same config and declares no checks.
- `dag_assets.py` constructs no `AssetCheckSpec`.

## Caveat: the component reads CACHED discover state

Adding `[checks.freshness]` to `rocky.toml` without refreshing the state leaves the spec undeclared while the engine emits the result. **Refresh the state after changing `[checks]`.** What happens next depends on the execution mode:

| mode | undeclared check result |
|---|---|
| `streaming` (default) | `_emit_results` drops it (`if spec_key not in declared_checks: continue`). The verdict disappears. |
| `pipes` | `_run_filters_pipes` yields `invocation.get_results()` with no filtering, so Dagster raises `DagsterInvariantViolationError` and the step FAILS. |

I probed the pipes half rather than assuming it: yielding an `AssetCheckResult` for a check absent from `check_specs` gives `Received unexpected AssetCheckResult. No checks currently being evaluated target asset 't' and have name 'freshness'.`

**Staleness is symmetric, and the other half is not fixable here.** The state key is `RockyComponent[<config_path>]` — the path, not the file's contents — so editing `rocky.toml` refreshes nothing:

```
freshness ADDED, no refresh     spec undeclared, engine emits
                                -> streaming drops it / Pipes fails the step
freshness REMOVED, no refresh   spec declared, engine emits nothing
                                -> placeholder reports passed=True
```

The second row is #1645 itself surviving through the cache. Gating on the discover projection cannot close it; only refreshing the state can. Called out in the CHANGELOG and the docstring rather than papered over.

That pipes gap is **pre-existing and wider than freshness** — `_run_filters_pipes` filters by asset key (`include_keys`), never by check name, so any check the component did not declare has the same effect there. `integrations/dagster/AGENTS.md` records the streaming half of this ("may emit check kinds (e.g. `null_rate`) the component never declared. `_emit_results` drops both"). Worth its own issue; not touched here.

`freshness_policy_from_checks` already behaves the same way against a `rocky` binary that predates the `checks` projection, so this change does not introduce that class.

## Evidence

Gates from `integrations/dagster/`: `uv run pytest -q` → 745 passed. `uv run ruff check .` → clean. `uv run ruff format --check .` → clean.

Mutation check (`scripts/mutation-check.sh`), mutating the gate away while keeping the parameter:

```
FAILED tests/test_freshness.py::test_freshness_check_spec_absent_when_there_is_no_checks_block
FAILED tests/test_freshness.py::test_freshness_check_spec_absent_when_checks_block_has_no_freshness
FAILED tests/test_freshness.py::test_materialized_table_reports_no_freshness_verdict_without_a_freshness_config
```

The mutated run's own log is the bug:

```
ASSET_CHECK_EVALUATION - Asset check 'freshness' on 'fivetran/acme/shopify/orders' passed.
```

One existing test moved: `test_pruned_placeholder_carries_prior_failure_end_to_end` used `freshness` as its "sibling check that carries forward green". Its discover payload declares no freshness, so it now asserts on `column_match` instead. Same mechanism, same proof.

Refs #1645


hugocorreia90:
**Red team: Codex (`codex exec`, GPT-5 class, read-only). Six rounds. Every round found a claim of mine that was stronger than the code. All accepted, all fixed — and no executable line changed after the first commit (`git diff 04df4a6b..HEAD -- integrations/dagster/src` is comments and docstrings only).**

| Round | Finding | Disposition |
|---|---|---|
| 1 | The helper was described as authoritative for engine emission. Stale state and an old binary are counterexamples. | Accepted. Reworded to say it reads a **projection**, and to name both counterexamples and what each does (streaming drops the result; Pipes fails the step). `75856f9a` |
| 2 | Staleness is **symmetric**. I documented only "freshness added, no refresh". "Freshness removed, no refresh" leaves the spec declared and the false green intact — and gating on the projection cannot close it. | Accepted. Verified myself first: the state key is `RockyComponent[<config_path>]`, the path and not the contents, so editing `rocky.toml` refreshes nothing. Both directions now documented, and the CHANGELOG headline is qualified to "a project whose `rocky discover` **state** reports no `[checks.freshness]`". `53aa2dc6` |
| 3 | "wrong whenever the projection and the live `rocky.toml` disagree" over-reaches — unrelated `[checks]` edits disagree without changing the answer. | Accepted. Narrowed to disagreement **about freshness**, with the counterexample named. `079989b6` |
| 4 | Still loose: a changed `threshold_seconds` IS a freshness disagreement that preserves the boolean. | Accepted. The body is `checks is not None and checks.freshness is not None`, so the sentence now says **whether freshness is configured — the boolean, not the settings**. `018b1e8e` |
| 5 | "or any other `[checks]` edit" contradicts the bullet below it — removing `[checks.freshness]` IS a `[checks]` edit and does change the answer. | Accepted, using the reviewer's own wording: "an edit that preserves whether `[checks.freshness]` is present". `952d2b34` |
| 6 | none — `LANE-S-R6 VERDICT: PASS` | Clean. |

**Found by me, before the review, in the same class:** I had written that the new gate is "the same condition that already decides whether the asset gets a `FreshnessPolicy`". `_build_group_contexts` does `model_policies.get(table.name, default_policy)`, so a per-model `[freshness]` frontmatter still attaches a policy with no `[checks.freshness]`. Corrected to **pipeline-level** in `245ae958`.

**Verified by probe, not assumption:** an `AssetCheckResult` for a check absent from `check_specs` raises `DagsterInvariantViolationError` and fails the step. That is what makes the Pipes half of the stale-state caveat a hard failure rather than a dropped verdict. `_run_filters_pipes` filters by asset key, never by check name — pre-existing and wider than freshness; worth its own issue, not touched here.

**Not reviewed by a second model.** Codex is one reviewer, not two. Round 1 also stalled in a reconnect loop after producing its finding and was killed; rounds 2-6 used a much smaller prompt.

The code is unchanged since `04df4a6b`. Mutation check, gates and the surface sweep in the PR description all still describe it.

hugocorreia90:
**Orchestrator review — verified against the diff, the log and the CI run. Ready for owner merge; I am not merging it because a declared check disappears from assets, which is the owner-merge class.**

What I checked myself, not from the report:

- `declare_freshness` is a **required** keyword-only parameter with no default — a `True` default would be the fail-open shape that lets a future call site re-introduce the bug. Every production and test call site passes it; a missed one would not import, and CI is green.
- The predicate is `checks is not None and checks.freshness is not None`, shared with the pipeline-level `FreshnessPolicy` derivation and pinned equivalent by `test_freshness.py`, so the two cannot drift.
- The gate is one `continue` on `FRESHNESS_CHECK_NAME`; the other three defaults are untouched.
- The changelog says what changes on upgrade in every direction, including the one the fix cannot close: freshness *removed* without a state refresh keeps the spec declared and the false green reachable. That caveat is correct — the state key is the config path, not its contents.
- Mutation check re-verified on the final head; 745 tests, ruff check and format clean.

Two follow-ups this review produced, so the tracker is honest: #1645 now carries what it still owns (`row_count`/`column_match` need the engine to expose toggles; the stale-state half; and the `passed=materialized` line in `_emit_placeholder_checks` that #1619, #1644 and #1645 are all symptoms of). And the Pipes-path gap this PR's author probed but did not file is now #1673.

One observation on process, not a defect: the six Codex rounds were all on prose, none on code — every round found a sentence stronger than the implementation, which is this repository's most common finding. The final text is accurate.

Issue #1521, opened 2026-08-23
bug(engine/rocky-cli): compile silently ignores malformed rocky.toml


mattfaltyn:
## Description

`rocky compile` silently treats a present but malformed `rocky.toml` as if no
configuration were supplied. The command exits successfully with
`has_errors: false` and empty diagnostics, while `rocky validate` correctly
reports the same file as `V001` and exits nonzero.

This disables every config-driven compile input for that invocation, including
the schema cache, masking/classification and freshness checks, portability
linting, and imported-contract diagnostics.

## Steps to reproduce

Create a valid one-model project:

```text
models/ok.sql
models/ok.toml
rocky.toml
```

`models/ok.sql`:

```sql
SELECT 1 AS id
```

`models/ok.toml`:

```toml
name = "ok"

[target]
catalog = "memory"
schema = "main"
table = "ok"
```

Use a malformed `rocky.toml`:

```toml
[portability
target_dialect = "bigquery"
```

Run:

```bash
rocky \
  --config ./rocky.toml \
  --output json \
  compile \
  --models ./models
```

Reproduced twice from a clean checkout of canonical `main` at
`98e7d385e5de5c217b077f346b6dee7069ed1b7d`.

## Expected behavior

A supplied config that exists but cannot be loaded should fail `rocky compile`
with the config error. The existing standalone behavior for a genuinely
missing config should remain unchanged.

## Actual behavior

Exit status is `0`; JSON reports:

```json
{
  "command": "compile",
  "models": 1,
  "diagnostics": [],
  "has_errors": false
}
```

As a control, `rocky --config ./rocky.toml --output json validate` exits `1`
and reports `V001` with the TOML parse error.

## Root cause

`engine/crates/rocky-cli/src/commands/compile.rs::compile_inner` reloads the
config at several feature seams and converts every `load_rocky_config()` error
to empty defaults using `Err(_)`, `.ok()`, or `if let Ok(...)`.

The fallback was introduced to preserve config-free standalone compilation,
but it does not distinguish a missing file from a malformed or unreadable file.
It can also weaken contract checking because an unavailable schema cache
degrades source types to `RockyType::Unknown`, which intentionally matches any
contract type.

## Acceptance criteria

- Load the optional config once per compile invocation.
- Preserve the current fallback when the config path does not exist.
- Propagate parse, validation, permission, and other load failures.
- Reuse the same parsed config for schema cache, masks/classifications,
  freshness, portability, and imports.
[...]

PR #1667, opened 2026-09-05
OWNER MERGE — fix(engine): refuse an invalid rocky.toml on every CLI and MCP entry point instead of treating it as absent

hugocorreia90:
Nine entry points each decided what a broken `rocky.toml` meant, and each read it as **no config at all**. They now refuse, carrying the loader's own error.

The sharpest case, reproduced in a test: `rocky mcp`'s `plan_preview` picks the SQL dialect from the configured adapter. A **Databricks** project whose config failed to load was previewed on the **DuckDB** dialect and returned another warehouse's SQL, with nothing said.

This refuses configs that "work" today only by silently compiling for the wrong warehouse, or against defaults.

## The shape

```
  rocky.toml is...      before                          after
  ────────────────      ──────────────────────────      ──────────────────
  absent                compile with defaults           unchanged
  loads                 use it                          unchanged
  unset ${VAR} in
    adapter creds       strict load failed → DuckDB     use it (#1536 rule)
  present, broken       swallowed → defaults/DuckDB     REFUSE (loader error)
```

One shared loader answers this for all of them:

```rust
rocky_core::config::load_optional_project_config(Option<&Path>)
    -> Result<Option<RockyConfig>, ConfigError>
```

Only `FileNotFound` becomes `None`. Everything else propagates.

## Entry points

| Site | Before | After |
|---|---|---|
| MCP `plan_preview` | `.ok()` → DuckDB dialect → emits SQL | refuses |
| MCP `lineage`, `inspect_schema`, `dependents`, `ai_contract`, `ai_test`, `explain_model` (via `compile_full` → `load_source_schemas`) | cold schema cache, every source column `Unknown` | refuses |
| MCP `breaking_change` | cold cache; zeroed counts + `has_breaking: false` | refuses |
| MCP `catalog` (via `compute_catalog_output`) | cold cache + default project name | refuses |
| `rocky lineage` | `Err(_)` → cold cache | refuses |
| `rocky catalog` | `.ok()` → cold cache + default project name | refuses |
| `rocky ci --diff`, `rocky lineage-diff` (via `compute_ci_diff`) | `Err(_)` → cold cache | refuses |
| `rocky audit` | `Err(_)` → cold cache | refuses |
| `rocky ai` | `Err(_)` → cold cache | refuses |
| `rocky ai-contract` | `Err(_)` → cold cache | refuses |
| `rocky plan --blueprint` (shares `plan_preview_output`) | `.ok()` → DuckDB dialect | refuses |
| `rocky compile` | already refused (#1521) | unchanged — **moved onto** the shared loader |
| `rocky emit-sql` | already refused (#1624) | unchanged — **moved onto** the shared loader |

The last two rows are the reason this holds. Both already had the rule written inline. Collapsing them onto the shared function is what makes "a new entry point inherits the answer" true rather than asserted.

Six of the CLI rows go through one new helper, `load_project_source_schemas`, so the schema-cache bootstrap has a single config leg instead of six copies of `Err(_) => HashMap::new()`.

## A correction to the issue

The issue's repro says a **Snowflake** project got DuckDB SQL. I could not reproduce a *visible* difference for Snowflake: on every model shape that renders offline, the Snowflake and DuckDB dialects emit byte-identical SQL (probed on `full_refresh`, `merge`, `materialized_view`). The wrong dialect was still being used — it just happened not to show.

**Databricks does show it**, and that is what the test asserts:

```
merge, Databricks:  ... WHEN MATCHED THEN UPDATE SET t.id = s.id
merge, DuckDB:      ... WHEN MATCHED THEN UPDATE SET id = s.id
materialized_view:  Databricks renders a statement; DuckDB and Snowflake render none
```

## Fail-before

`plan_preview_output`, against unmodified production code:

```
test plan_preview_invalid_config_refuses_instead_of_defaulting_to_duckdb ... FAILED
  a present but invalid rocky.toml must refuse the preview:
  PlanOutput { statements: [PlannedStatement {
      sql: "MERGE INTO c.s.m AS t ... WHEN MATCHED THEN UPDATE SET id = s.id" }] }
                                                              ^^^^^^^^^^^^^
                                          the DuckDB form, for a Databricks project

test plan_preview_tolerates_unset_credential_vars ... FAILED
  the project's own dialect must still be used: ... UPDATE SET id = s.id
```

The second failure is a **second route to the same defect**, not a separate issue: an unset `${DATABRICKS_HOST}` failed the strict load, the error was swallowed, and the project got DuckDB SQL. The credential-tolerant loader closes it.

## Mutation checks

One mutation — revert `load_optional_project_config` to the pre-fix `Ok(loader(path).ok())` — run against each new test group via `scripts/mutation-check.sh` (refuses a dirty tree; run on the committed state):

| Test group | Verdict |
|---|---|
| `rocky-core` `config::tests::optional_project_config` | fix-sensitive |
| `rocky-cli` `source_schemas::tests::project_source_schemas` | fix-sensitive |
| `rocky-cli` `commands::plan::tests::plan_preview` | fix-sensitive |
| `rocky-mcp` `tools::tests::mcp_` | fix-sensitive |

## Gates

| Gate | Result |
|---|---|
| `cargo test -p rocky-core -p rocky-cli -p rocky-mcp` | **3999 passed, 0 failed** |
| `cargo fmt --all --check` | pass |
| `just codegen` | **no-op** — see below |

`just codegen` ran because `rocky-core/src/config.rs` was touched. It produced no change to `schemas/`, the Pydantic models, the TypeScript bindings or `openapi.json` — a new `pub fn` cannot alter `RockyConfig`'s JSON schema. **There are no generated files to commit.** The only churn was `editors/vscode/package-lock.json`, reverted.

## Upgrade impact

A project whose `rocky.toml` does not parse or does not validate now gets an error from these commands instead of a plausible-looking answer. That config never worked — it only appeared to. The failure is new and can surprise someone.

Two things deliberately do not change:

- A project with **no** `rocky.toml` behaves exactly as before. A standalone `models/` directory still compiles. Tested on every changed path.
- An unset `${VAR}` in an adapter's connection field is still tolerated on these offline paths (#1536). Compiling opens no connection, so it must not require warehouse credentials.

## What remains

**Out of scope by decision** — `rocky lsp` and `POST /api/v1/compile` on `rocky serve` still degrade. Whether an editor should refuse a broken config or surface it as a diagnostic is a product decision, not an accident to patch. That is the issue's own deciding question and it is unanswered.

**Left deliberately** — the `rocky review` and `rocky branch promote` breaking-change gates and `rocky plan --semantic` still degrade. Each returns `Option` and already reasons in code about *skipping* the gate rather than blocking on a configuration issue. Turning a skip into a refusal changes what those gates mean, so it wants its own decision.

**Left on purpose, and it should stay that way** — the governance gate's own loader, `apply.rs::load_config_for_gate`, keeps the **strict** loader and is not folded into the shared one. A policy gate must require every variable to resolve.

**A known limit this PR makes heavier, and does not fix.** `FileNotFound` is produced in exactly one place — `read_config_file` maps `std::io::ErrorKind::NotFound` — and that kind also covers a `rocky.toml` that is a **dangling symlink**, or one deleted between the check and the read. Such a config reads as *absent*, so these commands compile against defaults: the very behaviour this PR closes. It is pre-existing, and six call sites already treat `FileNotFound` as absent (`compile`, `emit-sql`, `policy`, `gc`, `apply`, and the governance gate). But this PR makes `FileNotFound` the single load-bearing discriminator across nine more entry points, so the conflation now carries more weight. The fix belongs in `read_config_file` (distinguish "the path does not exist" from "the path exists and its target does not", via `symlink_metadata`), where it would fix every consumer at once — including the governance gate, in the fail-closed direction. Filed separately rather than widened into this diff.

**Not a defect** — `tools.rs::draft_marker_freezes` returns an empty freeze list on an unloadable config. I checked the consumer: `evaluate_apply_policy*` loads the config itself and returns `PolicyGate::Unloadable`, which every draft gate refuses. Defended independently of the marker list.

`Refs #1625` rather than `Closes` — the LSP and `serve` paths are still open.

## Standing questions

**Least confident:** whether refusing is right for `rocky ai` and `rocky ai-contract`. Those commands degrade to `Unknown` types on a cold cache anyway, so an unloadable config costs them less than it costs `plan_preview`. Refusing is consistent, but "consistent" is my reason, not a measured one.

**Most important thing possibly missing:** how many real projects have a `rocky.toml` that fails to load and have not noticed, because the commands they use are the ones that swallowed it. `rocky run` has always refused, so most projects cannot be in that state — but a project that only ever runs `rocky lineage` or drives the MCP tools could be, and for them this lands as a new hard failure.

Refs #1625


hugocorreia90:
## Red-team disposition — Codex (`gpt-5.4-codex`, read-only, high reasoning)

Two rounds. Round 1 was **blocked**, not by the code: the reviewer's tool display elided the middle of the 1083-line diff, and it correctly refused to judge from a partial capture. Round 2 supplied the same diff in four explicit `sed` ranges. It read all four, then stalled in a wait loop while trying to spawn its own sub-reviewer (`error=timeout_ms must be at least 10000`) and never printed the formatted verdict line. I stopped it there rather than run a third round.

What it did produce, before stalling:

> The primary pass finds the intended fail-closed boundary intact: config errors are propagated, while cache-read failures remain deliberately cold-map fallbacks.

**Finding 1 — dangling `rocky.toml` symlink reads as absent. Verified, accepted, not fixed here.**

> a dangling `rocky.toml` symlink can surface `ConfigError::FileNotFound`, yet the shared helper classifies the variant alone as "absent." That would restore the default-dialect behavior this change is meant to eliminate.

Correct. `read_config_file` maps `std::io::ErrorKind::NotFound`, which also covers a symlink whose target is gone. Pre-existing — six call sites already read `FileNotFound` as absent — but this PR makes it the single load-bearing discriminator across nine more entry points, so it now carries more weight. Filed as #1668 with the `symlink_metadata` fix, which belongs in `read_config_file` so it fixes every consumer at once, including the governance gate. Added to "What remains" in the description. Not widened into this diff.

### Questions I answered from primary source, since no verdict arrived

- **Can `FileNotFound` come from anything but the top-level `rocky.toml`?** No. It is constructed in exactly one place, `config.rs:6199` inside `read_config_file`, whose three callers all pass the top-level config path. An `[imports]` snapshot that cannot be read is a W012 warning at compile time, not a `ConfigError`. So the symlink case above is the *only* way `FileNotFound` means anything other than "no config here".
- **Does any error wrapping break a downcast?** No. `plan_preview_output`'s `ModelNotFound` downcast contract is a different error path and is still asserted by its existing test, which passes.
- **Is `draft_marker_freezes` a fail-open?** No. Its consumer `evaluate_apply_policy*` loads the config itself via `load_config_for_gate` and returns `PolicyGate::Unloadable`, which every draft gate refuses — independent of the marker list.

### Honesty note

Per `AGENTS.md`: **no independent red team completed a full pass on this change.** One reviewer read the entire diff and returned one finding, which is dispositioned above; it did not deliver the structured verdict. The three questions it was heading toward are answered above from the code, by me — the authoring model — which is not a substitute for independent review. Worth a second reviewer before merge.

hugocorreia90:
**Merge note.** This PR and #1675 conflict on one file only: `engine/CHANGELOG.md`. Both add a bullet as the first entry under `## [Unreleased]` → `### Fixed`, so whichever merges second needs a trivial resolution that keeps **both** bullets. Nothing else overlaps — verified with `git merge-tree --write-tree`, which reports `engine/CHANGELOG.md` as the sole conflicted path.

hugocorreia90:
Orchestrator review + the independent round #1667 did not get before.

**Codex (independent model), one full round on head `a9a96413` — BLOCK, 6 findings.** What it confirmed about THIS diff: all nine converted sites use the credential-tolerant loader; it found **no valid configuration that worked before and is newly refused** (only parses, validator errors and non-adapter placeholders, whose earlier "success" was a discarded error); no path widens `NotFound` handling; the two gate helpers the lane left alone (`branch promote`, `plan --semantic`) are protected because their public callers load strictly first (`plan.rs:2189`, `plan.rs:101`).

**Disposition:** findings 1–5 are sibling sites of the same class the PR did not touch — `review.rs:480` (approval marker with zero breaking on an unloadable config; verified), `backfill.rs:217` (`.ok()`; verified), `preview.rs:1167`, the `[ai]` secondary reads, and a false fail-closed comment at `plan.rs:1185`. Pre-existing, not regressions; filed as #1680 with the review site first. Finding 6 (six of nine callers lack a caller-coupled fail-before test) is real and carried into the same issue.

Merging as the strict improvement it is. The owner-merge title stands only for the exit-code change on configs that never loaded.

Issue #1670, opened 2026-09-05
A resumed run re-runs checks only for the tables it copies, so an earlier check failure is never re-evaluated


hugocorreia90:
`rocky run --resume` builds its check inputs only from the tables the resuming invocation copies. A check that failed on a table an earlier run already copied is never re-evaluated, so the resumed run can exit 0 while that table is still violating.

## The shape

```
  run 1   5 tables, table 3 fails to copy
          table 1's row_count check FAILS
          -> exit 2, record PartialFailure

  run 2   --resume  ->  skips tables 1,2,4,5 (already Success)
                        copies table 3
                        checks run for table 3 ONLY
          -> exit 0, and table 1 is still violating
```

The per-table check inputs (`source_batch_refs`, `target_batch_refs`, `freshness_batch_refs`, `assertion_targets`) are populated inside the per-table copy path in `engine/crates/rocky-cli/src/commands/run.rs`, so a table the resume filter removed contributes nothing to the check phase.

## Pre-existing, and deliberately left alone by #1598

This is not new. Checks have always been scoped to the current invocation's copies. #1598 wires `fail_on_error` into the replication exit path and closes the worst case — a resume of a run that copied **everything** would run no check at all, so `ensure_run_is_resumable` refuses it. What is left is the partial case above, where the resume does real copy work and does run checks, just not for the tables it skipped.

## Why it is worth a decision

The narrower the remaining copy work, the narrower the check coverage, and the resumed run's exit code reads as a clean recovery either way. An operator resuming a large replication cannot tell "the checks passed" from "the checks that ran passed".

## Directions

- Run the checks against every table in the pipeline's resolved set, not only the ones this invocation copied. A check reads the target, which exists for a skipped table, so this is a scope question rather than a dependency one. Costs check queries for tables that did not move.
- Or carry the previous run's failed check names forward and re-evaluate only those.
- Or leave the scope alone and refuse to let a resumed run report success while the run it resumes has an unresolved failed check on record. `RunRecord::check_outcomes` already carries the names, though not their severity.

Related: #1598, #1602, #1548.

hugocorreia90:
## Decision: a resumed run may not report success over an unresolved failed check

The check scope stays as it is. What changes is the exit code.

```
run 1   copies 5 tables, table 3 fails to copy
        table 1's row_count check FAILS
        ──▶ exit 2, PartialFailure recorded

run 2   --resume, copies table 3, checks table 3
        ──▶ exit 0    ✗ today
[...]

Issue #1669, opened 2026-09-05
A failed cross-source overlap check is invisible in Dagster


hugocorreia90:
A `cross_source_overlap` check that fails is recorded by the engine and then dropped by the Dagster component. Nothing in the Dagster UI shows it.

## The shape

Two rules meet and cancel the result out.

```
  component.py  declared specs   -> `candidate` names EXCLUDED
  component.py  emit results     -> a result with no declared spec is SKIPPED
                                    -> the failed overlap check reaches nobody
```

The exclusion is deliberate and its reason is sound: a cross-source overlap check is a **group** check the engine emits once, on one sibling asset, so pre-declaring it on every same-named sibling would leave the others showing a passing placeholder that reflects no evaluation. The emit side then drops any result whose `(asset_key, name)` pair was never declared, because Dagster rejects a check outside the declared `check_specs`.

Both live in `integrations/dagster/src/dagster_rocky/component.py`: the spec side filters `if not c.candidate`, and the emit side does `if spec_key not in declared_checks: continue`.

## Why it matters now

Raised on #1598 and unresolved by it. #1598 makes an error-severity check failure fail the run, so the failure does reach Dagster as a non-zero exit — `RockyResource` maps exit 2 to a partial failure. What stays missing is the *named* check: an operator sees the run degrade without seeing which check caused it, and an asset-check-based alert on the overlap check never fires.

## Directions

- Declare the group check on exactly one asset — the same one the engine emits it on — so the pair matches at emit time. Needs a stable rule for choosing that asset, shared by both sides.
- Or declare it on every sibling and emit a real result on the one the engine evaluated plus an explicit "not evaluated here" on the others, rather than a passing placeholder.
- Or surface it as an `AssetObservation`, which the emit path already treats as unconstrained by declared specs, so at least the value is visible even if it cannot gate.

Related: #1598, #1595.

hugocorreia90:
Fixed by #1709. Verified against `main` 27170b61.

Both halves of the cancellation are gone.

The spec side now declares the `candidate` names instead of excluding them (`integrations/dagster/src/dagster_rocky/component.py:1480-1500`).

The emit side no longer drops a result whose pair was never declared. `_undeclared_check_observation` (`component.py:3322`) logs a warning and carries the verdict as a `dg.AssetObservation`, which Dagster accepts outside the declared specs.

[...]

Issue #1666, opened 2026-09-05
column_match and row_count ignore a configured severity on the replication path


hugocorreia90:
A `column_match` check on a replication pipeline **fails on the first run of every table**, before the target exists. It passes on the second run. The failure is not a real schema mismatch — there is no target yet to compare against.

## Reproduce

The shipped POC `examples/playground/pocs/01-quality/02-inline-checks` does it. Run it twice against the same DuckDB file (engine 1.73.0):

```
run 1 (target absent)   column_match passed=false   missing: [event_type, event_id, user_id, occurred_at]
                        row_count    passed=true    source 500 / target 500
run 2 (target present)  column_match passed=true
                        row_count    passed=true
```

500 rows copied on run 1, and the check still says every source column is missing from the target.

## Where

`engine/crates/rocky-cli/src/commands/run.rs` resolves `target_cols` **before** the copy — the same probe that decides `target_exists`, which is `false` on a first run. `check_column_match(&source_cols, &target_cols, ...)` then compares a populated source against an empty target, so `missing` is the entire source column list and `passed` is `false`.

```
  describe target  ->  absent  ->  target_cols = []
  copy the table   ->  target created from source
  column_match     ->  source vs []  ->  every column "missing"  ->  FAIL
```

`check_row_count`, `check_freshness` and `check_null_rate` are unaffected — they run in the batched phase after every table has copied, so they measure the post-copy state. `column_match` is the only replication check computed from a pre-copy read.

## Second defect, same check

`AggregateCheckToggle::severity()` is read in exactly one place, `engine/crates/rocky-cli/src/commands/run_local.rs` (the quality runner). The replication runner never reads it, and `check_column_match` / `check_row_count` hard-code `TestSeverity::Error`. So on a replication pipeline this parses and is silently ignored:

```toml
[pipeline.x.checks.column_match]
enabled  = true
severity = "warning"
```

A user cannot downgrade either aggregate check on the replication path.

## Why it blocks #1598

#1598 wires `fail_on_error` into the replication exit path, so an error-severity check failure fails the run. With this defect in place, **every fresh replication project that sets `column_match = true` would exit 2 and record `PartialFailure` on its first run** — a false failure on the most ordinary case there is. The severity defect means the documented per-check escape hatch would not help; the only outs would be `fail_on_error = false` (which disables every check gate) or deleting the check.

[...]

Issue #1655, opened 2026-09-04
replication checks: a batched query error aborts the run, while the same error on the per-table path is reported as a check


hugocorreia90:
The batched and per-table replication check paths handle a query failure differently. The batched one aborts the whole run; the per-table one records the check as not evaluated and carries on. The comment above them says they are equivalent.

## Where

`engine/crates/rocky-cli/src/commands/run.rs`, in `run_batched_checks`. The comment reads:

> Otherwise, fall back to per-table queries via the generic `WarehouseAdapter` — same observable check results, just more round-trips.

The batched arm is a `tokio::try_join!` over `batch_row_counts` (source), `batch_row_counts` (target) and `batch_freshness`, followed by `.map_err(anyhow::Error::from)?`. Any one of those failing returns the error, so `run_batched_checks` never reaches the reporting code below and the run stops.

The per-table fallback catches each failure, records a reason, and emits a failing `not_evaluated` result for that table.

```
  batch adapter present   ->  one query fails  ->  the whole run errors out
  no batch adapter        ->  one query fails  ->  that check reports, the rest run
```

## Why it matters

Every warehouse adapter that implements `BatchCheckAdapter` takes the first path, so this is the common case, not the fallback. A single unreadable table takes down a replication run that has already written its data, and the other tables' checks never report at all. The per-table behaviour is the one the reporting contract describes.

Aborting is at least fail-closed — it does not manufacture a pass — so this is a consistency and blast-radius problem rather than a correctness hole.

## What would settle it

Have the batched arm catch a per-leg failure and fold it into the same failure maps the fallback populates, so one failed leg reports as `not_evaluated` for the tables it covered and the remaining checks still run. Then the comment claiming equivalent observable results becomes true.

## Provenance

Pre-existing on `main`. Found by an independent review of #1652, which makes the per-table side honest and so widens the gap between the two paths. #1652 already reconciles batch omissions for both row counts and freshness; this is the error case, which is separate.

hugocorreia90:
**Sequenced after #1598, deliberately.** Turning the batched `try_join!` abort into `not_evaluated` today would convert a hard error into **exit 0**, because replication status still ignores failed checks. The per-table path already reports honestly (#1602), so the two paths disagree — but the fix here must land inside the check-outcome contract #1598 establishes, or it opens the false green it is meant to close.

hugocorreia90:
Fixed by #1700. Verified against `main` 27170b61.

[...]

PR #1671, opened 2026-09-05
OWNER MERGE — fix(engine/rocky-cli): fail a replication run whose error-severity checks failed

### LINKED ISSUE #1598: rocky run exits 0 on a replication pipeline whose declared checks failed

hugocorreia90:
A replication run whose error-severity checks failed now exits 2 and records `PartialFailure`. It exited 0 and recorded `Success`, so an orchestrator reading the exit code and the scheduler reading the record both saw a healthy run.

`--resume` of such a run is refused: it would copy nothing, run no check, and exit 0 on data that is still violating.

## Draft: blocked on #1666

**Do not merge this before #1666.** Building it turned up a defect that this gate would turn into a false failure on an ordinary run.

`column_match` is computed from a read of the target taken **before** the copy, so on the first run of every table — when no target exists yet — it reports every source column as missing and fails. It passes on the second run. Under this gate that false failure would fail a fresh project's first run. The repo's own POC shows it, measured on both binaries:

```
examples/playground/pocs/01-quality/02-inline-checks   (column_match = true)
  1.73.0    run 1 exit 0    run 2 exit 0
  this PR   run 1 exit 2    run 2 exit 0     <- the false failure

examples/playground/pocs/00-foundations/01-replication-basics   (checks off)
  1.73.0    run 1 exit 0    run 2 exit 0
  this PR   run 1 exit 0    run 2 exit 0     <- healthy pipeline unchanged
```

Run 1's failed check is `column_match`, on a table that copied all 500 rows. `row_count` passes on both runs. #1666 has the detail, plus a second defect on the same check: `AggregateCheckToggle::severity()` is read only by the quality runner, so `severity = "warning"` on `column_match` or `row_count` parses and is ignored on the replication path — a user could not downgrade their way out.

**#1666 has to land first.** Shipping the gate over it is not a sequencing preference, because a user hit by the false first-run failure has no way out that keeps the gate. The per-check downgrade does not work on this path — that is #1666's second defect, `AggregateCheckToggle::severity()` is read only by the quality runner — so the only escape is `fail_on_error = false`, which turns off the whole gate this PR adds. A gate whose only escape hatch is disabling the gate is not shippable.

**When #1666 lands, delete the "Before you upgrade" paragraph from the CHANGELOG entry** — it is the only part of this PR that assumes the defect is still there.

## The four surfaces now give one answer

```
  copy every table  ->  an error-severity check fails, fail_on_error = true
                                    |
                        check_gate_failed = true      (stamped once, before any status)
                                    |
        +---------------+-----------+-----------+----------------+
        |               |                       |                |
   exit code 2     JSON status          RunRecord status      --resume
                   PartialFailure       PartialFailure         refused
                   check_gate_failed                          "would copy
                   tables_failed: 0                            nothing"
```

Before, those four read: exit 0, `Success`, `Success`, and a resume that copied nothing and exited 0.

## What changed

**The gate is its own field, not `tables_failed`.** `RunOutput.check_gate_failed` is a bool: at least one error-severity check failed **and** `fail_on_error` is on. `tables_failed` keeps its documented meaning — a count of tables and models — so one table with three failed assertions still reports `tables_failed: 0`. This is what #1599 was blocked for; the fold must not ship.

**Stamped before anything derives a status.** `merge_replication_compile_and_copy_errors` writes `output.status = derive_run_status()`, and `persist_run_record` and `finalize_idempotency` each call `derive_run_status` again. All three now read the same gate, so the stale-JSON-status defect from #1599 is closed at the function that had it.

**A check that could not be evaluated gates too**, at its own severity. Every `not_evaluated` result carries `passed: false` (#1602). Row-count, freshness, null-rate and custom checks hard-code error severity, so a broken one fails the run; an assertion or cross-source-overlap check carries the configured severity, so a warning-severity one that could not run stays advisory.

**A new exit sentinel, not `PartialFailure`.** `CheckGateFailure` maps to exit 2 in `main.rs` alongside `PartialFailure`. A separate type because the advice differs: a failed copy is resumable, a failed check is not, so the message must not say `use --resume`. The check failure also never reaches the compile-only branch — it does not touch `tables_failed`, which is that branch's condition.

**The resume gate learned a third case.** `ensure_run_is_resumable` admitted `Failure | PartialFailure` unconditionally. It now asks whether a resume would do any work:

```
  planned nothing                      -> resume: no checkpoint to reason about
  any entry Failed/Interrupted/Skipped -> resume: a resume re-attempts all three
  copied < planned                     -> resume: real copy work remains
  a model failed                       -> resume: the model phase re-runs
  otherwise                            -> refuse: a resume would do nothing
```

The discriminator is deliberately **not** `RunRecord::check_outcomes`, which records no severity: keying on it would refuse a legitimate resume over a warning-severity result, or over any result under `fail_on_error = false`. "Did a model fail" is severity-free — `to_run_record` writes a failed `models_executed` entry per `RunOutput::errors` entry, and a check failure adds none.

**One dependency, stated rather than hidden.** "Every planned table copied" is decided by a count against `total_tables`, and a count can only stand in for the planned *set* while every recorded key comes from the invocation's own plan. It does — one `init_run_progress` per run with `tables_to_process.len()`, both `record_table_progress` sites iterating that same list, per-table keying so no table is counted twice, and DAG sub-runs minting their own ids — but the gate cannot check any of it, and no arithmetic on counts can: a checkpoint holding `Success(a)` and `Success(x)` against a plan of `{a, b}` has the right count and the wrong set. The comment names every producer, reproduces that counterexample as the reason, and tells the next author that a new recording site must record a table from `tables_to_process`.

Proving completeness properly needs the planned table identities on the checkpoint, which is a `RunProgress` blob shape change — out of scope here, and named in the code as the real fix. **This is the one judgement call in the PR worth a second opinion.** The first two arms narrow the exposure: any non-`Success` entry means work remains whatever the count says, which is sound on its own because the resume filter skips only `Success` keys.

Known conservative edge, pinned by a test: `total_tables` is what the recorded run planned **after** its own resume filtering, so a resume of a resume reads as complete and is refused even though earlier tables would have been re-copied. The refusal is loud and says to re-run; the alternative direction is the silent exit 0 this gate exists to stop.

**Honest failure.** A warning-severity failure changes nothing. `fail_on_error = false` changes nothing. Quality, transformation and snapshot pipelines are untouched — the quality runner already read the flag, and it now shares one severity tally with the replication gate instead of keeping its own copy.

**The docs needed no change.** `data-quality-checks.md` and `failure-modes.md` already said an error-severity failure fails the run by default. They were describing the promise; the code was not keeping it.

## Evidence

Fail-before, one line per test — each mutation reverts one part of the fix on a clean tree via `scripts/mutation-check.sh`:

| Mutation, applied to a clean tree | Test that must fail | Result |
|---|---|---|
| drop `\|\| self.check_gate_failed` from `derive_run_status` | `derive_run_status_check_gate_on_a_fully_copied_run_is_partial` | failed under mutation |
| `replication_check_gate_failed` returns `false` | `the_check_gate_reads_severity_and_fail_on_error` | failed under mutation |
| resume gate's `copied < planned` relaxed to `<=`, so it never refuses | `resume_refuses_a_complete_checkpoint_whose_failure_was_not_a_copy_or_a_model` | failed under mutation |
| the call site stamps `check_gate_failed = false` | `a_failed_error_severity_check_fails_the_run_on_every_surface` (end to end, real binary) | failed under mutation |
| drop the resume gate's non-`Success` conjunct | `adversarial_checkpoints_stay_resumable` | failed under mutation |
| drop the resume gate's zero-plan conjunct | `adversarial_checkpoints_stay_resumable` | failed under mutation |

End to end through the real binary (`engine/rocky/tests/replication_check_gate.rs`), a DuckDB replication pipeline with a `not_null` assertion on a column the source leaves NULL:

- `a_failed_error_severity_check_fails_the_run_on_every_surface` — exit 2, JSON `status: "PartialFailure"` + `check_gate_failed: true` + `tables_failed: 0` + `tables_copied: 1`, `rocky history` reports `PartialFailure`, and both `--resume <id>` and `--resume-latest` refuse. On `main` this run exits 0 with `status: "Success"`.
- `fail_on_error_false_keeps_the_run_green` — exit 0, `Success`, field omitted.
- `a_warning_severity_check_failure_keeps_the_run_green` — exit 0, `Success`, field omitted, and the warning still appears in `check_results`.

Gates: `cargo +1.98.0 test -p rocky-cli -p rocky-core --lib` 1625 + 2079 pass; `clippy --all-targets -p rocky-cli -p rocky-core -p rocky -- -D warnings` clean; `fmt --all --check` clean; `uv run pytest` 241 pass.

Two existing resume tests changed. Both seeded a failure status over a run that had already copied everything, only to get past the #1548 gate — the comment says so — and neither is about complete-checkpoint resumability. They now seed the shape they are actually about: an unfinished copy, and a failed model phase. The refusal they used to cover incidentally has its own test.

## Generated files

`just codegen` ran. Committed: `schemas/run.schema.json`, `sdk/python/src/rocky_sdk/types_generated/run_schema.py`, `editors/vscode/src/types/generated/run.ts`, `docs/public/openapi.json`. `editors/vscode/package-lock.json` was reverted (the known spurious `libc` churn). The hand-written SDK shadow `sdk/python/src/rocky_sdk/types.py` gained the field by hand, which `test_schema_parity.py` requires.

`just regen-fixtures` ran and captured 40 fixtures with **no drift** — the fixture POC disables both aggregate checks, and the field is omitted from the wire when `false`, so no existing payload changes.

## Not in scope

- #1655, the batched `try_join!` abort. Sequenced after this on purpose: turning that abort into `not_evaluated` before this lands would make it exit 0.
- The Dagster component dropping undeclared `candidate` cross-source check results (noted on #1598). The exit code now carries the failure regardless, so the run is not silent there; the spec gap is filed separately as #1669.
- A resume of an **incomplete** checkpoint still re-runs checks only for the tables it copies, so a check that failed on an already-copied table is not re-evaluated. That is pre-existing — checks have always been scoped to the current invocation's copies — and is filed as #1670.

- Persisting the planned table identities on the checkpoint, so the resume gate compares sets instead of counts and stops depending on producer discipline. That is a `RunProgress` blob shape change, so it is filed as #1674 rather than made here. The independent review recommended it as the stronger follow-up while saying the gate should ship with the dependency stated.

Closes #1598
Refs #1599, #1602, #1666, #1669, #1670, #1674


hugocorreia90:
**Red team: Codex (`gpt-5.6-sol`), read-only, 3 rounds.** Rounds 1 and 2 returned BLOCK, round 3 ALLOW.

**Round 1 — the resume gate rested on an invariant it could not see.** It decided a checkpoint was complete by counting `Success` entries against `total_tables`, which is only sound while every recorded key comes from the run's own plan. That does hold, so the reviewer's concrete case was unreachable — but I fixed it rather than argue the invariant. Any entry recorded `Failed`, `Interrupted` or `Skipped` now means work remains whatever the count says, and a run that planned no tables stays resumable. Both are sound on their own: a resume skips only `Success` keys.

**Round 2 — the fix comment overclaimed, and that was the real defect.** It said the new guard made the gate independent of the producer invariant. It does not, and no arithmetic on counts can: `Success(a)` plus `Success(x)` against a plan of `{a, b}` has the right count and the wrong set. The reviewer offered two resolutions — persist the planned identities, or rely on the invariant explicitly and name it. The first is a `RunProgress` blob shape change, which this PR may not make, so I took the second. The comment now names every producer, reproduces the counterexample as the reason no count can work, points at the real fix, and tells the next author that a new recording site must record a table from `tables_to_process`. Filed as **#1674**.

**Round 2 also caught a CHANGELOG over-claim.** "Nothing that could be recovered stops being recoverable" was false given the resume-of-a-resume refusal this PR documents. Both that sentence and the `Failed`/`Interrupted`/`Skipped` one now give their reason instead of asserting a guarantee.

**Round 3 verdict: ALLOW**, and on the sequencing question it was explicit — ship the gate with the dependency stated, because deferring it preserves the known silent exit 0.

**Test gap on the other replication checks — accepted, and it is why this is a draft.** The reviewer could not rule out another check failing on an ordinary first run. One does: `column_match`, filed as **#1666**. Row-count, freshness and null-rate run after every table has copied, so they cannot repeat that mechanism.

**Confirmed sound across rounds:** status, exit code, record and idempotency agree; every `RunStatus` consumer change is the intended posture, not a regression; `tables_failed` and `execution.tables_failed` keep their documented meanings and no existing field changes its serialization; a warning-severity failure and `fail_on_error = false` change nothing.

Each conjunct of the resume gate is mutation-checked — reverting any one of them fails a named test.

hugocorreia90:

**Verified**
- `check_gate_failed` is a `bool` on `RunOutput`, stamped once (`run.rs`, before `derive_run_status`). Zero lines fold checks into `tables_failed`. No schema bump. Codegen committed, dagster fixtures show zero drift.
- The severity escape hatch is quality-only today: `AggregateCheckToggle::severity()` has exactly one reader, `run_local.rs:435`. Replication never reads it.
- Blocker #1666 is real in code: the `target_cols` passed to `check_column_match` (`run.rs:12343`) come from the pre-copy probe. Staying draft until #1666 lands is the right call.
- CI gap is real: `engine-ci.yml` runs no POC, and `sdk-ci.yml:54-58` installs the latest engine *release*. So no job would have caught #1666 on this PR. Filing separately.

**Not re-verified by me**
- #1666 states row_count / freshness / null_rate run post-copy, so column_match is the only first-run false failure. I did not re-trace those paths here.

**Owner merge** because a run with a failing check gate goes from exit 0 to exit 2 on upgrade.

hugocorreia90:
Orchestrator: rebased onto `main` (merge `0330871e`; changelog-only conflict, both entries kept). Marking ready.

**Interaction with #1678 (merged after the last review):** #1678 evaluates `column_match` after the copy and turns a rate-limited post-copy probe into `column_match_not_evaluated(..)`. This gate counts a `not_evaluated` result by its own severity (`check_failures_by_severity`, `output.rs`), and `column_match` is hard-coded error severity, so an unanswerable probe now fails the run: `PartialFailure`, exit 2. That is the same fail-closed posture #1602 chose for checks the engine could not evaluate and the one #1678 documented for `verify_after`; it is consistent, and it is the behaviour to know about on upgrade.

**Local gate on the rebased head:** the run module reported three failures in a parallel run (`run_trigger_from_env_maps_known_values`, `persist_run_record_stamps_trigger_submission_and_pipeline`, `a_run_adopts_an_upstream_traceparent`); all three pass with `--test-threads=1`. Two share `TRIGGER_ENV_LOCK`, so one panic poisons the lock for the others — a parallel-run artefact, not this branch. Clippy clean.

Independent review stands at `ac9dd957` (Codex R1 BLOCK, R2 BLOCK, R3 ALLOW); nothing but the changelog changed since. Codex is out of credits until Sep 7, so the rebase itself has no independent pass.

hugocorreia90:
Post-merge review done (Claude Opus 5 at max effort, at the owner's direction — Codex is out of credits, so this is not the independent pass `AGENTS.md` asks for). Verdict BLOCK. Filed as #1720; I verified the mechanism line by line on `main` df506990 first.

**The design is sound and the review confirmed a lot of it:** the stamping order holds, the exit-2 downcast works end to end, idempotency maps `PartialFailure` correctly so no false `SkippedIdempotent`, the scheduler does not retry a copy that succeeded, #1706's passing overlap result correctly does not gate, and the SDK shadow carries the field.

**The load-bearing gap is that `check_gate_failed` never reaches `RunRecord`** — zero occurrences in `rocky-core/src/state.rs`. The design rests on four surfaces giving one answer, and the fourth, the resume gate, is the one that cannot get it: it reconstructs the verdict from `models_executed`, which is orthogonal to check severity. Consequences, both verified:

- A check-gated run whose record carries any failed `models_executed` entry — including the one `run.rs:5435` writes for a failed `verify_after`, with no model involved — admits a resume at `run.rs:1724`. The resume copies nothing, pushes no `check_results`, so `check_gate_failed` is false, and `derive_run_status`'s `(_, false) => Success` arm records **`Success`**, exit 0.
- The higher-frequency shape: a run with one failed copy *and* a failed check on a copied table exits 2 printing `use --resume X to retry`. The resume legitimately copies the failed table, checks cover only that table, and the record goes **`Success`**. `latest_successful_run` matches exactly `Success`, so downstream `after` demands then fire on violating data.

That also makes the message at `run.rs:5630` actively harmful: the `tables_failed > 0` branch returns before the `CheckGateFailure` branch, so the sentinel whose own doc says "a failed check is not resumable … this message must not send the user there" never fires when both happened.

**Three more, all verified:**
- `AggregateCheckToggle::severity()` still has one reader (`run_local.rs:435`, the quality runner), so `severity = "warning"` on a replication check parses and is ignored. After #1700 and #1678 that is not academic: one failed batched leg or one rate-limited `DESCRIBE` gates the whole run at error severity, and the only escape is `fail_on_error = false`, which disables the gate entirely. The PR body's own bar — "a gate whose only escape hatch is disabling the gate is not shippable" — is still unmet for that class.
- `check_gate_failed`'s published description (`output.rs:268` and all four generated bindings) is unconditional, but the field is assigned in exactly one place, replication only. A quality pipeline with a failing check emits `status: "Success"`, `tables_failed: 0`, no `check_gate_failed`, while the record says `Failure` and the process exits 1 — the same three-answer shape #1599 was closed to prevent, on the pipeline type whose whole purpose is checking.
- Two `[Unreleased]` entries contradict each other: `:142` says a failed replication check does not fail the run, and `:156` still tells upgraders to audit `column_match` for a first-run defect that #1678 fixed hours earlier. The PR body pre-committed to deleting that paragraph when #1666 landed; #1666 landed first.

I am folding the changelog corrections into the #1718 lane and queueing #1720 behind the lanes now building.

Issue #1273, opened 2026-07-27
Shadow objects have no ownership or lifecycle contract (collisions, cleanup_after inert, repeat-run append)


hugocorreia90:
## Description

Rocky has no ownership or lifecycle contract for the tables shadow and branch runs create. Three symptoms fall out of that one gap, and fixing any of them in isolation risks contradicting the others.

### 1. Collision detection only sees Rocky's own model targets

`apply_shadow_rewrite` (`engine/crates/rocky-cli/src/commands/run.rs`) rejects a derived shadow target that overlaps a configured production target or another selected model's shadow target. It does not check whether the derived name already exists in the warehouse as an unrelated object.

Given a project whose only model targets `analytics.mart.orders`, and a warehouse that already contains an unrelated `analytics.mart.orders_rocky_shadow`, a shadow full-refresh replaces that table. Nothing in the project declares it, so no collision is detected.

### 2. `cleanup_after` has no consumer

`ShadowConfig::cleanup_after` (`engine/crates/rocky-core/src/shadow.rs`) defaults to `true` and is documented as "whether to drop shadow tables after comparison completes". No production code reads it — there is no drop anywhere. Every CLI construction (`commands/run.rs`, `commands/branch.rs`, `commands/apply.rs`) sets it to `false` regardless, so the default is unreachable and the field is inert.

Shadow tables therefore accumulate indefinitely.

### 3. Retained shadow tables corrupt repeat runs for append strategies

Because nothing is cleaned up, a second shadow run of an `incremental`, `merge`, `delete_insert`, or `microbatch` model finds its shadow table already populated and appends to it again. Row counts drift further from production with every run, so the comparison degrades rather than converging.

Separately, even a *first* shadow run of these strategies bootstraps into an empty table while production holds full history, so the comparison is not meaningful. `content_addressed` and `time_interval` are explicitly rejected under shadow; the incremental family is allowed but produces results that cannot be compared against production in the way the feature implies.

## Expected behavior

A documented contract answering: is a derived shadow target disposable, reusable across runs, or owned by a named branch? Once that is settled:

- Refuse to write a pre-existing shadow target unless it is provably Rocky-owned for the same branch or run namespace (warehouse tag or state-recorded ownership), or require a dedicated shadow schema whose contents Rocky owns wholesale.
- Honor `cleanup_after` for one-off shadows, deliberately preserve named branches, or remove the field if neither is intended.
[...]

PR #1672, opened 2026-09-05
test(engine): pin what a config refusal promises, and compile the init-adapter scaffold

### LINKED ISSUE #1607: rocky init-adapter: nothing compiles the scaffold, so the template can drift from the traits

hugocorreia90:
Four test-only changes. No shipped behaviour changes, no CHANGELOG entry.

| # | What it now proves | Issue |
|---|---|---|
| 1 | A refused `metadata_columns[].value` issues **zero** warehouse statements | Refs #1609, pins #1594 |
| 2 | `rocky init-adapter` emits a crate that compiles | Closes #1607 |
| 3 | A `_last_checkpoint` marker alone turns a provable removal into a hold | Refs #1121 |
| 4 | An unnamed `--shadow` run replaces a table it does not own | Refs #1273 (probe, `#[ignore]`) |

---

## 1. A recording warehouse adapter (#1609 point 3)

Before this, "the refusal landed before the warehouse" could only be shown by reading control flow, then checking one DuckDB observation: does the target schema exist? That observation cannot separate two different worlds.

```
   nothing was issued            ──┐
                                   ├─ both leave `staging` absent on DuckDB
   catalog created, schema not   ──┘
```

DuckDB's `create_catalog_sql` returns `None`, so there is no catalog statement on that adapter to sit between the guard and the assertion. `engine/rocky/tests/metadata_column_refusal.rs` says exactly this in its own header, and asks for a recording adapter.

`engine/crates/rocky-cli/src/testing.rs` adds one. It records every call in order, executes nothing, and its dialect **does** answer `create_catalog_sql`.

One test in `commands/run.rs` uses it, as a control/treatment pair on the same fixture:

```
  no metadata column   ──► preflight passes ──► CREATE CATALOG, CREATE SCHEMA, …
  with metadata column ──► preflight REFUSES ──► (no statement at all)
```

Three assertions carry it, and each one closes a way the test could pass while proving nothing:

- the control log is non-empty **and its first statement is the `CREATE CATALOG`** — that ordering is what makes an empty treatment log rule out a guard placed *after* catalog creation;
- the treatment error names `metadata_columns` and the offending component, so an unrelated earlier bail cannot pass as the refusal;
- the treatment log contains a `ListTables` call, so the adapter was provably reached — an empty statement log cannot pass by the adapter never having been constructed.

**Please read this bit — it is the one thing in the diff that looks like a production change.** `registry.rs` gains a `"recording"` arm. It sits behind `#[cfg(test)]`, as does the `testing` module in `lib.rs`. A released binary contains neither. `KNOWN_ADAPTER_TYPES` is untouched, so a user config with `type = "recording"` still fails with the ordinary unsupported-adapter message and the "did you mean?" suggestion.

**Scope, stated plainly.** The log counts SQL handed to `WarehouseAdapter::execute_statement` — every catalog, schema, pre-drop, CTAS, INSERT and MERGE. Tag, workspace-binding and grant emission also land there on a real warehouse, but the recording adapter resolves to the no-op governance adapter exactly as DuckDB does. So those three are absent from the *control* log rather than proven-absent from the treatment log.

## 2. Compile the `init-adapter` scaffold (#1607)

Nothing in CI compiled what `rocky init-adapter` writes. The `#[cfg(test)]` blocks inside the template are text, not code, and the render tests check that text against a hand-maintained list of method names.

`engine/rocky/tests/init_adapter_scaffold_compiles.rs` runs the **real binary** into a temp dir, then drives `rustc` once over the emitted crate root against the rlibs this test binary's own build already left in `deps/`. No synthesized workspace, no second dependency resolution, no rebuild of `rocky-core` — one metadata-only invocation.

Two halves, both required by #1607:

- as emitted, it must **fail**, naming the scaffold's own `compile_error!` text (`compile_error!` has no stable error code, and rustc prose is not stable — the message is ours);
- with `literal_escape` substituted for `LiteralEscape::Standard`, it must **succeed**.

The scaffold itself is unchanged, and no CLI option was added. The edition comes from `[workspace.package]` in `engine/Cargo.toml` rather than a literal, since the scaffold says `edition.workspace = true`.

One rlib set, one compile, no fallback — this is where the red team's two FAILs landed, see the disposition comment. `deps/` is shared and records no provenance, so the pick is the newest rlib per crate **that is not newer than this test binary**. Cargo writes every rlib it links before it links the binary, so anything newer cannot be what it linked. That is an eligibility bound, not proof of identity, and the file's header says so: a `deps/` doctored after the build can still fool it.

## 3. `CheckpointPresent` has a test (#1121)

Two `InMemory` stores differing by exactly one object:

```
  v0 bootstrap + own add ─┐
  v1 remove of that add   ├─► no marker  ──► ProvenRemoved { head_version: 1 }
                          └─► + marker   ──► Held(CheckpointPresent)
```

The control half is the point. A lone "this fixture holds" assertion passes on any shape the reader declines for some other reason. This one pins the marker as the discriminator. The kernel swap #1121 tracks stays deferred; this is its regression net.

## 4. Probe: shadow objects have no owner (#1273)

`engine/rocky/tests/shadow_target_ownership.rs` seeds `main.orders_rocky_shadow` with a column no model produces, runs `rocky run --shadow`, and observes the model's own columns in its place.

It is **`#[ignore]`d on purpose**. With no ownership or lifecycle contract for shadow objects there is no correct behaviour to assert, so a green CI lane must not depend on today's. It documents the defect until a contract lands. Run it with `cargo test -p rocky --test shadow_target_ownership -- --ignored`. Full findings, with the two line numbers and the grep that shows no ownership probe sits between them, are on #1273.

---

## Mutation evidence

Every test above was proven fix-sensitive with `scripts/mutation-check.sh`, which refuses a dirty worktree.

| Mutation | Result |
|---|---|
| Neuter **only** the preflight `resolve_metadata_columns(…)?` in `run.rs`, leaving the collection-loop call intact — the exact #1594 regression | exit 0. The test failed with `calls: [ListTables, ExecuteStatement("CREATE CATALOG …"), ExecuteStatement("CREATE SCHEMA …")]` |
| Delete `drop_table_sql` from the scaffold template | The test failed with its own "does not compile once `literal_escape` is chosen" panic plus `error[E0046]: not all trait items implemented, missing: drop_table_sql`. Re-run after each red-team fix, three times in total. **The script reported exit 3 (INCONCLUSIVE) and that verdict is wrong** — its heuristic reads the compile errors as a broken build. The log shows `test result: FAILED. 0 passed; 1 failed`, so the test ran and caught it. This is the known #1547 false verdict; read the log, not the exit code. |
| Replace the scaffold's `compile_error!` with `LiteralEscape::Standard` | exit 0. The test failed with "the emitted scaffold compiled — `literal_escape` answered the question for the adapter author instead of refusing" |
| `if false && (last == "_last_checkpoint" …)` in `proven_removed` | exit 0. The test failed on the treatment assertion |

The #1273 probe is not mutation-checked: there is no fix to revert.

## Red team

Codex (GPT-5.6 Terra / `gpt-5.6-sol`), three rounds: FAIL, FAIL, PASS. Both FAILs were real, both were the scaffold test's rlib selection, and both are fixed. Full disposition — including the two nits left unfixed on purpose and what each round was and was not shown — is in the comment below.

## Gates

`cargo +1.98.0` throughout.

- `clippy --all-targets --all-features -- -D warnings` — clean, whole workspace
- `fmt -- --check` — clean
- `test --all-features -p rocky-cli -p rocky-iceberg -p rocky` — green

Closes #1607
Refs #1609
Refs #1121
Refs #1273


hugocorreia90:
## Red team: Codex (GPT-5.6 Terra / `gpt-5.6-sol`), 3 rounds. R1 FAIL, R2 FAIL, R3 PASS.

The two FAILs were the same defect, narrowing. Both were real. Both are fixed.

**R1 — accepted.** The scaffold test tried every candidate rlib combination and returned on the first that compiled. A `target/` holding a stale but self-consistent artifact set would let a scaffold broken against today's `SqlDialect` compile against yesterday's — green on the exact drift the test exists to catch. Fixed in `1f2c9a58`: one compile, no fallback.

**R2 — accepted, including the part about my own prose.** Newest-by-mtime can itself pick a stale set, so my claim "false FAILURE only, never false PASS" was an overclaim, and the header contradicted its own admission that provenance is unproven. Fixed in `2973f909`, two ways:

```
   deps/  ├─ librocky_core-A.rlib   (this build)        ◄── eligible
          ├─ librocky_core-B.rlib   (a later cargo run) ◄── skipped: newer than this binary
          └─ …
   test binary mtime  ── upper bound ──┘
```

Cargo writes every rlib it links before it links the binary, so anything newer than this test binary cannot be what it linked. That removes the state that arises on its own — a later `cargo` invocation with a different feature set. The header now states the residual instead of denying it: a `deps/` doctored after the build can still carry a newer mtime, and in that state the test could pass when it should fail.

**R3 — PASS**, with the decision question answered directly: cargo exposes no artifact-identity channel to an integration test, so a documented residual plus a stated clean-`target/` assumption is the right disposition for a CI trait-drift test.

### Two R1 nits, deliberately not fixed

- The recorder unit test uses a fixed key, `"testing::order-probe"`. Codex called the collision risk speculative. No other test shares it; the key names its own test.
- `is_crate_loading_failure` lists rustc codes E0460/E0461/E0463/E0464/E0514. A future code outside that list picks the wrong *message*, not the wrong verdict — the test still fails.

### One R3 caveat, grounded

Codex flagged that it could not verify CI uses a fresh target directory. It does not: `engine-ci.yml:155` restores `engine/target` from a shared `Swatinem/rust-cache`. That does not break the bound, and the reason is step ordering, not anything about the cache's internals — the restore runs before the build, so the freshly linked test binary is still the newest artifact in `deps/`.

### What was excluded from the reviews, stated

R1 got the complete `engine/**` diff, nothing filtered. R2 and R3 got only the delta for the file that changed, plus that file in full; the other six files were unchanged since R1 passed them. No CHANGELOG entry and no generated bindings: the change is test-only and touches no CLI JSON output struct.

Every finding was verified against the code before acting on it. Mutation checks were re-run after each fix; the scaffold test still fails when `drop_table_sql` is dropped from the template.

hugocorreia90:

**Verified**
- Every production-code touch is `#[cfg(test)]`-gated: the `testing` module in `lib.rs` and the `"recording"` arm in `registry.rs`. `KNOWN_ADAPTER_TYPES` is only mentioned in a comment, never edited, so `rocky validate` still rejects `type = "recording"`.
- The #1273 probe test carries `#[ignore = "documents open defect #1273 …"]`; CI does not run it, by design. Nobody should read this PR's green as coverage of #1273.
- Scope caveat from the lane stands: the #1594 pin proves zero statements for catalog and schema only. Tag / binding / grant are unproven either way, because the recording adapter resolves to the no-op governance adapter, exactly as DuckDB does.

Self-merge class (tests and test doubles only). Merging when the updated branch is green.

Issue #1620, opened 2026-09-03
checks.freshness overrides parses but nothing consumes it (N2 of #1620)


hugocorreia90:
#1435 named two `rocky.toml` surfaces that parse and validate while nothing reads them. A systematic sweep of all 24 top-level `RockyConfig` fields and their sub-fields found **three more**. All three are sub-fields, not whole sections, and each is documented as if it worked.

Verified on `main` at `c32f28f9`, with a second independent pass.

## N1 — `[cache.schemas] replicate` (strongest)

**Claim in the docs:** `docs/src/content/docs/reference/configuration.md:923` and `:930` — "Replicate the schema cache via `[state]` sync", `replicate = true # opt in to share cache via the remote state backend`. Also in `schemas/rocky_project.schema.json:3631`.

**What happens:** nothing. Setting it to `true` changes neither sync leg.

- **Upload.** `engine/crates/rocky-core/src/state_sync.rs:2341` — `upload_state()` hard-codes `crate::state::LOCAL_ONLY_TABLE_NAMES`, which contains `"schema_cache"` (`engine/crates/rocky-core/src/state.rs:288`). The override `upload_state_with_excluded_tables` (`state_sync.rs:2391`) exists and its own doc at `:2338` says *"Use `upload_state_with_excluded_tables` to override the default when `[cache.schemas] replicate = true` is configured"* — and it has **no caller outside `state_sync.rs`** (one internal call and one test).
- **Download.** `engine/crates/rocky-cli/src/api.rs:1250` documents that `download_state` unconditionally preserves local-only tables.

```
$ grep -rn --include='*.rs' "\.replicate\b" engine/crates engine/rocky
# every hit is a #[cfg(test)] assertion inside config.rs, or a struct
# literal WRITING `replicate: false` (source_schemas.rs:147,
# schema_cache_writer.rs:189). Zero production reads.
```

The implementation hook was built and never wired: the same shape as #1435's grace-period half.

## N2 — `[pipeline.<name>.checks.freshness] overrides`

**Claim in the docs:** `configuration.md:386` shows `{ threshold_seconds = N, overrides = { ... } }` with no caveat. In `schemas/rocky_project.schema.json:1267`.

**What happens:** the check reads only the scalar. `engine/crates/rocky-cli/src/commands/run.rs:5057`:

```rust
let mut check = checks::check_freshness(lag, freshness_cfg.threshold_seconds);
check.severity = freshness_cfg.severity;
```

[...]

PR #1675, opened 2026-09-05
OWNER MERGE — fix(engine/rocky-core): wire [cache.schemas] replicate on both state-sync legs

hugocorreia90:
`[cache.schemas] replicate = true` parsed, validated, shipped in the project JSON schema, and was documented as *"Replicate the schema cache via `[state]` sync"*. It did nothing at all.

It works now. **The default is `false`, so nothing changes for anyone who has not opted in.**

## Why it did nothing

`schema_cache` was hard-coded into `LOCAL_ONLY_TABLE_NAMES` at every sync leg. Local-only tables are stripped from the remote copy on upload and overwritten from the local file on download — whatever the setting said.

The override helper the doc comment pointed at had **no caller outside its own module**:

```rust
// state_sync.rs — the comment on upload_state, before this PR
/// Use `upload_state_with_excluded_tables` to override the default when
/// `[cache.schemas] replicate = true` is configured.
```

Nothing ever did. The hook was built and never wired — the same shape as #1435's grace-period half.

## Four legs, not two

The issue names the upload and the download. There are **four** sites reading the constant, and wiring a subset is worse than wiring none — a project on `concurrency_control = "cas"` would have set the key, seen the non-CAS path honour it, and still replicated nothing.

| Leg | Site | Before | After |
|---|---|---|---|
| End-of-run upload | `upload_state` | `LOCAL_ONLY_TABLE_NAMES` | posture-aware |
| CAS upload | `upload_state_cas` | `LOCAL_ONLY_TABLE_NAMES` | posture-aware |
| Mid-run periodic uploader | `RemoteStateSession::start_periodic_uploader` | `LOCAL_ONLY_TABLE_NAMES` | posture-aware |
| Download merge + publish | `publish_merged` | `LOCAL_ONLY_TABLE_NAMES` | posture-aware **per arm** — see below |

One function answers it:

```rust
// state.rs
pub fn local_only_table_names(replicate_schema_cache: bool) -> &'static [&'static str]
//   false -> LOCAL_ONLY_TABLE_NAMES                          (unchanged)
//   true  -> the same list minus "schema_cache"
```

A table is replicated **precisely by not being local-only**, so that set difference is the whole feature.

## The download arm that must NOT change

`publish_merged` has three arms and they are not symmetric. The `Absent` arm — no remote object, but a local store exists — rebuilds the local file from the local-only set alone.

```
  remote Restored  -> take local-only FROM LOCAL, remote wins for the rest
                      replicate=true: schema_cache not taken -> remote's cache survives  ✅

  remote Absent    -> rebuild local file from the local-only set
                      replicate=true with the replicating set would OMIT schema_cache
                      -> the warm local cache is DELETED every time the remote is missing  ❌
```

That arm therefore keeps the **full** local-only set even when replicating. Replication is about sharing the cache, never about discarding it. A test pins it.

## Why the posture is an explicit argument

It is threaded through as a `bool` rather than derived inside the sync layer or carried as a `#[serde(skip)]` field on `StateConfig`. A field that silently defaults on any path that forgot to set it is the same shape as the defect being fixed here — and `RockyConfig` round-trips through plan snapshots, where a skipped field would be dropped. Explicit costs signature churn; it cannot silently not work.

## Fail-before / pass-after

Mutation: revert `local_only_table_names` to ignore the posture and always return the full list — exactly the pre-fix behaviour.

| Test | Under mutation | Why |
|---|---|---|
| `schema_cache_crosses_pods_when_replicate_is_enabled` | **FAILS** | the only test asserting the NEW behaviour |
| `state::tests::local_only_table_names_replicating_drops_only_the_schema_cache` | **FAILS** | asserts the posture is read |
| `schema_cache_stays_local_by_default` | passes | asserts what must NOT change |
| `replicating_upload_alone_does_not_reach_the_other_pod` | passes | asserts what must NOT change |
| `replicating_download_tolerates_a_remote_with_no_schema_cache_table` | passes | honest-failure guard |
| `replicating_download_keeps_the_local_cache_when_the_remote_is_absent` | passes | guards the arm deliberately left alone |
| `local_only_table_names_default_is_the_unchanged_list` | passes | that IS the mutation's behaviour |

Stated precisely because the first draft of this section claimed all five new tests were fix-sensitive, and the log says one is. That is correct: the others assert behaviour that must SURVIVE the change, so passing under mutation is what they are for.

**Per-leg mutations, added after the red team.** The whole-function mutation above cannot tell whether each of the four legs is independently covered. Two narrower mutations, each disabling the posture at ONE leg, answer that:

|---|---|
| `upload_state_cas` hard-codes the old set | `schema_cache_crosses_pods_through_the_cas_upload_leg` |
| the periodic uploader hard-codes the old set | `schema_cache_crosses_pods_through_the_periodic_uploader` |

Each is caught by exactly one test, and only that one. Both legs now have coverage that a subset-wiring regression cannot slip past.

The five new cross-pod tests drive the real upload → shared in-memory S3 → download chain:

- `schema_cache_stays_local_by_default` — the guard. `replicate = false` keeps the cache node-local, as before.
- `schema_cache_crosses_pods_when_replicate_is_enabled` — the fix. Pod A's cached types reach pod B.
- `replicating_upload_alone_does_not_reach_the_other_pod` — pins that both legs are needed, so the fix cannot be half-applied unnoticed.
- `replicating_download_tolerates_a_remote_with_no_schema_cache_table` — the honest-failure case (below).
- `replicating_download_keeps_the_local_cache_when_the_remote_is_absent` — the `Absent`-arm rule above.

Three more, added after the red team, cover the two upload legs the first round missed:

- `schema_cache_crosses_pods_through_the_cas_upload_leg` — `upload_state_cas` honours the posture.
- `cas_upload_leg_keeps_the_schema_cache_local_by_default` — and its default is unchanged.
- `schema_cache_crosses_pods_through_the_periodic_uploader` — the mid-run snapshot honours it too.

Plus three unit tests asserting `local_only_table_names(false)` **is** `LOCAL_ONLY_TABLE_NAMES`, that `true` drops `schema_cache` and nothing else, and that the two consts cannot drift apart.

## Honest failure

A remote state object written before anyone opted in has no `schema_cache` table at all. Turning the key on must not break that download, and it does not: `copy_named_tables` already treats a table absent from the source as a legitimate no-op, and `StateStore::open` eagerly creates every expected table, so the next open materialises it empty. Asserted end to end by `replicating_download_tolerates_a_remote_with_no_schema_cache_table`.

## Gates

| Gate | Result |
|---|---|
| `cargo test -p rocky-core -p rocky-cli` | **3857 passed, 0 failed** (plus the 3 tests added after review) |
| `cargo fmt --all --check` | pass |
| `cargo build --workspace --all-targets` | clean |

**No codegen.** `rocky-core/src/config.rs` is untouched — the config *field* already existed and already had a schema entry. Nothing generated changes, so there is nothing to commit under `schemas/`, the Pydantic models, the TypeScript bindings or `openapi.json`.

**No docs change either**, and that is the point: `docs/src/content/docs/reference/configuration.md:923` already promised this behaviour. The code now matches the documentation, rather than the documentation being walked back.

## Out of scope

N2 (`[pipeline.<name>.checks.freshness] overrides`) and N3 (`[schedule] poll_interval_seconds`) from the same issue are untouched and remain open. `Refs #1620`, not `Closes`.

## Standing questions

**Least confident:** whether `replicate = true` is safe on a *tiered* backend under concurrent writers. The end-of-run upload is last-writer-wins for the whole blob (`interleaved_writers_last_writer_wins_documented` pins that), so two pods replicating caches can clobber each other's entries. That is pre-existing for every replicated table and not made worse here — but the schema cache is the first table a user opts *into* replicating, so they meet the property deliberately rather than by inheritance.

**Most important thing possibly missing:** whether a replicated schema cache should carry a machine or warehouse identity. Today the cache key is `catalog.schema.table` with no notion of which warehouse produced it. Two pods pointed at different warehouses but sharing one `[state]` prefix would now share types across them. That is a real configuration, and the key shape is what would have to change — so it wants its own decision rather than a caveat here.

Refs #1620


hugocorreia90:
## Red-team disposition — Codex (`gpt-5.4-codex`, read-only, high reasoning)

One round, full verdict: **BLOCK**, with two findings. One accepted and fixed, one rebutted with evidence.

### Finding 1 — `[TEST-GAP]` the four-leg invariant lacked per-leg coverage. **Accepted, fixed.**

> The new cross-pod tests exercise unconditional upload and download, but existing periodic-uploader and CAS tests pass `false`. A mutation hard-coding the old table set specifically at `state_sync.rs:1207` or `state_sync.rs:2766` would escape.

Correct, and it is exactly the failure this PR argues against — wiring some legs and not others. The first round of tests drove `upload_state` and `publish_merged` only, so the two other upload paths were wired but unproven.

Three tests added (`a6132559`), then verified by two narrower mutations, each disabling the posture at one leg:

|---|---|
| `upload_state_cas` hard-codes the old set | `schema_cache_crosses_pods_through_the_cas_upload_leg` — FAILED |
| the periodic uploader hard-codes the old set | `schema_cache_crosses_pods_through_the_periodic_uploader` — FAILED |

Each is caught by exactly one test, and only that one.

### Finding 2 — `[CORRECTNESS]` public source-compatibility break. **Rebutted.**

> `LedgerSeamSession::new`, `RemoteStateSession::new`, `download_state`, … gain required parameters. Existing downstream callers will not compile. … This finding falls away only if `rocky-core` is explicitly private and carries no supported API contract.

That condition holds. `rocky-core` is workspace-internal and never published: there is no `cargo publish` in any workflow, and the four release artifacts are the engine **binary**, the `rocky-sdk` and `dagster-rocky` **wheels**, and the VS Code **VSIX** — no Rust crate. Every caller is in this repo, and the workspace compiles with 3857 tests passing. The reviewer flagged this as its own least-confident point and named the evidence that would settle it; this is that evidence.

The suggested fix would also be the wrong shape here. Keeping the old signatures as wrappers that default to `false` means a future caller can pick the old name and silently not replicate — the same silent-default defect this PR exists to remove. The parameter is required on purpose.

### Nit — accepted

> `state.rs:301` says the duplicate constants prevent drift; the tests prevent drift, while duplication itself permits it.

Right, and the comment now says so.

### Answers it confirmed

Byte-identical at `replicate = false` on every leg (a); an absent remote `schema_cache` table does not fail the download (b); no other table's sync behaviour changes, the second const is the first minus `schema_cache` only (c); no mechanical `false` landed in a wrong argument position and every production caller reads `cfg.cache.schemas.replicate` (d); the periodic uploader cannot disagree with the terminal leg, both read one stored posture (e).

I re-verified (a) and (d) independently before the review: exactly one production use of the raw `LOCAL_ONLY_TABLE_NAMES` constant remains in `state_sync.rs`, the deliberate `Absent` arm.

**Note:** the reviewer did not run a further independent pass, because the prompt forbade delegation — a previous run of this task deadlocked waiting on a sub-reviewer it had spawned. So this is one independent model, not two.

hugocorreia90:
**Merge note.** This PR and #1667 conflict on one file only: `engine/CHANGELOG.md`. Both add a bullet as the first entry under `## [Unreleased]` → `### Fixed`, so whichever merges second needs a trivial resolution that keeps **both** bullets. Nothing else overlaps — verified with `git merge-tree --write-tree`, which reports `engine/CHANGELOG.md` as the sole conflicted path.

hugocorreia90:
Orchestrator review. Checked against the diff and the branch, not the description.

**Verified**
- Every sync leg reads `cfg.cache.schemas.replicate` (apply, fulfill_api, gc, load, policy, run) instead of the constant; the trybuild fixture `discarded_authority.stderr` was regenerated for the new arity.
- The two sites that fold an `Option` into `replicate = false` (`apply.rs` backfill, `gc.rs` apply) only ever see `None` for a *missing* `rocky.toml`. Both loaders map `ConfigError::FileNotFound` to `None` and refuse every other error, so a config that fails to load cannot silently under-replicate an opted-in CAS project. That was the one thing I went looking for.
- `publish_merged`'s `Absent` arm keeps the full local set, and a test pins it; using the replicating set there would delete the warm local cache when the remote object is missing.

**Owner-merge** because a project that already set `replicate = true` starts uploading its schema cache on upgrade. Merging when the updated branch is green, after #1672. Conflicts with #1667 on `engine/CHANGELOG.md` only; I resolve that on whichever lands second.

PR #1678, opened 2026-09-05
fix(engine/rocky-cli): evaluate column_match against the target after the copy

hugocorreia90:
`column_match` on a replication pipeline failed on the **first run of every table** and passed on the second. Both column sets are now read **after** the copy, so the run that creates a table can also certify it.

The second half of #1666 — `severity` being ignored on the replication path — is deliberately **not** in this PR and stays open. See the last section.

## What was wrong

`process_table` resolved `target_cols` from the pre-copy probe, the same read that decides `target_exists`. On a first run that list is empty:

```
  describe target  ->  absent  ->  target_cols = []
  copy the table   ->  target created from source
  column_match     ->  source vs []  ->  every column "missing"  ->  FAIL
```

500 rows copied, and the check reported every column absent.

## What changed

`engine/crates/rocky-cli/src/commands/run.rs` — after the copy statement succeeds, the check probes **both** sides concurrently and compares those:

```
  before:  source_cols (pre-copy, maybe a batch pre-fetch)  vs  target_cols (pre-copy)
  after:   describe(source)  ||  describe(target)      both after the copy
```

`source_cols` and `target_cols` are untouched. Every other consumer still sees the pre-copy read: `target_exists`, `drift::detect_drift`, `use_full_refresh`, the `full_refresh_needs_predrop` pre-drop, and `resolve_merge_update_columns`.

Three false failures close with it, all real:

- **First run.** The target did not exist at probe time, so every column read as missing.
- **Drift repair.** A column added to the source and ALTERed into the target *this run* read as missing, because the pre-copy read predates the ALTER.
- **Stale pre-fetch.** Under the batch schema pre-fetch, `source_cols` is a snapshot taken before the table was scheduled. A source `ADD COLUMN` landing in that window made the copy's own `SELECT *` output read as an `extra` column.

A target that ends the run genuinely short of a column still fails — and is now caught even when the pre-copy read looked right.

A probe that will not answer (an error, or zero columns after a successful copy) records the reason in `not_evaluated` and fails, rather than comparing against a list that was never read. It never fails the table: the copy has already succeeded, and on `incremental` a retry would insert the same rows twice. A probe refused with a rate limit (`429`, `UC_REQUEST_LIMIT_EXCEEDED`) is carried out on `TableResult::probe_rate_limited` so the collector calls `on_rate_limit()` instead of `on_success()` — swallowing that error must not read to the adaptive throttle as "the warehouse is happy". `run()` has **two** collectors for a materialized table (`process_completed_result`, drained interleaved while adaptive concurrency is on, and an inline arm for the remainder); both carry the branch.

`engine/crates/rocky-core/src/checks.rs` — new `column_match_not_evaluated(reason)`, matching the `row_count_not_evaluated` / `freshness_not_evaluated` family from #1602.

Cost: two `DESCRIBE` calls per table, only when `column_match` is on (it defaults to off).

**On exit codes.** No counter, status or exit-code logic was changed, and a failed replication check still does not fail `rocky run` by itself (#1598). One consumer does read the result and is worth naming: a `[[policy.rules]]` entry with `verify_after = ["column_match"]` gates a run that auto-applied additive drift, and treats a failed check as a failed verification. That gate **could not pass before this change** — the pre-copy read always reported the just-ALTERed column as missing — and now passes on the ordinary path. It still fails, by design, when a probe cannot answer.

### 1. Fail-before / pass-after tests

Eight new tests in `engine/crates/rocky-cli/src/commands/run.rs`, driving the real `process_table` against an in-memory DuckDB.

| Test | Shape | Asserts |
|---|---|---|
| `column_match_passes_on_the_first_run_that_creates_the_target` | absent target | `passed`, empty `missing`/`extra` |
| `column_match_excludes_rocky_metadata_columns_on_a_first_run` | absent target + `_rocky_loaded_at` | the metadata column is written, and excluded |
| `column_match_sees_the_column_drift_repair_added_this_run` | existing target, source gained a column | `passed` |
| `column_match_survives_a_source_ddl_after_the_prefetch_snapshot` | both pre-fetches stale, live source gained `region` | `passed` |
| `column_match_still_fails_when_the_post_copy_target_differs` | existing target, a `DROP COLUMN` after the copy | `!passed`, `missing == ["region"]` |
| `column_match_records_a_reason_when_the_post_copy_probe_fails` | probe errors after the copy | table materializes, `not_evaluated` set, `!passed`, **not** rate-limited |
| `column_match_records_a_reason_when_the_source_probe_returns_nothing` | source probe fails | `not_evaluated` naming the source, `!passed`, both lists empty |
| `a_rate_limited_post_copy_probe_still_reaches_the_throttle` | `429` on the post-copy probe, then through `process_completed_result` | table materializes, `probe_rate_limited`, throttle backed off (`rate_limits_total() == 1`, `current() < 8`), `tables_copied == 1`, no table error |

The control matters: `column_match_still_fails_when_the_post_copy_target_differs` fails on `main` for the *opposite* reason — `main` reads the stale pre-copy target and reports `passed = true` on a target that really is a column short.

### 2. Mutation checks

Two runs of `scripts/mutation-check.sh` on the committed tree, because the two defects need different mutations.

**Mutation A** — restore `main`'s pre-copy comparison, keeping the new plumbing so it still builds:

```
mutation-check: applied mutation to engine/crates/rocky-cli/src/commands/run.rs
test result: FAILED. 0 passed; 7 failed; 0 ignored; 0 measured; 1633 filtered out
```

Exit 0, not the known-false exit 3 (#1547). Discriminating values from the log, so the mutation provably applied:

```
column_match_passes_on_the_first_run_that_creates_the_target
  got missing=["event_id", "event_type", "user_id"] extra=[]
column_match_sees_the_column_drift_repair_added_this_run
  got missing=["region"] extra=[]
column_match_still_fails_when_the_post_copy_target_differs
  a target that really is a column short must fail the check
```

Mutation A deliberately does **not** kill `column_match_survives_a_source_ddl_after_the_prefetch_snapshot`: `main` compares two equally-stale snapshots, so it passes there too. That test needs its own mutation.

**Mutation B** — keep the post-copy target read, feed the pre-copy `source_cols` back in (the shape the first commit had):

```
a column the copy itself wrote must not read as `extra` against a stale snapshot;
  got missing=[] extra=["region"]
```

**Mutation C** — drop the rate-limit branch from `process_completed_result` only, leaving the inline collector intact (exactly the half the red team found missing):

```
assertion `left == right` failed: the collector must report the throttled probe as a rate limit
```

All three were re-run against the final commit. Each script asserts its target is present exactly once and that the replaced code is gone afterwards (C additionally asserts exactly one rate-limit branch remains), so a silently-unapplied mutation cannot pass as a verdict.

### 3. POC

`examples/playground/pocs/01-quality/02-inline-checks`, run twice against a fresh DuckDB file with a debug binary.

**Before** — run 1:

```json
{"name": "column_match", "passed": false,
 "missing": ["event_id", "user_id", "occurred_at", "event_type"], "extra": []}
{"name": "row_count", "passed": true, "source_count": 500, "target_count": 500}
```

**After** — run 1:

```json
{"name": "column_match", "passed": true, "missing": [], "extra": []}
{"name": "row_count", "passed": true, "source_count": 500, "target_count": 500}
```

Run 2 passes in both, unchanged. Run metrics (`statements_executed`, retries, anomalies) are identical before and after; only wall-clock timings differ. Provenance note: the "before" binary was built in this worktree before the change and its base commit is not pinned, so the authoritative fail-before evidence is the mutation checks above, which revert these exact hunks on the committed tree.

Healthy control — `examples/playground/pocs/00-foundations/01-replication-basics` (which sets `column_match = false`) run once before and after. Status, `tables_copied`, `tables_failed`, `check_results`, `drift`, `permissions`, non-timing metrics, strategies and target names are **identical**.

### 4. Changelog

`engine/CHANGELOG.md`, under `## [Unreleased]` → `### Fixed`, asserted to sit inside the Unreleased section (the entry precedes the next `## [` heading). Ends with `(#1666)`.

### 5. No codegen

No `*Output` struct was touched. `not_evaluated` already exists on `CheckResult` (#1602), and `TableResult` / `probe_rate_limited` are private to `run.rs`. No `JsonSchema` type or field changed, so `just codegen` and `just regen-fixtures` were correctly not run — CI's `Check codegen is up to date` agrees. `git diff --stat origin/main...HEAD` is three files:

```
 engine/CHANGELOG.md                         |  10 +
 engine/crates/rocky-cli/src/commands/run.rs | 867 +++++++++++++++++++++++++++-
 engine/crates/rocky-core/src/checks.rs      |  18 +
```

### 6. Local gate

- `cargo +1.98.0 fmt --all --check` — clean
- `cargo +1.98.0 clippy -p rocky-cli --all-targets -- -D warnings` — clean
- `cargo +1.98.0 test -p rocky-cli --lib -- column_match a_rate_limited_post_copy_probe` — 8 passed
- `cargo +1.98.0 test -p rocky-cli --lib commands::run::tests` — 268 passed, 1 ignored (the whole module the change lives in)
- `cargo +1.98.0 test -p rocky-core --lib checks` — 45 passed

## Red team

Reviewer: **Codex** (`codex exec`, `model_reasoning_effort=high`), which ran its own independent second model (`gpt-5.6-terra`) inside round 1. Every finding was verified against the code before acting.

### Round 1 — verdict: BLOCK. 3 correctness findings + 1 test gap.

| # | Finding | Disposition |
|---|---|---|
| 1 | `[CORRECTNESS]` The check mixed a pre-copy **source** snapshot with a post-copy **target** snapshot. Under the batch pre-fetch, `source_cols` is captured before the task is scheduled, so a source `ADD COLUMN` in that window makes the copy's own `SELECT *` output read as `extra`. | **Fixed.** Both sides are now probed together after the copy (`tokio::join!` → `post_copy_column_match`). The pre-copy `source_cols.is_empty()` guard is gone, subsumed. New test + mutation B. |
| 2 | `[CORRECTNESS]` "Exit codes unchanged" was false for a configured `verify_after`. `finalize_drift_verify_after` reads the check's `passed`; a failure increments `tables_failed` and re-persists the run as a Failure. | **Confirmed; claim narrowed, code unchanged** — which is what the reviewer asked ("the policy behavior is intentionally fail-closed, so do not make the check pass"). The changelog and the section above now name that consumer, and note the gate **could not pass before this change**. |
| 3 | `[CORRECTNESS]` A rate-limited post-copy probe was swallowed, so the collector still called `on_success()` and the AIMD throttle could widen the semaphore while the warehouse throttled. | **Fixed.** `TableResult::probe_rate_limited`, set through the collector's own `is_rate_limit_error` predicate, drives `on_rate_limit()`. New test + mutation C. |
| 4 | `[TEST-GAP]` The six tests could not detect any of the three. | **Closed for the two that changed code** (the stale-pre-fetch test and the throttle test). Not closed for finding 2: there is still no full-run test through `run()` combining auto-applied additive drift, `verify_after`, and a failing probe. `finalize_drift_verify_after` has its own tests (`passing_verify_after_records_allow_and_succeeds`, `failing_verify_after_halts_and_records_deny`) and its semantics did not change. |

### Round 2 — incomplete. The reviewer hit its usage limit mid-verdict.

It delivered one partial conclusion before stopping: *"the post-copy comparison is materially improved, but the rate-limit fix only covers one of two reachable collector paths."*

Verified and **fixed**. `run()` has two collectors for a materialized table: `process_completed_result`, which it drains through while adaptive concurrency is on, and a near-identical inline arm for the remainder. Only the inline arm had the new branch — so the throttled-probe signal was dropped on exactly the path adaptive concurrency uses. Both arms now carry it, and the rate-limit test drives its outcome through `process_completed_result` rather than stopping at the carrier field.

The same round's file reads surfaced a second, smaller defect: the new helper had been inserted **between** `is_rate_limit_error`'s doc block and its `fn`, so that block documented the wrong function. Fixed; each doc is back on its own item.

### Round 3 — not available.

Codex credits are exhausted (`You've hit your usage limit … try again at Sep 7th`). **No independent reviewer has seen the third commit** (`04dcedb0`) — the second collector, the doc un-splice, and the extended throttle test. That is the one open gap in this PR's review. Its evidence is mutation check C plus the local gate, not a second pair of eyes.

### One question the reviewer left open, answered

*"Whether maintainers define concurrent source DDL as an intentional unstable-schema failure rather than a false `column_match`."* This PR treats it as a false failure, because that is what #1666 is about: the check exists to say whether the copy produced the expected columns, and a column the copy itself wrote is not a mismatch. The two reads are now taken microseconds apart and both after the copy, instead of one from before the table was scheduled. A source DDL landing inside that remaining gap would skew them the other way — the source read would be newer than the copy, so the new column reports as `missing`, and the next run heals it — but the window is a pair of concurrent `DESCRIBE` calls rather than the whole pre-fetch-to-copy interval, which can be minutes across a large table set.

## Left open in #1666

`AggregateCheckToggle::severity()` is read in exactly one place, `engine/crates/rocky-cli/src/commands/run_local.rs` (the quality runner). The replication runner never reads it, and `check_column_match` / `check_row_count` hard-code `TestSeverity::Error`. So on a replication pipeline this parses and is silently ignored:

```toml
[pipeline.x.checks.column_match]
enabled  = true
severity = "warning"
```

That is a behaviour change with exit-code implications — it decides whether a check failure can gate a run — so it belongs in its own owner-reviewed change, not bundled with a false-failure fix. #1666 stays open for it.

Refs #1666


hugocorreia90:
Red-team disposition, with the line each finding landed on (at `04dcedb0`).

| Round | Finding | Disposition | Where |
|---|---|---|---|
| 1 | Mixed pre-copy source / post-copy target snapshots | **fixed** | `engine/crates/rocky-cli/src/commands/run.rs:12372` (both probes, `tokio::join!`), helper at `run.rs:12805` |
| 1 | "Exit codes unchanged" false under `verify_after` | **claim narrowed, code unchanged** (the reviewer asked not to make the check pass) | `engine/CHANGELOG.md`, Unreleased → Fixed, final paragraph |
| 1 | Rate-limited probe read as `on_success()` | **fixed** | field `run.rs:84`, set at `run.rs:12372`, consumed at `run.rs:4274` and `run.rs:12610` |
| 1 | Test gap on all three | **closed for the two that changed code**; open for the `verify_after` tail, whose own gate is unchanged and separately tested | `drift_governance.rs` — `passing_verify_after_records_allow_and_succeeds`, `failing_verify_after_halts_and_records_deny` |
| 2 | Rate-limit fix reached only one of two collectors | **fixed** — `process_completed_result` is the one `run()` drains through while adaptive concurrency is on | `run.rs:12610` (was missing), `run.rs:4274` (already had it) |
| 2 | Doc block detached from `is_rate_limit_error` | **fixed** | `run.rs:12781` |

New helper: `column_match_not_evaluated` at `engine/crates/rocky-core/src/checks.rs:416`.

Round 3 was **not run**: the reviewer's credits are exhausted. No independent reviewer has seen the third commit (`04dcedb0`) — the second collector, the doc fix, and the extended throttle test. Its evidence is mutation check C plus the local gate.

hugocorreia90:
Post-merge review done (Claude Opus 5 at max effort, at the owner's direction — Codex is out of credits, so this is not the independent pass `AGENTS.md` asks for; it is a different session and prompt, same family). Verdict BLOCK. I verified every structural finding against `main` df506990 before filing.

**The fix itself is right.** `column_match` is evaluated after the copy, both probes are taken together, and `source_cols` / `target_cols` are byte-identical inputs for `target_exists`, drift, the pre-drop and merge-column resolution — the review found no consumer whose input changed. Every `on_success()` site is accounted for: two pruned arms, two error arms already gated, and the two patched arms. There is no fourth consumer.

**What it exposed, filed as #1718 — the more serious of the two, and it predates this PR:** `run()` has *three* collectors for a materialized table, not two, and two of them drop check inputs. `process_completed_result` never pushes `assertion_targets`, which is the only input to the assertion, null-rate, custom-check and cross-source-overlap loops — and that collector is the drain the spawn loop uses inline whenever a throttle is present, which is the default (`ConcurrencyMode::Adaptive` is `#[default]`). The auto-retry arm pushes `assertion_targets` but never `column_match_check`, `freshness_batch_ref` or the drift hook, and `table_retries` defaults to 1. This PR's body asserted there were two collectors that both carried the branch, and mutation check C was scoped to that enumeration.

**Two changelog corrections needed in the same follow-up.** This entry says "a failed replication check still does not fail `rocky run` by itself (#1598)". That was true when written and false on `main`: #1671 merged 22 minutes later, and `column_match_not_evaluated` is error-severity, so it now exits 2. In the other direction, #1671's own entry still tells the reader to check whether they set `column_match = true` because it "fails on the first run of every table" — which is exactly what this PR fixed. Both entries ship in the same `[Unreleased]` section describing opposite realities.

**Operational finding worth an owner's eye:** a *transient* post-copy `DESCRIBE` failure now exits the run 2 on data that landed correctly, and `--resume` then refuses because every table was copied. There is no retry of the probe, and the pre-copy probe fifty lines up already asks the adapter's own `classify_failure(..).is_retryable()` while the post-copy one only substring-matches four rate-limit phrasings. The knob that would bound this — `severity = "warning"` on a replication check — parses and is silently ignored (still open in #1666). I am folding the probe-retry into the #1718 fix lane.

Issue #1646, opened 2026-09-04
Decimal type strings without a parseable (p,s) become a fabricated Decimal(38,0) in both type normalisers; bare BigQuery NUMERIC fails a correct load


hugocorreia90:
Both type normalisers turn a bare or malformed decimal type string into a concrete `Decimal { precision: 38, scale: 0 }`. That is a made-up type, not an unknown one, so it is compared as if it were true.

- `engine/crates/rocky-core/src/contracts.rs`, `warehouse_type_to_rocky`: any string that starts with `DECIMAL`, `BIGNUMERIC`, `NUMERIC` or `NUMBER` and has no parseable `(p,s)` falls through to `Decimal { 38, 0 }`. `DECIMAL`, `NUMERIC`, `NUMERIC(10)`, `DECIMAL(a,b)` and `NUMBERWANG` all land there.
- `engine/crates/rocky-compiler/src/compile.rs`, `default_type_mapper`: the same shape for `DECIMAL` and `NUMERIC`.

## Where it is live

BigQuery. `describe_table` reads `data_type` from `INFORMATION_SCHEMA.COLUMNS` (`engine/crates/rocky-bigquery/src/dialect.rs`). BigQuery reports a default-precision column as a bare `NUMERIC` (which is `NUMERIC(38, 9)`) or `BIGNUMERIC` (`BIGNUMERIC(76, 38)`), and only spells out `(p, s)` for a parameterized column.

So on `rocky load` with a contract:

```
landed   NUMERIC        -> Decimal(38, 0)     (real type: 38, 9)
declared NUMERIC(38,9)  -> Decimal(38, 9)
is_assignable(Decimal(38,0), Decimal(38,9))
  scale   0 <= 9        ok
  integer 38 <= 29      NO   -> required_column_type violation -> staging dropped, load refused
```

A correct load is refused, and the message tells the user the column "has type NUMERIC (DECIMAL(38,0))", which is not what the warehouse has. The reverse also holds: a contract written `NUMERIC(38,0)` passes a landed bare `NUMERIC` that can hold nine decimal places.

DuckDB, Databricks and Snowflake always print `(p,s)` in their `DESCRIBE` output (DuckDB `DECIMAL(18,3)`, Databricks `decimal(10,0)`, Snowflake `NUMBER(38,0)`), so they are not affected by the bare case. Every warehouse is affected by the malformed-string case, which today reads as a valid `(38,0)`.

## What a fix looks like

Two parts, because the two cases differ:

[...]

PR #1682, opened 2026-09-05
OWNER MERGE — fix(engine/rocky-core): map bare warehouse decimal names to their real defaults, and unknown ones to Unknown

hugocorreia90:
> **Owner merge.** One compile that passes today can fail: a contract declaring `Decimal(38,0)` over a source column the warehouse reports as `DECIMAL(10)`, which now means precision 10. No load that promotes today is refused.

A bare `DECIMAL`, `NUMERIC` or `BIGNUMERIC` is no longer read as `DECIMAL(38,0)`. That was a made-up type, compared as if the warehouse had reported it. On BigQuery it refused a correct load.

## The defect

BigQuery's `INFORMATION_SCHEMA.COLUMNS.data_type` reports a default-precision column as a bare `NUMERIC`. Its real type is `NUMERIC(38, 9)`.

```
  landed   'NUMERIC'        ──► Decimal(38, 0)   ← made up; the column holds (38, 9)
  declared 'NUMERIC(38,9)'  ──► Decimal(38, 9)

  is_assignable(Decimal(38,0), Decimal(38,9))
      scale    0 <= 9      ok
      integer 38 <= 29     NO  ──► required_column_type violation
                                   ──► staging dropped, a correct load refused
```

The message then told the user the column "has type NUMERIC (DECIMAL(38,0))", which the warehouse never said. The reverse also held: a contract written `NUMERIC(38,0)` accepted a landed bare `NUMERIC` that holds nine decimal places.

## What changed

| type string | `warehouse_type_to_rocky` (load gate) | `default_type_mapper` (compile gate) |
|---|---|---|
| `DECIMAL(10,2)`, `decimal(10,2)` | `Decimal(10,2)` — unchanged | `Decimal(10,2)` — unchanged |
| `DECIMAL(10)` | `Decimal(10,0)` — unchanged (#1643) | **`Decimal(10,0)`** — was `Decimal(38,0)` |
| `DECIMAL`, `NUMERIC`, `BIGNUMERIC` | **`Unknown`** — was `Decimal(38,0)` | **`Unknown`** — was `Decimal(38,0)` |
| `NUMBER` | `Decimal(38,0)` — unchanged | `Unknown` — unchanged |
| `DECIMAL(a,b)`, `NUMERICWANG` | `Unknown` — unchanged (#1643) | **`Unknown`** — was `Decimal(38,0)` |
| `NUMBER(p,s)`, `BIGNUMERIC(p,s)` | `Decimal(p,s)` — unchanged | `Unknown` — unchanged |

`Unknown` is not a refusal. #1643 made `validate_contract_typed` branch on it before the matcher: the column goes to `ContractResult.warnings`, `passed` is unchanged, and the load promotes. The compile gate reports `I003` — "not checked" (#1240).

## I did not do one half of what the issue asks, on purpose

The issue asks for bare `NUMERIC` → `Decimal(38,9)` and bare `BIGNUMERIC` → `Decimal(76,38)`, BigQuery's real defaults. I did not map them. Two reasons.

**`warehouse_type_to_rocky` has no dialect.** `validate_contract_typed` calls it on the landed type *and* on `req.data_type`, the type a person hand-writes in a `.contract.toml` (`engine/crates/rocky-core/src/contracts.rs:234-235`). One reading has to serve every warehouse. `NUMERIC` is not a BigQuery-only name: `engine/crates/rocky-snowflake/src/types.rs:24-26` groups `NUMERIC` with `NUMBER` and `DECIMAL` in one Snowflake numeric family, and it is an alias for `DECIMAL` on DuckDB and Databricks, with different digits on each. Reading `(38,9)` globally would invent a type from a name — the defect this issue closes — and would refuse a correct load on a warehouse that is not BigQuery.

**The numbers are not in this repository.** I was asked to ground them here and not take them from memory. `NUMERIC` precision 38 is grounded (`engine/crates/rocky-bigquery/src/dialect.rs:388-389`). Scale 9 and `BIGNUMERIC(76,38)` are nowhere in the tree; the only `76,38` is a test input inside the file being changed, so citing it would be circular.

Bare `NUMBER` survives as the one grounded, unambiguous case, and its reading is unchanged from `main`:

- `engine/crates/rocky-snowflake/src/adapter.rs:219-220` — `INTEGER` is documented as an "alias for `NUMBER(38, 0)`"
- `engine/crates/rocky-snowflake/src/loader.rs:140` — `"BIGINT" => "NUMBER(38,0)"`
- no other adapter Rocky ships has a `NUMBER` type

**The live BigQuery defect is fixed either way.** A landed bare `NUMERIC` now reaches the `Unknown` branch: the column is named in `ContractResult.warnings` and the load promotes, instead of being refused for a type the warehouse never reported.

Reading `NUMERIC` correctly per warehouse needs a dialect-aware normaliser. That is a larger change than this one and I have not built it — I will file it separately, `Refs #1646`.

## Compile and load still disagree on two names

`default_type_mapper` does not know `NUMBER` or `BIGNUMERIC`; `warehouse_type_to_rocky` does. That gap predates this PR — it is the last row of the table the #1643 reviewer posted. I did not close it here. Adding those names would give a Snowflake source column a concrete type where it has none today, turning an informational `I003` into a hard `E011` and newly failing compiles for a reason unrelated to #1646.

`test_two_normalisers_agree_on_the_decimal_family` now pins both halves: the strings both mappers claim must agree, and the load-only names are asserted `Unknown` in the compiler, so adding one becomes a deliberate act. The "Mirrors the compiler's `default_type_mapper`" sentence at the top of `warehouse_type_to_rocky` is corrected — the #1643 reviewer asked for that.

### Fail-before, pass-after

Both by `scripts/mutation-check.sh` on the committed tree (it refuses a dirty one), reverting the bare-name fix in each file. Both **exit 0** — the tests failed under mutation, so they are fix-sensitive. Neither hit the exit-3 "INCONCLUSIVE" path; each log shows a real `test result: FAILED`.

`engine/crates/rocky-core/src/contracts.rs` — `cargo test -p rocky-core --lib contracts::`

```
test result: FAILED. 30 passed; 3 failed
    contracts::tests::test_typed_bare_numeric_is_reported_not_refused
    contracts::tests::test_typed_bare_numeric_against_narrow_contract_is_reported
    contracts::tests::test_warehouse_type_to_rocky_decimal_family_grammar
```

`engine/crates/rocky-compiler/src/compile.rs` — `cargo test -p rocky-compiler --lib compile::tests`

```
test result: FAILED. 19 passed; 2 failed
    compile::tests::test_default_type_mapper_decimal_family_grammar
    compile::tests::test_two_normalisers_agree_on_the_decimal_family
```

The two BigQuery cases from the issue are end-to-end through `validate_contract_typed`: a landed bare `NUMERIC` against a `NUMERIC(38,9)` contract now promotes with one warning, and against a `NUMERIC(38,0)` contract it is reported rather than accepted in silence.

### The owner-merge case is asserted, not described

`precision_only_source_decimal_no_longer_satisfies_a_38_digit_contract` (`engine/crates/rocky-compiler/src/contracts.rs`) drives `default_type_mapper("DECIMAL(10)")` into `validate_contract` against a contract declaring `Decimal(38,0)` and asserts exactly one `E011` at `Severity::Error`. The same contract over a source column that really is 38 digits still passes, so the reading is narrower, not broken. `bare_source_decimal_reports_i003_not_a_fabricated_match` pins the other direction: bare `DECIMAL` and `NUMERIC` report `I003` and never an error.

### Honest failure: the new `Unknown` does not fire on a healthy run

`test_healthy_describe_decimals_stay_concrete` (rocky-core) and `test_default_type_mapper_healthy_describe_decimals_stay_concrete` (rocky-compiler) assert the exact strings the adapters emit today still map to concrete decimals.

| string | where it comes from |
|---|---|
| `DECIMAL(10,2)`, `DECIMAL(18,3)` | `engine/crates/rocky-duckdb/src/adapter.rs:420`, `src/seed.rs:276` |
| `decimal(10,2)` (lowercase) | `engine/crates/rocky-databricks/src/adapter.rs:593,604`, `src/unity_catalog_client.rs:1056` |
| `NUMBER(38,0)`, `NUMBER(19, 0)` | `engine/crates/rocky-snowflake/src/loader.rs:140`, `src/dialect.rs:372,879` |
| `NUMERIC(38,9)`, `BIGNUMERIC(76,38)` | BigQuery's parameterized spellings |

That BigQuery reports the bare name is itself grounded in the repo: the live sweep at `engine/crates/rocky-bigquery/tests/dialect_sweep_live.rs:338-377` asserts `data_type` reads exactly `"NUMERIC"` after an `ALTER … SET DATA TYPE NUMERIC`.

### Local gate

- `cargo +1.98.0 fmt --all -- --check` — clean
- `cargo +1.98.0 clippy -p rocky-core -p rocky-compiler --all-targets -- -D warnings` — clean
- `cargo +1.98.0 test -p rocky-core -p rocky-compiler --lib` — **2556 passed, 0 failed**

### Codegen

No `*Output` struct changed and no `JsonSchema` description text changed. `RequiredColumn.type`'s doc comment — the one #1643 regenerated — is untouched and still accurate. `just codegen` / `just regen-fixtures` are not required, and there is no `editors/vscode/package-lock.json` churn.

### Changelog

`engine/CHANGELOG.md`, inside `## [Unreleased]` → `### Fixed`, with an **On upgrade** paragraph. The #1614 entry lower in the same section is amended: it said the bare names still read as `DECIMAL(38,0)`, and that dropping a contract to a bare `DECIMAL` matches any type. This change makes both false.

Refs #1646


hugocorreia90:
Orchestrator review. The lane that wrote this PR was cut off by an API limit before it could post its review record, so here it is.

**Evidence that exists**
- Mutation checks re-run on the final head `41dea8eb`: rocky-core mutation → `30 passed; 3 failed` (fix-sensitive), rocky-compiler mutation → `465 passed; 3 failed` (fix-sensitive). Both real verdicts, not the #1547 exit-3 shape.
- Reviewer: a Claude Sonnet pass — **same model family as the author, so not the independent red team `AGENTS.md` requires.** It traced every call site, ran the 24 decimal tests (all pass), cleared the regression hunt (`Unknown` structurally never reaches `violations`, so no load that promotes today is refused), cleared honest-failure for DuckDB / Databricks SQL / Snowflake / Trino DESCRIBE strings, and raised one nit — the DuckDB `(18,3)` / Databricks `(10,0)` digits were ungrounded — which `41dea8eb` removed.
- **Codex, the repository's independent red team, is out of credits until Sep 7 10:07.** Two attempts (the lane's and mine) returned only the usage-limit error. No independent verdict exists for this PR. Saying so per the rule rather than skipping it.

**The deviation from #1646, and why I accept it**
The issue asked for bare BigQuery `NUMERIC` / `BIGNUMERIC` to map to their real defaults. The PR maps them, and bare `DECIMAL`, to `Unknown`, which the load gate reports (#1614) instead of comparing; only bare `NUMBER` keeps a concrete default, `(38,0)`, grounded in the Snowflake adapter. Reason: `warehouse_type_to_rocky` takes no dialect and is fed both DESCRIBE strings and hand-written contract types, and the BigQuery digits appear nowhere in this repository — hard-coding them there would be the same fabrication this fix removes. Net effect on the issue's scenario: the correct load that is refused today promotes with a named warning; the reverse case (bare `NUMERIC` landed against a `NUMERIC(38,0)` contract) is reported rather than refused. Weaker than the issue asked, but honest, and `test_two_normalisers_agree_on_the_decimal_family` pins the two mappers together.

**Open question the reviewer could not settle:** whether Unity Catalog's REST `describe_table` ever omits `type_text` in production (`unity_catalog_client.rs:474-497` falls back to a bare `type_name`); if it does, that column becomes `Unknown` (reported) instead of the old fabricated `Decimal(38,0)`.

Owner-merge title stands for the one compile that newly fails (`DECIMAL(10)` against a 38-digit contract). Merging when green under the owner's "merge all when green" instruction; flagged for a Codex pass when credits return.

hugocorreia90:
Post-merge review done (Claude Opus 5 at max effort, at the owner's direction — Codex is out of credits, so this is not the independent pass `AGENTS.md` asks for). Verdict BLOCK, filed as #1721. I verified each mechanism on `main` df506990 before filing.

**What held up:** "no load that promotes today is refused" is true — the `Unknown` branch pushes to `warnings` and never to `violations`, and presence and nullability are untouched, so there is no path from `warehouse_type_to_rocky` to a newly-firing violation. The two string mappers agree on every construction the review could build (case, inner whitespace, `+10`, negative scale, `DECIMAL(300,0)`, a `NOT NULL` suffix, a trailing comment, `BIGNUMERIC` against a `NUMERIC` prefix). Incremental recompute is clean: the schema cache persists the raw warehouse string, not a derived type, so no stale `Decimal(38,0)` survives the upgrade. Codegen genuinely was not required. The Snowflake grounding citations are exact.

**What it missed is the direction of harm.** The self-review reasons about whether a *reading* is invented and never about which *gate outcome* changes. The shape it does not enumerate: the fabricated `(38,0)` was also refusing loads that genuinely did not fit. A BigQuery bare `NUMERIC` holding `(38,9)` against a `NUMERIC(10,2)` contract was refused; it now promotes with a warning. On the compile side the same column moved `E011` → `I003`, and `dagster_rocky/contracts.py:88` maps `I003` into the contract check as `passed=True`, so a red asset check goes green. `rocky load` has no strict mode. The changelog documents the two false-refusal shapes this fixed and the one compile it breaks, and not this third direction.

**Two mechanisms worth reading in the issue:** the `I003` suggestion tells the user a CAST only helps when the input type is known, but the CAST arm takes its type purely from the target — so following the advice re-creates the exact `(38,0)` this PR removed, on a column that holds `(38,9)`. And there is a *third* normaliser: `sql_type_to_rocky`'s `ExactNumberInfo::None` arm still returns `Decimal(38,0)`, and `default_type_mapper` is that function's `_ =>` fallback, so one typecheck pass now reads one column two ways.

**Also resolved: the Sonnet reviewer's open question is a yes.** `unity_catalog_client.rs:488` falls back to `type_name` when `type_text` is absent, `type_name` for a decimal column is the bare `DECIMAL` (Rocky writes that string itself), and the REST path runs *before* the SQL `DESCRIBE` on the load gate's `describe_table`. Whether a live Unity response ever omits `type_text` is the one thing static review cannot settle — one request against the Databricks sandbox would.

The issue lays out three options for what a gate should do with an uncomparable **landed** type. My read: reporting is right for a declared type Rocky does not understand, because that is the user's file; a landed type is the warehouse's own answer, and failing to parse it is not evidence the data conforms. That is a product call, so it is yours.

PR #1688, opened 2026-09-05
chore(ci): allow the pinned download-artifact action in release jobs

hugocorreia90:
The release-step allow-list gains `actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c` (v8.0.1).

## Why

#1687 (`rocky serve --ui`) builds the browser UI once per release and hands `dist/` to every matrix target as an artifact. The matrix reads it with `actions/download-artifact`, which the release-step allow-list did not know, so the containment policy refused `engine-release.yml`.

The policy job runs the checker from `main`, not from the candidate, so the allow-list change cannot ride in #1687 itself; it lands here first. One pinned first-party action, the same family (`actions/upload-artifact`) the list already trusts.

## Checks

- `python3 .github/scripts/check_credential_containment.py .` passes on this branch and on #1687's tree.
- The 110 policy regression tests pass.

## Review

Codex, the independent red team, is out until 2026-09-07. A one-line allow-list addition of a pinned first-party action; disclosed rather than skipped.


hugocorreia90:
The containment policy job fails on this PR by design: the checker is a frozen trust root, and a candidate that changes it reports `frozen trust root differs from trusted base`. That is the rule working, not a violation to fix. The previous allow-list changes (#1265, #1277, #1395) landed the same way: reviewed as a policy change and merged by the repository owner past the check. The other gates on this PR (the 110 regression tests, engine CI) run on the candidate and pass.

Once this is on `main`, the policy job on #1687 re-runs against the new trusted checker and turns green.

PR #1693, opened 2026-09-05
feat(engine): publish a container image per engine release

hugocorreia90:
Every `engine-v*` release now publishes one container image, `ghcr.io/rocky-data/rocky`, for `linux/amd64` and `linux/arm64`. The image is the release, bit for bit: nothing is compiled in the image build.

**Merge order.** The three pinned Docker actions the new `image` job uses must be on the containment allow-list before this merges, so the `Credential containment policy` and `Credential-containment regression` checks on this PR are red until the allow-list PR (policy-only, admin merge, the same shape as #1688) lands and this branch is updated. Every other check is expected green.

## What changes

| Piece | Change |
|---|---|
| `engine/Dockerfile` | Base `gcr.io/distroless/cc-debian12:nonroot`, pinned by digest (glibc, libstdc++, CA certificates; no shell). Copies `<context>/${TARGETARCH}/rocky` to `/usr/local/bin`. `/data` is created nonroot-owned in a `scratch` stage and copied in with `--chown`, because `WORKDIR` alone makes a root-owned directory the server cannot write its state store into. `WORKDIR /data`, `VOLUME /data`, `EXPOSE 8080`, `STOPSIGNAL SIGTERM`, `ENTRYPOINT rocky`, `CMD serve --host 0.0.0.0`. `rocky-lsp` stays out: an editor speaks to it on the host, and leaving it out halves the image. |
| `engine-release.yml` | New `image` job after `build`, before `publish` (`publish` now needs it, so a release never goes live without its image; an image failure keeps the draft unpublished, the same fail-closed rule as `checksums`). Downloads the two Linux archives from the draft, lays out the context, derives the tags from the git tag in a `run` step (`<version>`, `<major>.<minor>`, `latest`; a pre-release gets only its own tag), `setup-buildx` + `login` + `build-push` pinned by sha, `provenance: mode=max` and `sbom: true`, then pulls the pushed image and requires `rocky <version>` from `--version`. `permissions: contents: write, packages: write` on this job only. |
| Docs | New guide, `guides/run-the-image.md`: `docker run`, the mount and Linux ownership, running a pipeline once, the health probe (no `HEALTHCHECK`: no shell, so a kubelet `httpGet` is the in-cluster form), stop and drain, the scheduler rules, SBOM and provenance, upgrade and rollback across a state-schema change, a minikube example labelled an example (`replicas: 1`, `Recreate`, PVC, `fsGroup`), building it yourself. A pointer on the install page. |
| Changelog | The Added entry. |

## Acceptance (brief U5-P1, rocky-plans)

| # | Row | Evidence |
|---|---|---|
| 1 | Builds locally from a release archive; `--version` prints it; uid 65532; `/data` is the working directory | Built for `linux/arm64` from the `engine-v1.73.0` archive. `docker run --rm rocky:u5p1-local --version` → `rocky 1.73.0`. `docker image inspect` → `user=65532 workdir=/data cmd=[serve --host 0.0.0.0] stop=SIGTERM volumes=map[/data:{}] exposed=map[8080/tcp:{}]`, size 40 MB. |
| 2 | The `image` job is wired after `build`, before `publish`, `packages: write` on itself only, only pinned sources, policy green on `main` after the allow-list PR | `actionlint` clean. The checker reports exactly three `release step action source is not allow-listed` violations, one per Docker action, and nothing else; the allow-list PR clears them. |
| 3 | On the next `engine-v*` tag: both platforms, three tags, SBOM and provenance, the smoke printed the version | Waits for the next tag. **Owner action after that first push:** the package `rocky` appears under the org as private; set it public in the package settings once. |
| 4 | The guide's `docker run` starts the server; the health probe answers 200 | On the 1.73.0 image: `rocky playground pg` into a named volume (proves uid 65532 writes `/data`; `ls -ln` shows `65532 65532`), `rocky run --output json` → `status Success, 3 materializations` (DuckDB runs on distroless `cc`), state store at `/data/pg/models/.rocky-state.redb`, `serve` reachable through the published port, `GET /api/v1/health` → 200, `/api/v1/meta` with the bearer, `docker stop` → drain log line, exit 0. `--ui` itself waits for #1687. |

## Out of scope, named

Kubernetes support, a Helm chart, a `cosign` signature (one more trust-root action and a key or OIDC decision for the owner), the compose example (U5-P2), the deployment-contract page (U5-P3).

## The two questions

**Least confident.** The base image for the warehouse connectors. DuckDB runs (row 4). A connector that shells out or loads a plugin at runtime would find no shell and no loader; nothing in the tree does today, and the smoke runs `--version`, not a warehouse. The first user who runs the image against Databricks or Snowflake settles it.

**Most likely missing.** Signing. The attestations say what BuildKit built, not who published. Until a signature exists, the guide tells a puller to pin the digest they inspected.

## Review

Codex, the independent red team this repo requires, hit its usage limit on 2026-09-04; credits return on 2026-09-07. A self-probe stood in: the root-owned `WORKDIR` trap (found and fixed with the `scratch` stage), the loopback-bind trap (the default command carries `--host 0.0.0.0`), the Linux bind-mount ownership trap (documented, two remedies), the missing `HEALTHCHECK` (documented, with the kubelet form), the pre-release `latest` trap (excluded by the tag derivation), and the draft-asset read (the job keeps `contents: write` like every job in the file rather than guessing `read` works on a draft). The owed Codex pass covers this PR when credits return.


hugocorreia90:
Follow-up filed as #1915: the minikube example's `kubectl cp` recipe added on this page cannot work against the image this PR publishes. `kubectl cp` runs `tar` inside the target container and the image is distroless, so the copy fails with `exec: "tar": executable file not found in $PATH`.

This page is not in `engine-v1.73.0`, so the next engine release is the first one where a reader can reach the recipe and hit it. #1915 carries a verified replacement (a throwaway pod mounting the same claim, plus an init-container alternative) and stays open until the page is fixed.

Issue #1416, opened 2026-08-10
perf(engine/serve): resident scheduler still grows ~228 KB/h on Linux/glibc after the #1399 cache cap — second source unattributed


hugocorreia90:
## Summary

A 24h soak of `rocky serve --scheduler` on Linux/glibc, with the request-local history cache capped (#1399 / #1407), still fails the memory gate: **final-quarter RSS slope +228.1 KB/h against a 200 KB/h ceiling** (+335.8 KB/h full-run, +7.77 MB across 24h). Correctness in the same run was perfect. The growth is therefore **not** the request-local `/runs` cache behaviour that #1399 addressed, and a second source is unattributed.

## Evidence

Run: 24h, aarch64 Linux container (glibc, redb 2.6.3), one scheduled run per minute, `GET /api/v1/runs` harvested every 600s, `RUN_HISTORY` growing 0 → 1,440 rows.

```
final quarter  +228.1 KB/h over 6.00h — 720 readings, 0 dropped,
               0 off-grid spacings, worst spacing 31s (grid 30s)
full run       +335.8 KB/h, +7.77 MB / 24h
correctness    ticks 5745, executions 1440, harvested_fires 1440,
               sum_failed 0, child_failures 0, state_busy 0,
               lock_overridden 0, drained 0, max_zombies 0, fd 11 -> 11 (all classes delta 0)
```

The measurement is grid-validated with no sampling gaps, so this is a real slope rather than a disturbed-host artifact (three earlier attempts were correctly rejected as `INVALID` for exactly that reason).

## Why it is not the #1399 path

- The request-local cache cost was measured directly and is now small: at 50,000 rows one `/runs` request retains 4.0 MB with the cap versus 211 MB without.
- A comparison that runs the *wrong way* for the observer hypothesis: an **unfixed** 4h arm polled at the identical 600s cadence measured **+168 KB/h**, which is *lower* than this **fixed** build's +228 KB/h. Durations and history sizes differ, so it is not controlled — but leftover `/runs` cost cannot explain a larger slope on the build where that cost was removed.
- An earlier 4h arm that was never polled on `/runs` at all measured −58 KB/h, but only reached 216 rows — too small and too short to bound the behaviour seen here.

## Suggested next step (a discriminating experiment, not a guess)

[...]

PR #1695, opened 2026-09-05
fix(engine/rocky-cli): queue the server's own state-store reads, and measure the ceiling

hugocorreia90:
U5-P4 from the UI-layer plan: a load test that produces the one-process ceiling of `rocky serve`. Its first baseline found the ceiling at **one client**, so the fix ships with it.

## The finding

Every state-backed read of the HTTP API opens the store for the request. redb holds an exclusive `flock` on the file for the life of a handle, and a concurrent open polls that lock five times, 50 ms apart, before giving up as `Busy` → `503 engine_busy`. Two of the process's **own** reads raced for the lock routinely:

| Before (main + U5-P1, Apple M3 Pro, playground) | Requests | req/s | p99 | 503s |
|---|---|---|---|---|
| 1 client | 1,850 | 92 | 19 ms | 0 |
| 2 clients | 1,982 | 99 | 234 ms | **76** |

The poll exists so a `rocky run` in a terminal holding the store surfaces as a retryable 503. A poll cannot tell that case from "the request next to me is reading"; this process can.

## The fix

`ServerState::store_reads`, one permit. Every request-local read in `api.rs` goes through `store_read(&state, || ...)`, which takes the permit **inside** the blocking task (a client that disconnects drops the handler's future, not the running read, so the next read cannot start on a still-open store) and then runs the read off the async runtime as before. Sixteen call sites converted mechanically (`spawn_blocking` → `store_read`, the join-error map folded into the helper); the mutation paths (`persist_job`, the webhook spool) are untouched, and `503 engine_busy` keeps its one meaning: another process holds the store.

| After (same machine, same project) | Requests | req/s | p99 | 503s |
|---|---|---|---|---|
| 1 client | 1,807 | 90 | 20 ms | 0 |
| 2 clients | 1,897 | 95 | 45 ms | 0 |
| 4 clients | 1,937 | 97 | 69 ms | 0 |
| 8 clients | 1,923 | 96 | 140 ms | 0 |
| 16 clients | 1,960 | 97 | 274 ms | 0 |

Throughput saturates near 95 store-backed reads per second (about 10 ms per read); latency then grows with the queue. The verdict at the default 250 ms threshold: 8 clients. RSS did not move across the ladder. Routes served from the compile result (`/models`, `/dag/layers`) never take the queue and answer in under a millisecond.

**Pinned by** `concurrent_store_reads_never_answer_busy`: 64 concurrent reads over six store-backed routes, all `200`. Run first against the unconverted handlers: 51 of 64 answered 503. With the queue: passes.

## The instrument

`scripts/serve-ceiling.py`, stdlib only, next to `soak_verdict.py`. A staircase of concurrent clients over the reads the UI makes, 20 s per step; per step: requests, req/s, p50/p95/p99, errors, RSS before/after. Two guards keep the number honest: every route must answer `200` in a warm-up pass (a route the server lacks makes the run INVALID and is named, never skipped — verified against 1.73.0, which lacks four of the routes), and the top step is re-run against `/api/v1/health` alone to measure the harness's own limit (a ceiling within 80 % of it is `CLIENT_BOUND`, exit 2 — verified with `--mix /api/v1/health`: 22.8k req/s, correctly refused). `--soak-minutes` holds one load and reports the RSS slope in KB/h, the scheduler soak's statistic, for #1416.

## What else changes

- Embedding guide: "Reads queue, one at a time" and "What one process carries" with the table, the viewer arithmetic and the reproduce command; the sidecar section's "briefly races the state lock" sentence now says what the 503 means after this change.
- Changelog: Added (the script) and Fixed (the queue).
- The named follow-up, not in this PR: a short-TTL result cache per route so many viewers cost one store open per interval; the queue makes the ceiling honest, the cache would raise it.

## Acceptance (brief U5-P4, rocky-plans)

| # | Row | Evidence |
|---|---|---|
| 1 | Produces `ceiling.json` and a table; exits 2 naming a missing route | Both runs recorded above |
| 2 | INVALID when client-bound | `--mix /api/v1/health` → `CLIENT_BOUND`, exit 2 |
| 3 | One measured ceiling on a stated machine | The after table, on the embedding guide |
| 4 | The 30-minute soak slope at the ceiling's concurrency, on #1416 | Running at 8 clients; posted on #1416 when it completes |

## The two questions

**Least confident.** The queue serialises the whole blocking read, not just the open, so one slow read (a `custody` chain that compiles a large project) holds every other read behind it. On the playground a read is 10 ms; on a large project the queue wait is the new tail. The result cache is the answer if that shows up; the load test now exists to show it.

**Most likely missing.** The mix's weights. Equal weights over-count the per-call compiles (`/review/queue`, `/brief`); the real polling pattern comes from the U3 screens once built.

## Review

Codex, the independent red team this repo requires, hit its usage limit on 2026-09-04; credits return on 2026-09-07. A self-probe stood in: the permit's lifetime (inside the task, not the handler), the mutation paths left outside the queue, the join-error mapping preserved in the helper, a full `rocky-cli` and `rocky-server` unit run (1,649 + 102 passing), clippy clean on both crates. The owed Codex pass covers this PR when credits return.


hugocorreia90:
Acceptance row 4, the 30-minute soak at the ceiling's concurrency, is done and posted on #1416: 8 clients, 159,174 requests, 0 errors, RSS 46.9 MB → 28.9 MB, slope −33.6 MB/h over the run and +0.0 KB/h in the final quarter (61 grid samples, 0 dropped). The worst single-minute p99 was 886 ms, during minutes when this machine was also compiling; the idle-machine ladder is the number on the guide.

PR #1700, opened 2026-09-05
OWNER MERGE — fix(engine/rocky-cli): report a failed batched check leg as not_evaluated instead of aborting the run

hugocorreia90:
A failed batched replication check query no longer ends the run. It reports as a check, the way the per-table path already did.

The exit code on that failure changes, so this needs an owner merge. See "Exit codes" below.

```
before:  target COUNT leg fails -> try_join! -> `?` -> run ends, exit 1,
                                   no run JSON, no table's checks reported
after:   target COUNT leg fails -> every table that leg covered:
                                   row_count not_evaluated + the reason
                                   freshness leg still measures every table
```

## What changed

`run_batched_checks` in `engine/crates/rocky-cli/src/commands/run.rs` ran the three batched queries — source row counts, target row counts, freshness — under `tokio::try_join!` and `?`-ed the joined error. One failing leg ended the run, after the data had already been written.

It now uses `tokio::join!`, so every leg runs to completion, and folds a failed leg into the same `source_count_failures` / `target_count_failures` / `freshness_failures` maps the per-table fallback fills. Everything below that point is unchanged: the reporting code already reads those maps and calls `checks::row_count_not_evaluated` / `checks::freshness_not_evaluated` (#1602).

Untouched: the batch queries, which tables each leg covers, and the whole per-table fallback.

The comment claiming "same observable check results, just more round-trips" is rewritten to say what is true. Both paths report a query that could not answer as `not_evaluated` with the reason. They differ in round-trips **and in blast radius** — one batched leg is a single query over every table it was handed, so its failure names them all, while a per-table query names only its own table.

### 1. Fail-before test, control and treatment

`commands::run::tests::a_failed_batched_row_count_leg_is_not_evaluated_and_the_other_legs_still_report`.

Two tables, because a batched leg is one query over every table it is handed — with one table there is nothing left to show still reporting. A new `cfg(test)` double, `LegFailingBatchCheck`, sits beside the existing `EmptyBatchCheck`. It tells the source call from the target call by the **schema** of the tables it is handed, never by arrival order, so the test does not depend on how `join!` polls its branches.

- **Control** (`LegFailingBatchCheck::healthy()`): both tables measured on both check kinds. All pass, `not_evaluated` absent.
- **Treatment** (`FailingLeg::RowCountsIn("tgt")`): the run completes. Both tables report `row_count` `not_evaluated` = `target: the batch row count query failed: injected tgt batch COUNT failure`. Both tables' `freshness` is still a passing measurement. The anomaly history gains no row, so #1602's "a failed count must not enter the baseline as a 0" stays closed.

**The assertion that fails on `main`** is the `expect` on the returned `Result` — the abort itself:

```
a failed batched leg must not abort the run — the data has already landed (#1655): injected tgt batch COUNT failure
```

Three more tests:

- `a_failed_batched_freshness_leg_leaves_the_row_count_legs_reporting` — the mirror case.
- `both_batched_row_count_legs_failing_name_both_sides` — the joined `source: ...; target: ...` reason, built by the same formatter the per-table path uses.
- `a_batched_freshness_leg_that_answers_null_emits_no_check` — honest failure, pinned. A leg answering `max_timestamp: None` is an **empty table**, not a failed leg, and still emits no check.

**The stronger control is the twelve existing #1602 batched-check tests, unchanged byte for byte.** `join!` and `try_join!` are identical when every leg is `Ok`, and those tests still pass on the healthy path. The fixture gained a `try_run` that returns the `Result`; `run` is now a thin wrapper that `expect`s it, so no existing test needed editing.

### 2. Mutation check

Mutation — restore the abort for the target leg only:

```
let tgt = fold_row_count_leg(tgt, target_batch_refs, "target", &mut target_count_failures);
->
let tgt = tgt.map_err(anyhow::Error::from)?;
```

`scripts/mutation-check.sh engine/crates/rocky-cli/src/commands/run.rs <mut.py> bash -c 'cd engine && cargo +1.98.0 test -p rocky-cli --lib commands::run::tests::a_failed_batched_row_count_leg_is_not_evaluated_and_the_other_legs_still_report'`

```
test commands::run::tests::a_failed_batched_row_count_leg_is_not_evaluated_and_the_other_legs_still_report ... FAILED
test result: FAILED. 0 passed; 1 failed; 0 ignored
```

Exit 0. Not the exit-3 inconclusive shape from #1547.

### 3. Exit codes and status, stated exactly

`run_batched_checks` returns `Result<()>`. Its error was neither `Interrupted`, `PartialFailure` nor `CheckGateFailure`, so it fell through to `main() -> Result<()>`, which is **exit 1**. The `RunOutput` JSON is written after this call, so nothing was printed on stdout.

| setting | before | after |
|---|---|---|
| `fail_on_error = true` (the default), tables copied | exit 1, no run JSON | **exit 2**, `status: "PartialFailure"`, `check_gate_failed: true`, full `check_results` on stdout |
| `fail_on_error = false` | exit 1, no run JSON | **exit 0**, `status: "Success"`, the failing `not_evaluated` check in `check_results`, `check_gate_failed` omitted |

Both `status` values were read off `RunOutput::derive_run_status`, not assumed. `has_problem` is `interrupted || tables_failed > 0 || check_gate_failed`; on `fail_on_error = false` the gate is `false`, no table failed, so the status is `Success`.

The second row is why this is an owner merge: a failure that used to stop the run now exits 0, and the run records `Success`. That is already what any failing check does on `fail_on_error = false` — a violated `row_count` reports the same way today — but this failure class used to stop the run instead.

The exit-2 path is #1671's `CheckGateFailure`, reached because `row_count_not_evaluated` carries `passed: false` at error severity and `derive_run_status()` returns `PartialFailure` once data has landed. `RunStatus`, `tables_failed` and `replication_check_gate_failed` are untouched.

### 4. Changelog

`engine/CHANGELOG.md`, `## [Unreleased]` -> `### Fixed`, first entry, ending `(#1655)`. Verified by script that the insertion sits after the `## [Unreleased]` heading and before the next `## [` heading, under `### Fixed` with no heading in between. The **On upgrade** sentence states both exit-code rows above.

### 5. Codegen

No `*Output` struct changed, so no `just codegen` and no `just regen-fixtures`. No generated file, schema or lockfile touched. Two files changed in total.

### 6. Local gate

- `cargo +1.98.0 fmt --all --check` — exit 0
- `cargo +1.98.0 clippy -p rocky-cli --all-targets -- -D warnings` — exit 0 (confirmed by a separate invocation; the combined-gate run lost its exit code to a shell `PIPESTATUS` reset)
- `cargo +1.98.0 test -p rocky-cli --lib -- commands::run::` — `281 passed; 0 failed; 1 ignored`

No `--test-threads=1` re-run was needed. The three mutex-sharing tests (`run_trigger_from_env_maps_known_values`, `persist_run_record_stamps_trigger_submission_and_pipeline`, `a_run_adopts_an_upstream_traceparent`) all passed under default parallelism.

No independent red team ran on this change. What follows is my own adversarial pass — the inputs a reviewer would use to break the fix, and what I checked for each.

**A leg that fails for only SOME tables.** The trait cannot express it. `batch_row_counts(&[TableRef])` is one call returning one `Result`, so a leg either errors wholesale or answers. The real partial shape is a leg that **answers** and leaves tables out — not a failure, and it already reports "the batch row count query returned no result for this table". `a_table_the_batch_row_count_left_out_is_not_evaluated` and `a_table_the_batch_freshness_left_out_is_not_evaluated` pin those two reason texts apart and are unchanged by this PR.

**A leg whose error is a rate limit.** No error-class branch, deliberately. `AdapterError`'s `Display` carries the warehouse's own text (`429`, `UC_REQUEST_LIMIT_EXCEEDED`) into the reason string, so the operator reads it in `not_evaluated`. I checked whether swallowing the error robs the adaptive concurrency throttle, which is the trap #1678 called out for the `column_match` probe. It does not: `run_batched_checks` never received the throttle, and it is called only after every copy task has joined, so there is no remaining concurrency for this run to widen or narrow.

**Two legs failing at once.** Covered by `both_batched_row_count_legs_failing_name_both_sides`. The source and target failure maps are separate, and the reporting arm joins them into `source: ...; target: ...`.

**A table in no leg.** Unchanged and pre-existing. The row-count arm emits only for a `batch_asset_keys` entry it can pair with a source ref; freshness is driven by `freshness_batch_refs`. A table in neither list gets no check from either leg, before or after.

**A leg that fails while its check is disabled.** `row_count_enabled` / `freshness_enabled` still short-circuit to `Ok(vec![])` inside each branch, so a disabled check issues no query, cannot fail, and records nothing. The failure maps stay empty and the reporting arms are gated by the same two flags.

**Legs that used to be cancelled now run.** `try_join!` dropped the other two futures on the first error; `join!` does not. That is the point — it is how the other legs report — and it costs at most the two queries the run was already going to issue.

**What I am least confident about.** That no shipped `BatchCheckAdapter` uses `Err` to mean something a run should stop for. I read all three (`rocky-databricks`, `rocky-snowflake`, `rocky-bigquery`): each maps a query or transport failure to `AdapterError` and nothing else. If a future adapter used `Err` to signal "the session is gone", this change would report a check instead of stopping. The reporting contract is the right default, but it is an assumption about implementations, not something the trait enforces.

**What a reader might miss.** The `fail_on_error = false` row. Everything else here makes a silent failure louder; that one row makes a loud failure quieter — exit 1 becomes exit 0, and the record says `Success`. It is the same rule every other replication check on that setting already follows, and the failed check is in `check_results` either way, but a script that reads only the exit code or the status will stop noticing this class of failure.

Refs #1655


hugocorreia90:

**Verified:** `tokio::join!` replaces `try_join!`; each leg's error folds into the same `source_count_failures` / `target_count_failures` / `freshness_failures` maps the per-table fallback fills, via `fold_row_count_leg` and the freshness fold; every non-test hunk is inside `run_batched_checks`; the reporting below it is untouched. Test double `LegFailingBatchCheck` + `a_failed_batched_row_count_leg_is_not_evaluated_and_the_other_legs_still_report` pin the seam; the mutation (restore the abort on the target leg) is caught with a real `test result: FAILED`, not the #1547 exit-3 shape.

**The row to know about on upgrade** (the lane's own flag, and the owner-merge reason): with `fail_on_error = false`, this failure used to abort the run (exit 1, no run JSON) and now completes as `Success`, exit 0, with the failed check recorded as `not_evaluated` in `check_results`. That is what every other replication check already does on that setting; an operator reading only the exit code stops seeing this class. With the default `fail_on_error = true` it is `PartialFailure`, exit 2 (#1671).

**Independent review: not available today** (Codex credits exhausted until Sep 7). Merging under the owner's "merge all when green" instruction; queued for a Codex pass on Sep 7 alongside #1678, #1682 and the other lanes' PRs.

hugocorreia90:
Post-merge review done (Claude Opus 5 at max effort, at the owner's direction — Codex is out of credits, so this is not the independent pass `AGENTS.md` asks for). Verdict BLOCK. I verified each structural finding against `main` df506990 before filing #1719.

**The fold itself is right**, and the review confirmed the freshness fold is covered (`a_failed_batched_freshness_leg_leaves_the_row_count_legs_reporting`) and that a source-leg mutation would also fail — the suite is stronger than the single chosen mutation showed.

**What it got backwards is the question it named as its own least-confident item.** The PR body says: "does any shipped `BatchCheckAdapter` use `Err` to mean something a run should stop for? I read all three: each maps a query or transport failure to `AdapterError` and nothing else." Two of the three map an unimplemented **capability** to `AdapterError`, unconditionally, with no query issued — and `rocky-snowflake/src/batch.rs:64` says so in a comment naming `run.rs` as the intended fallback site, a fallback that was never built. So a Snowflake or BigQuery replication run with `row_count = true` now records a gating check failure on every table, on healthy data, and `--resume` refuses because every table copied. Filed as #1719 with the fix shape (a capability predicate, or a registry that returns `None` for adapters that do not batch).

**Second, a new fail-open this PR introduced:** `run.rs:6276` applies the configured freshness severity to `measured.chain(unevaluated)`, clobbering the `Error` that `freshness_not_evaluated` hard-codes so an unevaluated check gates. With `severity = "warning"` a failed `batch_freshness` leg exits **0** with `status: "Success"`, where before this PR it exited 1. The changelog's exit-code table keys on `fail_on_error` alone; the gate keys on `fail_on_error` **and** per-check severity. `row_count_not_evaluated` is not clobbered, which is why the tests do not see it.

**Two more, both pre-existing and both worth a line in the follow-up:** `run_batched_checks` still has two `?` early-returns (`run.rs:6020`, `:6056`) that abort before `pending_checks` is drained — the same class #1655 closed, on the per-table path, with the correct pattern sitting thirty lines below at `:6301`. And the rewritten parity comment at `run.rs:5907` claims both paths report an unreadable count as `not_evaluated`; the per-table path routes through `cell_as_u64`, while the Databricks batch parser's `.unwrap_or(0)` turns an unparseable cell into a measured zero that passes and is written to the anomaly history. Fix the parser or the comment, but not neither.

**Test gap that would have caught the severity finding before merge:** `run_status_for` reads the real gate and the real `derive_run_status`, and is used by three tests — all three from #1706. None of #1655's four tests asserts an exit code, a status, or `check_gate_failed`, which is the contract that made this an owner merge.

PR #1703, opened 2026-09-05
fix(engine/rocky-cli): load surrogate keys only for the models a narrowed preview renders

hugocorreia90:
A malformed `[[surrogate_key]]` on one model no longer fails a preview of another.
`rocky plan --model orders` refused when model `customers` had a bad key spec. It now renders.
Nothing that renders today is refused, and nothing that refuses today renders differently.

## What changed

`plan_preview_output` loaded the surrogate-key tree for the whole project before rendering
anything. `validate_surrogate_key_spec` returns a `ModelError` for the tree, not for the model
asked about, so one bad spec anywhere refused every narrowed preview. `rocky run` never behaved
this way. `rocky emit-sql` was narrowed the same way in #1556; this core inherited that unfiltered
load rather than choosing it.

The load now uses `load_surrogate_keys_from_tree_filtered` over the same derivation `run` and
`emit-sql` use — the `file_path` set of the compiled models this invocation selected:

```
before   compile ──► load EVERY model's key specs ──► render selected model
                          │
                          └── one bad spec anywhere ──► refuse

after    compile ──► load the SELECTED models' key specs ──► render selected model
                          │
                          └── a bad spec on a selected model ──► refuse
```

Selecting on the file name instead was rejected: a sidecar that renames its model with
`name = "..."` would silently lose its key, which renders wrong SQL instead of refusing loudly.

## What the unnarrowed preview does (kept)

**Unchanged — it still refuses, and it still names the model at fault.** `filter = None` selects
every compiled model, so the load covers every compiled model, so every compiled model's spec is
still validated up front. The issue's counter-argument (an unfiltered load reports a bad spec
early) is preserved for the invocation that has no reason to be narrow. The message already names
the model: `ModelError::InvalidSurrogateKey { model, reason }` formats as
`model 'customers' has an invalid surrogate_key: ...`. It reaches the user because
`fn main() -> anyhow::Result<()>` terminates through `Debug`, which prints the whole cause chain
under `Caused by:`; only `ConfigError` is intercepted for a miette rendering. Nothing was added.

**The refusal for a selected model is untouched, and it cannot be weakened by this change.** The
injection guard #1539 kept is `apply_surrogate_keys`, which calls `validate_surrogate_key_spec` on
every spec it interpolates, in every build (not a `debug_assert!`). The loader is early reporting;
the applier is the guard. A spec that reaches the SQL is always validated, whatever the load
selected.

### 1. Tests (six, all in `commands::plan::tests`)

| Test | What it pins |
|---|---|
| `plan_preview_narrowed_ignores_a_malformed_key_on_another_model` | **Treatment.** Bad key on `customers`, preview `--model orders` renders. |
| `plan_preview_narrowed_to_the_bad_model_still_refuses_and_names_it` | **Control.** Narrowing to the offending model still refuses; the message names `customers`. |
| `plan_preview_unnarrowed_still_refuses_a_malformed_key_and_names_the_model` | The unnarrowed behaviour above, pinned. |
| `plan_preview_narrowing_does_not_change_a_keyed_models_sql` | **Parity.** A validly keyed model's narrowed SQL is byte-equal to its statement in the unnarrowed preview, and carries `AS order_key` / `__rocky_keyed`. |
| `plan_preview_narrowing_keeps_a_nested_models_own_key` | The same, for a model in a subdirectory — where the compiled `file_path` and the sidecar walk's path could plausibly differ in form. |
| `plan_preview_unknown_filter_reports_not_found_not_another_models_bad_key` | An unknown `--model` (and a glob string) reports `ModelNotFound`, not another model's bad key. |

`cargo +1.98.0 test -p rocky-cli --lib -- commands::plan:: commands::emit_sql::` → **65 passed, 0 failed.**

### 2. Fail-before / mutation check

`scripts/mutation-check.sh` was run twice against the committed tree, once per direction. Both
mutations replace the predicate only, so the build is unaffected and the verdict is about the test,
not the compiler.

**A. Revert the fix** (predicate → `true`, byte-equivalent to the load on `main`):

```
test result: FAILED. 0 passed; 2 failed
```

The assertion that fails on `main`:

```
a malformed key on `customers` must not fail a preview of `orders`: invalid surrogate_key configuration

Caused by:
    model 'customers' has an invalid surrogate_key: key 'k' must list at least one input column
```

and

```
expected ModelNotFound for "nope", got: invalid surrogate_key configuration:
model 'customers' has an invalid surrogate_key: key 'k' must list at least one input column
```

**B. Over-filter** (predicate → `false`, the silent-wrong-SQL direction) — run over **both**
parity tests, the flat one and the nested one:

```
test result: FAILED. 0 passed; 2 failed
```

```
the narrowed preview must carry the selected model's own key:
CREATE OR REPLACE TABLE c.s.keyed AS
-- model: keyed
...
```

Exit 0 in both runs, so the #1547 false-`INCONCLUSIVE` case does not apply; the logs show
`test result: FAILED` either way.

### 3. Changelog

`engine/CHANGELOG.md`, `## [Unreleased]` → `### Fixed`, first entry, ending `(#1537)`. Asserted
programmatically that the entry's offset falls between the `## [Unreleased]` heading and the next
`## [` heading.

### 4. Codegen

Not needed. No `*Output` struct changed — the diff is `commands/plan.rs` and `engine/CHANGELOG.md`
only.

### 5. Local gate

`cargo +1.98.0 fmt --all --check`, `cargo +1.98.0 clippy -p rocky-cli --all-targets -- -D warnings`
and the test command above all pass.

## Deliberately out of scope

`compute_embedded_capabilities` keeps its whole-tree `load_surrogate_keys_from_tree(...)`. That map
feeds `ExecutionExtras::build`, which is hashed into the plan/apply TOCTOU fingerprint. Narrowing a
gate's input set would weaken the gate; it must keep hashing the resolved whole. Untouched here.

No independent red team ran on this change. Stating it rather than skipping it silently. What
follows is my own adversarial pass — the inputs a reviewer would reach for, and what was checked
for each.

**A model selected through a glob.** `plan_preview_output`'s filter is an exact `==` on the model
name; there is no pattern matching in this core. `rocky plan`'s flag is `model: Option<String>`,
"execute a single compiled model by name", handed to this core unexpanded — checked in the clap
definition, not assumed. A pattern therefore arrives as a literal, matches no compiled model, the
selected set is empty, no sidecar is read, and the caller gets `ModelNotFound` — pinned with
`"cust*"` in the unknown-filter test. Checked that the
selection predicate (`m.config.name == f`) and the render predicate (`model_ir.name == f`) and the
not-found predicate (`candidate.name == model`) all read the same name, so the three cannot disagree
about what "selected" means.

**A key spec that references a column of another model.** `validate_surrogate_key_spec` checks
identifier validity, never column existence, and specs are looked up per model by name
(`surrogate_keys.get(model_name)`). A spec on `customers` was never applied to `orders` before this
change either — only loaded and validated. So this narrows what is *validated*, never what is
*applied*, and a spec naming a column the model does not have fails at the warehouse exactly as it
did before. No behaviour moved here.

**A `--model` that names nothing.** Now reports `ModelNotFound` even when another model has a bad
key; before, the whole-tree load ran first and masked the real answer. Pinned. Note the ordering is
incidental, not load-bearing: with the filtered load an empty selection reads no sidecar at all, so
the not-found error is reached whichever check runs first.

**The over-filtering direction — the failure worse than the bug.** If
`selected_model_paths.contains(path)` ever missed, a keyed model would render silently *without* its
key. Two tests cover it, and mutation B kills both: a flat project and a nested one, each asserting
the key column is present, the flat one by byte-equality with the unnarrowed render. Both sides derive the
path from the same `models_dir` root (`path.display().to_string()` on the loader side,
`PathBuf::from` on this side), which is why the nested case is worth a test rather than an argument.

**What I am least confident about.** The claim "the unnarrowed preview is unchanged". It is exact
for compiled models, and that is what the preview renders — but the selected set is built from
`result.project.models`, so a `[[surrogate_key]]` block on a `.sql` file that is *not* a compiled
model is no longer validated by any preview. The whole-tree load did validate it. I could not
construct a project where that file's key is ever applied by any path (the applier looks up the
compiled model's name), so the lost validation is of a spec nothing reads. Naming it because
"unchanged" is a stronger word than the code earns. The same statement is true of #1556's
`emit-sql` narrowing, which shipped a week ago.

**What a reviewer might see that I did not.** The MCP `plan_preview` tool shares this core, so its
behaviour changes with it. I read the served text in `rocky-mcp/src/tools.rs` and it makes no claim
about surrogate-key validation timing, so nothing there goes stale. I did not re-run the MCP
roundtrip golden.

Refs #1537


hugocorreia90:

**Verified:** `plan_preview_output` loads surrogate keys with `load_surrogate_keys_from_tree_filtered` over the selected compiled models' file paths — the same derivation `run` and `emit-sql` (#1556) use; the only non-test hunks are that load and its comment; the `apply_surrogate_keys` validation on interpolation (the guard #1539 kept) is untouched. Six tests, including the two that fail on `main` and the two byte-parity ones the over-filter mutation kills. Both mutation runs are real `FAILED` verdicts.

**Kept as-is:** the unnarrowed preview still loads and validates every compiled model's spec and the refusal already names the model (`model 'customers' has an invalid surrogate_key: …`). The honest residual the lane states: a `[[surrogate_key]]` on a `.sql` that is not a compiled model is no longer validated by any preview; no path applies such a spec, and `emit-sql` has had the identical narrowing since #1556.

Self-merge class: a narrowed preview that refuses today renders; nothing that renders today is refused. **Independent review: not available today** (Codex credits exhausted until Sep 7); queued for the Sep 7 pass with the rest.

hugocorreia90:
Post-merge review done (Claude Opus 5 at max effort, at the owner's direction — Codex is out of credits, so this is not the independent pass `AGENTS.md` asks for). Verdict BLOCK, filed as #1730. Rank it third: both findings need a path component that is not valid UTF-8, so they fire on the Linux targets only.

**The narrowing is right and the review cleared it thoroughly.** The `file_path` join is exact for every path Rust can round-trip: `Path`'s `PartialEq` compares components, and both sides start from the same `models_dir` through `walk_model_dirs`, which keeps the root's form. Duplicate model names are rejected upstream, a glob arrives as a literal and yields `ModelNotFound`, the unnarrowed preview is genuinely unchanged, and the injection guard is untouched. It also **closes** a hijack: a rogue `.sql` whose sidecar claims a real model's name no longer wins the first-wins whole-tree load.

**The defect is the join itself.** `Model::file_path` is a `String` from `path.display()`, which is `to_string_lossy`; the loader compares a raw `Path`. Under a non-UTF-8 path component they can never be equal, so the preview renders without its surrogate-key column — silently, which is the direction this PR's own comment calls worse than the bug. The same join sits in `run.rs`, `emit_sql.rs` and, more seriously, in the incremental typecheck's affected-set seed, where a changed model is never marked affected and the previous typed result survives the edit. Making `Model::file_path` a `PathBuf` closes all four.

**Also in the issue:** `compute_embedded_capabilities` still uses the unfiltered load with `.unwrap_or_default()` while the apply-side recompute uses the filtered strict map, so the two sides of a TOCTOU fingerprint are now built by different functions over different sets.

Issue #1702, opened 2026-09-05
Seven more sites still read an unloadable rocky.toml as absent, and rocky review --queue re-swallows the refusal #1667 added


hugocorreia90:
#1667 and #1680 between them converted fourteen offline config reads onto `load_optional_project_config`, so a `rocky.toml` that is present and does not load refuses instead of reading as absent. Seven sites of the same class are still open. None is a regression; all are pre-existing.

```
site                                what it does with an unloadable config
review.rs (queue path)              `compile_project(..).ok()` — swallows the refusal #1667 ADDED at
                                    audit.rs::compile_project, so `rocky review --queue` still ranks
                                    every escalation with "blast radius unknown" on a broken config
plan.rs compute_embedded_capabilities  `load_rocky_config(..).ok()` for the routing identity, the
                                    governance identity, the env-resolved mask AND the schema cache
plan.rs (--semantic leg)            `Err(_) => HashMap::new()` for the schema cache
branch.rs (promote gate)            `Err(_) => HashMap::new()` for the schema cache
brief.rs (two sites)                `load_rocky_config(..).ok()`
cost.rs                             `match load_rocky_config(..)` degrading
restore.rs (five sites)             `load_rocky_config(..).ok()`
```

The first row is the cheapest and the most clearly wrong: `rocky review --queue` re-swallows an error the same helper now raises, so the conversion does not reach that command.

The `plan.rs` and `branch.rs` rows are the decision #1667 deferred and #1680 declined to pre-empt: each already reasons in code about *skipping* a gate rather than blocking on a configuration issue, and turning a skip into a refusal changes what those gates mean. #1680 corrected the false comment at `plan.rs` (an erased type change leaves the model out of the change map, and `EmbeddedCapabilities::touched` gates an absent model under the weaker `apply` capability, not `schema_change.breaking` — fail-open, not fail-closed) but deliberately left the code.

Two smaller items from the same review:

- `rocky compile`, `rocky emit-sql` and `plan.rs`'s `plan_preview_output` bare-`?` the loader, so their refusals do NOT name the config file. `load_project_source_schemas` and `catalog.rs` now write `failed to load config from <path>`; those three should match.
[...]

PR #1704, opened 2026-09-05
OWNER MERGE — fix(engine): refuse an unloadable rocky.toml in review --queue, brief and cost

hugocorreia90:
`rocky review --queue` called `compile_project(..).ok()`, which threw away the refusal #1667 added one function below it. A broken `rocky.toml` ranked every escalation with **"blast radius unknown"** — the same ranking a healthy project with no downstream models produces.

`rocky brief` and `rocky cost` had the same swallow. All three now refuse and name the file.

Owner merge: exit codes change for configs that never loaded.

## The shape

```
  rocky.toml is...      before                          after
  ────────────────      ──────────────────────────      ────────────────────
  absent                defaults / cold cache           unchanged
  loads                 use it                          unchanged
  unset ${VAR} in
    adapter creds       strict load failed → degrade    use it (#1536)
  present, broken       swallowed → degrade             REFUSE, naming the file
```

### 1. `rocky review --queue` — the priority

`build_queue` ranks each escalation by blast radius, from one compile of the project. The compile is seeded by the schema cache, and reading the cache needs the config. `compile_project(..).ok()` collapsed **two different failures into one degrade**:

```
  config present, does not load  ->  was: blast_radius = None   now: Err
  compile fails (bad model SQL)  ->  was: blast_radius = None   now: unchanged
```

The second degrade is deliberate and documented — a broken model must not take down the whole queue. So `audit::compile_project` is split: `compile_project` still loads-then-compiles, and a new `compile_project_with_schemas` takes the seed already resolved. `build_queue` loads the schemas (propagating), then compiles (degrading). Nothing else changes about the ranking.

**The config is read before the outstanding-selection short-circuit.** `rocky review --queue` on a broken config refuses whether or not the ledger holds a pending row today. Putting the read after that short-circuit would have made the contract "refuses, *unless* the queue happens to be empty" — an exception a user cannot see, and one that first bites on the day something is actually pending. The cost is one read-only schema-cache scan on an otherwise-empty queue. Pinned by its own test.

Credential-tolerant loader: `rocky review --queue` reads the state store and compiles offline, and opens no warehouse connection.

Three production callers share `compute_review_queue` and all three now refuse: the CLI, `GET /api/v1/review/queue` on `rocky serve`, and the governor's `review_queue` MCP tool (whose *approve* action computes the queue first, as its pending-escalation guard).

### 2. `rocky brief`

Three sections read `rocky.toml`, each for itself, each swallowing its own failure:

| section | on a broken config, before |
|---|---|
| cost | `availability: available`, dollar figures silently absent |
| autonomy | note "rocky.toml not loaded — budget degradations unavailable" |
| scheduler | `unavailable`, note "config unreadable: …" |

The cost row is the defect: an `available` section reporting no money, which is also what a project that never configured `[cost]` reports.

`compute_brief` now loads the config **once** and passes `Option<&RockyConfig>` to all three. Threading it to only the two `.ok()` sites the issue named would have *created* an inconsistency that does not exist today: on a config with an unset `${CREDENTIAL}`, cost and autonomy would render (tolerant load succeeds) while scheduler still said "config unreadable" (strict load fails). Today all three degrade together; after this they all work together.

The read is placed **before** the absent-state-store short-circuit, for the same reason as the queue. Pinned by its own test.

Credential-tolerant loader. `rocky brief` does reach an object store, through `[state]`'s durable freeze-marker tier — the answer to "why is that safe" is that tolerance is scoped to adapter **connection** fields and nothing else, so an unresolved `${VAR}` under `[state]` is still refused and no literal placeholder can reach `durable_tier_provider`. Adapters are read for their `type` only, to price runs.

One note text changed. With no `rocky.toml`, the scheduler section said `config unreadable: <FileNotFound>`; it now says `no rocky.toml — scheduled pipelines unknown`. The old label described a case it could not be describing — the only test that covered it passed a `missing.toml`, and a config that is present and does not load no longer reaches that function at all.

### 3. `rocky cost`

`[cost] warehouse_size` and `compute_cost_per_dbu` come from the config, and `resolve_warehouse_type` needs the adapter list. A load error logged a `warn!` and returned `None`, producing a rollup with no dollar figures — which is **also** what a config-less project prints, and what a non-billed source (Fivetran, Airbyte) prints. Three situations, one answer.

The config leg is now `adapter_pricing(config_path) -> Result<Option<…>>`, so the three outcomes are distinct and unit-testable. Credential-tolerant loader: `rocky cost` reads the state store and prices from the run record, and opens no warehouse connection.

### 4. `restore.rs` — a correction, no code change

The issue lists **five `.ok()` sites in `restore.rs`**. All five are inside `#[cfg(test)] mod tests` (the module opens at line 1083; the sites are at 1581, 1622, 1962, 2189, 2218). They are not swallows — each builds the `Option<RockyConfig>` argument of `run_restore_apply_in_with` from a config the test itself just wrote.

The production path already fails closed, and has since the finding that introduced that parameter:

```
  run_restore_apply_in
    -> load_rocky_config(config_path).with_context("loading … to resolve the warehouse …")?
    -> AdapterRegistry::from_config(&cfg)?
    -> run_restore_apply_in_with(…, Some(cfg))      the SAME snapshot, threaded down
```

It keeps the **strict** loader, correctly: it builds a live warehouse adapter to re-derive the artifact on its recording engine, so a literal `"${DATABRICKS_HOST}"` must not get through.

So five of the issue's nine rows are a grep artifact. Nothing to change.

### 5. Two more corrections to the issue

- **`brief.rs` has three config reads, not two.** The third (`build_scheduler`) used the strict loader and degraded to `unavailable`. It is converted here for the consistency reason above.
- **The "smaller item" about `rocky compile` / `rocky emit-sql` / `plan_preview_output` is already done on `main`.** All three write `failed to load config from <path>`, landed by #1539 (`f5b3e529`). Verified with `git log -L` on `compile.rs`, and by reading `emit_sql.rs` and `plan.rs` on `main`. Not re-done.

## Deliberately unchanged

`plan.rs`'s `compute_embedded_capabilities`, `plan.rs`'s `--semantic` leg, and `branch.rs`'s promote gate. Each reasons in code about **skipping a gate** rather than blocking on a configuration issue, and turning a skip into a refusal changes what those gates mean. #1667 deferred them and #1680 declined to pre-empt them; they should move together, in a PR that decides what "skip the gate" means. They remain open in #1702, which is why this PR says `Refs`, not `Closes`.

### Fail-before tests

Every test below is new and drives the **caller**, not the shared helper. The column is the observed behaviour with the fix reverted, measured by the mutation checks below.

| Site | Test | With the fix reverted |
|---|---|---|
| review `--queue` | `review_queue_refuses_a_present_but_unloadable_config` | **FAIL** — `Ok`, one entry, `blast_radius: None` |
| review `--queue` | `review_queue_refuses_an_unloadable_config_with_an_empty_ledger` | **FAIL** — `Ok`, `total: 0` (pins the read's placement) |
| review `--queue` | `review_queue_still_ranks_without_any_config` | **FAIL** on its discriminating half |
| review `--queue` | `review_queue_tolerates_an_unset_credential_var` | passes (honest-failure guard) |
| `rocky brief` | `brief_refuses_a_present_but_unloadable_config` | **FAIL** — `Ok` |
| `rocky brief` | `brief_refuses_an_unloadable_config_without_a_state_store` | **FAIL** — `Ok` (pins the read's placement) |
| `rocky brief` | `brief_still_renders_without_any_config` | **FAIL** on its discriminating half |
| `rocky brief` | `brief_tolerates_an_unset_credential_var` | passes (honest-failure guard) |
| `rocky cost` | `cost_refuses_a_present_but_unloadable_config` | **FAIL** — `Ok`, rollup printed |
| `rocky cost` | `cost_still_reports_without_any_config` | **FAIL** on its discriminating half |
| `rocky cost` | `cost_prices_a_run_with_an_unset_credential_var` | passes (honest-failure guard) |

The three `still_…_without_any_config` tests each carry a second half that writes the **broken** config into the same project and asserts it refuses. That is what stops an honest-failure guard from passing for the wrong reason — "absent works" is also true when the config leg has been deleted entirely.

Both `tolerates_an_unset_credential_var` tests pin **both** loader directions: the fixture must fail `load_rocky_config` (strict) and pass `load_optional_project_config` (tolerant). A fixture that failed neither would prove nothing.

Fixtures are the same vocabulary #1701 uses: a config that parses as TOML and fails a validator (`fivetran` without `kind = "discovery"`), and a quoted `${VAR}` in an adapter connection field.

### Mutation checks

`scripts/mutation-check.sh`, on the committed tree, each mutation restoring only the **config leg** and leaving the compile degrade and the new signatures intact.

| Target | Mutation | Test group | Verdict |
|---|---|---|---|
| `commands/review.rs` | the schema load back to `.ok().unwrap_or_default()` | `commands::review::tests::review_queue_` | **fix-sensitive** (exit 0, 3 of 4 failed) |
| `commands/brief.rs` | the single shared load back to `.ok().flatten()` | `commands::brief::tests::brief_` | **fix-sensitive** (exit 0, 3 of 4 failed) |
| `commands/cost.rs` | `adapter_pricing`'s load back to `.ok().flatten()` | `commands::cost::tests::cost_` | **fix-sensitive** (exit 0, 2 of 3 failed) |

No exit 3. The review run prints the defect verbatim:

```
a present but unloadable rocky.toml must refuse `rocky review --queue`: ReviewQueueOutput {
    total: 1,
    pending: [ReviewQueueEntry { model: "fct_orders",
                                 blast_radius: None,        <-- the gate never ran
                                 classification_weight: 3,
                                 score: 4358.42 }] }
```

`blast_radius: None` next to a `SchemaChangeBreaking` escalation is the defect, exactly: the queue ranked a breaking change as if it touched nothing downstream.

### Honest failure

| Site | (a) no `rocky.toml` | (b) valid config, unset `${CREDENTIAL}` in an adapter connection field |
|---|---|---|
| `rocky review --queue` | unchanged — ranks against a cold schema cache, blast radius unknown. Proven. | **now works better**: the cache is read. The strict loader would have cooled it. Proven, both loader directions. |
| `rocky brief` | unchanged — no dollar figures, the autonomy note, the scheduler section `unavailable` (new wording). Proven. | **now works better**: every section reaches its real answer. Proven. |
| `rocky cost` | unchanged — durations and bytes from the run record, no dollar figures. Proven. | **now works better**: the run is priced. Proven by building the rollup and asserting `total_cost_usd`, not just that the pricing parameters were read. |

**No previously-`0`-exit path on a project with no `rocky.toml` changes exit code.** That is the whole (a) column, and each row has a test. The only exit-code change is a `rocky.toml` that is present and does not load.

### Codegen

No `*Output` struct changed and `output.rs` is untouched, so `just codegen` and `just regen-fixtures` do not apply. `BriefSchedulerSection.note` is a `String` whose *value* changes for a project with no config; the corpus has no `brief` fixture and the engine holds no `brief` golden (`scheduled_pipelines` appears only in the producer and the struct definition), and `cost.json` is generated from a project whose config loads, so neither drifts. No `editors/vscode/package-lock.json` churn.

### Local gate

| Gate | Result |
|---|---|
| `cargo +1.98.0 test -p rocky-cli -p rocky-mcp --lib` | **1742 passed, 0 failed** (rocky-cli) + **70 passed, 0 failed** (rocky-mcp), on the merged tree |
| `cargo +1.98.0 clippy -p rocky-cli -p rocky-mcp --all-targets -- -D warnings` | pass |
| `cargo +1.98.0 fmt --all --check` | pass |

### Merged with #1701

#1701 landed on `main` while this was in flight, and `origin/main` is merged into this branch (merge commit in the history, not a rebase). Two conflicts, both resolved keeping **both** sides:

- `engine/CHANGELOG.md` — this entry now sits below #1701's and #1700's, still inside `## [Unreleased]` → `### Fixed`.
- `commands/review.rs` — the test module only. #1701 appended `compute_review` tests, this appends `compute_review_queue` tests. Both blocks are kept; the fixture constants are suffixed `_1680` and `_1702` respectively, so nothing collides.

The one shared production edit — the `failed to load config from <path>` context inside `load_project_source_schemas` — was written here **byte-identical to #1701's**, comment included, so `git diff origin/main -- source_schemas.rs` is now **empty**. That was the point of copying it verbatim rather than adding the context at the call site: a divergent version would have produced `failed to load config from P: failed to load config from P: …`.

The whole gate was re-run on the merged tree: **1742 passed** (rocky-cli) + **70 passed** (rocky-mcp), clippy and fmt clean.

The repository's independent reviewer is out of credits. A same-family reviewer would not satisfy the independence requirement, so none was run and none is presented as one. What follows is my own adversarial self-review, per site.

**review `--queue`.** The strongest objection is the extra work on the empty-queue path: the config parse and the schema-cache scan now run before the short-circuit that used to skip them. I accepted that; `rocky review --queue` is an interactive/CI command, not a hot loop, and a rule with an invisible exception is worse than one read-only redb open. The second objection is that splitting `compile_project` adds a function whose only caller is the queue — true, and the alternative (propagating every error from `compile_project`) would have turned a broken model into a failed queue, which the existing comment explicitly rejects. The third: `compile_project`'s own callers are unchanged, so nothing else moved.

**`rocky brief`.** I extended scope by one function beyond what the issue named. The justification is in the body: the narrow fix would have shipped a *new* inconsistency. The risk I took on is the `note` text change, which is a user-visible JSON string; I checked it appears in exactly two places in the repo (the producer and its one test) and in no fixture. The remaining wart: `build_scheduler` returns `unavailable` for a project that simply has no config, when `no_data` would arguably be the honest answer. I left the availability alone deliberately — changing it is a separate judgment about what the digest promises, and this PR is about config errors.

**`rocky cost`.** `adapter_pricing` is a new private function, so its three outcomes are testable without capturing stdout. The `warn!` is gone; `tracing::warn` is no longer imported. Someone relying on that warning line to notice a broken config now gets a non-zero exit instead, which is strictly louder.

**What I am least confident about.** The `rocky brief` scoping call. The issue named two sites; I converted three, on the argument that a partial conversion creates a new inconsistency rather than removing one. If the owner disagrees, the third site reverts on its own — it is one `match` and six mechanical test call sites — and the honest-failure table gains a row saying cost and autonomy price an unset-`${CREDENTIAL}` project while scheduler does not. I would rather be told to shrink it than ship the inconsistency silently.

**What I think is most likely being missed.** That `rocky brief` is the surface an operator reads *when something is already wrong*. Every other command here is asked a question and can decline to answer it; the brief is the morning digest, and a broken config is exactly the kind of thing it exists to surface. Refusing is the right call — but it means a project whose `rocky.toml` breaks loses its digest entirely, including the run failures and escalations that have nothing to do with the config. The alternative shape (keep rendering, mark all three config-fed sections `unavailable` with the loader's own error, exit non-zero) would tell the operator both things at once. I did not build it because it is a bigger change to what `compute_brief` promises, and because #1667's rule is "refuse, do not proceed against defaults". It is worth deciding on the record rather than by default.

Refs #1702


hugocorreia90:

**Verified:** review `--queue`, brief and cost now load through `load_optional_project_config` (tolerant — none of the three opens a warehouse) with the file named; the config leg propagates while the compile degrade in the queue path stays as it was; no `.ok()` on a config load survives in the touched files. Three mutation runs are real `FAILED` verdicts (the queue one prints the defect: `blast_radius: None` on a `SchemaChangeBreaking` escalation). Honest-failure table holds: no config-less path changes exit code; an unset `${CREDENTIAL}` now works *better* at all three.

**Corrections accepted:** `restore.rs`'s five `.ok()` sites are inside `#[cfg(test)]`; the production path already loads strictly and names the file — the issue row was wrong, and the title dropped "and restore" accordingly. The compile / emit-sql / plan naming item was already done by #1539.

**For the owner, on the record:** `rocky brief` is the surface read when something is already wrong. Refusing on a broken `rocky.toml` loses the whole digest, including run failures and escalations unrelated to the config. The alternative — keep rendering, mark the three config-fed sections unavailable with the loader's error, exit non-zero — tells the operator both things at once. This PR follows #1667's rule (refuse, never proceed against defaults) and leaves that alternative as a decision, not a default. Still open in #1702: `plan.rs` embedded capabilities / `--semantic`, `branch.rs` promote gate.

hugocorreia90:
Post-merge review done (Claude Opus 5 at max effort, at the owner's direction — Codex is out of credits, so this is not the independent pass `AGENTS.md` asks for). Verdict BLOCK, filed as #1727. Verified on `main` df506990.

**Notably clean, and stated so nobody re-audits it:** this PR contains **no fail-open**. The reviewer looked for the shape that has bitten five sibling PRs in this batch and found every direction-of-harm change here is pass→refusal. `load_project_source_schemas` can only `Err` from its config leg — every cache leg degrades to an empty map — so `build_queue`'s new `?` added exactly one refusal input. The `compile_project` split is faithful, with the whole remainder verbatim. The strict→tolerant swap is not a widening: the tolerant loader runs the same validator chain, and `tolerated_placeholder_path` matches only a depth-3 adapter field that is neither `type` nor `kind`, so `[state]`, `[cost]`, `[policy]` and nested sub-tables are still refused — the lane's claim that no literal `${VAR}` reaches the durable tier holds. The `restore.rs` correction is right (`#[cfg(test)]` opens at line 1082; the production path already refuses strictly). MCP fails closed and keeps the full chain.

**The finding is the one I raised as a decision, and it is worse than I framed it.** I said refusing costs the operator the digest. The review found the redundancy I was implicitly leaning on does not exist: `select_outstanding` has exactly two production callers, `brief.rs:359` and `review.rs:823`, and this PR converted **both**. On a project whose `rocky.toml` will not load there is now no command that lists what is waiting for a human — and the trigger can be as inert as a leftover `[schema_evolution]` section, which Rocky refuses at deserialization with a message saying nothing ever read it.

Six of the brief's nine sections take no config argument, the refusal at `brief.rs:168` precedes all of them, and the degrade machinery is already used one crate-file away: `audit.rs:608` renders its config-dependent section `Unavailable` with the error text and prints the rest. So `rocky audit` answers on a broken config and `rocky brief` does not.

**Two more in the issue:** `build_scheduler` returns `NoData` with `incident_count: 0` whenever no pipeline declares `[schedule]`, before the incident scan — so a webhook-only project (which needs no `[schedule]`) reports zero incidents while its incident directory has bundles. Pre-existing, but this PR widened the input set that reaches it and its new test pins that outcome. And both HTTP surfaces answer `500 internal_error` with the cause stripped by `to_string()` where the MCP equivalents use `{e:#}`.

The fix is the section-level degrade the owner already accepted for `rocky brief`; #1727 carries the shape, and `review --queue` and `cost` keep their whole-command refusal.

Issue #1705, opened 2026-09-05
cross-source overlap: a group with one keyless sibling stops checking the siblings that do carry the key


hugocorreia90:
After #1654, a cross-source overlap group holding a sibling with no key column is reported as a passing, not-applicable check. That is right for a two-sibling group: with one member unable to carry the key, there is nothing to compare.

It is coarse for a larger group. Three siblings, one of them keyless, and the two that **do** carry the key stop being compared with each other — a real duplicate key shared by those two is no longer flagged.

## What would settle it

Narrow the group to its key-carrying members and run the check on those when two or more remain:

```
  siblings A, B, C   C has no key column
  today   -> group not applicable, nothing measured
  better  -> check A vs B, report C as excluded
```

This was left out of #1654 on purpose. It changes `contributing_tables` and the meaning of the
emitted check, so it is a design decision rather than a bug fix, and it needs its own answer for
what the result should say about the excluded sibling.

Refs #1654

hugocorreia90:
Reclassifying this from "design decision" to a **fail-open regression**, with the evidence.

A post-merge review of #1706 (Claude Fable 5.1, at the owner's direction; not the independent Codex pass) found it, and I verified every line on `main` df506990.

**The behaviour.** `overlap_group_missing_key` (`run.rs:5810`) returns on the FIRST sibling missing the key, and the caller `continue`s (`run.rs:6489`), so the overlap query never runs for the group.

```
3 siblings, keys = ["customer_id"]
  tgt_a.orders  customer_id = 5     ─┐ a real overlap
  tgt_b.orders  customer_id = 5     ─┘
  tgt_c.orders  no customer_id       → classifier returns here
result: cross_source_overlap passed=true, overlap_count=0, exit 0
before #1706: exit 2, failed check
```

**Why it is not a design call.** The published contract already says per-sibling:

- `docs/src/content/docs/reference/configuration.md:499` — "Sibling tables whose key can't be evaluated (missing column / keyless) are **skipped** with a logged reason rather than erroring."
- `docs/src/content/docs/concepts/data-quality-checks.md:327` — "Rocky **skips them** with a logged reason rather than failing the check."
- The commit that added the check (`543bb3bf`) lists "per-table (vs per-group) keyless skip" in its own follow-ups.

The tenant fan-out this check exists for has three or more siblings by construction, so the fail-open case is the common one, not the edge.

[...]

Issue #1654, opened 2026-09-04
cross-source overlap: a keyless or missing-key sibling is reported as a failed check, though that case is tolerated by design


hugocorreia90:
A cross-source overlap check on a sibling table that legitimately has no key column, or is missing the configured key, records a **failed** check. The code says that case is tolerated; the result it produces says the opposite.

## Where

`engine/crates/rocky-cli/src/commands/run.rs`, the overlap query's `Err` arm. Its own comment states the intent:

> A missing key column / keyless table surfaces as a query error, and that case is tolerated by design (FR acceptance criterion) — but the ERROR TYPE does not say which case this is.

It then calls `rocky_core::checks::cross_source_overlap_not_evaluated`, and that constructor sets `passed = false` (`engine/crates/rocky-core/src/checks.rs`). So the tolerated case is emitted as a check that did not pass.

## Why the arm exists, and why it is still not right

Before it, the check was dropped entirely: no entry in `check_results`, nothing in the tally, nothing in the JSON. A group that was never evaluated read as a group with nothing to report. Reporting it is the right direction.

The problem is that one arm now serves two different situations that the error type cannot tell apart:

```
  keyless / missing-key sibling   -> tolerated by design   -> reported as FAILED
  syntax, permission, transport   -> a real failure        -> reported as FAILED
```

An ordinary project with two qualifying sibling tables, one of which has no such key, gets a failing check on every run for a configuration the design accepts.

## What would settle it

Distinguish the two before choosing the result. Either probe the key columns against the sibling's schema before building the query (`describe_table` is already available on the adapter at that point), or classify the adapter error. A tolerated sibling should then either emit no check, or emit one that passes and says the sibling carries no key — and a real query failure keeps the failing `not_evaluated`.

## Provenance

Pre-existing on `main`; not introduced by #1652, which moves this code into its own function without changing it. Found by an independent review of that PR, which read the moved lines as new. Verified against `main` before filing.

hugocorreia90:
Fixed by #1706 and #1717. Verified against `main` 27170b61.

The tolerated case is now settled **before** the query runs, instead of being inferred from a query error afterwards.

`partition_overlap_key_carriers` (`engine/crates/rocky-cli/src/commands/run.rs:5994`) describes each sibling and splits the group into carriers and proven non-carriers.

```
1 carrier   -> run.rs:6770  cross_source_overlap_not_applicable  passed, not gating
[...]

PR #1706, opened 2026-09-05
OWNER MERGE — fix(engine/rocky-cli): stop reporting a keyless overlap sibling as a failed check

hugocorreia90:
A cross-source overlap check no longer fails a run because one sibling table has no key column.
That shape is accepted by design, but it was reported as a failed check — and since #1671 a failed
error-severity check fails the run, so a healthy pipeline exited 2.

**Owner merge: the exit code on that run changes from 2 back to 0.**

## What was wrong

The overlap check groups sibling tables — same source type, same table name, different target
schemas — and flags a business key that appears in more than one. One `Err` arm served two
different situations, and the error type cannot tell them apart:

```
  keyless / missing-key sibling  -> tolerated by design  -> reported as FAILED  <- the defect
  syntax, permission, transport  -> a real failure       -> reported as FAILED  <- correct
```

## The fix

`overlap_group_missing_key` settles it **before** the query runs, from the sibling's own columns —
not by reading an error string afterwards.

```
  for each sibling of a qualifying (>=2 member) group:
      describe_table(sibling)
        Ok(cols), non-empty -> is every configured key column present (case-folded)?
                                 no  -> group is NOT APPLICABLE  -> passing check, reason carried
                                 yes -> fall through, run the query
        Ok(empty) | Err(_)  -> classify nothing -> fall through, run the query
```

### The representation, and why

A tolerated group emits a **passing** `CheckResult` whose `not_evaluated` carries the reason and
names the sibling. `overlap_count` stays `0` and remains a placeholder.

Omission was rejected. Dropping the check entirely is the bug #1652 fixed — a group that was never
evaluated read as a group with nothing to report — and Dagster's `emit_check_results` turns each
`CheckResult` into one `AssetCheckResult`, so an omitted check produces no event at all. Given
#1669 already makes a failed overlap check hard to see there, omission would make it worse.

Reporting `passed: true` with `not_evaluated: None` was also rejected: `overlap_count: 0` with no
reason reads as "no overlap found", which is exactly the defect #1524 closed.

This is the first `CheckResult` that sets `not_evaluated` while `passed` is `true`, so the field's
contract is widened, not repurposed: read `passed` to decide whether a check failed — it is what
`check_failures_by_severity` and `replication_check_gate_failed` actually read — and read
`not_evaluated` to learn whether the numbers beside it were measured. No code in the engine, SDK,
Dagster integration or VS Code extension branches on `not_evaluated`; it is declared and described,
never tested. A consumer that did treat it as "failed" would now see a not-applicable group as
failed — noisy, not silently green.

### The discriminator and its bound

It is a positive column check, so **no adapter error wording is involved anywhere**. That trap is
avoided by construction rather than by auditing five adapters' error strings.

Two deliberate limits:

- **`keys` only.** A `key_expr` is free-text SQL over arbitrary columns, and a column list cannot
  say whether it applies. A `key_expr` group is never classified, so its query failure keeps the
  failing `not_evaluated` it has today. Fail-closed.
- **A `DESCRIBE` that will not answer classifies nothing.** An error or an empty column list leaves
  the group alone to run its query. "Rocky could not read the schema" must never become "the key is
  tolerated" — that direction fails open on a key that really is there.

Cost: one `DESCRIBE` per sibling of a qualifying group, only when `cross_source_overlap` is
configured with `keys`. The schemas the copy path prefetches live on `TableTask` and are not in
scope at this seam.

## Evidence

Eight tests at the `run_batched_checks` seam, which exists so a test can drive it with a warehouse
that fails a specific query. There were no tests on the overlap runner path before this change.

| Test | Asserts | On `main` |
|---|---|---|
| `a_keyless_overlap_sibling_is_not_a_failed_check_and_does_not_gate_the_run` | passing, reason names the keyless sibling, gate not tripped, run `Success` | **FAILS** |
| `a_failing_overlap_query_still_fails_and_still_gates_the_run` | `not_evaluated`, `passed = false`, `PartialFailure` | passes (control) |
| `an_overlap_query_that_finds_no_shared_key_is_a_measured_pass` | passing, `not_evaluated` is `None` | passes (control) |
| `an_all_null_overlap_key_column_is_measured_not_treated_as_keyless` | measured pass, not swept into the tolerated case | passes (control) |
| `a_real_overlap_still_fails_the_check` | failing, and it gates | passes (control) |
| `an_overlap_key_column_differing_only_in_case_is_still_the_key` | `CUSTOMER_ID` matches `customer_id`, group is measured | passes (control) |
| `a_key_expr_overlap_group_is_never_classified_as_keyless` | keeps failing `not_evaluated` | passes (control) |
| `an_unreadable_overlap_sibling_schema_is_not_treated_as_keyless` | a failing `DESCRIBE` does not become "tolerated" | passes (control) |

Only the first assertion set fails before the fix. That is the point: the change is narrow, and the
seven controls pin behaviour it must not disturb.

**Which assertions fail on `main`**, with the fix's decision point removed:

```
a_keyless_overlap_sibling_is_not_a_failed_check_and_does_not_gate_the_run ... FAILED
  a keyless sibling is tolerated by design and must not fail:
  CheckResult { name: "cross_source_overlap:test.orders", passed: false, severity: Error,
    not_evaluated: Some("DuckDB error: Binder Error: Referenced column \"customer_id\"
    not found in FROM clause!"), details: CrossSourceOverlap { overlap_count: 0, ... } }
```

`result.passed` is false, the gate trips, and the run is `PartialFailure` instead of `Success`.

The run-level half of that assertion calls the real `replication_check_gate_failed` and the real
`derive_run_status`, fed the **fixture's own** `pipeline.checks` — the same value the runner passes
at its single call site — rather than a separately parsed config, and asserts `fail_on_error` is on
so the tests cannot quietly become a test of the disabled gate.

### Mutation check

`scripts/mutation-check.sh` on the runner file, mutation = delete the pre-check block so the group
falls through to the query exactly as on `main`:

```
MUT_EXIT=0
```

Exit 0, not the known-false exit 3 (#1547); the log shows `test result: FAILED. 10 passed; 1 failed`.

### Local gate

```
cargo +1.98.0 fmt --all --check                                          FMT_EXIT=0
cargo +1.98.0 clippy -p rocky-cli -p rocky-core --all-targets -D warnings CLIPPY_EXIT=0
cargo +1.98.0 test -p rocky-cli --lib -- commands::run::   289 passed, 0 failed, 1 ignored
cargo +1.98.0 test -p rocky-core --lib checks               46 passed, 0 failed
```

The three shared-mutex tests in `commands::run::` (`run_trigger_from_env_maps_known_values`,
`persist_run_record_stamps_trigger_submission_and_pipeline`,
`a_run_adopts_an_upstream_traceparent`) passed under default parallelism — no cascade, no
`--test-threads=1` rerun needed.

### Codegen

`just codegen` after the doc change on `CheckResult.not_evaluated` and `RunOutput.check_gate_failed`.
Description text only — no field added or removed. So the hand-written SDK runtime models are
unchanged, and `just regen-fixtures` is not required: fixtures capture CLI JSON values rather than
schema descriptions, and no fixture POC configures `cross_source_overlap`. Spurious
`editors/vscode/package-lock.json` churn was reverted.

No independent red team ran. What follows is my own adversarial self-review — the inputs a reviewer
would use to break the discriminator, and what I checked for each.

- **A sibling with the key column under a different case.** The break: comparing case-sensitively
  reads `CUSTOMER_ID` as keyless, so the group is written off and a real key silently stops being
  checked. That is fail-open and strictly worse than #1654. Checked: folding reuses
  `column_map::build_column_name_set`, the same lowercase derivation `column_match` compares with,
  and the configured key is folded the same way — not a hand-rolled `eq_ignore_ascii_case`, which
  would diverge on non-ASCII identifiers. Warehouses that upper-case identifiers (Snowflake) are the
  live version of this. Pinned by `an_overlap_key_column_differing_only_in_case_is_still_the_key`.
- **A key present but all-NULL.** The break: treating "no usable key values" as "no key column"
  stops measuring a group the engine can measure. Checked: the classifier reads the column list
  only, never the data. The column exists, the query runs, `IS NOT NULL` filters every row, and zero
  rows back is an honest measured pass with `not_evaluated: None`. Pinned by
  `an_all_null_overlap_key_column_is_measured_not_treated_as_keyless`.
- **A warehouse whose "column not found" error wording differs.** Not applicable, and that is the
  design: no adapter error text is read anywhere in the classification. I did not need to audit
  DuckDB, Databricks, Snowflake, BigQuery and Trino error strings, because a substring match on any
  of them was never introduced.
- **A `DESCRIBE` that fails or returns nothing.** The break: treating an unreadable schema as
  "keyless, therefore tolerated" turns a permission error into a silently green run. Checked: both
  `Err(_)` and `Ok(empty)` classify nothing and fall through. Pinned by
  `an_unreadable_overlap_sibling_schema_is_not_treated_as_keyless`.
- **A `key_expr` group with a keyless sibling.** The break: assuming the classifier covers every
  configuration. It does not, deliberately. Such a group still reports the failing `not_evaluated`.
  Pinned by `a_key_expr_overlap_group_is_never_classified_as_keyless`.
- **Severity.** A `severity = "warning"` overlap group is unaffected either way; it never gated. The
  not-applicable result carries the configured severity like every other overlap result.

### Least confident about

Whether "the whole group is not applicable" is the right granularity. A group of three siblings
where only one is keyless is now reported as not applicable in full, so the two that *do* carry the
key stop being compared with each other. Filtering the group to its key-carrying members and
checking those would be strictly more useful. I left it out on purpose: it changes
`contributing_tables` and the meaning of the emitted check, which is a design change the issue did
not ask for. Filed as #1705 rather than smuggled in here.

### Most important thing I might be missing

That a keyless sibling is a *configuration mistake* worth surfacing, not a shape to tolerate
quietly. The code comment and the issue both say the FR accepts it, which is what I built to — but
the FR text itself was not available to me, so that acceptance is taken from the code's own claim
about it rather than verified at the source. If the FR actually says something narrower, the right
answer is a warning-severity failed check, not a pass. That is a one-line change to which
constructor the arm calls; the classification and its tests stay as they are.

## Out of scope

- Dagster's `check_metadata` surfaces no `not_evaluated`, `overlap_count`, `contributing_tables` or
  `sample`, so the reason is visible in `--output json` and the logs but not in the Dagster UI. That
  is #1669's territory and is not made worse here.
- The three-sibling narrowing described above, filed as #1705.

Refs #1654


hugocorreia90:

**Verified:** the keyless / missing-key sibling is classified by a positive `describe_table` pre-check (case-folded through `column_map::build_column_name_set`, the same derivation `column_match` uses), never by adapter error text; only a `keys` tuple is classified, `key_expr` groups keep the failing `not_evaluated` (pinned by `a_key_expr_overlap_group_is_never_classified_as_keyless`); the new `cross_source_overlap_not_applicable` result is `passed: true` with the reason in `not_evaluated`, so #1671's gate (which counts `!passed` only) no longer trips. That is the first passing result carrying `not_evaluated`; the field's description was widened and the schema, Pydantic and TypeScript bindings regenerated with it. One test fails on `main` (`PartialFailure` instead of `Success`); seven controls unchanged; the mutation verdict is a real `FAILED`.

**Judgement calls I agree with:** reporting rather than omitting (omission is the defect #1652 closed, and Dagster would emit nothing), and reason-carrying rather than a bare pass (a bare `overlap_count: 0` is the #1524 shape). **Open, filed as #1705:** a three-sibling group with one keyless member now stops comparing the two that carry the key.

**The residual worth an owner's eye:** "tolerated by design" is taken from the code comment's claim about an FR acceptance criterion; if the FR is narrower, the honest form is a warning-severity failed check — a one-line constructor swap, tests unchanged.

**Independent review: not available today** (Codex credits exhausted until Sep 7); queued for the Sep 7 pass. Owner-merge title stands for the exit-code change (2 → 0 on that run). Merging when green under the owner's instruction.

Issue #1707, opened 2026-09-05
A dangling symlink still reads as "nothing there" for the incidents directory and the scheduler spool


hugocorreia90:
#1668 fixed the `rocky.toml` read: `ErrorKind::NotFound` no longer conflates "no entry at this path" with "an entry is here that cannot be read". Two other sites still make that conflation, on directories rather than files.

`std::fs::read_dir` on a symlink whose target is gone fails with `NotFound`, exactly as it does for a directory that was never created. Both sites map that to "empty".

**1. The incidents count in `rocky brief`** — `engine/crates/rocky-cli/src/commands/brief.rs:1253`

```rust
Err(e) if e.kind() == std::io::ErrorKind::NotFound => (0, None),
Err(e) => return unavailable(format!("incidents directory unreadable: {e}"), ...),
```

A `.rocky/incidents` symlink whose target is gone reports **0 incidents** in the governance brief, instead of taking the `unavailable` branch that already exists one line below. The brief is a report a human reads to decide something, so under-reporting to zero is the wrong direction.

**2. The scheduler's pending demand** — `engine/crates/rocky-core/src/schedule/spool.rs:318`

```rust
Err(e) if e.kind() == io::ErrorKind::NotFound => return Ok(out),
```

A `.rocky/spool` symlink whose target is gone yields an empty pending list, so the scheduler sees no demand rather than an error.

## What is already correct

Checked while fixing #1668, so nobody re-audits them:

- `engine/crates/rocky-mcp/src/tools.rs:7087` (`LeftBehind::inspect`) already uses `symlink_metadata` and reports `Present { symlink: true }`.
- `engine/crates/rocky-cli/src/commands/product.rs:1639` skips a dangling link on purpose, and says so in a comment.
- `engine/crates/rocky-cli/src/commands/review.rs:110` maps `NotFound` to `ReviewMarkerState::Absent`, which means "not approved". That is fail-closed, so a dangling marker symlink blocks rather than approves.

## Suggested fix

Same discriminator as #1668: `std::fs::symlink_metadata` on the directory path. It succeeds for a dangling link, so the two sites can take their existing error branch instead of the empty one. Neither needs a new error type.

Refs #1668

hugocorreia90:
Fixed by #1710. Verified against `main` 27170b61. Both named sites take their error branch on a dangling link.

[...]

Issue #1668, opened 2026-09-05
A dangling rocky.toml symlink reads as no config at all, because FileNotFound covers both


hugocorreia90:
`ConfigError::FileNotFound` is the single signal every caller uses to mean **"this project has no `rocky.toml`"**. It is constructed in exactly one place:

```rust
// engine/crates/rocky-core/src/config.rs — read_config_file
std::fs::read_to_string(path).map_err(|e| {
    if e.kind() == std::io::ErrorKind::NotFound {
        ConfigError::FileNotFound { path: path.to_path_buf() }
    } else {
        ConfigError::ReadFile(e)
    }
})
```

`ErrorKind::NotFound` does not only mean "no such path". It also covers:

- a `rocky.toml` that **is** a symlink whose target was deleted or was never created,
- a `rocky.toml` deleted between the moment a caller decided to load it and the read itself.

In both cases the project plainly *has* a config — it just cannot be read. Every consumer treats that as **absent** and carries on against defaults.

## Why it matters more now

Six call sites already branch on `FileNotFound` to mean absent: `compile`, `emit-sql`, `policy`, `gc`, `apply`, and `apply.rs::load_config_for_gate` (the governance policy gate). #1667 routes nine more offline entry points through one shared loader whose whole contract is *"only `FileNotFound` becomes `None`; everything else refuses"*.

That makes this the single load-bearing discriminator for the class of defect #1625 is about. A dangling symlink walks straight through it and gets the silent-defaults behaviour back:

```
  rocky.toml -> /mnt/shared/configs/prod.toml   (target deleted)

  read_to_string -> ErrorKind::NotFound
                 -> ConfigError::FileNotFound
                 -> "this project has no config"
                 -> compile against defaults / preview on the DuckDB dialect
```

The governance gate has the same shape in a worse direction: `load_config_for_gate` maps `FileNotFound` to `Ok(None)`, and no config means no `[policy]` block to enforce.

## Suggested fix

One place, `read_config_file`. `std::fs::symlink_metadata` succeeds for a dangling symlink because it stats the link rather than the target, so the two cases are distinguishable:

- the path does not exist at all → `FileNotFound` (unchanged)
- the path exists but cannot be resolved or read → a non-`FileNotFound` error

Fixing it there fixes every consumer at once, and moves the governance gate in the fail-closed direction.

## Worth deciding

[...]

PR #1708, opened 2026-09-05
OWNER MERGE — fix(engine/rocky-core): a dangling rocky.toml symlink refuses instead of reading as no config

hugocorreia90:
A `rocky.toml` that is a symlink to a missing file used to read as "this project has no config".
Every command then carried on against defaults. It now refuses, naming the file and the target.

One function changed — `read_config_file` in `engine/crates/rocky-core/src/config.rs`. No call site
was touched, so every loader inherits the fix.

## The defect

`ConfigError::FileNotFound` is the one signal every caller reads as "no `rocky.toml` here". #1667 made
it the single discriminator for the offline entry points that share one loader. It was built from
`std::io::ErrorKind::NotFound`, which does not only mean "nothing is at this path":

```
  rocky.toml -> /mnt/shared/configs/prod.toml   (target deleted)

  read_to_string -> ErrorKind::NotFound
                 -> ConfigError::FileNotFound
                 -> "this project has no config"
                 -> compile against defaults / preview on the DuckDB dialect
                 -> no [policy] block for the governance gate to enforce
```

## The fix

`std::fs::symlink_metadata` stats the link instead of following it, so it still succeeds for a
dangling symlink. That is the discriminator:

```
  read_to_string(path) -> NotFound
           |
           +-- symlink_metadata FAILS with NotFound
           |        nothing is at this path  -> FileNotFound     (absent, unchanged)
           |
           +-- symlink_metadata SUCCEEDS
                    an entry IS at this path -> UnreadableFile   (refuse)
```

`ConfigError::UnreadableFile { path, detail }` is new. Its `detail` names the symlink target it could
not resolve, so the refusal is diagnosable from the message alone.

What each shape yields now:

| Path shape | Read errno | Before | After |
|---|---|---|---|
| Nothing at the path | ENOENT | `FileNotFound` | `FileNotFound` (unchanged) |
| Parent directory missing | ENOENT | `FileNotFound` | `FileNotFound` (unchanged) |
| Dangling symlink | ENOENT | `FileNotFound` | **`UnreadableFile`** |
| Broken symlink chain | ENOENT | `FileNotFound` | **`UnreadableFile`** |
| Symlink to a real file | — | loads | loads (unchanged) |
| Directory at the path | EISDIR | `ReadFile` | `ReadFile` (unchanged, now pinned by a test) |
| Symlink to a directory | EISDIR | `ReadFile` | `ReadFile` (unchanged) |
| Permission denied | EACCES | `ReadFile` | `ReadFile` (unchanged) |

The message a user sees also improves. `try_upgrade_config_error` special-cases `FileNotFound` to
suggest `rocky init`. A dangling symlink no longer takes that arm, so it stops being told to create a
project it already has.

### Tests

Seven new tests in `engine/crates/rocky-core/src/config.rs`. Four are `#[cfg(unix)]`: creating a
symlink on Windows needs Developer Mode or `SeCreateSymbolicLinkPrivilege`, so on Windows those tests
would fail for a reason that has nothing to do with the discriminator. The non-symlink tests run
everywhere.

| Test | Asserts |
|---|---|
| `a_dangling_config_symlink_refuses_instead_of_reading_as_absent` (unix) | not `FileNotFound`; the message names the path, the target, and the word `symlink` |
| `an_absent_config_path_is_still_file_not_found` | the control: absence keeps its own variant |
| `a_config_path_under_a_missing_directory_is_file_not_found` | a missing parent is still absence |
| `a_symlink_to_a_real_config_loads_normally` (unix) | a resolvable symlink is an ordinary config |
| `a_broken_symlink_chain_refuses_and_names_the_immediate_hop` (unix) | a chain refuses, and the message names the hop as written |
| `a_directory_at_the_config_path_refuses_as_a_read_error` | `ReadFile`, at the read site and through the shared loader |
| `optional_project_config_refuses_a_dangling_symlink_but_not_absence` (unix) | the caller-level pairing: dangling is `Err`, absent is `Ok(None)`, same directory |

The assertion that fails on `main` is the first one: on `main` a dangling symlink IS `FileNotFound`.

```
test result: ok. 284 passed; 0 failed; 0 ignored; 1833 filtered out
```

### Mutation check

The mutation reverts the CLASSIFICATION, not the message text: it widens the "nothing is here" arm to
a catch-all, so every `NotFound` reads as absence again — exactly what `main` does.

```
-        Err(stat_error) if stat_error.kind() == std::io::ErrorKind::NotFound => {
+        _ => {
```

```
scripts/mutation-check.sh engine/crates/rocky-core/src/config.rs <mut.py> \
    cargo +1.98.0 test --manifest-path engine/Cargo.toml -p rocky-core --lib config

mutation-check: applied mutation to engine/crates/rocky-core/src/config.rs
test result: FAILED. 281 passed; 3 failed; 0 ignored; 1833 filtered out
    config::tests::a_broken_symlink_chain_refuses_and_names_the_immediate_hop
    config::tests::a_dangling_config_symlink_refuses_instead_of_reading_as_absent
    config::tests::optional_project_config_refuses_a_dangling_symlink_but_not_absence
mutation-check: OK — the test failed under mutation, so it is fix-sensitive.   (exit 0)
```

Exactly the three dangling-symlink tests failed. The four control tests — absent path, missing parent
directory, resolvable symlink, directory at the path — still PASSED under the mutation, which is what
makes them controls rather than co-signers.

### Local gate

All four ran from `engine/`, on `1.98.0`, and passed:

```
cargo +1.98.0 fmt --all --check                                          clean
cargo +1.98.0 clippy -p rocky-core -p rocky-cli --all-targets -- -D warnings   exit 0
cargo +1.98.0 test -p rocky-core --lib config       ok. 284 passed; 0 failed; 1833 filtered out
cargo +1.98.0 test -p rocky-cli --lib -- commands::compile:: commands::emit_sql::
                                                    ok.  39 passed; 0 failed; 1708 filtered out
```

The `rocky-cli` set is the one that exercises the absent-config path — it includes
`emit_sql::tests::a_missing_config_still_emits` and
`emit_sql::tests::emit_refuses_a_malformed_config_but_tolerates_an_absent_one`.

### Codegen

No `*Output` struct changed, so `just codegen` and `just regen-fixtures` were not needed. The diff is
`engine/crates/rocky-core/src/config.rs` and `engine/CHANGELOG.md` only.

The repo requires an independent red team by a different model. Codex is out of credits, and a
same-model Claude reviewer would not satisfy the independence requirement, so none was run. My own
adversarial pass instead, one line per shape I tried to break the fix with:

- **A symlink chain broken at the far end.** `read_link` returns the IMMEDIATE hop, and that hop
  exists — so "target X does not exist" would have been a false statement. The message says the target
  "cannot be resolved", which is true for both the simple and the chained shape. Pinned by a test.
- **A symlink to a directory.** Reading it fails `EISDIR`, not `ENOENT`, so it never reaches the new
  discriminator and stays a `ReadFile` refusal. Verified by syscall probe.
- **A path whose parent directory is missing.** Read and stat both report `NotFound`, so it stays
  `FileNotFound`. This is the shape a wrong fix breaks — `rocky compile --models <dir>` on a project
  with no config file would start refusing. Pinned by its own test.
- **Permission denied on the link target.** The read fails `EACCES`, not `ENOENT`; unchanged, and it
  already refused.
- **A race where the file vanishes between the read and the stat.** The read fails `NotFound`, the
  stat then also fails `NotFound`, and the result is `FileNotFound`. That is the conflation #1668 says
  cannot be resolved by any stat call, and it is not resolved here. The reverse race — the file
  appears between the two calls — lands the non-symlink `Ok(_)` arm and refuses, which is the
  fail-closed direction.
- **`symlink_metadata` failing for some reason other than `NotFound`.** Unreachable in practice: every
  errno that would produce it (`EACCES`, `ELOOP`, `ENOTDIR`) makes the read fail with that same errno
  rather than `NotFound`. It is fail-closed by construction and deliberately not tested — a contrived
  test for it would prove only that the arm compiles.
- **Consumers of `ConfigError`.** Every match on `ConfigError` outside `config.rs` ends in `_ =>`
  (`validate.rs::config_error_diagnostic`, `error_reporter.rs::try_upgrade_config_error`), and the six
  `FileNotFound`-means-absent sites use `matches!` / an `Err(...)` arm. Adding a variant changes no
  coded contract. `rocky validate` reports a dangling symlink under V001 before and after, with a
  better message.
- **The claim that only one place constructs `FileNotFound`.** Verified by grep: `config.rs:6227` was
  the only constructor, and it is now inside `classify_read_not_found`.

### What I am least confident about

That `Ok(_)` non-symlink arm. I cannot construct it on a normal filesystem — it needs the entry to
reappear between two syscalls — so its message ("an entry exists at this path, but reading it reported
the file missing") is unexercised prose. If it ever fires, it refuses, which is the direction I want,
but the wording has never been read by a user.

### What I think is most likely being missed

I assumed the same defect ran through every other `NotFound`-means-absent site and checked instead of
asserting it. Most of them are already correct: `LeftBehind::inspect` in the MCP tools already uses
`symlink_metadata`; `product.rs` skips a dangling link deliberately and says so; and the review
marker maps `NotFound` to "not approved", which is fail-closed. Two sites do still conflate the two —
the incidents directory behind `rocky brief` and the scheduler spool, both on `read_dir` rather than
`read_to_string`. Out of scope here, filed as #1707.

Refs #1668


hugocorreia90:

**Verified:** the discriminator lives in the one constructor of `FileNotFound` (`read_config_file` → `classify_read_not_found`, via `std::fs::symlink_metadata`, naming the link target through `read_link`); no call site changes, so every caller of the shared loader inherits it. Shapes pinned by tests: absent → `FileNotFound` (unchanged), missing parent → `FileNotFound` (unchanged), dangling link and broken chain → the new `UnreadableFile`, link to a real file → loads, directory → `ReadFile` (unchanged, now pinned). One test fails on `main`; the mutation (widen the absent arm back to `_ =>`) is caught with a real `FAILED`.

**Decision the issue left open, accepted:** a file deleted between the decision to load and the read stays conflated with absence — no stat can separate them; the reverse race refuses, which is the fail-closed side.

**Owner-merge reason:** a project whose `rocky.toml` is a dangling symlink ran on defaults and now stops. #1707 carries the two sibling sites (`brief.rs` incidents dir, `schedule/spool.rs`) that still read a dangling link as "nothing there".

hugocorreia90:
Post-merge review done (Claude Opus 5 at max effort, at the owner's direction — Codex is out of credits, so this is not the independent pass `AGENTS.md` asks for). Verdict BLOCK, filed as #1729.

**The discriminator itself is correct, and the review proved it thoroughly:** all eight `ConfigError` consumers match `FileNotFound` narrowly and send every other variant down a bailing arm, so `UnreadableFile` is fail-closed at every one. The MCP surface is a genuine improvement it confirmed: `load_config_for_gate` now maps a dangling link to `PolicyGate::Unloadable`, which the tools refuse — where `FileNotFound` used to resolve to `NotConfigured` and **allow** the draft write.

**The defect is one layer earlier: two commands decide whether to call the loader at all with `Path::exists()`, which follows symlinks.** So a dangling `rocky.toml` answers `false` and the new refusal is unreachable. `rocky state retention sweep` (`state.rs:148`) then sweeps with `StateRetentionConfig::default()`, deleting run history, DAG snapshots and quality snapshots that the project's own `[state.retention]` said to keep — exit 0, no warning, irreversible. `rocky serve` (`main.rs:4347`) is handed `config_path: None`, which silences W004 and W005 on every served compile, defaults the schema-cache posture, and relocates the scheduler's spool: an explicit `--config /etc/rocky/prod.toml` becomes `./rocky.toml`, so webhook demands spool to the current directory.

The self-review grepped every `ConfigError` match arm and the `FileNotFound` constructor. It never grepped for existence probes, so it never saw the gate that runs before its gate. The `[Unreleased]` changelog claim should be narrowed to "every command that loads the config through the shared loader" — the LSP and both `rocky plan` seams also `.ok()` it away.

Issue #1673, opened 2026-09-05
dagster: the Pipes path filters run results by asset key only, so a check result the component did not declare fails the step


hugocorreia90:
When a `rocky` run emits a check result for a check the Dagster component did not declare, the two execution paths do different things. The streaming path drops it. The Pipes path yields it, and Dagster raises `DagsterInvariantViolationError` and **fails the step**.

`integrations/dagster/src/dagster_rocky/component.py`, `_run_filters_pipes` (~line 2766): it filters `invocation.get_results()` by asset key (`include_keys`) and never by check name. Any `AssetCheckResult` whose `(asset_key, check_name)` is not in the multi-asset's `check_specs` therefore reaches Dagster unfiltered.

Probed rather than assumed (by the author of #1664): yielding a result for an undeclared check gives

```
Received unexpected AssetCheckResult. No checks currently being evaluated target asset 't' and have name 'freshness'.
```

## How a user reaches it

Any check the engine produces that the component did not declare. Today that is the stale-state case #1664 documents — `[checks.freshness]` added to `rocky.toml` without a component-state refresh, so the engine emits `freshness` and the component never declared it — and it becomes every check kind as soon as declaration is gated on configuration (#1645's remaining halves). A configuration edit turning into a hard step failure on one path and a silent drop on the other is the defect; neither is the right answer, but they should at least agree.

## What would settle it

Filter Pipes results by `(asset_key, check_name)` against the declared `check_specs`, the way the streaming path effectively does, and log what was dropped so the mismatch is visible. Or declare the mismatch as a check failure on the asset rather than a step failure. Either is a small change in `_run_filters_pipes`; the choice is which failure mode the integration wants.

Pre-existing and wider than freshness. Found by the independent review of #1664; refs #1645, #1619, #1644.

hugocorreia90:
Fixed by #1709. Verified against `main` 27170b61.

The Pipes path now applies the declared-check filter through the same helper the streaming path uses, instead of letting an undeclared result fail the step.

`integrations/dagster/src/dagster_rocky/component.py:2977-2993`

```
result is AssetCheckResult, has asset_key + check_name,
and (asset_key, check_name) not in declared_check_pairs
    -> _undeclared_check_observation(..)  -> yield an AssetObservation, continue
```

A result with no `asset_key` or `check_name` passes through unchanged, because Dagster resolves those itself when unambiguous.

[...]

PR #1709, opened 2026-09-05
fix(dagster): an undeclared check result warns instead of failing the step, and a failed cross-source overlap check is visible

hugocorreia90:
A check result the component never declared no longer fails the step in `pipes` mode, and no longer vanishes in `streaming` mode.
A failed `cross_source_overlap` check is now a named, failed asset check in Dagster instead of reaching nobody.
Python only, under `integrations/dagster/`. 764 tests pass, ruff clean.

```
                       BEFORE                          AFTER
  streaming   result with no spec -> dropped     -> warning + AssetObservation
  pipes       result with no spec -> STEP FAILS  -> warning + AssetObservation
  overlap     never declared, then dropped       -> declared; real verdict on the
                                                    engine's sibling, nothing (a
                                                    Dagster SKIP) on the others
```

## What was wrong

Two defects met at one seam in `component.py`.

**#1673.** The two execution paths disagreed about a check result whose `(asset_key, check_name)` the component never declared. `_emit_results` dropped it silently. `_run_filters_pipes` filtered `invocation.get_results()` by asset key only, so the result reached Dagster, which raised `DagsterInvariantViolationError` and failed the step. Adding `[checks.freshness]` to `rocky.toml` without refreshing the component state is enough to reach it.

**#1669.** `cross_source_overlap` is a *group* check. The engine evaluates it once per `(source_type, table)` sibling group and reports the verdict on one member. The component excluded those names from its declared specs, and the emit side then dropped the one real result. A failed overlap check reached nobody.

## Design choice 1 — undeclared results: warn and observe, on both paths

Both paths now filter against the same `(asset_key, check_name)` set — the one `_asset` already builds for the compliance path — and both call one helper, `_undeclared_check_observation`. It logs a warning naming the asset, the check, the verdict and the likely cause (stale state; refresh with `dg defs state refresh`; a non-default check also needs `surface_configured_checks: true`), and returns an `AssetObservation` carrying the verdict.

An observation is the right primitive because it is unconstrained by the declared specs. It cannot gate a run and no asset-check alert fires on it — declare the spec for that — but the value lands on the asset's timeline instead of nowhere.

`declared_check_pairs` is a **required** keyword argument on `_run_filters_pipes`, with no default, following `declare_freshness` from #1664: an empty default would read as "nothing is declared" and drop everything, and a permissive default is the bug itself.

Rejected: **fail the step as a check failure instead**. That keeps a configuration edit fatal, which is the half of #1673 worth removing.
Rejected: **drop silently but log**. #1669 is exactly a dropped verdict nobody sees; a log line in a compute log is not a record on the asset.

## Design choice 2 — the overlap check gets a home

The candidate names are now declared, and the non-carrier siblings get an explicit non-evaluated result.

Rejected: **declare it on one deterministic sibling**. Read the engine first: `run.rs` attaches the result to `siblings[0]`, the first entry of `assertion_targets`, which is populated in the order tables are materialized. That is run order, not a sortable rule, and under a Dagster subset the sibling set differs per run — with fewer than two siblings copied the check does not run at all. The component cannot predict that asset when it builds specs.

Rejected: **fan the verdict out to every sibling**. `discover.checks.configured_checks` is keyed by table name alone, so `cross_source_overlap:duckdb.orders` is projected under `orders` and reaches a `postgres.orders` asset too. Fanning out would stamp the duckdb group's verdict on an asset that is not in that group.

Rejected: **an `AssetObservation` only**. It is what an *undeclared* result gets. A configured check that the engine names and evaluates deserves to be a check, so alerts and check selections can see it.

Chosen instead:
- Declare the candidate name on every same-named sibling, stamped with `rocky/group_check` in the spec metadata, and only on assets whose Rocky source type the name targets. `_group_check_targets_source_types` reads the name only to decide targeting — the spec name still comes verbatim from discover — and **fails open**: a name whose shape it does not recognise is declared on every same-named asset, never silently dropped.
- The engine's real result lands on the sibling the engine reported it on, as a named check with the engine's own severity.
- Every other sibling gets **nothing**, plus one INFO line naming the sibling that carried the verdict (or saying the group was not evaluated in this run at all).
- That branch runs before the pruned / contained branches: those explain why the *asset* was skipped, which is not why this *check* has no verdict, and the pruned branch would otherwise carry a prior verdict forward as though it were this run's.

Rejected (after orchestrator review of the first push): **an explicit `passed=False` / WARN "not evaluated on this asset" result on the other siblings.** On a three-sibling group that is two warning failures on every run, permanently, for every opt-in user. It reads as a broken pipeline, and it is the same "placeholder that reflects no evaluation" the original exclusion existed to avoid — only amber instead of green.

**Yielding nothing is Dagster's own state for this, not the absence of one.** Dagster writes an `ASSET_CHECK_EVALUATION_PLANNED` record for every declared check in the step. A check with no evaluation keeps status `PLANNED`, and `AssetCheckExecutionRecord.resolve_status` maps `PLANNED` on a finished, non-failed run to `AssetCheckExecutionResolvedStatus.SKIPPED` — the enum's own comment reads "the run finished, didn't fail, but the check didn't execute". Probed on dagster 1.13.17: with two declared checks and one yielded, the unyielded one has exactly one record, status `PLANNED`, event `ASSET_CHECK_EVALUATION_PLANNED`. So the run records a skip for that check, and the last real verdict stays attached to the earlier run id it came from — visible as an earlier run in the UI, not presented as this run's answer.

**What an operator now sees for a FAILED overlap check.** `check_metadata()` was dropping every field the overlap result carries. It now surfaces `overlap_count`, `contributing_tables` and a sample of overlapping keys — and `not_evaluated`, the reason the engine gives when it declines to run a check, on *every* check kind. That is the difference between "no overlap" and "never measured", and the keyless-table / misconfigured-key case is the most likely real failure. `check_metadata()` is shared, so these fields also reach every other consumer of it — `emit_check_results()` and the hand-rolled asset surfaces (`load_rocky_assets`, `dag_assets`), not just the component. Additive in every case.

**Scope, stated plainly.** The named-check half of #1669 reaches only components with `surface_configured_checks: true`; that flag gates the whole configured-check surface and defaults to `false`. Default-config users get the #1673 half: the verdict as an observation with a warning, instead of silence. #1664's rule is untouched — freshness is still declared only when the pipeline configures it. `RockyResource`'s exit-code mapping and `rocky_sdk/types.py` are untouched.

## Evidence

**Fail-before.** New file `tests/test_undeclared_checks.py`, 13 tests. Run against `origin/main`'s `component.py` / `contracts.py` / `checks.py`: **13 failed**. The four the brief asks for:

| Test | On `main` |
|---|---|
| (a) pipes, undeclared `freshness` | `DagsterInvariantViolationError: Received unexpected AssetCheckResult. No checks currently being evaluated target asset 'fivetran/acme/us_west/shopify/orders' and have name 'freshness'` — step fails |
| (b) streaming, same input | step succeeds, `assert 0 == 1` — the verdict was dropped, no observation |
| (c) FAILED overlap on the engine's asset | `KeyError: ('fivetran/acme/us_west/shopify/orders', 'cross_source_overlap_fivetran_orders')` — no such check |
| (d) an unevaluated sibling yields no result, step succeeds | fails — the overlap check is not declared at all on `main`, so the carrier's own result is missing too |

Honest reading of (d): on `main` the spec is not declared, so there is no result of any kind to assert against. It fails trivially. It is a regression guard for this change, not a fail-before test — mutation 3 below is what actually exercises it. The helper-level tests also fail trivially on `main` (`ImportError` — the helpers are new). The file is 15 tests now; all 15 fail against `main`'s sources.

**Mutation checks** (manual; `scripts/mutation-check.sh` is Rust-oriented). Committed first, then reverted one fix at a time, then `git checkout HEAD --` and confirmed a clean tree.

1. Revert #1673 — pipes back to `yield from invocation.get_results()`, streaming back to a bare `continue`:
   ```
   FAILED test_pipes_undeclared_check_does_not_fail_the_step
   FAILED test_streaming_undeclared_check_is_observed_not_dropped
   2 failed, 11 passed
   ```
2. Revert #1669 — restore `if not c.candidate` at the declare site:
   ```
   FAILED test_failed_overlap_check_is_a_named_failed_check_on_the_engines_asset
   FAILED test_sibling_the_engine_did_not_evaluate_gets_no_passing_placeholder
   FAILED test_not_evaluated_overlap_result_carries_the_engines_reason
   FAILED test_group_check_spec_is_not_declared_on_a_foreign_source_type
   FAILED test_execution.py::test_configured_checks_predeclared_when_opt_in
   5 failed, 38 passed
   ```
3. Make the group-check branch emit a placeholder again (the rejected `passed=False` / WARN result), keeping the declaration:
   ```
   FAILED test_unevaluated_sibling_yields_no_result_and_the_step_succeeds
   FAILED test_unevaluated_sibling_logs_where_the_group_was_evaluated
   FAILED test_unevaluated_group_check_with_no_carrier_says_so
   3 failed, 12 passed
   ```
   All three mutations were re-run on the final head.

**Probes, not assumptions.** Against Dagster 1.13.17, with `@multi_asset(can_subset=True)`:
- yielding a result for an undeclared check raises `DagsterInvariantViolationError` and fails the step (confirms #1673);
- a declared check that yields *no* result on a selected asset is tolerated, and its record stays `PLANNED` — which resolves to `SKIPPED` for that run;
- a failed, ERROR-severity, non-blocking check does **not** fail the step;
- an `AssetObservation` for an asset outside the multi-asset is tolerated — this change still refuses to emit one.
- `PipesMessageHandler._handle_report_asset_check` queues a **top-level** `AssetCheckResult`; pipes never nests one inside a `MaterializeResult`, so there is no second place to filter.

**Gates.** `uv run pytest` → 766 passed. `uv run ruff check src/ tests/` → clean. `uv run ruff format --check src/ tests/` → 53 files already formatted. No `cargo`, no `engine/` changes, no codegen surface touched (no `output.rs` edit, so no `just codegen` cascade).

**Changelog.** `integrations/dagster/CHANGELOG.md`, `## [Unreleased]` → `### Fixed`, one entry per issue. `integrations/dagster/AGENTS.md`'s "declared-check filtering" design note said `_emit_results` drops undeclared results; it is now accurate for both paths, plus a new group-check note.

No independent red team ran on this change. Codex is out of credits, and a same-model Claude reviewer would not satisfy the independence requirement, so none was used. Self-review of the cases most likely to break it:

- **A result whose asset key is not in the multi-asset at all.** The helper logs the warning and returns `None` — no observation. Dagster tolerates a foreign-key observation (probed), but it would put an event on a timeline this step does not own. Covered by `test_undeclared_check_on_a_foreign_asset_warns_without_an_observation`.
- **Two siblings both emitting the group check.** Two filters in one op can each carry an overlap result on a different sibling. Each is a real verdict on its own asset, and both are yielded. The same `(asset_key, check_name)` twice is dropped by the existing `yielded_checks` guard, so Dagster never sees the "output returned multiple times" error. The placeholder pass skips anything already yielded, so a sibling with a real verdict never also gets a "not evaluated" one.
- **A check name the engine changes case on.** `sanitize_check_name` does not fold case, so `Cross_Source_Overlap:…` sanitizes to a name that does not match the declared spec. The result is then undeclared: a warning and an observation, not a step failure — which is exactly the failure mode this PR is for. The targeting helper compares the prefix case-sensitively and fails open, so a case change declares the spec on every same-named sibling rather than nowhere.
- **Component state refreshed mid-run.** The specs a step filters against are the ones its own process built, so the filter is always self-consistent within a step. A refresh between runs is the intended way to fix a stale-state warning. A refresh cannot make this change fail a step it did not fail before.
- **Noise, the real cost.** With `surface_configured_checks: false` and many configured checks (say `null_rate` on 20 columns), every run now logs one warning and emits one observation per dropped result where it used to be silent. That is deliberate — "not silent" is the requirement — but it is a volume change for that configuration, and a two-tier rule (warn on failures, aggregate the passes) was rejected as an inconsistency between the two paths. If it proves noisy in practice it should be an aggregate line, filed as a follow-up rather than guessed at now.
- **A new check appears for opt-in users.** On a component with `surface_configured_checks: true` and `cross_source_overlap` configured, the overlap check is declared where it was not declared before. Siblings that do not carry a run's verdict record a skip, not a failure, so no alert fires on them. The one asset that does carry the verdict can now show a failed check — which is the entire point of #1669 — and a failed non-blocking check does not fail a step (probed).
- **What I am least confident about.** How a skipped check reads in the Dagster UI's check badge. The resolved status is `SKIPPED` and the underlying record is per-run, but I have not looked at the rendered badge for a check that alternates between an evaluated run and a skipped one — only at the storage-level status the API returns. If the badge shows the last *evaluation* without marking the skip, an operator glancing at a non-carrier sibling could read an old verdict as current. That is the trade the orchestrator's call accepts against two permanent warnings per run, and it is a UI question, not a code one.
- **The most important thing I could be missing.** Whether real projects hit the sibling case at all through the component. The engine groups siblings by `(source_type, table)` within one `rocky run`; the Dagster component groups assets by the translator's group name and subsets per source. If siblings routinely land in different Dagster groups, or subsetting routinely leaves one sibling per run, the overlap check rarely runs under Dagster at all and the visible outcome is mostly the "not evaluated" result. Settling that needs a live multi-sibling project, not a fixture — the tests here pin the mechanics, not the frequency.

Out of scope, not fixed here: `docs/src/content/docs/dagster/checks.md` lists `cross_source_overlap` among the check kinds but says nothing about the group semantics or the opt-in; and the table-name-keyed projection in `rocky discover` is what forces the targeting helper to exist at all — a `(source_type, table)` key would remove the guesswork. Neither is touched.

**Second push.** The first version of this PR gave the unevaluated siblings a `passed=False` / WARN result. Orchestrator review rejected it as two permanent warnings per run on a three-sibling group. The second commit replaces it with silence plus an INFO line, and the section above records the reasoning and the Dagster `SKIPPED` probe behind it.

Refs #1669, #1673


hugocorreia90:
Orchestrator review. First version yielded a `passed=False` WARN placeholder on every sibling the engine did not evaluate — two permanent warning failures per run for a three-sibling group, the original exclusion's "placeholder that reflects no evaluation" inverted from green to amber. Asked for silence instead; the second push found the Dagster-native form: a declared check with no evaluation stays `PLANNED` and resolves to `SKIPPED` on a finished run (probed on 1.13.17), while the last real verdict stays attached to the run that produced it.

**Verified:** both paths filter on the same `(asset_key, check_name)` set through one helper; an undeclared result logs a warning naming asset, check, verdict and cause and lands as an `AssetObservation` — never fails the step, never vanishes; the group check is declared on every same-named sibling (`rocky/group_check`), the engine's real result lands on its carrier, the others yield nothing plus one INFO line naming the carrier. Three mutations, including the inverted one (re-emit the placeholder → 3 tests fail). 766 pytest, ruff clean; Python only, no codegen cascade.

**Residual, stated honestly by the lane:** the rendered UI badge for a check that alternates evaluated / skipped was not inspected — storage-level status only.

Self-merge class (no step that succeeds today fails). **Independent review: not available today** (Codex credits exhausted until Sep 7); queued for the Sep 7 pass.

hugocorreia90:
Post-merge review done (Claude Opus 5 at max effort, at the owner's direction — Codex is out of credits, so this is not the independent pass `AGENTS.md` asks for). Verdict BLOCK, filed as #1728. Verified on `main`.

**Confirmed clean, so nobody re-audits it:** every caller of `_run_filters_pipes` is updated and it is module-private; the two paths genuinely cannot be given different declared sets (`_asset` builds `declared_check_pairs` from the same `check_specs` it hands to `_emit_results`); the check-name byte-match holds end to end through both sanitizers; `_group_check_targets_source_types` fails open as documented and its matrix covers the collapsed-key case; the foreign-key guard on the observation works, so no event lands on a timeline the step does not own; `check_metadata()`'s new fields are genuinely additive. And the lane's Dagster probe was right at the storage level — an unyielded declared check does resolve to `SKIPPED`.

**Where it breaks is one filter above the one this PR fixed.** `_emit_results` drops any result whose asset is not in `selected_keys` at `component.py:3531`, nine lines before `_undeclared_check_observation` at `:3541`. The overlap verdict attaches to `siblings[0]` — materialization order — so the carrier is not knowable at spec-build time, and a partial selection that spans every source while omitting the carrier sibling discards the whole result. No check, no observation, no log, successful step. That is #1669 reproduced through a different filter, and the Pipes path drops it one layer earlier with the comment "Silent by design".

**Compounding it: nothing reads `check_gate_failed`.** Its only two occurrences in `integrations/dagster/src` and `sdk/python/src` are the model fields themselves. Since #1671 an error-severity failure exits 2 with the flag set, `allow_partial=True` returns the parsed result rather than raising, and `_log_run_diagnostics` logs only table failures. So the engine's own "I failed this run on a check" is discarded at the same seam. Together, a check-gated run can finish as a green Dagster step with no failed check anywhere.

**Three more in the issue.** #1706's not-applicable result renders green with `overlap_count: 0` beside it, which reads as "measured, found nothing" — and `checks.py:166` still asserts `not_evaluated` implies failed. The SKIPPED claim in `AGENTS.md` and the CHANGELOG is wrong at the consumer: `AssetCheckHealthState` falls back to the last completed record, so a sibling that once carried a failing verdict stays degraded on every run it is not the carrier, including after the overlap is fixed. And the engine's pipes emitter hardcodes error severity, so `severity = "warning"` behaves differently in the two modes this PR claims to have unified.

**The load-bearing test gap:** nothing pins the carrier-outside-selection case, because the existing overlap test materializes both siblings. The suite is green on the defect.

PR #1710, opened 2026-09-05
OWNER MERGE — fix(engine): a dangling symlink at the webhook spool refuses instead of reading as no pending demand

hugocorreia90:
A scheduler spool that is a broken symlink read as "no pending demand". It now refuses, naming the path and the link target.

The incidents directory #1707 also names needed no change. It has been correct since #1339. A test now pins that.

## What #1707 asked for, and what was actually there

Line numbers below are on `main` at `4dc1cdb4`. #1707 quoted `brief.rs:1253` and `spool.rs:318`; #1704 has since shifted `brief.rs` by +23.

| Site | On `main` | This PR |
|---|---|---|
| `engine/crates/rocky-core/src/schedule/spool.rs:318` — `list_pending_files` | `NotFound` -> empty list. A dangling `.rocky/pending-demands` link reads as no demand. | Fixed. |
| `engine/crates/rocky-cli/src/commands/brief.rs:1276` — the incidents count | **Already correct.** Lines 1224-1234, added by #1339, call `symlink_metadata` and return `unavailable` for *any* symlink at `.rocky/incidents`, dangling or not, before `read_dir` runs. `incidents.rs:127` guards the sweep the same way. | Not touched. One test added to pin the dangling half. |

**Deviation from the assigned title.** The title I was given said "at the incidents dir or the spool". That half is false, so I retitled to name only the spool. Flagging it here rather than shipping a claim the diff does not contain.

The issue also says the spool lives at `.rocky/spool`. The real path is `.rocky/pending-demands` (`spool_dir`).

## The change

`std::io::ErrorKind::NotFound` names two different worlds and only one of them is "there is none of this yet". `read_dir` on a symlink whose target is gone fails exactly as it does on a directory that was never created.

```
  read_dir(.rocky/pending-demands) -> NotFound
           |
           +-- symlink_metadata FAILS with NotFound
           |        nothing is there -> Ok(empty)   <- unchanged, the ordinary case
           |
           +-- symlink_metadata SUCCEEDS
                    an entry IS there -> Err(...)   <- new
```

The discriminator #1708 wrote for the `rocky.toml` read was private to `config.rs`, so it is lifted, not copied. It now lives in `engine/crates/rocky-core/src/path_presence.rs`, a crate-private module returning a neutral `PathPresence::{Absent, Present { detail }}`. `config.rs::classify_read_not_found` is reduced to mapping that onto `ConfigError`; `spool.rs` maps it onto `io::Error`.

One detail string changed, because the helper now backs a directory read as well as a file read: "reported the file missing" -> "reported nothing there". No test or caller asserts on that text, and it has not shipped (it is in `[Unreleased]`).

The spool's refusal uses `io::Error::other`, so its kind is `Other`, **not** `NotFound`. A future caller matching on the kind cannot reinstate the conflation this closes.

Downstream: `consume_webhook_demands` (`reconcile.rs:584`) is the only production caller. It already logs a warning and skips the webhook step for that tick. So the tick still runs its cron and freshness work. Silent-and-idle becomes logged-and-idle. No panic, no aborted tick, no state change.

### 1. Fail-before tests

Site 2, `schedule::spool::tests` (all 4 new):

| Test | Shape | Asserts | On `main` |
|---|---|---|---|
| `a_dangling_spool_symlink_refuses_instead_of_reading_as_empty` `#[cfg(unix)]` | dangling link | `Err`, kind != `NotFound`, message names the spool path **and** the target | **fails** |
| `an_absent_spool_directory_is_still_no_pending_demand` | no spool at all | `Ok(empty)` | passes (control) |
| `a_spool_whose_parent_is_missing_is_still_absence` | `.rocky` itself absent | `Ok(empty)` | passes (control) |
| `a_symlink_to_a_real_spool_directory_lists_as_before` `#[cfg(unix)]` | link -> real dir | empty when empty; lists 1 after `accept` through the link | passes (control) |

Site 1, `commands::brief::tests`:

- `scheduler_section_refuses_a_dangling_incidents_symlink` `#[cfg(unix)]` — a dangling `.rocky/incidents` link yields `SectionAvailability::Unavailable`, `incident_count == 0`, note contains "symlink". This **passes on `main`**; it is a pinning test, not a fail-before test, and it fails only if the #1339 guard is removed. Said plainly because site 1 needed no fix.
  - Note for evidence item (c): at site 1 a *resolvable* symlink also returns `Unavailable` (`scheduler_section_refuses_a_symlinked_incidents_dir`, already on `main`). The incidents guard is deliberately stricter than the spool's. No test claims a resolvable incidents link reads through, because it does not.

The three symlink tests are gated `#[cfg(unix)]`, so they do not run on the Windows target.

### 2. Mutation check

Committed first, clean worktree, ran on the committed state:

```
scripts/mutation-check.sh engine/crates/rocky-core/src/schedule/spool.rs <mut.py> \
  cargo +1.98.0 test --manifest-path engine/Cargo.toml -p rocky-core --lib schedule::spool
```

The mutation reverts only the new arm (`PathPresence::Present` -> `return Ok(out)`), leaving every import used so the build cannot break and hand back a false exit 3.

```
---- schedule::spool::tests::a_dangling_spool_symlink_refuses_instead_of_reading_as_empty stdout ----
panicked at crates/rocky-core/src/schedule/spool.rs:965:14:
a dangling spool symlink must not read as no pending demand: []

test result: FAILED. 20 passed; 1 failed; 0 ignored
mutation-check: OK - the test failed under mutation, so it is fix-sensitive.
```

Exit **0** - a clean verdict, not the known-false exit 3 of #1547. The `[]` in the panic is the pre-fix answer: an empty list. The three control tests **passed** under the mutation, which is what makes them controls rather than duplicates of the first.

### 3. Changelog

`engine/CHANGELOG.md`, under `## [Unreleased]` -> `### Fixed`, with an **On upgrade** sentence, ending `(#1707)`. Asserted in place: `[Unreleased]` is line 8, the entry line 46, the next `## [` heading line 148.

### 4. Codegen

No `*Output` struct changed. `brief.rs` gained a test only; `BriefSchedulerSection` is untouched. No `just codegen`, no `just regen-fixtures`, no lockfile churn.

### 5. Local gates

| Gate | Result |
|---|---|
| `cargo +1.98.0 fmt --manifest-path engine/Cargo.toml --all --check` | clean |
| `cargo +1.98.0 clippy --manifest-path engine/Cargo.toml -p rocky-core -p rocky-cli --all-targets -- -D warnings` | clean |
| `cargo +1.98.0 test ... -p rocky-core --lib schedule::spool` | 21 passed, 0 failed |
| `cargo +1.98.0 test ... -p rocky-core config::` | 254 passed, 0 failed |
| `cargo +1.98.0 test ... -p rocky-cli --lib commands::brief` | 16 passed, 0 failed |

`AGENTS.md` requires an independent red team for a substantive change. Codex is out of credits, and a same-model Claude reviewer would not satisfy the independence requirement, so none was run and none is presented as one. My own adversarial pass, per shape:

**Spool (changed)**

- **Symlink to a real directory.** `read_dir` succeeds; the new branch is never reached. Tested, including a demand written through the link.
- **Symlink to a real *file*.** `read_dir` fails `NotADirectory`, not `NotFound`, so it takes the pre-existing `Err(e) => return Err(e)` arm. Already an error before this PR, still one. Untested; unchanged by the diff.
- **A broken chain (link -> link -> gone).** `read_link` names the *immediate* hop, which exists, so the message says the target cannot be **resolved** rather than claiming a name is missing. `config.rs::a_broken_symlink_chain_refuses_and_names_the_immediate_hop` still exercises exactly this arm of the shared helper, so the behaviour stays pinned after the lift.
- **Parent missing.** Both calls report `NotFound` -> `Absent` -> `Ok(empty)`. Tested. This is the ordinary case and it must not move.
- **Permission denied on the spool.** `read_dir` fails `PermissionDenied`, never `NotFound`; unchanged. The residual sub-case — `read_dir` says `NotFound` but `symlink_metadata` fails another way — lands in `Present` and refuses. Fail-closed, which is the direction I want.
- **Race, create-after-read.** The spool is created between the two calls: `symlink_metadata` succeeds, not a symlink, so we refuse a healthy spool. Cost is one warned tick; the next tick lists it normally. Self-healing and loud. I accept it over the alternative, which is a silent zero.
- **Race, delete-after-read.** The link is removed between the two calls: `Absent`, `Ok(empty)`, same as pre-fix. The fix fails open here, but there is nothing left behind the link to under-report, so no demand is lost.
- **Windows.** The discriminator should hold (a reparse point still stats without following), but the tests are `#[cfg(unix)]`, so the fix is **unproven by test on Windows**. Stating it rather than implying coverage.

**Incidents (unchanged)**

- Every symlink shape, dangling or resolvable, already returns `Unavailable`. Nothing to change.
- **Residual, and I did not close it.** `brief.rs:1276` still maps a bare `NotFound` to `(0, None)`. After the `is_symlink` guard, that arm is reachable only by a race (a non-symlink entry appearing between the stat and the `read_dir`). Closing it would mean making `path_presence` public across the crate boundary for one race arm, at a site the issue reports as already fixed. Out of scope here; noted so nobody reads the asymmetry as an oversight.
- The same conflation also sits at `spool.rs::count_corrupt` and `spool.rs::sweep_tombstones`. **Neither is reachable in production** (below), so a fix there would be unprovable. Left alone on purpose.

**Least confident about:** the create-after-read race. It turns a transient into a refusal, and I am asserting the cost is one tick from reading the caller, not from running a scheduler against a spool being created concurrently.

**Most important thing I may be missing:** whether any embedder outside this repo calls `list_pending_files` and treats every `Err` as fatal. In-repo the only production caller degrades gracefully, but `list_pending_files` is `pub` in `rocky-core`, so an external caller's error handling is not something I can see from here.

## Out of scope — filed separately, not fixed here

1. **Tombstones are never swept in production.** `reconcile.rs:518` calls `spool::sweep_tombstones(&spool::spool_dir(&opts.rocky_dir), now)`, but `sweep_tombstones` itself calls `spool_dir(rocky_dir)`. The path double-nests to `.rocky/pending-demands/pending-demands`, which never exists, so the sweep always returns `Ok(0)`. The id-dedup window is unbounded and the spool grows without limit. The unit test `sweep_drops_tombstones_past_the_ttl_only` passes because it calls the function directly with the correct `rocky_dir`; no test covers the caller.
2. **`spool::count_corrupt` has no production caller.** Its doc says "for a `rocky doctor` warning", but `doctor.rs` never mentions the spool. Only tests call it. Coded, not wired.
3. **`rocky-core/src/{intern.rs, poison.rs}` are not declared in `lib.rs`.** Neither is compiled. Likely dead files. Noticed while placing the new module; not investigated further.

Out-of-scope items 1 and 2 are filed as #1711.

Refs #1707


hugocorreia90:
Merged `main` in after #1704 landed (`4dc1cdb4`), which changed `build_scheduler` to take `Option<&RockyConfig>`. The conflict in `brief.rs` was purely additive — both sides appended tests — so both blocks are kept and the new pinning test now calls the #1704 signature (`Some(&loaded(&config))`), matching its sibling `scheduler_section_refuses_a_symlinked_incidents_dir`.

`spool.rs` and `path_presence.rs` are byte-identical to the commit the mutation check ran against (`git diff 1bf6cfbc HEAD --` on both is empty), so that evidence stands unchanged.

Re-ran on the merged tree: `fmt --check` clean, `clippy -p rocky-core -p rocky-cli --all-targets -D warnings` clean, `schedule::spool` 21 passed, `config::` 254 passed, `commands::brief` 20 passed. The changelog entry is still inside `[Unreleased]` (line 48; the next `## [` heading is line 156).

hugocorreia90:

**Verified:** the discriminator is lifted, not copied — `path_presence.rs` (crate-private, `PathPresence::{Absent, Present{detail}}`) now backs both `config.rs` (→ `UnreadableFile`, unchanged behaviour) and the spool's `list_pending_files` (→ an `io::Error` that is deliberately not `NotFound`, naming path and target). Absent and missing-parent stay `Ok(empty)`; a link to a real directory lists as before. One test fails on `main`; the mutation verdict is a real `FAILED` with the pre-fix `[]` in the message; three controls pass under mutation.

**Corrections accepted:** the incidents-dir site in `brief.rs` was already guarded (#1339, `symlink_metadata` before `read_dir`) — pinned with a test, not touched — and the title now names only the spool; the issue's `.rocky/spool` path is really `.rocky/pending-demands`.

**Found on the way, filed as #1711 (not fixed here):** `reconcile.rs` passes `spool_dir(rocky_dir)` into `sweep_tombstones`, which calls `spool_dir` again, so production sweeps `.rocky/pending-demands/pending-demands` and always returns `Ok(0)` — tombstones are never swept and the spool grows without bound. The unit test passes because it calls the function with the right path. That is the more important defect of the two.

**Independent review: not available today** (Codex credits exhausted until Sep 7); queued for the Sep 7 pass. Owner-merge title stands (a project with a dangling spool link ran as empty and now stops). Merging when green under the owner's instruction.

hugocorreia90:
Post-merge review done (Claude Opus 5 at max effort, at the owner's direction — Codex is out of credits, so this is not the independent pass `AGENTS.md` asks for). Verdict BLOCK, filed as #1731.

**The spool fix is right and the refusal degrades the tick rather than killing it** — verified at `reconcile.rs:588`: no demand is dropped, dequeued or double-run, and the cron and freshness legs still run. `io::Error::other` gives it kind `Other`, so a future `e.kind() == NotFound` caller cannot reinstate the conflation. The lift is faithful, with no arm lost.

**The defect is that the refusal reaches only a `tracing::warn!`.** It pushes no `SkippedDemand` and `TickReport` has no field that could carry it, so `TickOutput` comes back identical in every health-bearing field to a healthy idle tick — `skipped` empty, `counts` zero, exit 0. That is a codegen'd schema, so it is what the Python SDK and the VS Code extension read, and the resident scheduler's metrics derive from the same struct. The changelog says the reconciler skips the step "rather than reporting zero pending demand"; it logs, and it still reports zero pending demand everywhere a machine looks. The precedent for the fix is already in the enum: `TickSkipReason::HistoryUnavailable` exists for exactly this shape.

Issue #1713, opened 2026-09-05
The spool's other two read sites still read a dangling symlink as an empty directory


hugocorreia90:
Found while fixing #1711. Not fixed there, to avoid a second PR editing the same lines as #1710.

`.rocky/pending-demands` is read from three places in `engine/crates/rocky-core/src/schedule/spool.rs`. Each maps `read_dir` returning `NotFound` to a benign empty answer:

| Function | `NotFound` becomes | After #1710 |
|---|---|---|
| `list_pending_files` | empty list | asks `classify_not_found`, refuses a present-but-unreadable path |
| `sweep_tombstones` | `Ok(0)` | unchanged |
| `count_corrupt` | `Ok(0)` | unchanged |

#1707 is precisely that `NotFound` means two different things: nothing at the path, and a symlink whose target is gone. #1710 closes it for the scan. The other two read sites keep the conflation.

## How much it matters

Less than for the scan, which is why this is separate rather than urgent.

- `sweep_tombstones`: a spool it cannot open has nothing to sweep, so `Ok(0)` is not a wrong answer about the world. It is only a wrong answer about why. After #1710 the tick already refuses the webhook step for that tick, so the operator is told something.
- `count_corrupt`: has no production caller at all (#1712), so it under-reports to nobody today. That changes the moment something wires it.

## What to do

Once #1710 has landed, run the same `classify_not_found` discriminator at both remaining read sites, or say in the module docs why the scan is the only site that needs it. Right now the file has one read site that discriminates and two that do not, with nothing recording the difference as deliberate.

Refs #1707
Refs #1711

hugocorreia90:
Fixed by #1757. Verified against `main` after the merge.

`count_corrupt` and `sweep_tombstones` now ask `classify_not_found` at their `NotFound` arms, so all three spool read sites tell "nothing is at this path" from "something is here and cannot be read".

The three share one `unreadable_spool()` helper, and the test asserts the three error strings are **equal** — not merely that each contains "cannot be read". That matters because the reconciler surfaces the scan's text verbatim as the tick's `spool_unreadable` reason (#1731 / #1752), so a wording difference would read as a cause difference.

No new signal for the sweep: on a dangling link the same tick's `list_pending_files` already fails, so the tick already reports `spool_unreadable`. This makes the sweep consistent with that rather than inventing a second way to say it.

[...]

Issue #1712, opened 2026-09-05
`spool::count_corrupt` has no production caller, and the obvious doctor site is silent where the spool fills


hugocorreia90:
`spool::count_corrupt` counts quarantined webhook demands. Nothing in production reads it — only tests do.

Its doc comment used to say the count backs "a `rocky doctor` warning". That was never true: `doctor.rs` never mentions the spool. #1711 corrected the doc rather than wiring the caller, because the obvious wiring site is wrong in a way worth deciding on its own.

## Why the obvious site does not work

`scheduler_check` in `engine/crates/rocky-cli/src/commands/doctor.rs` already computes `rocky_dir`, so the call itself is two lines. But that check returns `None` — emits nothing at all — when no pipeline declares a `[schedule]`:

```rust
if scheduled.is_empty() {
    return None; // Nothing scheduled — stay silent.
}
```

A webhook does not need a schedule. `webhook_trigger` accepts a demand for any pipeline in `config.pipelines`, and the reconciler consumes it. So a project with zero `[schedule]` blocks can still fill `.rocky/pending-demands` with `.corrupt-*` files, and the scheduler check would stay silent on exactly that project.

Wiring `count_corrupt` into `scheduler_check` would therefore be a half-gate: it warns where webhooks are scheduled, and says nothing where they are not.

## What to decide

Pick one, then wire it and test it:

1. A separate `spool` doctor check, emitted whenever `.rocky/pending-demands` exists. Independent of `[schedule]`.
2. Keep it inside `scheduler_check`, and change that check's silence condition so it also runs when the spool directory exists.
3. Warn from the reconciler tick instead, and drop `count_corrupt` from the doctor idea entirely.

Whichever wins, a corrupt file today is invisible outside a directory listing. A quarantined demand is payload the engine could not parse and deliberately did not delete, so nobody being told about it is the gap.

Refs #1711

hugocorreia90:
Fixed by #1794. Took option 1 — a separate `spool` doctor check, emitted whenever the spool exists, independent of `[schedule]`.

Not folded into `scheduler_check`: that returns `None` when no pipeline declares a `[schedule]`, and a webhook needs no schedule, so it would have warned where it was not needed and stayed silent where it was — the half-gate this issue warned about.

[...]

Issue #1711, opened 2026-09-05
The scheduler never sweeps webhook tombstones: the caller double-nests the spool path


hugocorreia90:
Found while fixing #1707. Not fixed there — it is a different defect with a different blast radius.

## The bug

`sweep_tombstones` resolves the spool directory itself:

```rust
pub fn sweep_tombstones(rocky_dir: &Path, now: DateTime<Utc>) -> io::Result<usize> {
    let dir = spool_dir(rocky_dir);   // rocky_dir.join("pending-demands")
```

Its only production caller passes a path that is **already** resolved (`engine/crates/rocky-core/src/schedule/reconcile.rs:518`):

```rust
if let Err(e) = spool::sweep_tombstones(&spool::spool_dir(&opts.rocky_dir), now) {
```

So the sweep scans `.rocky/pending-demands/pending-demands`. That directory is never created. `read_dir` returns `NotFound`, the function returns `Ok(0)`, and the `if let Err` never fires. The sweep has never removed a tombstone on a real tick.

## Why it matters

`TOMBSTONE_TTL` is 24 hours, and the sweep is what bounds the id-dedup window. Without it:

- `.done` tombstones accumulate in `.rocky/pending-demands` forever, so the spool grows without limit on any project taking webhook demand.
- Every `list_pending_files` scan gets slower as they pile up.
- The dedup window is unbounded rather than 24 hours. That direction is safe (a redelivery is dropped rather than double-run), so this is a growth and drift problem, not a correctness one.

## Why no test caught it

`spool.rs::sweep_drops_tombstones_past_the_ttl_only` calls `sweep_tombstones(dir.path(), ...)` with the correct `rocky_dir`, so the function is proven and the caller is not. A fix should add a reconciler-level test that ticks past the TTL and asserts the tombstone is gone from `spool_dir(&opts.rocky_dir)` — that test fails today.

## Suggested fix

Drop the `spool_dir(...)` at the call site so it passes `&opts.rocky_dir`, matching every other `spool::` call in `reconcile.rs`. Consider whether `sweep_tombstones` should take the spool directory instead, so the two conventions in the module stop being confusable.

## Related, same file

`spool::count_corrupt` has no production caller. Its doc comment says "(for a `rocky doctor` warning)", but `doctor.rs` never mentions the spool; only tests call it. Either wire it into `doctor` or drop the stale claim.

Refs #1707

hugocorreia90:
[...]

PR #1714, opened 2026-09-05
fix(engine/rocky-core): sweep tombstones from the spool directory the scheduler actually uses

hugocorreia90:
The scheduler's webhook tombstone sweep has never removed a tombstone. It swept `.rocky/pending-demands/pending-demands`, a path that never exists, and read the resulting "not found" as zero swept.

The caller now passes the `.rocky` directory, and `spool_dir` returns a `SpoolDir` type so resolving the spool twice is a compile error.

## The defect

`sweep_tombstones` resolves the spool itself. Its one production caller passed a path that was already resolved:

```rust
// before
if let Err(e) = spool::sweep_tombstones(&spool::spool_dir(&opts.rocky_dir), now) {
```

So the sweep ran `read_dir` on `.rocky/pending-demands/pending-demands`. That returns `NotFound`, which the function reports as `Ok(0)` — the branch that exists for a project with no spool at all. The caller's `if let Err` never fired, so nothing was logged either. Every `.done` tombstone written since the feature shipped is still on disk.

The direction is safe. An unbounded id-dedup window drops a redelivery rather than running it twice. This is unbounded growth and a drifted TTL, not a double run.

The existing unit test could not catch it: `spool.rs::sweep_drops_tombstones_past_the_ttl_only` calls `sweep_tombstones(dir.path(), ..)` with the correct argument, so it proved the callee while the caller was wrong.

## The contract, and which side changed

The `.rocky` dir wins, and the caller changed.

Every spool function that resolves the spool for itself already took the `.rocky` dir: `accept`, `accept_journaled`, `list_pending_files`, `count_corrupt`, `sweep_tombstones`. The reconciler's own pending scan, six lines away, passes `&opts.rocky_dir`. Line 518 was the only site passing anything else.

`spool_dir` now returns a `SpoolDir` newtype instead of a `PathBuf`, so the two conventions are no longer the same type:

- Passing a resolved spool dir where the `.rocky` dir is wanted is a compile error, not a silent empty scan.
- No `Deref<Target = Path>`, on purpose. A deref coercion would let a `SpoolDir` slide straight back into a `&Path` parameter and re-arm the confusion. `as_path()`, `join()` and `display()` are the ways out, and each is an explicit act.
- The functions that take one file's path (`dispose`, `is_tombstoned`, `read_pending`, `quarantine`, `drop_pending`) are untouched — they were never part of this.

## 1. Fail-before test, through the production caller

Three tests in `reconcile.rs` drive `tick_once`, not `sweep_tombstones`.

`a_tick_sweeps_an_expired_tombstone_from_the_spool_the_scheduler_writes` — accept a delivery, tick once so it runs and leaves a `.done` tombstone, age that tombstone past the 24-hour TTL, tick again. The assertion that fails on an unfixed tree:

```
a tick must sweep an expired tombstone out of <tmp>/.rocky/pending-demands
```

One detail matters for anyone editing these: `dispose` stamps the tombstone's mtime with real wall-clock, while the sweep compares it against the `now` handed to the tick. `webhook_now()` is a fixed past instant, so a tombstone aged against wall-clock would sit in the tick's *future* and nothing would expire. The helper ages relative to the tick instant, and the function's doc now says so.

Two honest-failure controls:

- `a_tick_sweep_keeps_a_fresh_tombstone_and_never_touches_another_file` — a tombstone one hour inside the window survives (the dedup window must stay closed), and a `.corrupt-` quarantine sitting beside it is untouched.
- `a_tick_sweep_keeps_a_pending_demand_while_removing_an_expired_tombstone` — a paused pipeline's deferred delivery is still pending after the tick that sweeps the expired tombstone. An accepted `202` is never swept.

## 2. Mutation check

Committed first, then:

```
scripts/mutation-check.sh engine/crates/rocky-core/src/schedule/spool.rs <mutation.py> \
  cargo +1.98.0 test --manifest-path engine/Cargo.toml -p rocky-core \
  --lib schedule::reconcile::tests::a_tick_sweep
```

The mutation targets the callee, not the call site, because with the newtype the literal old call site no longer compiles and a build failure is not evidence. It reproduces the pre-fix resolved path exactly:

```rust
let dir = spool_dir(rocky_dir);  →  let dir = spool_dir(rocky_dir).join("pending-demands");
```

Result:

```
running 3 tests
test ..a_tick_sweeps_an_expired_tombstone_from_the_spool_the_scheduler_writes ... FAILED
test ..a_tick_sweep_keeps_a_fresh_tombstone_and_never_touches_another_file ... ok
test ..a_tick_sweep_keeps_a_pending_demand_while_removing_an_expired_tombstone ... FAILED
test result: FAILED. 1 passed; 2 failed
```

The control passing under mutation is the point: it does not assert the sweep ran, so it cannot be a restatement of the fix.

## 3. Changelog

`engine/CHANGELOG.md`, inside `## [Unreleased]` → `### Fixed`, asserted by the insertion script to precede the next `## [` heading. It says what an operator sees on upgrade: nothing printed (the sweep logs only on failure and never logs a count), `.rocky/pending-demands` shrinking once and then staying flat, the one-off delete cost of the first sweep and that it is unmeasured at scale, and the one behavior change — the id-dedup window is 24 hours again, so a redelivery of an id consumed more than a day ago now runs instead of being acked as a duplicate forever.

## 4. Codegen

No `*Output` struct changed. `just codegen` and `just regen-fixtures` are not needed, and no generated file is in this diff.

## 5. Local gate

```
cargo +1.98.0 fmt --all --check                                  pass
cargo +1.98.0 clippy -p rocky-core --all-targets -- -D warnings   pass
cargo +1.98.0 test -p rocky-core --lib schedule                   130 passed, 0 failed
cargo +1.98.0 check --workspace --all-targets                     pass
```

The workspace check is load-bearing here, not routine: `spool_dir`'s return type changed, so a build that passes is proof that no other crate was doing the same double-resolve.

No independent red team ran on this change. What follows is my own adversarial pass, which is not a substitute.

**Is any other caller doubling the path?** No. `grep -rn 'spool_dir' engine --include='*.rs'` finds the definition, four internal resolutions (each on its own `rocky_dir` argument), eight test uses that compute a path for direct filesystem work and never feed it back, and `reconcile.rs:518`. `grep -rn 'pending-demands' engine --include='*.rs'` adds only a doc comment in `rocky-cli/src/api.rs`. The workspace build closes it properly: after the type change, any other doubling site would fail to compile, and none does.

**A tombstone newer than the retention window.** Kept, and pinned by a control. The mtime is stamped at *consumption*, not at accept — `dispose` opens the file, sets the modified time, `sync_all`s it, and only then renames. So the window runs from the moment the demand ran, which is what makes a demand accepted long before it is consumed safe.

**A tombstone whose demand is still pending.** Safe by tick ordering. The webhook consume phase is step 6 and the sweep is step 7. Within one tick, `is_tombstoned` drops the duplicate pending file *before* the sweep can remove the tombstone, so there is no tick in which the same delivery is both dropped as a duplicate and re-run. Once the tombstone expires and is swept, a later redelivery of that id does run — that is the 24-hour window doing its documented job, and it is the one direction where this fix causes a run that would not have happened before.

**A spool that is a symlink.** `read_dir` follows a symlink, so a spool relocated behind a link that resolves sweeps normally. A *dangling* spool symlink makes `read_dir` fail with `NotFound`, which `sweep_tombstones` and `count_corrupt` still map to `Ok(0)` — the same fail-open shape #1710 is closing for `list_pending_files`. I did not change it here: #1710 edits these lines and closing the same conflation twice would conflict. For the sweep it is benign (an unreachable spool has nothing to sweep). Filed as #1713 so the file does not end up with one read site that discriminates and two that do not, with nothing saying whether that is deliberate. Note for the merge: when #1710 lands, its `list_pending_files` calls `classify_not_found(&dir)`, and that function takes `&Path` (verified against the PR diff), so it needs `dir.as_path()` — one line. Its `dir.display()` in the same block compiles unchanged, because `SpoolDir` carries a `display()`.

**The first post-upgrade sweep on thousands of tombstones.** One non-recursive `read_dir`, plus one `remove_file` per expired tombstone, once. The two halves are not the same, so take them apart.

The listing is not new work. `list_pending_files` already lists that same directory on *every* tick and walks every tombstone in it to skip them, so on a large spool that recurring cost is exactly what this fix ends.

The unlinks are new. A spool holding a hundred thousand tombstones pays a hundred thousand metadata writes on that first sweep — work the tick has never done — which is seconds of local filesystem time against roughly a tenth of a second for the listing beside it. Once, and then the directory is small.

It runs last in the tick, after every demand has been consumed, so a slow first sweep delays only the tail of that one tick. Its error is warned and the next tick retries; it never fails the tick. It is synchronous inside the async tick, like every other filesystem step there, so it is not a new *kind* of blocking, but on a very large spool it is a longer one, once. I have not measured it at that scale — the reasoning is structural, not empirical.

**Merge class.** Self-merge. The first sweep cannot fail the tick — its error is warned, never propagated — and its cost is one-off and bounded by what accumulated. Two things are worth a second opinion rather than a blocker: the restored dedup window, which is a behavior change on upgrade rather than a failure mode, and that one-off unlink cost on a very large spool, which I reasoned about and did not measure.

## `count_corrupt`

Doc corrected, caller not wired. The doc said the count backs "a `rocky doctor` warning"; `doctor.rs` never mentions the spool, and only tests call the function.

I did not wire it because the obvious site is wrong. `scheduler_check` already has `rocky_dir` in hand, so the call is two lines — but it returns `None` and emits nothing when no pipeline declares a `[schedule]`, and a webhook needs no schedule: `webhook_trigger` accepts for any pipeline in `config.pipelines`. Wiring there would warn where webhooks are scheduled and stay silent where they are not, which is a half-gate. The placement decision is #1712, and the doc now points at it rather than claiming a caller that does not exist.

## Also filed

- #1712 — `count_corrupt` has no production caller, and the obvious doctor site is silent where the spool fills.
- #1713 — the spool's other two read sites still read a dangling symlink as an empty directory.

## The two AGENTS.md questions

**What am I least confident about?** The restored 24-hour dedup window. After upgrade, a redelivery of a delivery id consumed more than a day ago runs the pipeline where it was previously acked as a duplicate forever. That is the documented `TOMBSTONE_TTL` and the whole point of the sweep, but it is the one place where fixing a growth bug causes a run. A sender that retries on a multi-day cadence would feel it. What would settle it: knowing whether any deployment has been leaning on the unbounded window without knowing it. I cannot check that from here.

**What is the most important thing I am missing?** Most likely the merge with #1710, which edits the same function family and adds `classify_not_found(&dir)` against what is now a `SpoolDir`. I have named the exact line, but I have not built the merged tree. Second: I reason about the first sweep's cost structurally and never measured it against a spool holding a hundred thousand tombstones.

Refs #1711


hugocorreia90:

**Verified:** `spool_dir` now returns a `SpoolDir` newtype (no `Deref`), so the callee cannot be handed an already-resolved path — the only wrong caller (`reconcile.rs`) is the side that changed, and a clean `check --workspace --all-targets` is the proof no other caller doubled it. The fail-before test drives `tick_once`, not `sweep_tombstones`, and asserts the tombstone leaves `.rocky/pending-demands`; the mutation reproduces `main`'s resolved path in the callee (the old call site no longer compiles, which is correctly not treated as evidence) and is caught with two real failures while the fresh-tombstone control passes. `count_corrupt`'s doc no longer claims a caller; its placement is #1712. #1713 carries the dangling-symlink conflation in the two sweep/count paths that #1710 does not cover.

**On upgrade, for the owner:** the documented 24-hour `TOMBSTONE_TTL` is restored. A webhook redelivery of an id consumed more than a day ago used to be acked as a duplicate forever (unbounded tombstones); it now runs. That is the intended contract, but it is the one place this growth fix causes a run — treating it as owner-merge in spirit even though the title is self-merge.

**Merge order:** #1710 is ahead; once it lands, this branch's merge of `main` needs `dir.as_path()` at `classify_not_found(&dir)` (named by the lane). I will apply that when the watcher reports the branch behind or failing.

hugocorreia90:
Post-merge review done (Claude Fable 5.1, at the owner's direction — a different model from the authoring lane but the same family, so not the independent Codex pass `AGENTS.md` asks for). Verdict BLOCK, and I verified each structural finding on `main` df506990.

**The fix itself is correct**: the sweep now runs on `.rocky/pending-demands`, the `SpoolDir` newtype makes the double-resolve a compile error, and the tick-level test holds on the merged tree. Tombstone identity is sound — a hex blake3 of the length-prefixed `(pipeline, kind, token)` — so no forged, colliding or escaping id.

What the review found is what making the sweep real exposes, filed together above: the pause check runs before the tombstone drop, so a paused pipeline can outlive a duplicate's tombstone; `rocky tick --now <future>` is not gated on `--dry-run` and erases the dedup window in one silent pass; and the first post-upgrade sweep is unbounded while the tick lock has no heartbeat.

**On the `--idempotency-key` idea from my decisions memo — the review found five reasons not to implement it as written**, all verified: a deduped child exits 0 with no run record, so the tick would report a skipped redelivery as a successful execution; the key is not pipeline-scoped, so one delivery id fanned out to two pipelines makes the second skip; `dedup_on = "success"` deletes the key on failure, so a failed webhook run would re-run on redelivery where the tombstone dedups on any outcome; the `s3`-only and `gcs`-only backends refuse the flag at parse time, turning every webhook child into a failure there; and an inherited `ROCKY_IDEMPOTENCY_KEY` in the scheduler's environment is passed to every child today, which would make every run after the first skip. Also: idempotency keys **are** pruned (`sweep_idempotency_best_effort`, 30-day default), so that route trades a 24-hour window for a 30-day one rather than closing it. I withdraw the recommendation in its memo form.

Issue #1715, opened 2026-09-05
A never-measured check is persisted as passed, so verify_after gates on it


hugocorreia90:
`RunRecord.check_outcomes` stores `CheckOutcome { name, passed }` (`engine/crates/rocky-core/src/state.rs:2333`). The reason is dropped. Two policy gates then read that record and treat `Some(true)` as "ran and passed":

```
engine/crates/rocky-cli/src/commands/apply.rs:3385   outcomes AND-aggregate c.passed
                                            :3405   Some(true)  => {}                  // gate passes
                                            :3406   Some(false) => "(failed)"
                                            :3407   None        => "(absent — did not run)"
engine/crates/rocky-cli/src/commands/drift_governance.rs:507   the same AND-aggregation
```

The gate is written fail-closed for the absent case: "a named check that did not run cannot be confirmed". A check that *ran and was not measured* is exactly that case, and it now reads as a pass.

## How it is reachable today

#1706 introduced the first passing result carrying `not_evaluated`: `cross_source_overlap_not_applicable` (`passed: true`, reason in `not_evaluated`). A policy rule with `verify_after = ["cross_source_overlap:duckdb.orders"]` on an apply whose group has one keyless sibling now passes the gate on a check that measured nothing. With the mis-typed-key shape in #1705, it passes on every group.

Before #1706 no passing result carried `not_evaluated`, so `passed` was a sufficient proxy. It no longer is.

## What would settle it

Carry the distinction into the persisted outcome — an `evaluated: bool` (serde default `true`, so old blobs read as evaluated) or an `Option<String> not_evaluated` on `CheckOutcome` — and have both gates treat a never-evaluated pass as unverifiable, with the same wording as the absent case. That is a state schema bump (v24 → v25) because the record shape changes; the field is serde-additive, so a v24 blob forward-reads.

Cheaper interim, if the schema bump is not wanted now: make the gates refuse when the named check's result is missing from the run's `check_results` in the JSON as well, or keep #1706's constructor from ever emitting a pass for a group that had carriers (see #1705). Neither closes the general hole.

Found by a post-merge review of #1706. Verified on `main` df506990.

Refs #1706, #1705, #1654.

hugocorreia90:
Two claims in this issue have gone stale since it was filed against `df506990`. Re-verified against `main` 33a73e06.

[...]

PR #1717, opened 2026-09-05
OWNER MERGE — fix(engine/rocky-cli): compare the key-carrying siblings instead of passing the whole overlap group

hugocorreia90:
**Owner merge: runs that pass today will fail.** That is the point — they are hiding a real overlap.

A cross-source overlap group holding one keyless sibling stopped comparing the siblings that *do* carry the key. Three siblings, two sharing a `customer_id`, the third keyless: `passed: true`, `overlap_count: 0`, exit `0`. Before #1706 that exited `2`.

The split is now per sibling. Rocky drops the siblings that carry no key column, names them in the log, and measures the rest.

```
  keys = ["customer_id"]      tgt.orders   customer_id = 5  ┐ a real overlap
                             tgt2.orders   customer_id = 5  ┘
                             tgt3.orders   no customer_id     excluded

  before  ->  cross_source_overlap  passed: true   overlap_count: 0   exit 0
  after   ->  cross_source_overlap  passed: false  overlap_count: 1   exit 2
              contributing_tables: [.tgt.orders, .tgt2.orders]
```

## The partition rule

`partition_overlap_key_carriers` splits a group into carriers and excluded siblings, using the same case-folded derivation the old classifier used (`rocky_core::column_map::build_column_name_set`, both sides lowercased). A sibling is a **carrier** when it carries every configured key column, **or** when the classifier could not judge it. Then:

| carriers | result | `passed` | gates? |
|---|---|---|---|
| 2 or more | the query runs over the carriers; `contributing_tables` is the set measured | measured | on a real overlap |
| exactly 1 | `cross_source_overlap_not_applicable`, reason names what was excluded | `true` | no |
| 0 | `cross_source_overlap_not_evaluated`, reason names the key nothing carries | `false` | per `severity` |

Two things stay exactly as they were, and both are the fail-closed direction:

- A **`key_expr`** group is never classified. `keys` is empty on that path, so every sibling comes back a carrier and the whole group is queried, byte for byte as before. Its query failure keeps the failing `not_evaluated` it has today.
- A **`describe_table` that errors or returns nothing** classifies nothing. That sibling stays a carrier, so the query still runs over the whole group and a permission-denied `DESCRIBE` can never become "the key is tolerated".

Unchanged: the check name, the asset the result attaches to (`siblings[0]`), the severity passthrough, `RunStatus`, `check_gate_failed`, and the persisted `CheckOutcome` (#1715 owns that question). No `*Output` struct or `CheckResult` shape changed, so no codegen cascade.

## What an operator sees

| case | result | reason |
|---|---|---|
| 3 siblings, 2 carriers overlapping, 1 keyless | `passed: false`, `overlap_count: 1`, `contributing_tables` = the 2 carriers | none — it was measured |
| 3 siblings, 2 carriers not overlapping, 1 keyless | `passed: true`, `overlap_count: 0`, `contributing_tables` = the 2 carriers | none — it was measured |
| 2 siblings, 1 keyless (unchanged) | `passed: true`, `overlap_count: 0` | `.tgt2.orders carries no customer_id column; only 1 of 2 siblings carries the configured cross_source_overlap key, so there is nothing to compare` |
| a key no sibling carries | `passed: false`, `not_evaluated` | `.tgt.orders carries no custmer_id column; .tgt2.orders carries no custmer_id column; no sibling in this group carries the configured cross_source_overlap key (custmer_id), so the key cannot be evaluated` |

**One clause I did not fully meet, on the record.** The issue asks for the excluded siblings to be named in the *result* of a measured group. `CheckResult` has one free-text field, `not_evaluated`, and setting it on a measured result would tell every consumer the check never ran — the dagster integration surfaces it as exactly that distinction. So on a measured group the excluded siblings are named in the log (`warn!`) and shown by a `contributing_tables` narrower than the group. Enumerating them in the JSON needs a new `CheckDetails::CrossSourceOverlap` field and the whole codegen cascade; this fix does not take it.

### 1. Fail-before tests

All in `engine/crates/rocky-cli/src/commands/run.rs`, driving the real `run_batched_checks` against a real DuckDB, and reading the real gate (`replication_check_gate_failed`) and the real status derivation.

| test | case | on `main` |
|---|---|---|
| `a_keyless_sibling_does_not_stop_the_carriers_being_compared` | (a) 3 siblings, 2 carriers with a real overlap, 1 keyless | **FAILS** — `passed: true`, not gated |
| `two_carriers_with_no_overlap_are_a_measured_pass_over_the_carriers` | (b) 3 siblings, 2 carriers with no overlap, 1 keyless | **FAILS** — `not_evaluated` is `Some`, never measured |
| `a_key_no_sibling_carries_is_a_failed_not_evaluated_check` | (c) a mistyped key | **FAILS** — `passed: true`, exit 0 |
| `a_keyless_overlap_sibling_is_not_a_failed_check_and_does_not_gate_the_run` (existing) | (d) 2 siblings, 1 keyless | passes — control, unchanged |
| `a_group_where_every_sibling_carries_the_key_runs_the_same_sql` | (e) 3 carriers, nothing excluded | passes — control; asserts the issued SQL is **byte-identical** to `generate_cross_source_overlap_sql` over the whole group |
| `a_key_expr_overlap_group_is_never_classified_as_keyless` (existing) | (f) `key_expr` | passes — control, unchanged |

Plus three from the self-review below: `the_excluded_sibling_may_be_the_attachment_target`, `siblings_missing_different_key_columns_leave_one_carrier`, and a strengthened `an_unreadable_overlap_sibling_schema_is_not_treated_as_keyless`.

`an_unreadable_overlap_sibling_schema_is_not_treated_as_keyless` needed strengthening: it asserted only `!passed` and `not_evaluated.is_some()`, which a zero-carrier skip would *also* satisfy. It now asserts the overlap query was actually issued over both siblings and that the reason is the warehouse's own error, so it discriminates "ran and failed" from "never ran".

### 2. Mutation check

`scripts/mutation-check.sh engine/crates/rocky-cli/src/commands/run.rs <mutation.py> cargo +1.98.0 test --manifest-path engine/Cargo.toml -p rocky-cli --lib -- commands::run::tests::`

The mutation puts the group-wide skip back, in two edits that together reproduce `main` exactly: the skip arm fires on **any** excluded sibling rather than on the carrier count, and the inner arm can never reach `not_evaluated`, so every skipped group is the passing `not_applicable` result.

```
mutation-check: baseline — running the test command against the CLEAN tree
mutation-check: applied mutation to engine/crates/rocky-cli/src/commands/run.rs
...
failures:
    commands::run::tests::a_key_no_sibling_carries_is_a_failed_not_evaluated_check
    commands::run::tests::a_keyless_sibling_does_not_stop_the_carriers_being_compared
    commands::run::tests::the_excluded_sibling_may_be_the_attachment_target
    commands::run::tests::two_carriers_with_no_overlap_are_a_measured_pass_over_the_carriers

test result: FAILED. 291 passed; 4 failed; 1 ignored; 0 measured; 1479 filtered out
---
```

Verdict **OK** (exit 0), not exit 3 — the log shows `test result: FAILED`, so the mutation compiled and the tests ran. The four that die are (a), (b), (c) and the attachment case. The 291 that survive include every control: (d) the two-sibling keyless group, (e) the all-carriers group and its byte-identity SQL assertion, (f) the `key_expr` group, and the unreadable-`DESCRIBE` fall-through. Under the mutation, (c) reports `passed: true` with `overlap_count: 0` — the production fail-open, reproduced.

The worktree was clean before the run and clean after (the script's restore trap verified).

### 3. Local gate

All from the worktree root, on the pushed commit.

```
cargo +1.98.0 fmt --manifest-path engine/Cargo.toml --all --check                                  -> 0
cargo +1.98.0 clippy --manifest-path engine/Cargo.toml -p rocky-cli -p rocky-core \
    --all-targets -- -D warnings                                                                   -> 0
cargo +1.98.0 test --manifest-path engine/Cargo.toml -p rocky-cli --lib -- commands::run::
    test result: ok. 295 passed; 0 failed; 1 ignored                                               -> 0
cargo +1.98.0 test --manifest-path engine/Cargo.toml -p rocky-core --lib checks
    test result: ok. 46 passed; 0 failed; 0 ignored                                                -> 0
```

The three mutex-sharing tests (`run_trigger_from_env_maps_known_values`, `persist_run_record_stamps_trigger_submission_and_pipeline`, `a_run_adopts_an_upstream_traceparent`) all passed on the default thread count; no `--test-threads=1` re-run was needed.

### 4. Codegen

Not needed. No `*Output` struct and no `CheckResult`/`CheckDetails` shape changed; the fix reuses `check_cross_source_overlap`, `cross_source_overlap_not_applicable` and `cross_source_overlap_not_evaluated` as they stand.

### 5. Docs

Both published pages already describe the fixed behaviour, so they are quoted, not edited:

- `docs/src/content/docs/reference/configuration.md:499` — "Sibling **tables** whose key can't be evaluated (missing column / keyless) are **skipped** with a logged reason rather than erroring."
- `docs/src/content/docs/concepts/data-quality-checks.md:327` — "Some sibling **tables** cannot be evaluated: the key column is missing, or the table is keyless. Rocky **skips them** with a logged reason rather than failing the check."

Both say *tables*, plural, skipped individually — per sibling, and via a **logged** reason, which is what this fix does. The same page also says the result "carries the overlap count, the **contributing tables**", which is why `contributing_tables` is the measured set.

One place the code goes past the docs: the zero-carrier case. Neither page covers "no sibling carries the key". `CheckResult.not_evaluated`'s own doc comment does — "a `keys`/`key_expr` misconfiguration … always fails" — and that is the behaviour implemented. No doc edit needed; nothing on either page is now false.

Codex is out of credits, and the Fable pass that found this defect is the same model family as the author. No independent red team ran. What follows is the author's own adversarial self-review, which does not substitute for one.

**A sibling whose key column differs only in case.** Both sides are lowercased — the sibling's columns through `build_column_name_set`, the configured key through `k.to_lowercase()`. `CUSTOMER_ID` against `customer_id` is a carrier. This derivation is unchanged from #1706, and `an_overlap_key_column_differing_only_in_case_is_still_the_key` still pins it. Reading that sibling as keyless would now be worse than before: it would silently drop a table from a comparison that still reports a pass, rather than skipping the group loudly.

**A carrier set of exactly two where one is empty.** The empty table contributes no rows to the `UNION ALL`, so `COUNT(DISTINCT _src)` never exceeds 1 and the group is a measured pass. That is correct — an empty table shares no key with anything — and it is an honest measurement, not a skip: `not_evaluated` stays `None`. No new behaviour; two carriers is two carriers whether or not a third sibling was excluded.

**A group where the excluded sibling is `siblings[0]`.** This is the asset the result attaches to. I kept the attachment there deliberately, and `the_excluded_sibling_may_be_the_attachment_target` pins it: the result still lands on `tgt.orders`'s asset while `contributing_tables` is `[.tgt2.orders, .tgt3.orders]`. The target is still valid — `siblings[0]` comes from `assertion_targets`, so it is a table this run actually copied, and its asset key is the one the runner already had. Moving the attachment to the first *carrier* would be worse: the same group's result would migrate between assets as schemas change, breaking a Dagster check declared against a fixed asset. And the declared side is unaffected either way — `output.rs:4705` keys the declared name by the **table name**, not by a sibling's asset, so the discover/run byte-match guarantee is untouched. The cost is that the JSON can show a check hanging off an asset absent from its own `contributing_tables`; that is the honest reading, and the reason line says why.

**A `keys` list of two columns where different siblings miss different ones.** A carrier must hold *every* key column — `missing` is empty or the sibling is excluded, never partially matched. So with `keys = ["customer_id", "region"]`, a sibling with only `customer_id` and a sibling with only `region` are both excluded, leaving one carrier and a not-applicable result. There is no path that compares a partial key. `siblings_missing_different_key_columns_leave_one_carrier` pins it, including that the reason names each sibling with the column *it* lacks.

**100 siblings — the DESCRIBE cost.** The partition describes every sibling; the old classifier returned on the first missing one. Worst case therefore moves from 1 `DESCRIBE` to N, sequential. Bound: one per sibling, per qualifying (≥2 member) group, per run, and only when `keys` is configured — a `key_expr` group short-circuits before any call. A **healthy** 100-sibling group already paid 100 `DESCRIBE`s on `main`, so this is not a new ceiling, only a new floor for the unhealthy case; and naming *all* the excluded siblings is not possible without it. Not batched, and not concurrent. If a very wide fan-out makes this hurt, the schemas the copy path already prefetches onto `TableTask` are the place to reuse — out of scope here, and I have filed nothing because no such report exists.

**What I am least confident about.** That "zero carriers fails" is right for every existing user. It flips two shapes to failing, not one: the mistyped key, and a two-sibling group where *neither* sibling ever carried the key. The second may have been running green since #1706 and reading as tolerated. The issue and `CheckResult.not_evaluated`'s own description both call a key nothing carries a misconfiguration, and `severity = "warning"` still reports without gating, so I implemented it — but it is the one arm where a user could reasonably have wanted the old silence. It is called out in the changelog.

**What I might be missing.** Whether any consumer treats `contributing_tables` as the declared group rather than the measured set. I checked every reader: `dagster_rocky/checks.py:157` renders it as a metadata string, the SDK and TS bindings are plain lists, and nothing branches on its length or membership. But `contributing_tables` is also the field an operator eyeballs, and a shorter list in a *passing* result is a quiet signal. That is the same gap as the unmet clause above, and #1715's persisted-outcome question is adjacent to it.

Refs #1705, #1654


hugocorreia90:
Follow-up commit `9b2db0b7`: the changelog named only the mistyped-key half of the zero-carrier failure. It now names both shapes and their different remedies — a typo you correct in `rocky.toml`, versus a group where no sibling ever carried the key, which has been passing quietly since 1.73.0+#1706 and needs a decision about whether the check belongs on that group at all. The test that pins this arm, `a_key_no_sibling_carries_is_a_failed_not_evaluated_check`, is the broader shape.

Also re-checked `an_all_null_overlap_key_column_is_measured_not_treated_as_keyless` for lost discriminating power, the way the DESCRIBE test had lost it. It keeps it: both siblings carry the column, so both are carriers and the group is measured; excluding one would give a not-applicable result and trip `not_evaluated.is_none()`, excluding both would trip `passed` as well. No change needed.

hugocorreia90:

**Verified:** `partition_overlap_key_carriers` splits per sibling; `carriers.len() < 2` branches to `not_evaluated` (`passed: false`) when the set is empty and `not_applicable` (`passed: true`) at exactly one, and otherwise the query runs on the carriers with `contributing_tables` narrowed to the measured set. A sibling the classifier could not judge — `describe_table` errored or returned nothing — is counted as a carrier, so it is queried and a real overlap still surfaces. That is the fail-closed default and it is the right one. `key_expr` groups carry empty `keys`, so every sibling is a carrier and the SQL is byte-identical. Two files changed, no `CheckResult` shape change, no codegen cascade. The mutation restores the group-wide skip and kills the three treatment tests plus the attachment test while every control survives — a real `FAILED` verdict, not the #1547 exit-3 shape.

**The disclosed gap, accepted:** a measured group names its excluded siblings in a `warn!` and implicitly through the narrowed `contributing_tables`, not in the JSON. Putting them in `not_evaluated` would tell every consumer the check never ran, which is exactly the distinction Dagster surfaces, and a new details field pulls in the whole codegen cascade. The right call for this PR.

**The zero-carrier decision, on the record:** it flips two shapes, not one — a mistyped key, and a group where no sibling ever carried the key, which has been reading as tolerated since #1706. Both are misconfigurations by the docs' own account and `severity` still governs whether they gate, so failing is right; the changelog names both shapes and their different remedies.

**Independent review: not available** (Codex out of credits; the pass that found this defect is the same family as the lane). Merging under the owner's standing instruction, because this restores documented behaviour that regressed yesterday.

Issue #1716, opened 2026-09-05
Webhook dedup after #1714: a paused pipeline can lose a tombstone, and tick --now can erase the whole window


hugocorreia90:
#1714 made the webhook tombstone sweep real (it had been sweeping a path that never existed, so tombstones were immortal). A post-merge review found two consequences of the tombstone now being able to expire. Verified on `main` df506990.

## 1. A paused pipeline can lose a duplicate's tombstone, and the duplicate runs on resume

`consume_webhook_demands` checks `paused` **before** it checks `is_tombstoned`:

```
engine/crates/rocky-core/src/schedule/reconcile.rs:656   if paused { ...; continue; }      ← deferred here
engine/crates/rocky-core/src/schedule/reconcile.rs:667   if kind == Id && is_tombstoned    ← never reached while paused
```

Sequence: a delivery is consumed and tombstoned; a redelivery slips through the accept-side race the code already documents (`spool.rs:303-306`: the tombstone fast-path passes before `dispose`'s rename, the hard link lands after it) and leaves a second pending file; the pipeline is paused for 24 hours or more; a tick sweeps the tombstone (`spool.rs:577`); on resume `is_tombstoned` is false and the duplicate spawns a second run.

Before #1714 the tombstone was immortal, so the duplicate was dropped. The comment at `spool.rs:306` — that the race can "never" cause a second run — is now stronger than the code.

Fix: drop a tombstoned duplicate **before** the pause check. A duplicate of a consumed id is not a demand to defer.

## 2. `rocky tick --now <future>` erases the whole dedup window, silently

`--now` is documented for "tests, replay, catch-up previews" and is not gated on `--dry-run`. It flows into the sweep:

```
engine/crates/rocky-cli/src/commands/tick.rs:79        let now = now_arg.unwrap_or_else(Utc::now);
engine/crates/rocky-core/src/schedule/reconcile.rs:522 spool::sweep_tombstones(&opts.rocky_dir, now)
engine/crates/rocky-core/src/schedule/spool.rs:577     now - modified >= TOMBSTONE_TTL → unlink
```

`rocky tick --now 2030-01-01T00:00:00Z` (no `--dry-run`) unlinks every tombstone in one pass. The swept count is discarded at `reconcile.rs:522`, so nothing records it. Any redelivery afterwards runs.

[...]

PR #1722, opened 2026-09-05
OWNER MERGE — fix(engine/rocky-core): drop a duplicate webhook demand before the pause check, and stop tick --now from erasing the dedup window

hugocorreia90:
A webhook redelivery could run twice after a long pause. And one `rocky tick --now 2030-01-01T00:00:00Z` could forget every consumed delivery id.

Both are consequences of #1714, which made the tombstone sweep real for the first time. Items 1 and 2 of #1716. Item 3 (the unbounded first sweep, and the tick lock it holds with no heartbeat) is untouched and stays open in #1716.

```
consume_webhook_demands, per spool file

  before                              after
  ─────────────────────────────       ─────────────────────────────
  config-orphan?  → dispose           config-orphan?  → dispose
  paused?         → DEFER  ◄──┐       tombstoned dup? → DROP
  tombstoned dup? → drop      │       paused?         → DEFER
  run                         │       run
                              │
        a duplicate deferred here outlived its own
        tombstone once the sweep became real

tombstone age, two clocks

  dispose  ──stamps mtime── host wall clock
  sweep    ──compared to── the tick's `now`   ← --now <future> unlinks everything
  sweep    ──compared to── host wall clock    ← after this change
```

## 1. A paused pipeline lost a duplicate's tombstone

`consume_webhook_demands` checked `paused` before `is_tombstoned`. A duplicate pending file — the accept-side race `spool.rs` documents, where the tombstone fast-path passes just before `dispose`'s rename and the hard link lands after it — was therefore deferred, not dropped. Deferring keeps the file pending. Once the pause outlasted the 24-hour `TOMBSTONE_TTL`, a tick swept the tombstone out from under it, and on resume `is_tombstoned` was false and the duplicate ran a second time.

The tombstone drop now runs first, and the two blocks are adjacent so the order is one readable statement. A duplicate of an already-consumed id is not a demand to defer.

Nothing else about the pause changes. A non-duplicate demand is still deferred, still recorded in `report.skipped` as `TickSkipReason::Paused`, still fires on resume. The config-orphan check stays ahead of both.

## 2. `--now` could erase the whole dedup window

`sweep_tombstones` aged tombstones against the instant the tick was handed. `--now` is documented for "tests, replay, catch-up previews" and is not gated on `--dry-run`, so one `rocky tick --now 2030-01-01T00:00:00Z` unlinked every tombstone on disk. The swept count was then thrown away, so nothing recorded it.

**Chosen: option (a) — the sweep compares against wall-clock.** `sweep_tombstones` no longer takes an instant at all. Dropping the parameter makes handing it a logical clock a compile error, the same shape #1714 used with `SpoolDir`.

Why not (b), stamp `dispose` with the injected `now`? Two independent reasons, and the first is decisive:

- **It does not close the reported defect.** Every tombstone already on disk was stamped at wall-clock by a real tick. Under (b), `--now 2030` still computes `2030 - mtime(2026) >= 24h` and unlinks all of them. (b) only makes the two clocks agree going forward inside one clock regime.
- It moves the hazard to a past `--now`. `rocky tick --now <two days ago>` would write an already-expired tombstone, which the next ordinary tick sweeps — reopening the exact F5 double-run window `dispose`'s own comment exists to prevent.

Why not (c), skip the sweep when `now` was injected? "Injected" is a CLI notion, so it would have to be threaded into `TickOptions` as a flag. The `(d)` test would then assert the sweep on a configuration that cannot occur in production — a fake clock that is marked not-injected. Under (a) that test is exactly production behaviour. (c) also leaves the ordinary path still comparing a logical instant against a filesystem stamp; they coincide today, but nothing enforces it.

**On the module doc the issue points at.** `reconcile.rs:6-7` and `tick.rs:7-8` said the core reads no wall clock. That claim was already false for this subsystem: `spool::dispose` calls `SystemTime::now()` to stamp the tombstone, and is called from the core reconciler. Rather than argue it away, the claim is narrowed to what is true — every scheduling decision reads the injected `now`; spool file-age GC reads the clock that stamped the mtimes. `grep -rn 'SystemTime::now()\|Utc::now()' engine/crates/rocky-core/src/schedule/` now returns, in production code, only `dispose`'s stamp, this sweep, `spawn.rs`'s `Instant` duration, and `lock.rs`'s heartbeat — which already ages a file the same way, `SystemTime::now().duration_since(modified)`. This change makes the sweep consistent with the lock.

Every doc site that carried the old clock claim was corrected with it:

1. `reconcile.rs` module doc — "the core reads no wall clock"
2. `tick.rs` module doc — the same claim, from the CLI side
3. `tick_once`'s doc — the same claim again
4. `run_tick`'s doc — "the sole clock read in the whole reconciler"
5. the sweep call-site comment in `tick_once`
6. `sweep_tombstones`'s doc — the "age relative to the `now` you will pass here" paragraph, which inverts
7. `consume_webhook_demands`'s doc — the tombstone is consulted before the pause gate, not only before claiming
8. `accept`'s fast-path "can never cause a second run" claim, which item 1 restores
9. the `--now` help text in `main.rs`
10. the `age_tombstone` test helper, whose stated rationale inverted

**The sweep stops being invisible.** It returns `TombstoneSweep { swept, kept, failed }` and the tick logs it: `info` when it removed or failed to remove one, `debug` otherwise. A non-zero `failed` is not automatically an incident — two overlapping ticks racing the same expired tombstone produce one `swept` and one `failed`, and the field's doc names that case first.

Robustness that came with it: the age is now `SystemTime::duration_since`, so a tombstone stamped in the future is `Err` and kept, and the sweep no longer converts an mtime through `DateTime<Utc>`, whose chrono `From<SystemTime>` can panic on an absurd stamp.

### 1. Fail-before tests, driving `tick_once`

All four drive the production entry point, not the spool helpers.

| Test | What it pins |
|---|---|
| `a_paused_pipeline_drops_a_duplicate_instead_of_outliving_its_tombstone` | (a) the item-1 defect, end to end |
| `a_paused_pipeline_still_defers_the_non_duplicate_beside_a_dropped_duplicate` | (b) pause semantics and `report.skipped` are unchanged |
| `a_far_future_now_does_not_sweep_a_tombstone_written_seconds_ago` | (c) the item-2 defect |
| `a_tick_sweeps_an_expired_tombstone_from_the_spool_the_scheduler_writes` | (d) an ordinary tick still sweeps a genuinely expired tombstone |

(a) reproduces the accept-side race the code documents: consume the delivery, hide the tombstone outside the spool, accept the same id again so a duplicate pending file is linked, restore the tombstone. Then pause, age the tombstone past the TTL, tick twice, resume, tick. Its observations are captured rather than asserted mid-sequence, so the run count is what fails first. **On `main` it runs twice.**

(b) puts the duplicate and a genuinely new delivery on the *same* paused pipeline, so "any `Paused` skip" cannot tell them apart. It asserts the count is exactly one, that exactly one file is left pending, and that the file left pending is the non-duplicate (`token == "evt-2"`). Then resume fires it.

(c) asserts the tombstone survives `--now` four years ahead, and states the consequence directly: a redelivery of that id still answers `AcceptOutcome::Duplicate`.

(d) is #1714's test, re-aged. It aged the tombstone relative to the tick instant, because the sweep read that instant; it now ages relative to wall-clock, because the sweep reads the clock that stamps the file. The fresh-tombstone control is held at a deliberate `wall_now - TTL + 1h`, not "brand new", so it exercises the boundary rather than passing trivially.

**Which assertions fail on `main`:**

```
mutation 1 (the reorder)
  a_paused_pipeline_drops_a_duplicate_instead_of_outliving_its_tombstone
    "a delivery id consumed once must never run twice, pause or no pause"
    left: 2, right: 1
  a_paused_pipeline_still_defers_the_non_duplicate_beside_a_dropped_duplicate
    "exactly one deferral: the new delivery, not the dropped duplicate"
    left: 2, right: 1

mutation 2 (the injected clock)
  a_far_future_now_does_not_sweep_a_tombstone_written_seconds_ago
    "a tombstone stamped seconds ago survives `--now` four years ahead"
  a_tick_sweeps_an_expired_tombstone_from_the_spool_the_scheduler_writes
    "a tick must sweep an expired tombstone out of <spool>"
  a_tick_sweep_keeps_a_pending_demand_while_removing_an_expired_tombstone
    "the expired tombstone is swept"
```

Two `spool.rs` unit tests moved to the new signature, and one was added: `sweep_keeps_a_tombstone_whose_mtime_is_in_the_future`.

### 2. Mutation check

Committed first, both against `reconcile.rs`, both scripts asserting the replacement actually changed the file — "a mutation that silently did not apply" is a failure shape `AGENT_REVIEW.md` names, and exit 3 would not distinguish it.

**Mutation 1 — the reorder.** The fix puts the two blocks adjacent, so the mutation is one exact swap, `src.replace(DROP + PAUSE, PAUSE + DROP)`, reproducing `main`'s order byte for byte.

```
scripts/mutation-check.sh engine/crates/rocky-core/src/schedule/reconcile.rs <mut.py> \
  cargo +1.98.0 test --manifest-path engine/Cargo.toml -p rocky-core \
  --lib schedule::reconcile::tests

test result: FAILED. 34 passed; 2 failed
```

**Mutation 2 — the injected clock.** `main`'s literal call site no longer compiles, because `sweep_tombstones` has no instant parameter, and a build failure is not evidence (the reasoning #1714 used). The mutation instead injects a helper reproducing `main`'s comparison exactly — the tombstone's mtime against the instant the tick was handed — and routes the call through it.

```
test result: FAILED. 33 passed; 3 failed
```

The honest-failure control passes under both mutations: `a_tick_sweep_keeps_a_fresh_tombstone_and_never_touches_another_file` asserts survival, which the mutation also produces, so it cannot be a restatement of the fix.

### 3. Changelog

`engine/CHANGELOG.md`, inside `## [Unreleased]` → `### Fixed`, with the insertion asserted to precede the next `## [` heading. It carries an **On upgrade** line: `rocky tick --now <a future instant>` no longer deletes webhook tombstones, and anyone who used it to clear the spool should remove `.rocky/pending-demands/*.done` themselves. Ends `(#1716)`.

### 4. Codegen

Not needed, and checked rather than assumed. No `*Output` struct changed. `TickReport` gained no field, and `build_tick_output` reads only `report`, `now`, `dry_run` and `duration_ms`. The sweep counters reach a log line, not the JSON. So `just codegen` and `just regen-fixtures` do not apply and no generated file is in this diff.

### 5. Local gate

```
cargo +1.98.0 fmt --all --check                                        pass
cargo +1.98.0 clippy -p rocky-core -p rocky-cli --all-targets -D warnings  pass
cargo +1.98.0 test -p rocky-core --lib schedule                        138 passed, 0 failed
cargo +1.98.0 test -p rocky-cli --lib -- commands::tick::              5 passed, 0 failed
```

All four ran and passed. One note for the next lane: `fmt` must be invoked from `engine/`, or with `--manifest-path engine/Cargo.toml`. The repository root has no `Cargo.toml`, so a bare `cargo fmt --all --check` there exits 1 without checking anything — a non-zero exit that is not a formatting failure.

No independent red team ran on this change. The Codex plugin is out of credits, and a Claude reviewer is the same family as the authoring lane, so it would not satisfy the independence `AGENTS.md` asks for. What follows is my own adversarial pass. It is not a substitute.

**A duplicate whose pipeline is paused AND removed from the config.** The config-orphan check still runs before the tombstone drop, deliberately. Such a demand is finalized through `dispose` with a loud warning and a `ConfigError` skip, and never runs — the pause never enters it, exactly as on `main`. One detail: for an `Id` demand `dispose` renames the pending file onto the existing tombstone path, and Unix `rename` replaces the destination, so the tombstone survives with a refreshed mtime. That extends the dedup window by up to 24 hours. Fail-safe direction.

**A tombstone whose `dispose` failed halfway.** Three sub-states, none changed here. Failing at the open/stamp/`sync_all` leaves the pending file untouched and no tombstone; the claim is already terminal, so the next tick re-resolves to `Dedup` and retries the disposal. Failing at the rename is the same. Failing at the directory `fsync` leaves the tombstone in place with a correct consumption-stamped mtime; a retry hits `NotFound` and is idempotent. What this PR changes is that such a tombstone is now aged against the same clock that stamped it — on `main`, one `--now <future>` swept it instantly.

**Two ticks racing the same spool.** The tick lock makes a second tick skip, unless the holder's heartbeat is stale and the override lets both proceed. Two overlapping sweeps calling `remove_file` on one expired tombstone give one `swept` and one `failed`; nothing is deleted twice and no live tombstone is touched. On the consume side, `drop_pending` ignores `NotFound`, so a double drop is idempotent. If a racing sweep unlinks a tombstone between another tick's `is_tombstoned` and its `drop_pending`, the duplicate is still dropped, because `drop_pending` does not re-check. If it unlinks just before `is_tombstoned`, the duplicate runs — that is the 24-hour window genuinely expiring, unchanged.

**A `--now` in the PAST.** After this change `--now` cannot age a file in either direction, so a past `--now` is as harmless to the spool as a future one. It still moves scheduling decisions, which is the flag's documented job. This is the case option (b) would have made worse.

**A clock that goes backwards between the stamp and the sweep.** `wall_now.duration_since(modified)` returns `Err`, the tombstone is counted `kept`, and the dedup window stays closed for longer rather than shorter. Same direction as the pre-#1714 behaviour, and the safe one. Pinned by `sweep_keeps_a_tombstone_whose_mtime_is_in_the_future`.

**Deliberate non-changes, checked and left alone.** A dropped duplicate still records no skip — recording a `Dedup` skip would move `TickCounts` and ripple into the dagster fixtures, and it matches what `main` already does for a non-paused duplicate. `quarantine(&path, now)` still puts the injected instant into a corrupt file's name; cosmetic under `--now`, and out of scope. `docs/…/running-without-an-orchestrator.md` shows `--now` only with `--dry-run` and claims nothing about tombstones, so it needed no edit. Item 3 of #1716 is untouched.

## Merge class

`OWNER MERGE` because option (a) changes an observable effect of a documented flag: `rocky tick --now <a future instant>` no longer deletes webhook tombstones. Nothing on the ordinary path changes — without `--now` the sweep already compared wall-clock against wall-clock.

## The two AGENTS.md questions

**What am I least confident about?** That the reorder is safe for a demand whose pipeline is paused *and* whose store read faults. The pause read is deliberately fail-open — a cursor read failure falls through to normal consumption. The tombstone drop now runs before that read, so a duplicate is dropped even when the store is unreadable, which is the direction I want but is not covered by a test that forces a store fault. I did not find a way to inject one at the `tick_once` level without a new seam, and I judged a new seam out of scope for this fix.

**What is the most important thing I am missing?** The first sweep after upgrading is still unbounded, and it now deletes tombstones for real on every project that ever took a webhook. That is item 3, and it is the operational risk that ships with #1714, not with this PR — but this PR is the one that makes the sweep's outcome visible, so an operator reading a large `swept` count in the log will meet item 3 before it is fixed. It stays open in #1716.

Refs #1716, #1711


hugocorreia90:

**Verified:** the tombstone drop now runs before the pause check, so a duplicate of a consumed delivery is dropped rather than deferred; `sweep_tombstones` no longer takes an instant at all (`fn sweep_tombstones(rocky_dir: &Path) -> io::Result<TombstoneSweep>`), so handing it a logical clock is a compile error rather than a convention; and the sweep's outcome is logged with `swept` and `failed` counts instead of being discarded.

**The clock choice is right, for the reason the lane gives.** Stamping `dispose` with the injected clock would not have fixed anything: every tombstone already on disk was stamped at wall-clock, so `--now 2030` would still unlink all of them. It would also move the hazard to `--now <past>`, writing an already-expired tombstone and reopening the window `dispose`'s own comment exists to close.

**Owner-merge is the correct class:** `rocky tick --now <future>` no longer deletes webhook tombstones, and that is an observable effect of a documented flag.

**Still open in #1716, and worth knowing before this ships:** item 3, the unbounded first post-upgrade sweep with no lock heartbeat. This PR is what makes the sweep's outcome visible, so the first operator to hit it will meet it as a large `swept` count on a tick that held the lock for minutes.

**Independent review: not available** (Codex out of credits; a Claude reviewer is the same family as the lane). Merging under the owner's standing instruction once the branch is up to date.

Issue #1725, opened 2026-09-06
A source-text guard in run.rs stops scanning 1,600 lines early: a #[cfg(test)] helper sits in the production half


hugocorreia90:
`every_partition_status_write_covers_the_whole_batch` in `engine/crates/rocky-cli/src/commands/run.rs`
guards a real invariant: every partition-status write must cover the whole batch, never just the leading key
(#1488 / #1496). It scans the file's own source and asserts `.record_partition(` never appears in production
code.

It cuts the production half at the **first** `\n#[cfg(test)]`:

```rust
let full = include_str!("run.rs");
let src = &full[..full.find("\n#[cfg(test)]").expect("a test module")];
```

There is a `#[cfg(test)] fn build_replication_strategy` at `run.rs:11749`, roughly 1,600 lines above the test
module header at `run.rs:13337`. So `src` ends at 11749 and the guard has never seen the production code
after it. A `.record_partition(` call added below that line passes the test.

Measured on the tip of #1718's branch: cutting at the first `#[cfg(test)]` yields 2 occurrences of
`collect_materialized_table(`; cutting at `\n#[cfg(test)]\nmod tests {` yields the true 4.

#1718 added a second source-text guard (`one_collector_consumes_every_materialized_table`) and anchored it on
the module header instead, plus whitespace-stripped counting so `cargo fmt` wrapping a call cannot zero a
count. The older guard should use the same anchor. Whether it should also strip whitespace is worth a look —
`.record_partition(` is a method call that rustfmt can split.

Refs #1718, #1496

hugocorreia90:
Fixed by #1750. Verified against `main` f0dc26c1.

Both source-text guards go through one helper that owns the anchor and the whitespace stripping:

```
engine/crates/rocky-cli/src/commands/run.rs:13844  fn production_source()
engine/crates/rocky-cli/src/commands/run.rs:13883  the partition guard uses it
engine/crates/rocky-cli/src/commands/run.rs:33786  #1718's collector guard uses it
```

One helper rather than two correct call sites, because the anchor is the easy thing to get wrong and getting it wrong is silent — a third guard should not get to choose again.

## The measurement

Planting `.record_partition(` inside the blind spot:

```
old anchor  sees 0 occurrences  -> MISSED
new anchor  sees 1 occurrence   -> CAUGHT
```

[...]

Issue #1724, opened 2026-09-06
The inline drain cannot act on fail_fast or error_rate_abort_pct, so both settings are silently inert on a default run


hugocorreia90:
#1718 unified the **materialized** arm: one `collect_materialized_table` that the spawn loop's inline drain, the final drain and the auto-retry arm all call. It deliberately left the other arms alone, because they differ in four ways and folding them together would ship a regression rather than fix one.

The divergence is real and is the same class of defect.

```
                              final drain (~run.rs:4506)     process_completed_result (~run.rs:13090)
frame_warehouse_anyhow_error   yes (403/401 framed)          NO — raw `format!("{e:#}")`
materialize_error hook         fires                          does NOT fire
fail_fast -> abort_all         yes                            cannot (it cannot break the caller's loop)
error_rate_abort_pct           checked after the match        absent
```

`ConcurrencyMode::Adaptive` is `#[default]`, so the inline drain is the collector a default-configured run
uses for every table that completes while the spawn loop waits on the semaphore. For those tables:

- a warehouse 403/401 reaches `TableErrorOutput.error` as the raw error chain rather than the framed,
  actionable message;
- the `materialize_error` hook does not fire at all, so an alerting subscriber never hears about the failure;
- `fail_fast` does not abort the remaining tables from this path, and the error-rate abort does not see the
  completion.

The first two are the ones that change what an operator sees. The last two are structural: an inline drain
inside the spawn loop cannot break the spawn loop, so any fix has to return a decision to the caller rather
than act.

Found while fixing #1718.

Refs #1718

hugocorreia90:
The two operator-visible rows landed in #1754. Verified against `main` after the merge.

| | final drain | inline drain |
|---|---|---|
| `frame_warehouse_anyhow_error` | 403/401 framed | **now framed** |
| `materialize_error` hook | fires | **now fires** |
| `fail_fast` → `abort_all` | yes | still cannot |
| `error_rate_abort_pct` | checked | still absent |

One thing had to move with the framing, and it is worth recording because the fix would otherwise have introduced the bug it was fixing: the adaptive throttle's rate-limit detection now runs against the **raw** chain, not the reported message. Framing rewrites auth failures, so leaving `is_rate_limit_error` pointed at the reported text would have silently stopped detecting rate limits. The final drain already detects on `raw` for exactly this reason.

[...]

Issue #1718, opened 2026-09-05
Three collectors consume a materialized table, and two of them silently drop checks — the default concurrency path runs no assertion, null_rate, custom or overlap check


hugocorreia90:
`rocky run` has **three** places that consume a materialized table's `TableResult`, and each one forwards a different subset of the check inputs. Two of them silently drop checks. Verified on `main` df506990.

```
collector                          assertion_targets  column_match  freshness_ref  drift hook
inline arm            run.rs:4470         yes             yes            yes          yes
process_completed_result  :12952          NO              yes            yes          NO
auto-retry arm            :4855           yes             NO             NO           NO
```

## 1. The default concurrency path skips four check kinds

`assertion_targets` is pushed at exactly two production sites, `run.rs:4489` and `run.rs:4886`. It is the **only** input to four loops in `run_batched_checks`: assertions (`run.rs:6300`), null-rate (`:6348`), custom checks (`:6385`) and cross-source overlap (`:6426`).

`process_completed_result` does not push it. That collector is the drain the spawn loop uses inline whenever a throttle is present:

```
run.rs:4288   if throttle.is_some() {
run.rs:4289       while let Some(completed) = join_set.try_join_next() {
run.rs:4290           process_completed_result(...)
```

and `ConcurrencyMode::Adaptive` is `#[default]` (`config.rs:370`). So on a default-configured run, every table collected while the spawn loop waits on the semaphore gets **no** assertion, null-rate, custom or cross-source-overlap check, and nothing reports the omission. Once the semaphore saturates that is most of the table set, not a race.

The comments assert the opposite. `run.rs:6290` says `assertion_targets` is "populated per materialized table", `run.rs:6343` says "every materialized table". Neither is true under the default mode.

## 2. A retried table loses its column_match, freshness and drift

`table_retries` defaults to 1 (`config.rs:497`). The auto-retry arm (`run.rs:4855-4900`) forwards materializations, batch refs, `assertion_targets`, tags and watermarks — and never `tr.column_match_check`, `tr.freshness_batch_ref` or the drift hook. A table whose first copy attempt failed and whose retry succeeded therefore contributes no `column_match` result and no freshness reference.

Against the #1598/#1671 gate both are fail-open: a genuinely column-short target, on the class of table most likely to be unhealthy, passes silently. Against a `verify_after` rule it flips to fail-closed only if the retried table is the run's only table (`column_match` absent reads as "absent — did not run", `drift_governance.rs:535`).

## What would settle it

[...]

PR #1726, opened 2026-09-06
OWNER MERGE — fix(engine/rocky-cli): collect every materialized table through one path, so the default concurrency stops skipping checks

hugocorreia90:
On a default-configured `rocky run`, data-quality checks the user asked for were not running at all.
Three places consumed a copied table's result, and each forwarded a different subset of the check inputs.
All three now call one collector. Checks that were silently skipped now run, and can therefore fail — that is the owner-merge reason.

## What was wrong

```
collector                                    assertion_targets  column_match  freshness_ref  drift action
spawn loop's inline drain (adaptive default)        NO              yes            yes          counter only
final drain                                         yes             yes            yes          yes
auto-retry arm                                      yes             NO             NO           counter only
```

`assertion_targets` is the only input to four check loops — row-level assertions, null-rate, custom SQL
checks and cross-source overlap. The spawn loop drains finished tables inline whenever a throttle is
present, and `ConcurrencyMode::Adaptive` is `#[default]` (`config.rs:370`). So on a default run every table
collected while the loop waits on the semaphore ran none of those four checks, and nothing reported it.
Once the concurrency limit is reached that is most of a large run, not a rare race.

`table_retries` defaults to 1 (`config.rs:497`), so a table that failed once and succeeded on retry is an
ordinary shape. That arm dropped `column_match`, the freshness reference, and the drift action it had just
applied to the warehouse — on the class of table most likely to be unhealthy, and fail-open against the
#1598 check gate.

## The fix

`collect_materialized_table` is the one place a materialized `TableResult` is consumed. It destructures the
struct field by field, so a field added later stops compiling until the collector says what happens to it.
All three sites call it, and the duplicated bodies are gone.

Two differences survive between the callers. Both are named parameters, not omissions:

| parameter | the two drains | auto-retry arm | why |
|---|---|---|---|
| `throttle: Option<ThrottleSignal>` | `Some` | `None` | the spawn loop has joined; there is no permit left to issue. Signalling there would only move the reported `output.execution.final_concurrency`. |
| `emit_materialize_hooks: bool` | `true` | `false` | the retry arm fires no per-table hook today. Turning a retried table into an `after_materialize` / `drift_detected` emit would start outbound webhooks on a path that sends none. That is a separate decision, on its own evidence, not a consequence of this fix. |

`checkpoint_index` is also a parameter: the drains pass their position in completion order, the retry arm
passes the table's own index, which is the row its failed first attempt already wrote.

**One behaviour change beyond the checks.** `after_materialize` and `drift_detected` now fire for tables the
inline drain collects. Which of the two drains happened to collect a table used to decide whether its hook
fired at all — the same defect as the checks, in the hook surface.

**The pruned / failed / panicked arms are NOT unified.** They genuinely differ between the two drains, in
four ways (framed 403/401 auth errors, the `materialize_error` hook, `fail_fast`'s `abort_all`, the
error-rate abort). Folding them together would ship a regression rather than fix one. That divergence is a
real second defect of the same class — filed as #1724, not fixed here.

## The post-copy probe (#1678 review — same file, same failure class)

`post_copy_column_match` treated any adapter error that did not substring-match four rate-limit phrasings as
a settled `not_evaluated`. Since #1598 that fails the run — exit 2 on data that already landed — and
`--resume` then refuses, because everything was copied.

`probe_columns_after_copy` now asks what the pre-copy probe fifty lines up asks,
`warehouse.classify_failure(&e).is_retryable()`, and asks a retryable failure once more.

- **Budget: one extra attempt, no backoff.** The connectors already retry a transient request with backoff
  internally (`rocky-databricks`: `RetryBudget::from_config(config.retry.max_retries_per_run)`), so an error
  that reaches here has escaped that budget. A second layer of waiting would only lengthen the run.
- **It does not propagate, and that is deliberate.** The brief offered "retries, or propagates as it does
  pre-copy". Propagating post-copy would move a copied table into `tables_failed`, and on `incremental` the
  table retry would insert the same rows twice — the invariant #1666 states explicitly ("It never fails the
  table"). A second failure is a probe that genuinely will not answer, and stays `not_evaluated`, which is
  fail-closed for the check (`passed: false`) without failing the table.
- **A rate limit is still not retried** and still reaches the adaptive throttle. The retry returns the LAST
  attempt's error, so a retry that is itself rate-limited reaches the throttle too.

## Changelog corrections (`[Unreleased]`, same PR)

- The #1666 entry said "A failed replication check still does not fail `rocky run` by itself (#1598)".
  False since #1671 landed the #1598 gate. It now says the gate exists and that a failed `column_match` is
  one of the failures it catches.
- The #1598 entry still told the reader that `column_match` "is computed from a read of the target taken
  before the copy, so it fails on the first run of every table". #1666, in the same release, replaced that
  read. The paragraph now states it in the past tense, points at #1666, and keeps the pointer the reader
  needs: under the gate a target that really is short of a column now fails the run.

### 1. Fail-before tests

New tests, all in `engine/crates/rocky-cli/src/commands/run.rs`:

| test | covers |
|---|---|
| `the_default_concurrency_runs_every_table_assertion` | (a) 40 tables (cap is 32), DEFAULT adaptive, one `not_null` assertion each; asserts 40 `not_null:id` outcomes on the persisted `RunRecord`. Runs the same project under `concurrency = 4` as a control and asserts the two runs' check outcomes are identical. |
| `the_inline_drain_contributes_every_check_input` | (a) deterministic: drives `process_completed_result` — the collector a default run uses — and asserts `assertion_targets`, `column_match`, the freshness ref and both row-count refs all land. |
| `a_retried_table_contributes_its_column_match_and_freshness` | (b) drives the collector with exactly the retry arm's arguments (no throttle, no hook emit, the task's own checkpoint index) and asserts `column_match` and the freshness reference land, and the `Success` row is written at that index. |
| `a_retryable_post_copy_probe_is_asked_again` | (c) a `Transient(ServerBusy)` DuckDB failure on the post-copy probe is retried once, answers, the check passes, and the real gate (`replication_check_gate_failed`) does not trip — exit 0. |
| `an_unanswerable_post_copy_probe_is_still_not_evaluated` | (d) control: a permanent failure is asked once, stays `not_evaluated`, `passed: false`, and still gates the run. |
| `a_retry_that_is_rate_limited_still_reaches_the_throttle` | the retry returns the last attempt's error, so a 429 on the second attempt sets `probe_rate_limited`. |
| `a_rate_limited_post_copy_probe_is_not_retried` | a 429 on the first attempt is not asked again. |
| `one_collector_consumes_every_materialized_table` | structural: exactly one `.assertion_targets.push(` and exactly four `collect_materialized_table(` (one definition, three call sites) in the production half. Whitespace-stripped, so `cargo fmt` wrapping a call cannot zero a count. |

**(e) healthy runs are unchanged.** Stated as an invariant rather than a captured baseline, because it is
checkable inside one test: under `ConcurrencyMode::Fixed` the inline drain never runs (`throttle.is_some()`
is false), so every table goes through the final drain, which already carried the full body. The e2e test
asserts the Fixed and Adaptive runs produce identical `check_outcomes`. A single-table run also never takes
the inline path — the drain runs before each spawn, and nothing has been spawned at the first one.

**Which assertions fail on `main`:**

- Measured, under mutation (see §2 — the collector is now shared, so a mutation of it is the closest
  reachable reproduction of `main`'s inline drain):
  `the_default_concurrency_runs_every_table_assertion` fails at
  `assertion left == right … left: 0, right: 40` with `Outcomes: []`;
  `the_inline_drain_contributes_every_check_input` fails at "the inline drain must contribute the table's
  assertion target"; `a_retried_table_contributes_its_column_match_and_freshness` fails at "a retried table
  must still contribute its column_match result: []"; `a_retryable_post_copy_probe_is_asked_again` fails at
  "the retryable probe must be asked exactly twice … left: 1, right: 2".
- **Reasoned, not measured:** on `main` specifically, the adaptive run of the e2e test reports *fewer than*
  40 while the fixed run reports 40, so both the count assertion and the cross-mode equality assertion fail.
  The mechanism: the inline drain calls `try_join_next()` before `acquire_owned()`, and with 40 tables at a
  cap of 32 the loop must wait for a permit on the last 8 spawns, each wait guaranteeing a completed task
  for the next iteration's drain to collect. I did not run this against `main` — measuring it needs a branch
  switch, and `engine/target` is shared with two other active lanes.

### 2. Mutation checks

All via `scripts/mutation-check.sh`, on the committed tree, `cargo +1.98.0 test --manifest-path engine/Cargo.toml -p rocky-cli --lib -- <tests>`.

| # | mutation | verdict |
|---|---|---|
| M1 | delete `sinks.assertion_targets.push((target_ref, asset_key.clone()));` in the collector | **OK — fix-sensitive.** `the_default_concurrency_runs_every_table_assertion` (0 vs 40) and `the_inline_drain_contributes_every_check_input` both failed. |
| M2 | delete the `if let Some(check) = column_match_check { … }` block in the collector | **OK — fix-sensitive.** `a_retried_table_contributes_its_column_match_and_freshness` and `the_inline_drain_contributes_every_check_input` both failed. |
| M3 | delete the retryable branch in `probe_columns_after_copy` (always return the first error) | **OK — fix-sensitive.** `a_retryable_post_copy_probe_is_asked_again` failed ("asked exactly twice … left: 1"). The control `an_unanswerable_post_copy_probe_is_still_not_evaluated` still passed, which is correct: a permanent failure behaves the same either way. |

No `INCONCLUSIVE` (exit 3) verdicts, so #1547's false-negative shape did not arise.

### 3. Changelog

`engine/CHANGELOG.md`, `## [Unreleased]` → `### Fixed`, first entry. Verified to sit inside `Unreleased`
(line 8) and above the next `## [` heading (`## [1.73.0]`, line 192). It says plainly that runs on the
default concurrency have been skipping assertion, null-rate, custom and overlap checks for the tables
drained inline, and that a check that never ran may now fail. Ends `(#1718)`.

### 4. Codegen

No `*Output` struct changed. `git diff --stat` is `engine/CHANGELOG.md` and
`engine/crates/rocky-cli/src/commands/run.rs` only, so no `just codegen` and no `just regen-fixtures` are
needed and no generated file is touched.

### 5. Local gate

`cargo +1.98.0 fmt --all --check` clean. `cargo +1.98.0 clippy -p rocky-cli --all-targets -- -D warnings`
clean. `cargo +1.98.0 test -p rocky-cli --lib -- commands::run::` — 297 passed, 0 failed, 1 ignored
(the Databricks live-sandbox test). The three mutex-sharing trigger tests
(`run_trigger_from_env_maps_known_values`, `persist_run_record_stamps_trigger_submission_and_pipeline`,
`a_run_adopts_an_upstream_traceparent`) passed in the normal parallel run; no `--test-threads=1` rerun was
needed.

Codex is out of credits, and a Claude reviewer is the same model family as the author, so it would not
satisfy the independence requirement in `AGENTS.md`. Recording that rather than skipping it silently. My own
adversarial pass, on the five shapes most likely to break:

**A table that completes between the last spawn and the final drain.** It is never seen by the inline drain
(the spawn loop has exited), so it arrives at the final drain's `join_next()` and goes through the same
collector with the same arguments as any other drained table. Nothing about it is special after this change;
before it, it was one of the tables that happened to get the full body.

**A run with no throttle.** `throttle.is_some()` is false, so the spawn loop never calls the inline drain at
all and every table reaches the final drain. The final drain still constructs a `ThrottleSignal`, whose
inner `if let Some(t) = signal.throttle` is `None` and no-ops — identical to `main`'s `if let Some(t) =
&throttle`. Verified by the Fixed-concurrency half of the e2e test, which produces byte-identical check
outcomes to the adaptive half.

**A retry that is itself rate-limited.** `probe_columns_after_copy` returns the LAST attempt's error, so
`post_copy_column_match`'s `read()` computes `rate_limited` from the 429 rather than from the blip that
preceded it, and the throttle is still told. Pinned by
`a_retry_that_is_rate_limited_still_reaches_the_throttle`, which asserts both `probe_rate_limited` and that
the `not_evaluated` reason carries the 429 text.

**A hook that must not fire twice.** `after_materialize` fires at most once per table. A table has exactly
one outcome per attempt: attempt 1 either succeeds (fires `after_materialize`, is never retried) or fails
(fires `materialize_error`, and only then enters the retry arm — which fires nothing). The retry arm's
`emit_materialize_hooks: false` also means the change cannot introduce a second fire even if that reasoning
were wrong. `before_materialize` still fires once, from the spawn loop, unchanged.

**A run cancelled mid-drain.** The `if interrupted { … }` block (`run.rs:4605`) ends in
`return Err(anyhow::Error::new(Interrupted))` at `run.rs:4750`, which is above both the auto-retry arm
(`run.rs:4753`) and `run_batched_checks` (`run.rs:4952`). So an interrupted run reaches no check at all, and
the newly-populated `assertion_targets` changes nothing for it. What that block does read —
`deferred_watermarks` for the terminal flush, and `output.materializations` for the partial JSON — is
written by the collector at exactly the point the inline drain wrote it before. The checkpoint the collector
writes is the same `TableStatus::Success` row at the same index, so the block's
`plan \ {Success|Failed|Skipped}` diff sees the same set.

**What I am least confident about.** The e2e test's sensitivity on `main`. It is deterministic *after* the
fix (all 40, always) but its failure on `main` depends on at least one task surfacing through
`try_join_next()` during one of the 8 forced semaphore waits. Tokio does not contract the ordering between
a permit's release and the JoinSet's completion notification, so a single wait could in principle race. With
8 of them the probability that none surfaces is negligible, and the failure direction is the safe one: it
would fail to catch the bug, never fail spuriously. The deterministic guards are
`the_inline_drain_contributes_every_check_input` and `one_collector_consumes_every_materialized_table`.

**The most important thing I might be missing.** Whether firing `after_materialize` and `drift_detected` for
inline-drained tables is wanted. It is the same defect as the checks — whether your hook fired depended on
which loop collected the table — and I judged consistency the right call, but it is the one part of this
change that sends new outbound webhooks and commands on the default configuration. A run with 40 tables
under adaptive concurrency will now fire roughly 40 `after_materialize` hooks where it fired far fewer. If
that is not wanted, the fix is one argument: pass `false` at the `process_completed_result` call site, with
the reason written down. I would rather be told than guess.

## Out of scope, filed not fixed

- **#1724** — the pruned / failed / panicked arms diverge between the two drains in four ways. Same class,
  different arms; unifying them here would have shipped a regression.
- **#1725** — `every_partition_status_write_covers_the_whole_batch` cuts its source scan at the first
  `#[cfg(test)]`, which is a test-only helper at `run.rs:11749`, roughly 1,600 lines above the test module.
  Its `.record_partition(` guard has been blind to everything below that line. The new structural test in
  this PR anchors on the module header instead.

Refs #1718, #1666


hugocorreia90:

**Verified:** both production `assertion_targets.push` sites are gone from the arms and there is now one `collect_materialized_table`, called by the inline drain, the final drain and the auto-retry arm. The post-copy probe now asks `warehouse.classify_failure(&e).is_retryable()` rather than substring-matching four rate-limit phrasings, so a transient metadata error no longer settles as `not_evaluated` and exits the run 2 on data that landed.

**The test I like most is the structural one:** a source grep asserting a single `.assertion_targets.push(` in production. Three collectors diverged precisely because the same work was written three times, and a test that fails when a fourth appears is what stops that recurring. That is the right shape for this class.

**What this closes, for the record:** on the default `ConcurrencyMode::Adaptive`, every table drained inline was producing no assertion, null-rate, custom or cross-source-overlap result at all, silently, and the two comments in `run_batched_checks` asserted the opposite. A retried table (default `table_retries = 1`) was additionally losing `column_match` and its freshness reference. Both predate #1678; reviewing that PR is what surfaced them.

**Independent review: not available** (Codex out of credits; a Claude reviewer is the same family as the lane). Owner-merge is right: checks that silently did not run now run, and can fail. Merging when the updated branch is green.

hugocorreia90:
On the hook question you flagged twice: **keep it firing.** Whether a hook fired used to depend on which drain happened to collect the table, which is scheduling, so no consumer could have been relying on the old subset — and a consumer that cannot take a per-table `after_materialize` was already broken without knowing it. Consistency is the right call.

What the entry was missing is the size. "Tables the inline drain collects" is opaque; under the default adaptive concurrency it is most of the set once the semaphore saturates. I pushed `cc477702` adding that: a 40-table run can send up to 40 of each hook where it previously sent a handful, with a line to check a rate-limited or per-call-billed endpoint before upgrading. Nothing else changed.

Also noting your two out-of-scope finds are the good kind: #1725 is a guard that silently stopped guarding — `full.find("\n#[cfg(test)]")` truncates at the *first* match, and `run.rs` has one at 11813 and another at 13209, so the scan has been blind to ~1,400 lines. That is worth fixing before anything relies on it again.

Issue #1732, opened 2026-09-06
A failed verify_after gate still admits an empty resume when the check gate did not trip


hugocorreia90:
Found while fixing #1720, and deliberately left out of that fix because the cause is different and narrowing the branch is a behaviour change beyond that issue.

## What happens

A failed post-apply `verify_after` gate pushes a `<verify_after>` entry onto `output.errors` (`rocky-cli/src/commands/run.rs`, the `finalize_drift_verify_after` error branch). `to_run_record` writes one `models_executed` entry with `status: "failed"` per errors entry, so the record carries a failed "model" that is not a model.

`ensure_the_resume_would_do_work` admits a complete checkpoint whenever a model failed, on the stated reasoning that "the model phase re-runs on a resume". A `<verify_after>` entry is not a model, so nothing re-runs. The resume skips every `Success` key, copies nothing, builds no check input, and derives `Success`.

#1720 closes this for the case where the check gate ALSO tripped — `RunRecord::check_gate_failed` (state schema v25) is read before the failed-model branch and refuses. It does not close the case where `verify_after` fails and the gate does not:

- `finalize_drift_verify_after` fails closed on a required check that is **absent** from `RunRecord::check_outcomes`. Absence is not a failing check, so `replication_check_gate_failed` sees nothing to gate on.
- `[pipeline.<name>.checks] fail_on_error = false` makes every check advisory, so the gate is always `false` while `verify_after` can still fail.

In both, the record carries `check_gate_failed: false` and a failed `<verify_after>` model entry, the resume is admitted, and it records `Success` — over an auto-applied schema migration whose verification could not be confirmed.

## What would settle it

Either the failed-model branch stops counting synthetic `<…>` entries as "a model the resume re-runs", or the `verify_after` custody outcome is persisted the way #1720 persisted the check gate, and read by the same refusal. The first is smaller; the second matches the shape #1720 established and is checkable by a later reader.

Refs #1720, #1598

hugocorreia90:
Fixed by #1749. Verified against `main` 01958217.

Taken the second way this issue offered — the verdict is persisted the way #1720 persisted the check gate, and read by the same refusal — because the first way rests on a `<…>` naming convention nothing enforces.

```
engine/crates/rocky-core/src/state.rs:2445      RunRecord::verify_after_failed   (schema v26)
engine/crates/rocky-cli/src/commands/run.rs:5665   stamped in the failure branch
[...]

Issue #1720, opened 2026-09-05
A resume can launder a check-gated failure into a Success record, because check_gate_failed is never persisted


hugocorreia90:
A `rocky run` that correctly fails its check gate can be turned into a green `Success` record by a resume that copies nothing, runs no check, and re-evaluates nothing — and the failing run's own error message tells the operator to do it. Verified on `main` df506990.

## Root cause

`check_gate_failed` reaches `RunOutput` and stops. It appears **zero times** in `engine/crates/rocky-core/src/state.rs`, so it is never persisted on `RunRecord`. The resume gate therefore cannot ask "did the checks gate this run"; it infers resumability from `models_executed`, a signal with no relationship to check severity:

```
run.rs:1720   if copied < total_tables            -> resume (a real copy remains)
run.rs:1724   if models_executed.any(failed)      -> resume  ← admits the empty case
run.rs:1731   else bail "nothing to resume"
```

And an empty resume derives green:

```
output.rs:5297  has_progress = tables_copied > 0 || !materializations.is_empty()   // false
output.rs:5299  has_problem  = interrupted || tables_failed > 0 || check_gate_failed  // false
output.rs:5300  (_, false) => Success
```

## Two triggers

**A — no `--all` needed.** A `[[policy.rules]]` with `verify_after = ["column_match"]`. One rate-limited `DESCRIBE` makes `column_match` not-evaluated, which trips the check gate and fails the verify-after gate; `run.rs:5435` then pushes a `<verify_after>` entry onto `output.errors`, which `to_run_record` writes as a `models_executed` entry with `status: "failed"`. Every table is `Success` in the checkpoint. Then:

```
rocky run --resume <id>
  -> run.rs:1724 admits (a "failed" model entry exists)
  -> the resume filter drops every Success key, tables_to_process is empty
  -> no check_results push, so check_gate_failed = false
  -> derive_run_status -> Success, exit 0, record Success
```

**B — the higher-frequency one, and the tool suggests it.** Three tables, `row_count = true`; two copy, one fails, and one of the copied tables fails its row-count check. The run correctly exits 2 with:

```
1 table(s) failed during parallel execution (run_id: X, use --resume X to retry)
```

[...]

Issue #1549, opened 2026-08-29
bug(engine): resume checkpoints carry no pipeline scope, so --resume-latest can cross pipelines


hugocorreia90:
## Description

A resume checkpoint records no execution identity. `RunProgress` stores `run_id`, `started_at`, `total_tables` and a table list — no pipeline name, no `--filter`, no target routing, no branch or shadow config, no config fingerprint.

`--resume-latest` therefore selects the most recent checkpoint in the **project**, not the most recent checkpoint *for the pipeline you asked to resume*. In a multi-pipeline project the checkpoint you resume from may belong to a different pipeline than the one about to run.

`TableProgress.table_key` is the fully-qualified **target** name (`catalog.schema.table`, set from `tr.target_full_name`). That is what decides how the mismatch shows up, and it gives two distinct outcomes:

- **Disjoint targets** — the wrong pipeline's keys match nothing, so every table rebuilds, but the run reports `resumed_from: <a different pipeline's run>`. Misleading, not destructive.
- **Overlapping targets** — two pipelines whose `catalog_template` / `schema_template` resolve to the same names. Keys collide, and the run **skips tables based on another pipeline's progress**. Nothing forbids that config.

This predates #1544. What changed is the confidence placed in the lookup: `resolve_resume_progress` now hard-fails when no checkpoint is found, so a successful lookup reads as "my checkpoint was found". It means *a* checkpoint was found.

## Version

- Read on `54619294` (main, after #1544).

## Steps to reproduce

Config with two replication pipelines, `p1` and `p2`.

1. `rocky run --pipeline p2` — succeeds, writes the latest checkpoint.
2. `rocky run --pipeline p1 --resume-latest`.

Step 2 resolves p2's checkpoint. With disjoint targets it rebuilds all of p1 and reports `resumed_from` pointing at p2's run. With target templates that resolve to the same names, it skips p1 tables whose keys appear in p2's checkpoint.

I have not built the two-pipeline fixture; the finding is from reading `get_latest_run_progress` and the `RunProgress` / `TableProgress` definitions, and the second outcome depends on a target-name collision I have not exercised end to end. That fixture is the first thing to write.

## Expected behavior

A resume either recovers the run the operator meant, or refuses. Concretely: persist the replication scope with the checkpoint — at minimum pipeline name, filter, and target routing (shadow / branch) — and refuse a resume whose current execution identity does not match the checkpoint's.

## Actual behavior

[...]

PR #1734, opened 2026-09-06
OWNER MERGE — fix(engine): persist the check gate so a resume cannot launder a failed run into a success

hugocorreia90:
A `rocky run` that its checks correctly failed could be resumed into a green `Success` record, exit 0 — and the failing run's own message told the operator to do it.

The verdict was never written down. This persists it on the run record (state schema **v24 -> v25**), refuses a resume that would re-run none of the gating checks, carries a standing gate forward into a resume that is admitted, and fixes the message that pointed at `--resume`.

**Owner merge:** a resume that exits 0 today now refuses or reports failure, and the state schema moves.

## What was wrong

`check_gate_failed` reached `RunOutput` and stopped — zero occurrences in `rocky-core/src/state.rs`. The resume gate therefore inferred resumability from `models_executed`, which is orthogonal to check severity.

```
  run 1: checks gate it            -> record PartialFailure, exit 2
  run 2: --resume run 1
           copies nothing (or only the failed table)
           builds check inputs only from what IT copied
           -> check_gate_failed = false -> derive_run_status -> Success, exit 0
  latest_successful_run matches on exactly Success
           -> every downstream `after` demand fires on violating data
```

Two shapes reached it:

- **No `--all` needed.** A failed post-apply `verify_after` gate pushes a `<verify_after>` entry onto `output.errors` with no model involved; `to_run_record` writes it as a `models_executed` entry with `status: "failed"`. Every table is `Success`, so the resume filter drops every key and the run is empty.
- **The common one, and the tool suggests it.** One table fails to copy while a *different*, already-copied table fails its check. The run exits 2 printing `use --resume X to retry`. The resume legitimately re-copies the failed table; checks cover only that table, so the earlier violation is never re-examined.

## The fix

1. **Persist the verdict.** `RunRecord::check_gate_failed`, serde-additive with `#[serde(default)]`. State schema **v24 -> v25**: a doc stanza on `CURRENT_SCHEMA_VERSION`, the const, the `EXPECTED_VERSION` pin in the table-set test, and a forward-read test. No table added; `EXPECTED_TABLES` untouched; no blob walk.
2. **Refuse the empty resume.** `ensure_the_resume_would_do_work` refuses when the record carries the gate and no copy work remains — placed *before* the failed-model branch, which is what admitted the `<verify_after>` shape.
3. **Carry the gate forward** (see the choice below) into a resume that is admitted.
4. **Fix the advice.** The failed-copy branch returns before the check-gate branch, so a run where both failed only ever said "use --resume to retry". `PartialFailure` and `CheckGateFailure` now vary their message; the uninherited messages are byte-identical to before.

### Item 3: carry forward, not re-evaluate — and why

I carried the prior verdict forward rather than re-running the earlier run's checks.

- Carrying forward cannot report green over a violation nobody re-examined. It is a strictly fail-closed change, and small: one `||` behind a named helper.
- Re-evaluating is more useful — it can also *clear* the gate — but it needs a way to name the earlier run's copied tables, run their checks outside a copy, and decide what a check that now passes means for a record that already says it failed. That is a feature with its own release, not a bug fix.
- The resumed run states it, and here is exactly where, because two of these are narrower than they first read:
  - **JSON (stdout, what an orchestrator reads):** `status: "partial_failure"`, `check_gate_failed: true`, `resumed_from: "<prior id>"`. This is the reliable channel.
  - **A status line naming the prior run.** Goes to stderr under `--output json` (that is what `status_line!` does), so a consumer reading stdout does not see it. Operator-facing only.
  - **The exit-2 message.** Names the prior run *only* when the resume reaches the `CheckGateFailure` branch. A resume that also has a failed copy returns `PartialFailure` from the earlier branch, whose message says a check gate also failed but does not name which run raised it.

  `RunOutput::check_gate_failed`'s doc was rewritten to cover the inherited case, because a field that contradicts its own doc is worse than no doc. I did **not** add a new `RunOutput` field for this — that is four generated-binding drift surfaces for something the JSON already says.

### One thing I found while fixing it

The gate is stamped where the output is **created**, not at the check-gate stamp far below. The SIGINT path persists its `RunRecord` before that stamp runs, so a late stamp would let an *interrupted* resume of a gated run record `check_gate_failed = false` — and a resume of that interrupted run would inherit nothing and could go green. Stamping at creation covers every persist site by construction.

### Tests (all new; which fail on `main`)

| Test | Fails on `main`? |
|---|---|
| `resume_refuses_a_complete_checkpoint_whose_checks_gated_the_run` | **yes** — `main` admits it (the failed-model branch) and the run records `Success` |
| `a_resume_of_a_gated_run_cannot_record_success` | **yes** — `main` derives `Success`; also asserts the resumed run's own record carries the gate, so a resume of the resume inherits it |
| `an_interrupted_resume_still_records_the_standing_gate` | **yes** — the field does not exist on `main` |
| `the_inherited_gate_is_stamped_before_the_interrupt_path_persists` | **yes** — the stamp does not exist on `main`; asserts the ordering the interrupt path depends on |
| `test_v24_run_record_forward_deserializes_check_gate_failed_false` | **yes** — the field does not exist on `main` |
| `a_recorded_check_gate_survives_the_state_store` | **yes** — proves `record_run` -> redb -> `get_run` |
| `the_partial_failure_advice_names_the_check_gate_only_when_one_failed` | **yes** — `main`'s message has no gate half; the ungated half is asserted byte-identical |
| `the_check_gate_message_names_the_run_that_raised_an_inherited_gate` | **yes** — same |
| `resume_allows_a_gated_run_that_still_has_tables_to_copy` | no — control, the narrowing conjunct for the refusal |
| `an_ungated_failed_run_resumes_and_inherits_nothing` | no — **control**, a genuine copy failure still resumes |
| `a_resume_with_no_standing_gate_still_reaches_success` | no — **control**, an honest recovery still reaches `Success` |
| `a_run_with_fail_on_error_off_records_no_gate_to_inherit` | no — **control**, an advisory pipeline records no gate |
| `a_healthy_run_records_no_gate` | no — **control**, a healthy run's record and status are unchanged |
| `a_pre_v25_record_inherits_no_gate` | no — **control**, a v24 record resumes as it does today |

The existing `resume_allows_a_complete_checkpoint_when_a_model_failed` still passes: its record carries no gate.

**Which rows carry weight.** Several say "fails on `main`" only because the field does not exist there — true of any new test, and cheap. The two that matter are `resume_refuses_a_complete_checkpoint_whose_checks_gated_the_run` and `a_resume_of_a_gated_run_cannot_record_success`: `main` reaches both code paths, has all the inputs, and gets the answer wrong. Both are pinned by a mutation check below.

### Mutation checks

Three, via `scripts/mutation-check.sh` against the committed tree.

**1. The write wire** — `rocky-cli/src/output.rs`, `to_run_record`'s `check_gate_failed: self.check_gate_failed` -> `false`; test `a_resume_of_a_gated_run_cannot_record_success`.
**OK — fix-sensitive.** Panics on "the severity-resolved verdict must reach the record". Nothing the resume gate reads is populated without this line.

**2. The refusal** — `rocky-cli/src/commands/run.rs`, `if record.check_gate_failed {` -> `if false {`; test `resume_refuses_a_complete_checkpoint_whose_checks_gated_the_run`.
**OK — fix-sensitive.** The failure prints the admitted checkpoint: `Some(RunProgress { run_id: "run-1", total_tables: 2, tables: [… Success, … Success] })`. That is `main`'s behaviour verbatim — the resume that goes on to record `Success`.

**3. The early stamp** — `rocky-cli/src/commands/run.rs`, delete `output.check_gate_failed = resolved_check_gate(false, inherited_gate.as_ref());`; test: the whole `commands::run::` suite.
**FAILED the first time — 306 tests passed with the line deleted.** I ran this as a coverage probe rather than as evidence, and it found a real gap: no behavioural test reaches the SIGINT path, so the interrupt fix was covered by construction and by nothing else. Closed by adding `the_inherited_gate_is_stamped_before_the_interrupt_path_persists`, which asserts the ordering over this file's own source — the shape `rocky-mcp/src/tools.rs` already uses for a claim about its own text. Re-run after that commit: **OK — fix-sensitive** (306 passed, 1 failed).

Reported as it happened, not as it ended, because a probe that has to be fixed is the more useful half of the evidence.

### Honest failure

- A run with no checks configured: unchanged (`a_healthy_run_records_no_gate`).
- A run whose checks all pass: unchanged (`a_resume_with_no_standing_gate_still_reaches_success`).
- A resume of a genuine copy failure with no check failure: unchanged, still resumes, still reaches `Success` (`an_ungated_failed_run_resumes_and_inherits_nothing`, `a_resume_with_no_standing_gate_still_reaches_success`).
- `fail_on_error = false`: unchanged, and it never records a gate, so there is never one to inherit (`a_run_with_fail_on_error_off_records_no_gate_to_inherit`).
- The ungated `PartialFailure` and `CheckGateFailure` messages are asserted byte-identical to `main`.

One honest limit on the controls: exit codes are asserted **by derivation**, not by spawning a process. `run_status_exit_result` maps `Success` to `Ok(())` / exit 0 and `PartialFailure` to the sentinel `main.rs` maps to exit 2, and the tests assert the `RunStatus` those mappings read. No control spawns `rocky` and reads `$?`.

### Codegen and fixtures

No `*Output` field was added, but `RunOutput::check_gate_failed`'s **doc comment** changed, and a doc comment becomes the schema's `description`. `just codegen` therefore moves four files, and all four are committed:

```
schemas/run.schema.json                                 1 description
docs/public/openapi.json                                1 description
sdk/python/src/rocky_sdk/types_generated/run_schema.py  the same text
editors/vscode/src/types/generated/run.ts               the same text
```

No `editors/vscode/package-lock.json` churn. `sdk/python/src/rocky_sdk/types.py` needs no hand edit — no top-level field changed, so `test_schema_parity.py` still holds.

`just regen-fixtures` for the v24 -> v25 bump. **The diff is version-only**, five files:

```
doctor.json            "state schema v24 matches this binary"      -> v25
doctor_ci/doctor.json  "no on-disk state schema (binary supports v24)" -> v25
state.json                       schema_version_supported/_on_disk 24 -> 25
incremental/state_after_run1.json                                  24 -> 25
incremental/state_after_run2.json                                  24 -> 25
```

Codex is out of credits, and a Claude reviewer is the same family as the author, so nothing here has had an independent pass. My own adversarial review, on the five shapes I was asked to probe:

- **A resume of a resume.** Covered and asserted. The resumed run's own record carries the verdict, so run 3 inherits from run 2 exactly as run 2 did from run 1. The gate does not decay with depth. If the resume has no copy work left it is refused instead.
- **A v25 record read by a v24 binary.** The blob is backward-safe — one extra key, and serde ignores unknown keys — but a v24 binary never reaches it, because the version check runs at OPEN. Under `on_schema_mismatch = fail` it stops and names both versions. Under the default `recreate` it **deletes the local store**, which discards the run history and every verdict with it; a resume after that launders exactly as it does on `main`. That is a pre-existing property of `recreate`, not something this change introduces, but it is the honest limit of the fix and it is written into the CHANGELOG's "On upgrade" paragraph.
- **A run interrupted mid-check.** This is the hole I found and closed; see above. The residual: the *first* run being interrupted before its own check gate is stamped records no gate — correctly, because its checks never finished, and its checkpoint has non-`Success` tables so the resume re-copies and re-checks them.
- **`fail_on_error` flipped between the run and the resume.** A run under `fail_on_error = false` never records a gate, so there is nothing to inherit from one. A verdict that *was* severity-resolved while the gate was on is not erased by flipping the flag afterwards — letting a config flip erase it would reopen this bug through the config instead of through the resume. Re-running the pipeline is what clears it. Asserted.
- **A resume whose scope check passes but whose pipeline config changed.** The scope check (#1549) already refuses a changed pipeline name, filter or routing. A change it permits — a check removed from `rocky.toml`, a severity lowered — does not clear an inherited gate: the earlier run's verdict is read from its record, not re-derived from today's config. That is deliberate. The data that violated is still there, and deleting the check is not the same as fixing it.

Two more I was not asked about:

- **A resume whose `total_tables == 0`.** Admitted, gate or no gate. The old justification here said such a run "cannot have tripped the check gate"; persisting the verdict made that claim checkable, so I replaced it with one that does not depend on it — with no completed keys the resume re-plans every table, so every check runs again. The comment is corrected in the code, not just here.
- **The `<verify_after>` shape without a tripped gate.** A `verify_after` custody failure that does *not* trip the check gate (a named check absent from `check_outcomes`, or `fail_on_error = false`) still leaves a `models_executed` entry with `status: "failed"` and still admits an empty resume that records `Success`. That is a distinct defect with a different cause, and narrowing the failed-model branch is a behaviour change beyond this issue. **Filed as #1732.**

The `recreate` path. `[state] on_schema_mismatch = recreate` is the default, and a fleet running mixed versions during a rollout will have older pods delete a v25 local store and start fresh — which silently discards every persisted verdict and restores the laundering path for those pods. The CHANGELOG says so, and #1525 already tracks state loss not being harmless, but "the version gate protects the downgrade" is weaker here than it sounds: it protects the *read*, not the *data*.

## The most important thing I may be missing

Whether stopping the laundering at the *record* stops it everywhere. I went and checked rather than assuming. Every production reader of `RunRecord::status` treats `PartialFailure` as not-good:

- `StateStore::latest_successful_run` matches on exactly `RunStatus::Success` — the join an `after` / `freshness` demand uses.
- `schedule/claim.rs::terminal_state`: for `After` / `Freshness`, `Partial` releases the claim and leaves the demand standing, the same as `Failure`. Only `Success` completes it.
- `run.rs::finalize_idempotency` maps `PartialFailure | Failure` to `FinalOutcome::Failed`, so the key does not deflect a later retry.
- `catalog.rs` picks the newest run with `status == Success`, exactly.
- `brief.rs` counts `Failure | PartialFailure` as an incident.

The one place `Partial` completes a claim is `Cron` / `Webhook`, where any outcome completes by design — the next fire is a new key, and it says nothing about downstream freshness. So I do not have a consumer that would still read the resumed run as green. What I have not proven is the negative for *external* consumers: an orchestrator that keys on exit code alone still sees 2, which is the same code a partial copy failure gives, so it cannot distinguish "one table failed" from "the checks gated this". The JSON says which; the exit code does not.

Refs #1720, #1598


hugocorreia90:

**Verified:** `check_gate_failed` is now on `RunRecord`, `#[serde(default)]` so a v24 blob forward-reads as `false`; the schema moves to v25 with the const, the `EXPECTED_VERSION` pin and no table added; `ensure_the_resume_would_do_work` refuses on `record.check_gate_failed` when no copy work remains; and a carried-forward verdict rides `inherited_from` through the named `resolved_check_gate(this_run_gated, inherited)`, so a resumed run cannot report green over a violation nobody re-examined. Codegen and fixtures moved version-only.

**The best piece of work here is mutation check 3.** Deleting the early stamp left 306 tests passing — the lane ran the mutation as a coverage probe rather than a confirmation, found a real hole, and closed it with a source-order assertion. That is the right instinct, and it is how the collector divergence in #1718 should have been caught months ago.

**Also caught beyond the brief:** the SIGINT path persists its `RunRecord` before the check-gate stamp, so an interrupted resume would have recorded no verdict and the next resume would have inherited nothing. Stamping where the output is created closes it.

**Item 3, carrying the verdict forward rather than re-evaluating: agreed.** Re-evaluating can also *clear* a gate, needs a way to run checks outside a copy, and is a feature rather than a bug fix. Carrying it forward is honest and small.

**The residual, stated on the record:** `on_schema_mismatch = "recreate"` is the default, so an older binary in a mixed-version fleet deletes the local store and discards every persisted verdict, restoring the laundering path for those pods. The version gate protects the read, not the data. It is in the changelog's On upgrade, which is the right place for it.

**And the honest external gap:** an orchestrator keying on the exit code alone still sees 2 for both a gated run and a plain partial copy failure. The JSON distinguishes them; the exit code does not.

**Independent review: not available** (Codex out of credits; a Claude reviewer is the same family as the lane). Owner-merge is right — a resume that exits 0 today refuses or reports failure, and the state schema moves. Merging when the updated branch is green.

Issue #1735, opened 2026-09-06
severity = "warning" on null_rate and custom checks still silences a check the engine could not run


hugocorreia90:
#1719 fixed this clobber for freshness. The same shape is still in place for two other check kinds.

## What happens

`run.rs` writes `cfg.severity` over **every** null-rate result, measured or not:

```rust
// run.rs, run_null_rate_checks — both the every-column path and the per-column path
let mut check = checks::null_rate_not_evaluated(col, cfg.threshold, reason.clone());
check.severity = cfg.severity;
```

`run_local.rs::run_custom_checks` and `custom_checks_not_evaluated` do the same with `custom.severity` for `[[checks.custom]]`.

`null_rate_not_evaluated` and `custom_not_evaluated` hard-code `TestSeverity::Error` (`rocky-core/src/checks.rs`) precisely so a check the engine could not run still gates.

So with `[pipeline.<name>.checks.null_rate] severity = "warning"`:

1. the null-rate query fails,
2. every configured column is reported `not_evaluated`, `passed: false`, at **warning** severity,
3. `replication_check_gate_failed` counts error-severity failures only, so the gate stays clear,
4. the run exits 0 with `status: "Success"`.

A `[[checks.custom]]` entry with `severity = "warning"` whose query fails behaves the same way.

## Fix

The same shape as #1719: apply the configured severity only when `check.not_evaluated.is_none()`. The configured severity describes a violation the check measured. It is not a statement about a query Rocky could not run.

## Why this is not a one-liner

A committed test asserts the behavior the fix would change, and its doc comment says so: `a_null_rate_query_that_fails_reports_every_column_not_evaluated` in `run.rs` asserts `TestSeverity::Warning` on a failed null-rate query, under the doc line "at the configured severity". Both the assertion and the comment have to flip. The change also alters an exit code the way #1719 did, so it needs its own changelog entry saying so, and a gate assertion through `run_status_for`.

## Deliberately excluded

`cross_source_overlap_not_evaluated` and `assertion_not_evaluated` take the severity as a constructor parameter instead of overwriting it afterwards. That is a deliberate design, not this defect.

Refs #1719, #1700, #1655.

hugocorreia90:
Fixed by #1742. Verified against `main` 5fe75784.

The configured severity applies to a measured result only. All five sites are converted:

```
run.rs:6943       every_column_not_evaluated   the write is gone
run.rs:7022       the per-column path          guarded on not_evaluated.is_none()
[...]

Issue #1719, opened 2026-09-05
Snowflake and BigQuery batch adapters return Err as a capability sentinel, and the engine now gates on it — plus severity=warning silences a failed freshness leg


hugocorreia90:
Two of the three shipped `BatchCheckAdapter`s use `Err` as a **capability sentinel** — "not implemented, fall back" — and since #1655/#1700 the engine converts that sentinel into a gating check failure. Verified on `main` df506990.

## 1. Snowflake and BigQuery replication runs fail on healthy data

```rust
// engine/crates/rocky-snowflake/src/batch.rs:64
// Snowflake could batch these with a UNION ALL over the supplied
// tables, but it's not wired yet. Return an error so `run.rs` falls
// back to per-table row-count queries via `WarehouseAdapter`.
Err(AdapterError::msg("batch_row_counts not yet implemented for Snowflake"))
```

`batch_freshness` is the same, and `engine/crates/rocky-bigquery/src/batch.rs:102` mirrors both. **The fallback that comment promises was never implemented in `run.rs`.** `registry.rs:622` hands back a `SnowflakeBatchCheckAdapter` / `BigQueryBatchCheckAdapter` unconditionally, so `run.rs` takes the batched branch, the leg "fails", and #1700's fold records:

```
row_count  not_evaluated: "source: the batch row count query failed:
           batch_row_counts not yet implemented for Snowflake; target: …"
           passed: false, severity: Error   (hard-coded, checks.rs:406)
```

With the default `fail_on_error = true` that trips `replication_check_gate_failed` — exit 2, `PartialFailure`, every table, every run, on healthy data. There is no capability probe anywhere in the tree (`supports_batch` / `batch_supported` / `can_batch`: zero hits).

Before #1700 the same configuration exited 1 with no JSON, so this path has been broken for longer; #1700 changed the shape and made it worse by writing `passed: false` into `RunRecord.check_outcomes`, which the `verify_after` gate AND-aggregates — so a policy rule naming `row_count` now always denies, halts the run, writes a deny custody row and burns the granting rule's autonomy budget, on an unreadable *capability* rather than unreadable data.

Bounded by opt-in: `row_count` defaults to `false` and `freshness` to `None`, and Rocky's live-verified replication use is Databricks-centric. That bounds who is hit, not the defect — and the fact that the earlier exit-1 shape was never reported says nobody has exercised this path recently.

[...]

PR #1736, opened 2026-09-06
OWNER MERGE — fix(engine): let an adapter say it cannot batch, instead of failing every check on Snowflake and BigQuery

hugocorreia90:
Snowflake and BigQuery replication runs failed every row-count and freshness check on healthy data. Both adapters said "I do not batch this" by returning an error, and the runner reads an error as a query that failed.

They now say it through a capability, and the leg falls back to one query per table — the fallback their own comments promised and nobody had written.

Second fix: `severity = "warning"` on freshness also silenced a freshness query that failed. That severity now applies to measured results only.

## The defect

```
before   snowflake, row_count = true
         └─ batch_row_counts() ──► Err("not yet implemented for Snowflake")
            └─ #1700 fold ──► row_count not_evaluated, passed:false, severity:Error
               └─ check gate ──► exit 2, PartialFailure
                                 every table, every run, on healthy data

after    └─ supports_row_counts() == false ──► never asked
            └─ SELECT COUNT(*) per table ──► a real measurement ──► exit 0
```

`AdapterRegistry::batch_check_adapter` returns `Some(..)` for both warehouses unconditionally, and there was no capability probe anywhere in the tree. A `verify_after` policy rule naming `row_count` always denied, because the recorded outcome said `passed: false`.

The freshness half: `run.rs` wrote `freshness_cfg.severity` over **every** freshness result, measured or not. `freshness_not_evaluated` hard-codes `TestSeverity::Error` precisely so an unevaluated check gates. With `severity = "warning"` a failed freshness leg landed in the warning bucket, the gate stayed clear, and the run exited 0 with `status: "Success"`.

## The mechanism, and why this one

Two defaulted predicates on `rocky_core::traits::BatchCheckAdapter`:

```rust
fn supports_row_counts(&self) -> bool { true }
fn supports_freshness(&self) -> bool { true }
```

`run_batched_checks` reads them **per leg** and sends a leg the adapter cannot batch down the per-table path. The two per-table loops were lifted into `per_table_row_counts` / `per_table_freshness`, so "no batch adapter" and "cannot batch this leg" run the same code in the same order.

Two alternatives were rejected:

- **Registry returns `None` for Snowflake and BigQuery.** Wrong: the same trait object also serves `batch_describe_schema`, which both adapters genuinely implement, and which `discover --with-schemas` warm-up and the run loop's column pre-fetch use. `None` would cost those warehouses their batched schema describe to express something about two other methods.
- **A `BatchSupport` enum return.** It changes the signature of two required methods, which breaks every out-of-tree implementor. Defaulted predicates are additive, and match the default-method style the same trait already uses for `batch_describe_schema`.

On the required "remove the `Err` stubs": the methods stay required, so a new adapter must consciously write a body rather than inherit a silent "unsupported" default — the default predicate is `true`, so a defaulted body would recreate exactly this bug for an adapter that implements neither. What is removed is the stub's **meaning**: nothing in the tree reads that `Err` as a fallback signal any more, the message now says the caller ignored the capability, and a test asserts the method is never called when the predicate is `false`.

The SDK mirror (`rocky-adapter-sdk`) gained the same two defaults, because its module doc promises method-surface parity for additive defaults.

### Which assertions fail on `main`

Each mutation below restores `main`'s rule at exactly one site, so its failures are the fail-before evidence.

| Test | On `main` |
|---|---|
| `an_adapter_that_batches_neither_leg_measures_per_table_and_does_not_gate` (a) | `row_count` is `not_evaluated`, `passed: false`, severity `Error` — the Snowflake/BigQuery shape |
| `an_adapter_that_batches_only_row_counts_falls_back_for_freshness_alone` (a′) | `freshness` is `not_evaluated`, though row counts batched fine |
| `a_failed_freshness_leg_gates_even_when_the_configured_severity_is_warning` (c) | `gated` is **false**, severity `Warning`, status `Success` |
| `a_freshness_check_whose_query_fails_is_not_evaluated` (existing) | its committed assertion was `TestSeverity::Warning`; it is now `Error`, and its doc comment was updated with it |

Controls that pass before and after: `a_batched_leg_that_genuinely_fails_is_still_not_evaluated_and_still_gates` (b), `an_adapter_that_batches_both_legs_is_unchanged` (d), `a_warehouse_with_no_batch_adapter_is_unchanged` (e), `a_measured_stale_table_still_honours_severity_warning`.

### Honest failure: Databricks and DuckDB

- **Databricks** — `DatabricksBatchCheckAdapter` overrides neither predicate, so both are `true` and every leg still batches. Test (d) proves the batched answer is what the checks report: the double answers `7`, which the one-row seeded tables can never produce, so the count itself says which path ran. Under mutation 1 (main's dispatch restored) test (d) still passes, along with all four #1655 tests and both their doubles — the capability read changes behavior only for an adapter that declares it cannot batch.
- **DuckDB** — no `BatchCheckAdapter` at all, so both batched legs yield `None` and the per-table helpers run in the same order as before. Test (e) proves it end to end.
- **Not unchanged, on purpose**: a failed *freshness* query under `severity = "warning"` now reports at error severity on every path, DuckDB included. That is the second fix, and the changelog says so under "On upgrade".

### Mutation checks

Both via `scripts/mutation-check.sh`, against the committed tree, test command `cargo test -p rocky-cli --lib -- commands::run::` (302 tests).

1. **The capability read** — `batch_check.filter(|bc| bc.supports_row_counts())` → `batch_check` for both legs, which is `main`'s rule: any adapter batches every leg. Verdict **OK** (exit 0). 300 passed, 2 failed — (a) and (a′), both with `not_evaluated: Some("the batch row count query failed: batch_row_counts not yet implemented")`.
2. **The severity guard** — the `if check.not_evaluated.is_none()` guard removed, which is `main`'s unconditional line. Verdict **OK** (exit 0). 300 passed, 2 failed — (c) failing at the gate with `severity: Warning` printed, and the existing freshness test failing at `left: Warning, right: Error`.

### Local gate

- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — exit 0
- `cargo test -p rocky-cli --lib -- commands::run::` — 302 passed, 0 failed (re-run after merging `origin/main`)
- `cargo test -p rocky-core -p rocky-snowflake -p rocky-bigquery -p rocky-adapter-sdk` — all green (rocky-core: 2120 passed)

No `*Output` struct changed, so no `just codegen` / `just regen-fixtures` cascade.

## Other check kinds with the same clobber

`null_rate` (two sites in `run.rs`) and `[[checks.custom]]` (four sites in `run_local.rs`) write the configured severity over their `*_not_evaluated` results the same way. Not fixed here: freshness is the reported defect with a proven gate consequence, and flipping the other two changes behavior a committed test currently asserts, so it needs its own evidence and its own changelog line. Filed as #1735.

`cross_source_overlap_not_evaluated` and `assertion_not_evaluated` take severity as a constructor parameter instead of overwriting it afterwards. Deliberate design, not this defect.

No independent red team was run. Codex is out of credits, and a Claude reviewer is the same model family as the author, which does not satisfy the independence requirement in `AGENTS.md`. What follows is the author's own adversarial pass, so weight it accordingly.

**An adapter that supports row counts but not freshness.** Covered by construction, not by hope: the capability is read per leg, and test (a′) runs exactly that adapter — the row-count leg batches (count 7) while freshness falls back to a per-table `SELECT MAX(ts)` in the same run. The batched legs still share one `tokio::join!`, so a mixed adapter loses no concurrency.

**A capability that changes between two runs.** Nothing caches it. `AdapterRegistry::batch_check_adapter` builds the adapter inside `run()`, and `run_batched_checks` reads the predicate on every call. A capability that flips takes effect on the next run. The one visible consequence is that the check's *reason text* differs between the two paths ("the batch row count query failed" vs "the row count query failed"), which is pre-existing and intended — it names blast radius.

**A third-party adapter implementing the trait out of tree.** `rocky_core::traits::BatchCheckAdapter` is `pub`. This is **not a breaking change**: both methods are defaulted, so an existing implementor keeps compiling and keeps its current behavior (`true`, both legs asked). An out-of-tree adapter that used the same `Err`-as-sentinel trick keeps behaving exactly as it does today — its error still folds into `not_evaluated` — and can opt into the capability whenever it likes. Two caveats I cannot close from here: an implementor with an *inherent* method of the same name now shadows the trait method (Rust resolves inherent first, silently), and an implementor that never sets the predicate gets no warning. The `rocky-adapter-sdk` mirror is a contract copy the registry does not consume, so an SDK-only adapter is unaffected either way; the SDK's `AdapterCapabilities.batch_checks` is a whole-adapter flag, which is precisely why the in-tree signal had to be per-operation — Snowflake batches describe and nothing else.

**A leg that fails for a subset of tables.** It cannot, and that is unchanged. A batched leg is one query for every table it was handed: it answers or it fails, wholesale. The partial shape that does exist is a leg that answers and leaves a table out, which still reports "the batch query returned no result for this table" — pinned by `a_table_the_batch_row_count_left_out_is_not_evaluated` and `a_table_the_batch_freshness_left_out_is_not_evaluated`, both unmodified. The per-table path is the one with per-table granularity, and it is now reached by more adapters, which strictly improves that reporting.

**Where I would attack this next.** The `batch_describe_schema` default still returns `Err("not supported by this adapter")` as its own sentinel. It is not this defect — its caller in `discover.rs` genuinely implements the fallback (warn, skip the schema, carry on) — but it is the same anti-pattern one layer over, and it would become a real defect the moment someone makes that call site gate anything.

## The two standing questions

**Least confident about:** that "Databricks is byte-identical" is proven rather than argued. What is proven is that the adapter overrides neither predicate, that the batched answers reach the check results unchanged, and that no test of the batched path changed. What is *not* proven is a live Databricks run — this change has no live warehouse coverage, and neither did the code it fixes.

**Most important thing I might be missing:** that this defect went unreported at all. Before #1655 the same configuration exited 1 with no JSON, and nobody filed it. Either nobody runs `row_count = true` on Snowflake or BigQuery, or somebody does and reads the failure as normal. If it is the second, this fix changes a check they have learned to ignore into one that means something, and the changelog is the only place they will find out.

Refs #1719, #1655


hugocorreia90:
Orchestrator review. Checked against the diff, not the description. Merged `main` in and resolved the changelog conflict after #1734 landed.

**Verified:** `supports_row_counts()` and `supports_freshness()` are defaulted `true` on the trait, so an out-of-tree implementor keeps compiling and keeps today's behaviour — not a breaking change. `run_batched_checks` reads them **per leg**, so an adapter that batches one and not the other falls back for the right one only. Both unconditional `Err("not yet implemented for …")` stubs are gone from the Snowflake and BigQuery adapters, so `Err` can no longer mean "I cannot do this" where the engine reads it as "the query failed". The `not_evaluated.is_none()` guard stops the configured freshness severity from clobbering the `Error` that `freshness_not_evaluated` sets deliberately.

**The mechanism choice is right.** Returning `None` from the registry would have been simpler and wrong: the same trait object serves `batch_describe_schema`, which Snowflake and BigQuery genuinely implement and which `discover --with-schemas` and the run loop's column pre-fetch depend on. `None` would have cost them their batched schema describe to fix an unrelated leg.

**Found and filed, not fixed here (#1735):** `null_rate` and `[[checks.custom]]` clobber their `*_not_evaluated` severity the same way, at six sites. A committed test asserts the current behaviour and the change moves an exit code, so it needs its own evidence rather than riding along.

**The honest residual:** "Databricks is byte-identical" is argued, not measured — the adapter overrides neither predicate, no batched-path test changed, and all of them survive mutation 1, but there is no live warehouse run behind it. Neither was there behind the code this fixes.

**Worth reading twice, from the lane:** nobody reported this. Before #1655 the same configuration exited 1 with no JSON and no issue was filed. Either nobody runs `row_count = true` on Snowflake or BigQuery, or somebody does and has learned to ignore the failure — in which case this turns a check they ignore into one that means something, and the changelog is the only place they will find out.

**Independent review: not available** (Codex out of credits; a Claude reviewer is the same family as the lane). Owner-merge is right: exit codes change on those two warehouses, and a silenced freshness failure gates again.

Issue #1729, opened 2026-09-06
A dangling rocky.toml symlink still reads as absent in serve and state retention sweep, because Path::exists() follows the link — and the sweep then deletes history


hugocorreia90:
#1708 made a dangling `rocky.toml` symlink refuse instead of reading as absent. Two commands never reach that refusal, because they decide whether to load the config with `Path::exists()` — which follows the symlink and answers `false`. One of them deletes data. Verified on `main`.

## 1. `rocky state retention sweep` deletes history the config said to keep

```rust
// engine/crates/rocky-cli/src/commands/state.rs:148
let policy = if config_path.exists() {
    let cfg = load_rocky_config(config_path)...?;
    cfg.state.retention
} else {
    StateRetentionConfig::default()
};
```

Trigger: a project with `[state.retention] max_age_days = 3650` whose `rocky.toml` is a dangling symlink, then `rocky state retention sweep`.

The configured block is never read. The defaults apply, and run history, DAG snapshots and quality snapshots past the default age are deleted. Exit 0, no warning, irreversible. This is the exact class #1708 exists to close, one probe upstream of the code it changed.

The sweep must refuse here and name the link.

## 2. `rocky serve` silently relocates the spool

```rust
// engine/rocky/src/main.rs:4347
let config = if cli.config.exists() { Some(cli.config.as_path()) } else { None };
```

Trigger: `rocky serve --config /etc/rocky/prod.toml --scheduler`, where that path is a dangling symlink.

Three consequences, none a refusal:

- The served compile runs with empty `mask`, `allow_unmasked` and `project_freshness` (`rocky-server/src/state.rs:200-213`), so **W004 and W005 go silent on every compile**.
- The schema-cache posture falls back to the built-in default instead of `[cache.schemas]` (`state.rs:274-281`).
- `serve.rs:99` does `config_path.map(..).unwrap_or_else(|| PathBuf::from("rocky.toml"))`, and `scheduler/mod.rs:53` derives `.rocky` from that — so an explicit `--config /etc/rocky/prod.toml` becomes `./rocky.toml`, and the resident scheduler and webhook ingress spool to `./.rocky/pending-demands` in the current directory.

## 3. Cosmetic

`product.rs:255` uses `!config_path.is_file()`, so `rocky product verify` reports "rocky.toml not found at <path>" for a config that is plainly there. It returns `needs_input`, so the direction is safe; only the message lies.

## Fix

[...]

PR #1739, opened 2026-09-06
OWNER MERGE — fix(engine): stop deciding config presence with a link-following probe, which let a sweep delete kept history

hugocorreia90:
`Path::exists()` follows symlinks, so a dangling `rocky.toml` answered "absent" and three commands never called the loader at all. One of them deletes data.

The loader is now the probe: only `ConfigError::FileNotFound` means absent, and an entry that exists but cannot be read refuses, naming the path.

## The three sites

| site | before | after |
|---|---|---|
| `commands/state.rs` retention sweep | `exists()` false → `StateRetentionConfig::default()` → deletes run history, DAG and quality snapshots past the 365-day default | refuses, naming the path; a genuinely absent config still sweeps with defaults |
| `rocky/src/main.rs` serve | `exists()` false → `config_path: None` → W004/W005 silent on every served compile, default schema-cache posture, and an explicit `--config /etc/rocky/prod.toml` spooling webhooks to `./.rocky` | refuses via `resolve_serve_config_path`; `None` is still the honest no-config answer |
| `commands/product.rs` verify | `is_file()` false → "rocky.toml not found at &lt;path&gt;" for a config that is plainly there | reports the real reason; still `needs_input` |

The retention sweep is the one that matters: exit 0, no warning, irreversible.

## Evidence

**Fails on `main`** (all `#[cfg(unix)]`, 10 gates): `a_dangling_config_symlink_refuses_the_sweep_and_deletes_nothing` — on `main` the rows past 365 days are gone; `a_malformed_config_still_refuses`; `a_dangling_config_symlink_refuses_to_start` (serve). Controls that pass on both: no `rocky.toml` still sweeps with defaults, serve still starts config-less, a real readable config is unchanged everywhere.

**Mutation checks**, both `mutation-check: OK — the test failed under mutation, so it is fix-sensitive`, with real `test result: FAILED` in the log rather than the #1547 exit-3 shape:

```
1  state.rs  restore the exists() gate   3 passed; 2 failed
2  config.rs collapse the presence arm   3 passed; 1 failed
```

**Gates:** `cargo +1.98.0 test --workspace` 2130 passed, 0 failed; `fmt --all --check` clean; `clippy --workspace --all-targets -D warnings` clean.

**The `exists()` sweep asked for.** Every non-test hit that gates a config or state read is now closed. `commands/init.rs:21` keeps its `exists()` deliberately — it asks "is there already a config here" before writing one, which is a presence question, not a read gate.

## Changelog

The `[Unreleased]` entry for #1668 claimed "a project whose `rocky.toml` is a dangling symlink used to run against defaults and now refuses". That was false for these three commands, and remains false for the LSP (`lsp.rs:427`, `:647`) and both `rocky plan` seams (`plan.rs:1231`, `:1290`), which all `.ok()` the error away. The entry is now scoped to error propagation rather than the loader, and names the seams still open.

Codex is out of credits, and a Claude reviewer is the same model family as the author, so this carries no independent pass. Adversarial self-review of the requested shapes: a symlink chain to a real file resolves and loads normally (the loader follows it; only a *broken* chain refuses); `symlink_metadata` failing on a parent-permission error surfaces as a read error rather than absence, which is the fail-closed direction; a relative `--config` is resolved before the probe, so a different CWD changes which file is read but not whether the refusal fires; a config deleted between probe and load no longer has a window, because there is only one read now; on Windows, symlink creation needs privileges, so the tests are `#[cfg(unix)]` and the behaviour there is unchanged.

Refs #1729, #1708, #1668.


hugocorreia90:
CI was red on `E0063: missing field `check_gate_failed`` — #1734 added that field to `RunRecord` after this branch cut its retention fixture. Merged `main` in and stamped the field. Re-gated on the merged tree: `check --workspace --all-targets` clean, fmt clean, `clippy --workspace --all-targets -D warnings` clean, and the state/serve/product tests 86 passed / 0 failed. Nothing in the fix changed.

Issue #1741, opened 2026-09-06
severity = "warning" on cross_source_overlap silences the mistyped-key refusal #1717 created — and the same shape on [[assertions]]


hugocorreia90:
`#1719` and `#1735` narrowed the configured severity so it applies to a **measured** result only. `row_count`, `freshness`, `null_rate` and `[[checks.custom]]` now keep the `TestSeverity::Error` their `*_not_evaluated` constructor chose, so a check the engine could not run still gates.

Two check kinds are outside that rule, because they take the severity as a **constructor parameter** rather than overwriting it afterwards. `#1735` recorded that shape as deliberate and excluded it. Looking at what it now costs, one of the two defeats a fix that landed this week.

Verified on `main` 27170b61.

## 1. `severity = "warning"` silences the mistyped-key case #1717 created

`#1717` made a `cross_source_overlap` group whose key **no sibling carries** report a failing `not_evaluated`, on purpose. Its own comment says why (`engine/crates/rocky-cli/src/commands/run.rs:6741`):

> A key no sibling carries is a MISCONFIGURATION — a mistyped column name, say — not a tolerated shape. It fails, so the operator sees it on the first run instead of reading every group as a silent pass.

But the result is built with the configured severity (`run.rs:6750`):

```rust
rocky_core::checks::cross_source_overlap_not_evaluated(
    name, contributing, reason, overlap_cfg.severity,
)
```

`cross_source_overlap_not_evaluated` takes that severity verbatim (`engine/crates/rocky-core/src/checks.rs:338-347`). So:

```
[pipeline.bronze.checks.cross_source_overlap]
keys = ["order_idd"]      # typo
severity = "warning"      # a documented, schema-exported key

-> every group: not_evaluated, passed = false, severity = Warning
-> replication_check_gate_failed counts error-severity failures only
-> gate clear, exit 0, status: "Success"
```

The operator who wrote `severity = "warning"` meant "a duplicate across siblings is only a warning". The typo means Rocky is checking nothing at all, and the run says Success. That is the same reading #1717 set out to prevent — reinstated by a config key.

The two genuine query-failure sites (`run.rs:6817`, `run.rs:6896`) pass the configured severity too, so a permission or transport failure on the overlap query is silenced the same way.

## 2. The same shape on `[[assertions]]`

[...]

PR #1780, opened 2026-09-07
fix(engine): a check that could not run is never advisory, so a failed query stops clearing the gate

hugocorreia90:
Severity grades a **measurement**. It says how bad three null rows are, or how bad this much overlap is. A check whose query failed produced no measurement to grade — but two constructors carried the declared severity through anyway.

So a test declared `severity = "warning"` whose query failed came back as a warning-severity failure. `check_failures_by_severity` filed it in the warning bucket, `replication_check_gate_failed` reads only the error bucket, and the run exited 0 having verified nothing about that column.

```
severity = "warning", the query fails
  before   not_evaluated, severity=Warning  ->  warning bucket  ->  gate clear  ->  exit 0
  after    not_evaluated, severity=Error    ->  error bucket    ->  gate fires  ->  exit 2
```

## The change

`assertion_not_evaluated` and `cross_source_overlap_not_evaluated` drop the `severity` parameter and hard-code `TestSeverity::Error`. That matches the four constructors that already did it — `row_count_not_evaluated`, `freshness_not_evaluated`, `null_rate_not_evaluated`, `custom_not_evaluated`. 14 call sites lose the argument.

One rule, stated once: **severity governs a measured result, and only a measured result.**

`cross_source_overlap_not_applicable` is deliberately untouched. It sets `passed: true` for a keyless sibling, and `check_failures_by_severity` skips passing checks, so its severity never reaches a bucket at all.

## This was live, not theoretical

Three existing tests asserted the old behaviour. The first is the strongest evidence:

- **`assertions_and_custom_checks_on_an_unaddressable_table_are_not_evaluated`** (`run.rs`) runs a real DuckDB fixture with `severity = "warning"` and a table whose hyphen fails identifier validation. It asserted the result came back `Warning` — a real end-to-end path producing a non-gating failure on a table Rocky never addressed. It now asserts `Error`.
- **`check_failures_by_severity_buckets_not_evaluated_by_its_own_severity`** (`output.rs`) — its *name* stated the defect. Renamed to `..._buckets_every_unevaluated_check_as_an_error` and given a measured warning, so both directions are pinned.
- **`the_check_gate_reads_severity_and_fail_on_error`** (`run.rs`) used `assertion_not_evaluated` as its warning-severity fixture, which can no longer produce one. It now uses `check_assertion` — a measured violation, which is what that row was always meant to represent.

## Evidence

New test `an_advisory_assertion_that_could_not_run_still_gates` walks the real chain, `replication_check_gate_failed` -> `derive_run_status`, for both halves of the rule. Its assertions are ordered so a revert fails on the **exit code**, not on a severity field.

`not_evaluated_constructors_fail_and_round_trip_through_json` now asserts `severity == Error` across all six constructors, so a seventh added later is covered by construction.

Both halves mutation-probed with `scripts/mutation-check.sh` (flipping the hard-coded `Error` back to `Warning`):

```
mut_assertion.py  -> an_advisory_assertion_that_could_not_run_still_gates
   FAILED at "a check that never ran gates"   <- the gate, not the field
mut_overlap.py    -> not_evaluated_constructors_fail_and_round_trip_through_json
   FAILED: left: Warning, right: Error
both: "mutation-check: OK - the test failed under mutation, so it is fix-sensitive."
```

Full suites green: `rocky-core` 2142 passed, `rocky-cli` 1840 passed, 0 failed. Clippy `--all-targets --all-features` clean.

## Breaking

A project with an advisory check whose query has been failing will start failing the run. That is the fix: the old exit code said the data was fine when nothing had been read. Changelog entry included, stated in those terms.

**That `check_failures_by_severity` is the only consumer of a `not_evaluated` check's severity.** I grepped the field and found the bucketer plus the JSON output; if some other reader treats `severity` as "what the user configured" rather than "how this result grades", it now sees `Error` where the config said `warning`. The JSON `severity` field on a `not_evaluated` check is the visible surface of that, and it does change.

**What I may be missing:** whether anyone deliberately sets `severity = "warning"` on a check they expect to fail to evaluate — for example an assertion against a table that only exists in some environments. Under the old behaviour that was a silent pass; under this one it fails the run. The honest answer is that such a config was already broken, since the check was never verifying anything, but the failure will surface as new.

## Review

Red-teamed by the `codex:codex-rescue` plugin — findings and disposition in a comment below. Refs #1741.

### DIFF
--- a/engine/CHANGELOG.md
+++ b/engine/CHANGELOG.md
@@ -68,6 +68,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
      after    not_evaluated, severity=Error    ->  error bucket    ->  gate fires  ->  exit 2
   ```
 
+  The **quality pipeline's `row_count`** had the same defect a fourth time, and worse: its failure arm was hand-built rather than going through a constructor, so it set `not_evaluated: None` — telling every consumer the check had run — and described a row-count check as `CheckDetails::Custom` with a fabricated `result_value: 0, threshold: 1`. It now uses `row_count_not_evaluated`, which gets all three right.
+
   **Breaking:** a project with an advisory check whose query has been failing will start failing the run. That is the point — the previous exit code said the data was fine when nothing had been read. `cross_source_overlap_not_applicable` is unchanged: a keyless sibling passes, and a passing check never reaches the severity buckets. Refs #1741.
 
 - **A model sidecar that was a dangling symlink read as absent, so the model compiled against defaults — different strategy, possibly a different target table, silently.**
--- a/engine/CHANGELOG.md
+++ b/engine/CHANGELOG.md
@@ -56,6 +56,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   **On upgrade:** the state store schema moves v23 → v24. No table is added and no blob migration runs; a v23 store upgrades on the next read-write open. The bump is load-bearing: `observed_failing` is a new variant of the `fulfill_state` record's tagged enum, and a 1.73.0 binary cannot read a blob that carries it. **Rolling back is the part to plan for.** Remote state keys are qualified by schema version (`v23/state.redb` on an object store, `…v23:…` on Valkey), so a 1.73.0 binary never reads a v24 remote object. It reads whatever is under its own v23 key, which may be stale, and it cannot see progress recorded under v24. The incompatible-blob hazard is a **local** `state.redb` this version wrote. A 1.73.0 binary opening it either stops with a schema-mismatch error or, on the paths that honour `[state] on_schema_mismatch`, deletes the file and starts fresh under the default `recreate`. Do not point a 1.73.0 binary at a local state file written by this version, and do not share one local file between versions. `rocky doctor --check state_schema` reports the mismatch and exits 3. Which commands honour the key and which refuse is being pinned down in #1679. `fulfill_state` is local-only, so a re-run rebuilds the loop's position; see #1525 before relying on that loss being harmless.
 
 ### Fixed
+- **A check the engine could not run kept the severity you declared, so a `severity = "warning"` check that failed to execute cleared the gate and the run exited 0.**
+
+  Severity grades a *measurement*: it says how bad three null rows are, or how bad this much overlap is. A check whose query failed produced no measurement to grade, but `assertion_not_evaluated` and `cross_source_overlap_not_evaluated` carried the declared severity through anyway. `check_failures_by_severity` then filed the failure in the warning bucket, `replication_check_gate_failed` reads only the error bucket, and a run that verified nothing about a column reported success.
+
+  The declared severity is now ignored for a check that did not run — every `*_not_evaluated` constructor is `TestSeverity::Error`, matching the four (`row_count`, `freshness`, `null_rate`, `custom`) that already were. A measured violation is unaffected and still grades at the severity you set.
+
+  ```
+  severity = "warning", query fails
+     before   not_evaluated, severity=Warning  ->  warning bucket  ->  gate clear  ->  exit 0
+     after    not_evaluated, severity=Error    ->  error bucket    ->  gate fires  ->  exit 2
+  ```
+
+  **Breaking:** a project with an advisory check whose query has been failing will start failing the run. That is the point — the previous exit code said the data was fine when nothing had been read. `cross_source_overlap_not_applicable` is unchanged: a keyless sibling passes, and a passing check never reaches the severity buckets. Refs #1741.
+
 - **A model sidecar that was a dangling symlink read as absent, so the model compiled against defaults — different strategy, possibly a different target table, silently.**
 
   `Path::exists()` is `std::fs::metadata(..).is_ok()`, and `metadata` **follows** a symlink. A file that is a symlink to a deleted target therefore answers `false`, and every caller that gated a read on that probe read a present file as absent and took its "there is none of this" branch.
--- a/engine/crates/rocky-cli/src/commands/run.rs
+++ b/engine/crates/rocky-cli/src/commands/run.rs
@@ -21654,20 +21654,24 @@ backend = "local"
             Some("email".to_string()),
             "the assertion query failed: no such column: emial",
         );
-        assert_eq!(
-            never_ran.severity,
-            TestSeverity::Error,
-            "an unevaluated check is never advisory"
-        );
-
+        // The gate first, deliberately: this test has to fail on the EXIT
+        // CODE when the fix is reverted, not on a severity field a reader
+        // could dismiss as cosmetic.
         let mut gated = RunOutput::new(String::new(), 0, 1);
         gated.tables_copied = 1;
-        gated.check_results.push(failing_check_bag(never_ran));
+        gated
+            .check_results
+            .push(failing_check_bag(never_ran.clone()));
         gated.check_gate_failed = super::replication_check_gate_failed(&gated, &config);
         assert!(gated.check_gate_failed, "a check that never ran gates");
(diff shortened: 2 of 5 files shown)


hugocorreia90:
## Review status — read this before merging

I dispatched this to the independent red team (the `codex:codex-rescue` plugin). It accepted the task as `task-mtrej841-u3v7jz` and returned a **receipt, not a verdict**. The plugin's status command is reserved for explicit human invocation, so I cannot read the result myself.

**To surface it, run `/codex:status task-mtrej841-u3v7jz`.** Until that lands, the only review on this PR is the same-model pass below. That is not the independent review AGENTS.md asks for, and I am saying so rather than letting the PR read as reviewed.

## Same-model adversarial pass (Claude, not independent)

I attacked the five axes I had given the red team. Three are the ones that could actually sink the change.

### 1. Is `cross_source_overlap_not_applicable` really safe to leave alone?

My argument in the description was that it sets `passed: true` and the bucketer skips passing checks. That argument only holds if the bucketer is the **only** consumer of severity. It is.

Every read of a `CheckResult`'s severity in the workspace:

```
output.rs:5358   match check.severity      <- inside check_failures_by_severity,
                                              which does `if check.passed { continue; }` first
output.rs:2180   finding.severity          <- breaking-change findings, a different type
output.rs:4500   msg.severity == "error"   <- diagnostics, a String
run.rs:31738/31790/31832, run_local.rs:2476/2489/2562, checks.rs:1114
                                           <- all test assertions
```

One consumer, and it tests `passed` before reading severity. **Clear.**

### 2. Is there a second gate path that reads severity?

No, and for a structural reason worth stating: **`CheckOutcome`, the persisted type, has no `severity` field.** It carries `name`, `passed`, `not_evaluated` (`state.rs:2494`). So the `verify_after` gates, the resume gate and `unverified_required_checks` cannot read severity even in principle — they read `passed` and `not_evaluated`. This change cannot reach them. **Clear.**

### 3. Did I miss a site that re-applies the configured severity after construction?

This is the one that would have quietly defeated the whole fix, and it is the axis I would have bet on. There are exactly three post-construction writes to `severity` in the workspace:

```
run.rs:6606       check.severity = freshness_cfg.severity;   guarded: if check.not_evaluated.is_none()
run.rs:7105       check.severity = cfg.severity;             guarded: if check.not_evaluated.is_none()
run_local.rs:729  check.severity = severity;                 on the Some(measured) arm only;
                                                             the None arm deliberately omits it
```

All three are already guarded — by **#1719** (freshness) and **#1735** (null-rate and custom), which fixed exactly this shape on those three check kinds. Their comments state the rule in almost the words I used.

That reframes this PR: it is not a new rule, it is the last two constructors that had not been brought under an existing one. `assertion_not_evaluated` and `cross_source_overlap_not_evaluated` were the remainder. Worth knowing for whoever reviews it — the design question was settled twice already.

No struct-literal bypass either: the two sites that build a `not_evaluated` value directly (`apply.rs:10040`, `state.rs:7314`) are both inside `#[cfg(test)]` and both build `CheckOutcome`, which has no severity.

### 4 and 5, less certain

**Codegen and fixtures.** I did not run `just codegen` or `just regen-fixtures`. My reasoning is that this changes a *value*, not a *shape* — no field added, removed or retyped on `CheckResult`, and `CheckOutcome` is untouched — so the exported schema and the generated bindings cannot drift. CI's "Check codegen is up to date" is the real oracle and it is running on this PR; if it goes red, that reasoning was wrong and I will fix it rather than argue with it.

**The changed tests.** I re-aimed three, and I would want a second pair of eyes on whether any got weaker rather than redirected. My own read: `the_check_gate_reads_severity_and_fail_on_error` is the one to scrutinise, because I swapped its warning fixture from `assertion_not_evaluated` to `check_assertion`. The row it feeds is labelled "a warning-severity failure never gates" and still tests exactly that, now with a fixture that can actually be a warning. But it is a test edit made by the person whose change forced it, which is the shape that deserves suspicion.

That the JSON `severity` field on a `not_evaluated` check is not load-bearing for some consumer outside this repo. Inside it, the field has one reader. Outside it — the Python SDK, Dagster, anyone parsing `rocky run --output json` — a check that used to report `"severity": "warning"` will now report `"severity": "error"`. That is the intended, visible consequence, but it is a wire-visible change and nothing in this repo can prove nobody keys off it.

hugocorreia90:
## The red-team pass found a fourth site, and it was the worst one

The independent reviewer could not be reached (see below), so I ran the pass myself. Working axis 3 — "is there a check kind that reaches `not_evaluated` without going through a constructor?" — turned up a live defect this PR had missed.

`run_quality`'s `row_count` failure arm, `run_local.rs`, built its result by hand:

```rust
Err(e) => checks.push(CheckResult {
    name: "row_count".into(),
    passed: false,
    severity: row_count_severity,   // the CONFIGURED severity
    not_evaluated: None,            // claims the check ran
    details: CheckDetails::Custom { // a row-count check, described as custom
        query: ..., result_value: 0, threshold: 1,
    },
})
```

Three wrongs in one literal:

1. **The configured severity on a failed query** — the exact #1741 shape. A `severity = "warning"` quality pipeline whose `SELECT COUNT(*)` failed produced a warning-severity failure, never reached the error bucket, and the run exited 0.
2. **`not_evaluated: None`** — so every consumer, including the persisted `CheckOutcome` and therefore `unverified_required_checks`, read it as a check that ran and failed rather than one that never ran.
3. **`CheckDetails::Custom` with `result_value: 0, threshold: 1`** — a fabricated measurement, for a check that measured nothing, under the wrong variant.

`row_count_not_evaluated()` already existed for this case and gets all three right.

### Why it survived three previous fixes

This arm was **untestable where it sat**. `run_quality` does not return its `RunOutput`, and the persisted `CheckOutcome` has no `severity` field — so the distinction was observable from nowhere. That is why #1719, #1735 and the first commit on this branch all passed it by.

So I extracted `quality_row_count_check(queried, configured)`, mirroring exactly how #1735 made `run_custom_checks` reachable, and covered both arms:

```
a_quality_row_count_query_that_failed_keeps_error_severity
```

It fails the query the way production fails it — the table named in the config is not in the database — and asserts the measured arm still takes the configured severity, so the rule is narrowed rather than removed. Mutation-probed by restoring the hand-built literal: `mutation-check: OK — the test failed under mutation, so it is fix-sensitive.`

### Where the axes stand now

| Axis | Result |
|---|---|
| Only consumer of `CheckResult::severity` | `output.rs:5358`, inside `check_failures_by_severity`, which skips passing checks first. Clear. |
| A second gate path reading severity | Impossible: `CheckOutcome` (state.rs:2494) has no severity field. Clear. |
| Post-construction severity writes | Three, all guarded by `not_evaluated.is_none()` or on a measured-only arm — by #1719 and #1735. Clear. |
| Constructor bypass | **One found and fixed** (this comment). The only remaining direct `not_evaluated: Some(..)` literals are in `#[cfg(test)]` and build `CheckOutcome`. |
| Adapters / MCP / serve / LSP | The cross-crate `CheckResult` hits are all `TypeCheckResult` or `DraftCheckResult`, different types. Clear. |
| `passed` flipped after construction | The one `.passed = ` in the workspace is a `usize` count on `TestRunResult`. Clear. |

### On the review itself

The `codex:codex-rescue` plugin was tried twice. The first attempt dispatched a background Codex task (`task-mtrej841-u3v7jz`) and returned a receipt; its status command is reserved for human invocation, so I cannot read the result. The second attempt, asking for an inline pass, replied `CANNOT REVIEW INLINE`.

So **there is no independent review on this PR**, and per AGENTS.md I am saying so rather than skipping it silently. What is here is a same-model pass — which did find a real defect, but is not the independent check the policy asks for. `/codex:status task-mtrej841-u3v7jz` may still surface the original verdict if anyone wants it.

Issue #1785, opened 2026-09-07
dagster: check severity is mapped without reading passed or not_evaluated, so an unmeasured overlap group reports severity=warning


hugocorreia90:
`dagster_check_severity` (`integrations/dagster/src/dagster_rocky/checks.py:18`) maps a Rocky check's severity without looking at `passed` or `not_evaluated`:

```python
def dagster_check_severity(check: CheckResult) -> dg.AssetCheckSeverity:
    if getattr(check, "severity", None) == "warning":
        return dg.AssetCheckSeverity.WARN
    return dg.AssetCheckSeverity.ERROR
```

Its own test states this deliberately — the mapper "ignores `passed`" and a passing warning stays `WARN` (`tests/test_checks.py:132`).

## Where that becomes wrong

`cross_source_overlap_not_applicable` (`engine/crates/rocky-core/src/checks.rs:367`) is the one constructor that sets `not_evaluated: Some(..)` **while** `passed: true`, and it still takes the configured severity as a parameter. With two or more siblings where only one carries the key, the runner calls it with the operator's severity (`run.rs:6809`).

So Dagster receives a check that is passing, carries a `not_evaluated` reason, and reports `"severity": "warning"` in its metadata (`checks.py:164`) — for a group Rocky **never measured**.

## Why I did not fold this into #1741

The severity is inert for health: the check passes, and a passing check does not degrade asset health whatever its severity. So this is a reporting inaccuracy, not a gate bypass, and it does not belong in a fail-open fix.

It is filed because the *reason* it was missed matters more than the impact. On #1780 I claimed severity had exactly one consumer, and that the consumer tests `passed` first — which is true of `check_failures_by_severity` and true of nothing else. **I had grepped Rust only, and stated the result as a property of the system.** The engine is not the only reader of its own output; `dagster-rocky` and the SDK read the same JSON.

## What to decide

Three options, and I do not think the first is obviously right:

1. **Leave it.** A passing check's severity is inert in Dagster, so the metadata is cosmetic.
2. **Have `dagster_check_severity` consider `not_evaluated`** — anything unevaluated reports `ERROR`, matching the engine's #1741 rule. This makes the passing-but-unmeasured case report `ERROR` on a *passing* check, which is odd but honest.
3. **Have Dagster surface `not_evaluated` as its own metadata field** so "we did not measure this" is visible without overloading severity. This is probably the real answer — the information is already on the wire and simply is not rendered.

Option 3 is not exclusive with the others.

[...]

Issue #1784, opened 2026-09-07
Dagster Pipes reports every check at ERROR, discarding the configured severity


hugocorreia90:
`rocky run` under Dagster Pipes reports **every** check at `ERROR`, whatever the operator configured.

`engine/crates/rocky-cli/src/commands/run.rs:7327` passes the severity as a literal:

```rust
pipes.report_asset_check(
    &asset_key,
    &check_name,
    passed,
    crate::pipes::PipesCheckSeverity::Error,   // <- every check, always
    &check_value,
);
```

`check.severity` is right there in the loop — `check_value` is the serialized `CheckResult` — and it is not read.

## This is not a missing capability

`pipes.rs` already threads severity end to end. `report_asset_check` takes it as a parameter (`pipes.rs:212`) and writes it to the wire (`pipes.rs:237`), and there is a test asserting a `WARN` reaches the message:

```
pipes.rs:376  fn report_asset_check_includes_severity()
pipes.rs:394  assert_eq!(msg["params"]["severity"], "WARN");
```

So the plumbing is built and tested. The one production call site does not use it.

## What it costs

A `[[checks.assertions]]` entry with `severity = "warning"` — an advisory check the operator deliberately marked non-gating — surfaces through Pipes as an `ERROR` asset check. In Dagster, `ERROR` degrades asset health and can page an `ASSET_HEALTH_DEGRADED` alert; `WARN` does not. So the advisory setting is honoured by the engine's own gate and discarded on the Pipes wire.

Note the direction. This is the **opposite** of #1741, which was about an unevaluated check being treated as advisory when it should gate. Here a measured, deliberately-advisory violation is escalated to a hard failure. Both are the configured severity being ignored; they fail in opposite directions.

## The one thing to decide first

Whether the fix is simply `dagster_check_severity`'s rule in Rust — `warning` → `WARN`, everything else → `ERROR` — or whether it should also carry the #1741 rule, so a check with `not_evaluated: Some(..)` reports `ERROR` regardless of the configured severity.

I think it must carry both, or Pipes reintroduces exactly the hole #1741 closed: an unevaluated advisory check would go out as `WARN` and stop degrading health. `CheckResult` already carries `not_evaluated`, so both facts are available at the call site.

## Provenance

Found by an independent Codex review of #1780 (158 tool calls), reading the rollout rather than the plugin's return value. I verified the call site and the `pipes.rs` plumbing directly before filing.

Refs #1741, #1666.

[...]

PR #1786, opened 2026-09-07
fix(engine): a quality target whose schema cannot be listed stops reporting a clean run

hugocorreia90:
Two defects the **independent Codex review of #1780** found and I did not. I had reported that review as unavailable; it had in fact run — 158 tool calls — and written its verdict to the session rollout, which is readable. That correction is on #1741 in full.

## 1. A schema-wide quality target that cannot be expanded checks nothing and exits 0

The more serious of the two, and a level up from #1741: not a check reported wrongly, but a check **never created**.

A `[[pipeline.x.tables]]` entry with no `table` names a whole schema, which Rocky enumerates with `list_tables_sql`. Both failure arms — SQL generation and execution — did `warn!` then `continue`, emitting no `CheckResult` at all:

```
list_tables_sql fails / the query fails
   -> warn! + continue
   -> no CheckResult
   -> neither severity bucket
   -> error_failures == 0
   -> tables_failed untouched
   -> RunStatus::Success, exit 0
```

A dropped permission or a renamed schema read as a clean bill of health, and the schedule reconciler would treat that run as a satisfied `after` / `freshness` demand.

Both arms now record a failing `custom_not_evaluated` named `schema_expansion` against the schema's asset key. That constructor is hard-coded to `TestSeverity::Error`, so the result lands in the error bucket and the **existing** `error_failures > 0 && fail_on_error` path does the rest — failing the run, setting `tables_failed`, and persisting `Failure`. No new control flow.

`custom_not_evaluated` is the honest constructor here rather than a convenience: the list-tables statement genuinely *is* a query the engine could not evaluate, and `CheckDetails::Custom` carries it so the reason and the SQL both survive.

The test asserts through `check_failures_by_severity` — the same function `run_quality` calls to compute `error_failures` — so it pins the number that drives the exit code, not the presence of a struct.

## 2. The "every constructor" test covered six of seven

`not_evaluated_constructors_fail_and_round_trip_through_json` claims in its doc comment to cover every `*_not_evaluated` constructor and to make a future one safe by construction. There are **seven**. It listed six. The omission is `column_match_not_evaluated` (`checks.rs:453`).

I also wrote "all six constructors" and "a seventh added later is covered by construction" on #1780. Both wrong, and the seventh was the one missing.

Added, and the doc comment now states the count with the `grep` that checks it:

```
grep -c "^pub fn .*_not_evaluated" checks.rs
```

so the next reader can verify the claim instead of trusting it. A claim about completeness that cannot be checked is how this one survived.

## Evidence

```
cargo test -p rocky-core -p rocky-cli --features duckdb   all green, 0 failures
cargo clippy --all-targets --all-features                 clean
cargo fmt --all                                           clean
```

## Filed, not fixed here

Two further review findings, both of which move a severity users actually see, and both of which belong with #1666's "should a measured result honour the configured severity" question rather than with a fail-open fix:

- **#1784** — Dagster Pipes passes `PipesCheckSeverity::Error` as a literal for every check (`run.rs:7327`), discarding the configured severity. `pipes.rs` threads severity properly and has a test expecting `"WARN"`; the call site never uses it. This is the **opposite** direction from #1741: a measured, deliberately-advisory violation escalated to a hard failure.
- **#1785** — `dagster_check_severity` maps severity without reading `passed` or `not_evaluated`, so `cross_source_overlap_not_applicable` reports `"severity": "warning"` for a group Rocky never measured. This is the finding that showed my #1780 claim — "severity has exactly one consumer, and it tests `passed` first" — was a Rust-only grep stated as a property of the system.

**The check name `schema_expansion`.** It is a name no consumer pre-declares, so a Dagster user who has declared the four default specs will see this result dropped by `_emit_results` rather than surfaced. The engine's own gate is the primary consumer and it works, so the run does fail — but the orchestrator will not say why. Fixing that properly means declaring the name somewhere consumers can see it, which is the same projection question as #1645's second half.

**What I may be missing:** whether a schema that legitimately contains zero tables is now distinguishable from one that could not be listed. It is — an empty list yields no checks and no failure, while a failed list yields this result — but nothing tests the empty-schema case, and I have not added one because I could not tell from the config whether an empty schema is meant to be an error.

## Review

Reviewer: **Codex, independent**, via the plugin, read from the session rollout. Its findings on #1780 are dispositioned in full on #1741. This follow-up has not itself been re-reviewed. Refs #1741.

### DIFF
--- a/engine/crates/rocky-cli/src/commands/run_local.rs
+++ b/engine/crates/rocky-cli/src/commands/run_local.rs
@@ -438,6 +438,41 @@ fn quality_row_count_check(
     }
 }
 
+/// Records that a schema-wide quality target could not be EXPANDED, as a
+/// failing check the engine could not evaluate.
+///
+/// A `[[pipeline.x.tables]]` entry with no `table` names a whole schema, and
+/// Rocky enumerates it with `list_tables_sql`. When that SQL cannot be built or
+/// cannot be run, no table is known — so every check for that target is
+/// unevaluated.
+///
+/// Both arms used to `continue` after a `warn!`, emitting NO `CheckResult` at
+/// all. Nothing reached either severity bucket, `error_failures` stayed 0, and
+/// a run that had checked nothing persisted `Success` and exited 0. A dropped
+/// permission or a renamed schema read as a clean bill of health.
+///
+/// `custom_not_evaluated` is the right constructor rather than a
+/// convenience: the list-tables statement genuinely IS a query Rocky could not
+/// evaluate, and the constructor is hard-coded to `TestSeverity::Error`, so
+/// this lands in the error bucket and the existing
+/// `error_failures > 0 && fail_on_error` path fails the run (#1741).
+fn checks_for_unexpandable_target(
+    output: &mut RunOutput,
+    table_ref: &rocky_core::config::TableRef,
+    list_sql: &str,
+    reason: String,
+) {
+    output.check_results.push(TableCheckOutput {
+        asset_key: vec![table_ref.catalog.clone(), table_ref.schema.clone()],
+        checks: vec![rocky_core::checks::custom_not_evaluated(
+            "schema_expansion",
+            list_sql,
+            0,
+            reason,
+        )],
+    });
+}
+
 /// Execute `rocky run` for a quality pipeline.
 ///
 /// Runs data quality checks against the specified tables without any data movement.
@@ -500,7 +535,13 @@ pub async fn run_quality(
                             catalog = table_ref.catalog.as_str(),
                             schema = table_ref.schema.as_str(),
                             error = %e,
-                            "failed to build list-tables SQL — skipping"
+                            "failed to build list-tables SQL"
+                        );
+                        checks_for_unexpandable_target(
+                            &mut output,
+                            table_ref,
+                            "",
+                            format!("could not build the list-tables SQL: {e}"),
                         );
                         continue;
                     }
@@ -516,7 +557,13 @@ pub async fn run_quality(
                             catalog = table_ref.catalog.as_str(),
                             schema = table_ref.schema.as_str(),
                             error = %e,
-                            "failed to list tables in schema — skipping"
+                            "failed to list tables in schema"
+                        );
+                        checks_for_unexpandable_target(
+                            &mut output,
+                            table_ref,
+                            &list_sql,
+                            format!("could not list the tables in this schema: {e}"),
                         );
                         continue;
                     }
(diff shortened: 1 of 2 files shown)


hugocorreia90:
## A pre-existing hazard this change routes a new failure through

The independent Codex review of this PR is still running, but it has already surfaced something worth recording here, because it bears directly on whether this fix does what the PR description claims.

Its observation: *"the quality runner emits JSON before it stamps `tables_failed`, so the persisted record and process exit can say failure while the JSON still says `status: success`."*

I verified it, and it is worse than stated. **`run_quality` never derives `output.status` at all.** `RunOutput::new` defaults the field to `RunStatus::Success` (`output.rs:4935`), and across `run_quality`'s whole body (`run_local.rs:446-742`) there is no assignment to it — the only `output.status = output.derive_run_status()` in that file is at line 344, inside `run_transformation`.

```
a quality run with a failing error-severity check

  --output json     "status": "success",  "tables_failed": 0
  persisted record  RunStatus::Failure
  process exit      non-zero
```

Filed as #1788 with the full trace.

## What that means for this PR

**The fix still works.** The new `schema_expansion` result reaches the error bucket, `error_failures` becomes non-zero, `tables_failed` is stamped, the persisted record derives `Failure`, and the process exits non-zero. Every claim in the description holds.

**But the JSON payload for that run will say `"status": "success"`** — as it already does for every other quality check failure. So a consumer reading only the payload sees a clean run where previously it saw a clean run. For that specific consumer, this fix changes nothing.

I am not folding #1788 into this PR. It changes the payload for **every** quality run that was already failing, not just the case added here, and it needs the `#1604` justification in the existing comment traced before the ordering moves. That is a separate blast radius and deserves its own review rather than riding along.

Recording it here so the fix is not read as more complete than it is: this closes the engine-side gate, and the JSON contract is still wrong until #1788 lands.

hugocorreia90:
## Review disposition

Reviewer: **Codex, independent**, read from the session rollout. It ran to `task_complete` but answered **three of the five axes** it was given — axes 4 and 5 were not delivered. Recording that rather than presenting the pass as complete; I answered axis 5 myself below, and axis 4 remains open.

### Axis 1 — does `schema_expansion` reach a real failure? CONFIRMED, with a caveat already filed

It traced the chain and confirmed the local failure reaches exit 1. The caveat is #1788 (the JSON payload still says `status: success`), which I filed before this landed. It added one amplification I had not traced: the SDK accepts non-zero-exit JSON under `allow_partial=True` (`sdk/python/src/rocky_sdk/client.py:827`) and returns that success-shaped `RunResult` — so a consumer using that flag sees the wrong status *and* no exception. Added to #1788.

It also answered the question I flagged as least certain in the description: **an empty schema and an unlistable schema are now distinguishable.** A schema that lists successfully with zero tables returns an empty vector (`run_local.rs:549`) and succeeds with zero checks; a listing failure produces the new failing check. Different outcomes, as intended.

Two inert cases, both pre-existing and both correct: `fail_on_error = false` still emits the failed check and deliberately exits 0 (documented escape hatch, `config.rs:1216`), and `checks.enabled = false` bypasses expansion entirely.

### Axis 2 — does the new check name break anything downstream? NOTHING FOUND

This was my other stated uncertainty, and the answer is better than I guessed. The name is an unrestricted string; the result fits the existing custom-check variant in `run.schema.json`; both the handwritten and generated SDK models accept it. In `RockyComponent`, an undeclared selected check becomes an **observation**, not an invalid `AssetCheckResult` (`component.py:3345`, `3584`), and the Pipes path applies the same guard — so it cannot cause a `DagsterInvariantViolationError`. My caveat about the name was over-cautious.

One forward-looking note worth keeping: if quality execution is ever added to the Dagster component, a schema-level `[catalog, schema]` result will usually fall outside the selected table assets and be dropped with a warning, and the `check_gate_failed` backstop would not rescue it because this path leaves that flag false.

### Axis 3 — CONFIRMED, and it is a hole in the merged #1782, not in this PR

Filed as **#1790**. `PASS_BY_ABSENCE_CHECK_NAMES` assumes the event producers are unconditional. They are not: `run.rs:6499` runs anomaly detection only `if row_count_enabled && let Some(store) = state_store`, so with `row_count = false` the detector never runs and #1782 reports the check green. Verified directly.

### Axis 5 — not delivered; answered here

*Are there other `continue` or early-return sites in `run_quality` that skip work without recording a check?*

I audited the function's whole span. Three other sites, none of the same class:

| site | what it does | verdict |
|---|---|---|
| `format_table_ref(..).unwrap_or_else(..)` | falls back to an unquoted `cat.schema.table` | degrades, does not skip — checks still run |
| `"failed to compile quarantine SQL — skipping"` | skips a remediation | different class: quarantine is an action, not a check, and the assertions that drive it still report their own failures |
| `StateStore::open(..)?` | propagates | fatal by design, not a silent skip |

So the two arms this PR fixes were the only sites that skipped work **and** produced no check result. That is the claim the PR makes, and it holds.

### Axis 4 — not delivered, still open

*Can a transient unproduced check, now recorded as `passed=False`, get carried forward indefinitely by the prune path as a permanent red?* Unanswered by the review and not traced by me. It concerns #1782's merged behaviour rather than this PR, and I have noted it on #1790's thread as the remaining question there.

## Merging

Every axis that bears on **this** change is answered: the fix reaches a real failure (1), breaks nothing downstream (2), and covers every site of its class (5). The two confirmed holes are in other code and are filed — #1788 (JSON status) and #1790 (anomaly false-green).

Issue #1788, opened 2026-09-07
A quality pipeline's --output json always reports status: success, even when the run fails its check gate and exits non-zero


hugocorreia90:
`rocky run --output json` on a **quality** pipeline always reports `"status": "success"`, including on a run that fails its check gate, exits non-zero, and persists `Failure`.

## The three signals disagree

```
a quality run with a failing error-severity check

  --output json     "status": "success",  "tables_failed": 0     <- wrong
  persisted record  RunStatus::Failure                            <- right
  process exit      non-zero (anyhow::bail!)                      <- right
```

## Why

Two facts compose.

**1. `run_quality` never derives its status.** `RunOutput::new` defaults the field to `RunStatus::Success` (`engine/crates/rocky-cli/src/output.rs:4935`), and across `run_quality`'s whole body (`run_local.rs:446-742`) there is no assignment to `output.status`. The only `output.status = output.derive_run_status()` in that file is at `run_local.rs:344`, inside `run_transformation`. So the value emitted is the constructor default, never a derivation.

**2. The `tables_failed` correction lands after the JSON is already out.**

```
run_local.rs:681   crate::output::print_json(&output)?;          <- emitted here
run_local.rs:716   output.tables_failed = error_failures;        <- corrected here
run_local.rs:740   anyhow::bail!("quality pipeline failed: ...")  <- exits here
```

The ordering is deliberate and documented — the comment above line 716 says it is "Done after the JSON emit above so the `--output json` payload is unchanged — the persisted record, not the payload, carries the corrected status."

That comment is accurate about what the code does. The problem is that it treats an incorrect payload as the thing to preserve.

## Why it matters

`status`'s own doc comment (`output.rs:220-226`) makes the opposite promise:

> Always populated — non-skipped runs derive this field from `tables_failed` / materialization counts, so **JSON consumers no longer need to re-derive status from counts themselves**.

`rocky-sdk` and `dagster-rocky` are exactly those consumers, and they read this field. A consumer that trusts the documented contract reads `success` from a run that checked nothing or failed its checks. A consumer that also inspects the exit code catches it; one that parses the payload alone does not.

This is the same family as #1741 and #1645 — a signal that reports "fine" when nothing established that — but at the JSON contract rather than at a check.

## The fix, and the thing to decide first

Mechanically it is small: stamp `tables_failed` and set `output.status = output.derive_run_status()` **before** `print_json`, not after.

[...]

PR #1795, opened 2026-09-07
fix(engine): a quality run's JSON reports the status it actually reached

hugocorreia90:
`rocky run --output json` on a quality pipeline **always** reported `"status": "success"` — including on a run that failed its check gate, exited non-zero, and persisted `Failure`.

```
a quality run with a failing error-severity check

  --output json     "status": "success",  "tables_failed": 0     wrong
  persisted record  RunStatus::Failure                            right
  process exit      non-zero                                      right
```

## Two faults composed

**1. `run_quality` never derived `status` at all.** `RunOutput::new` defaults the field to `RunStatus::Success`, and across the whole function there was no assignment to it — the only `output.status = output.derive_run_status()` in the file is in `run_transformation`. So the emitted value was the constructor default, whatever the checks found.

**2. The `tables_failed` correction ran after the payload was printed.** The comment above it said so, and treated the uncorrected payload as the thing to preserve:

> Done after the JSON emit above so the `--output json` payload is unchanged — the persisted record, not the payload, carries the corrected status.

Accurate about what the code did. The problem is that the payload it preserved was wrong.

## Why this matters

`status`'s own doc comment makes the opposite promise:

> Always populated — non-skipped runs derive this field from `tables_failed` / materialization counts, so **JSON consumers no longer need to re-derive status from counts themselves**.

`rocky-sdk` and `dagster-rocky` are exactly those consumers. Worse, the independent review of #1786 traced one more step: the SDK accepts a non-zero-exit payload under `allow_partial=True` (`client.py:827`) and returns that success-shaped `RunResult` — so a caller using that flag saw the wrong status **and** no exception.

## The change

The correction now runs before the emit and stamps `status`, exactly as `run_transformation` already did. Extracted as `finalize_quality_status` so the logic is testable and the ordering is one visible call site rather than two scattered blocks.

**The justification I checked before moving it:** `#1604` governs *how* the payload is printed — through `print_json`, so the watch loop's `COMPACT_JSON` promise of one compact object per line is honoured — not *when* it is built. Nothing there depended on the stale value. That was the only thing standing behind the old ordering, and it turned out not to be about ordering at all.

## Tests

| test | covers |
|---|---|
| `a_failed_quality_gate_is_stamped_onto_the_payload_not_just_the_record` | the defect: payload agrees with exit code and with the persisted derivation |
| `fail_on_error_false_keeps_the_payload_successful` | **the discriminator** |
| `a_clean_quality_run_is_unchanged` | no invented `tables_failed` on a clean run |

The middle one is the one that matters. `fail_on_error = false` is a documented escape hatch that deliberately exits 0, so the payload must keep saying `success`. Without that assertion the fix could have been "always report `Failure` when any check failed" — a different and wrong change that the first test alone would not have caught.

## Evidence

```
cargo test -p rocky-cli --features duckdb --lib     all green
cargo clippy --all-targets --all-features           clean
```

Mutation-probed by removing the `status` derivation:

```
```

**Whether any consumer depends on the old value.** The change makes a previously-`success` payload report `failure` for runs that were already failing — visible to anyone parsing it. I traced the in-repo consumers (the SDK's `parse_rocky_output`, the Dagster resource) and neither switches on `status` in a way this breaks; both already treat a non-zero exit as the authority. But a user script that parses the JSON and ignores the exit code will start seeing failures it did not see before. That is the point of the fix, and it is still a behaviour change worth stating.

**What I may be missing:** whether `run_snapshot` — the third runner in this file, which also calls `print_json` at line ~1229 — has the same defect. I did not check it. If it does, it is the same fix and deserves the same treatment, but I would rather land this one and look than widen the change on a guess.

## Review

No independent red team on this PR. The defect itself came from the Codex review of #1786, read from the session rollout. Refs #1788.

### DIFF
--- a/engine/crates/rocky-cli/src/commands/run_local.rs
+++ b/engine/crates/rocky-cli/src/commands/run_local.rs
@@ -473,6 +473,36 @@ fn checks_for_unexpandable_target(
     });
 }
 
+/// Brings a quality run's `tables_failed` and `status` into agreement with the
+/// verdict its checks produced, before anything is emitted or persisted.
+///
+/// `RunOutput::derive_run_status` keys only on
+/// `tables_copied`/`tables_failed`/`interrupted` and IGNORES `check_results`,
+/// and nothing else in `run_quality` touches `tables_failed`. So a quality run
+/// with FAILING error-severity checks derived `Success`, and the schedule
+/// reconciler would treat a failed quality gate as a satisfied
+/// `after`/`freshness` demand. This maps the SAME condition the exit code uses
+/// (`error_failures > 0 && fail_on_error`) onto `tables_failed`.
+///
+/// It also stamps `status`, which `run_quality` never derived at all — the
+/// field kept the `RunStatus::Success` that `RunOutput::new` defaults it to,
+/// whatever the checks found.
+///
+/// **Call this before the JSON emit** (#1788). Doing the correction afterwards
+/// left the payload saying `"status": "success"` while the persisted record
+/// said `Failure` and the process exited non-zero — three signals, two of them
+/// right. `status`'s own doc comment promises consumers they "no longer need to
+/// re-derive status from counts themselves", and `rocky-sdk` and
+/// `dagster-rocky` are those consumers. `#1604` governs HOW the payload is
+/// printed (through `print_json`, so `COMPACT_JSON` is honoured), not when it
+/// is built, so nothing there depended on the stale value.
+fn finalize_quality_status(output: &mut RunOutput, error_failures: usize, fail_on_error: bool) {
+    if error_failures > 0 && fail_on_error {
+        output.tables_failed = error_failures;
+    }
+    output.status = output.derive_run_status();
+}
+
 /// Execute `rocky run` for a quality pipeline.
 ///
 /// Runs data quality checks against the specified tables without any data movement.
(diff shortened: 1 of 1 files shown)


hugocorreia90:
## Answering the open question in this PR's description

I wrote that I had not checked whether `run_snapshot` — the third runner in this file — shares the defect, and that I would rather land this and look than widen on a guess. Looked, and it does. Added it here rather than filing a follow-up, since it is one line in the same file and the branch was re-running CI anyway.

**It is a worse-reading version of the same omission.** `run_snapshot` *does* set `tables_failed` before it emits, so the counts in its payload were already right. But it never derived `status` either, so a failed snapshot shipped:

```json
{ "status": "success", "tables_failed": 1 }
```

A payload that contradicts itself inside one object. The quality case was at least internally consistent — both fields wrong together, disagreeing with the exit code.

So the tally for this one omission is three sites:

| runner | before |
|---|---|
| `run_transformation` | correct — derives `status` before the emit, with a comment saying why. This is where the pattern comes from. |
| `run_quality` | never derived; the `tables_failed` correction also ran after the emit |
| `run_snapshot` | never derived; counts were right, so the payload disagreed with itself |

New test: `a_failed_snapshot_payload_does_not_contradict_its_own_counts`.

```
cargo test -p rocky-cli --features duckdb --lib   all green, 0 failures
cargo clippy --all-targets --all-features         clean
```

Issue #1766, opened 2026-09-07
A backfill plan reaches the review queue with no blast radius, so two backfills cannot be ranked


hugocorreia90:
A backfill plan reaches the review queue with `blast_radius` unset. The queue and the UI both render it as `not computed`.

That is the one field that lets a reviewer rank two backfills against each other. Without it, two pending backfills are indistinguishable on the surface that exists to tell them apart.

## Reproduce

```bash
rocky playground /tmp/bf && cd /tmp/bf
rocky backfill --model raw_orders
rocky backfill --model revenue_summary
rocky --output json review --queue
```

Every pending row:

```json
{
  "capability": "backfill",
  "model": "backfill: 3 model(s)",
  "blast_radius": null,          ← here
  "classification_weight": 2,
  "score": 2.0
}
```

## What that does to the ranking

`queue_score` handles the missing value deliberately, and documents it:

```rust
// crates/rocky-cli/src/commands/review.rs:994
/// An unknown blast radius (compile failed / model removed) contributes as
/// zero so the entry still ranks on class and age rather than dropping out.
(blast_radius.unwrap_or(0) + 1) as f64 * f64::from(classification_weight) * staleness_factor
```

That fallback is right for the case it names — a plan whose blast radius *could not* be computed. This issue is that backfills take it every time, so their scores collapse onto `1 × 2 × (1 + hours)`. Three plans created seconds apart scored `2.0005`, `2.0`, `2.0`: identical to three decimal places, ordered by age alone. The class term is the same for every backfill, so age is the only signal left.

## Why it is odd

The information exists at the moment the escalation is recorded. `commands/backfill.rs` already computes `ordered` — the resolved closure of models the backfill would rebuild — and uses its length for the label:

```rust
// crates/rocky-cli/src/commands/backfill.rs:327
&format!("backfill: {} model(s)", ordered.len()),
```

The escalation is recorded in the same function, with the closure in hand.

## What I am not proposing

Not that `ordered.len()` *is* the blast radius — a backfill's downstream blast radius is not the same as the size of the set it rebuilds, and the schema-change path computes the field a particular way. The right derivation should reuse that one rather than being invented here. The point of this issue is narrower: the field is empty on a path where the input is available, and the "could not be computed" fallback is doing duty for "was never computed".

## Where it shows

- `rocky review --queue` — the `blast_radius` field and the `score` it feeds
- the UI review screen — `blast radius: not computed`, in the queue card and on the plan detail page

[...]

PR #1808, opened 2026-09-08
fix(engine/rocky-cli): record the models behind a plan-level review escalation, so the queue can rank it

hugocorreia90:
## What

A plan-level review escalation now records the **models** behind its human label, so `rocky review --queue` can rank it and `rocky audit --for <model>` can find it.

## Why

`record_plan_review_escalation` writes one ledger row per plan and puts a human summary in `PolicyDecisionRecord::model`. All three callers do it:

```
backfill.rs:322   "backfill: {N} model(s)"                    -- no model name at all
gc.rs:927         "gc: {N} artifact(s) across {M} model(s)"   -- no model name at all
restore.rs:315    "restore: {name} ({hash}...)"               -- a name wrapped in a sentence
```

None of those is a graph key. The queue ranks with `blast_radius_of`, which takes one model name and starts with `result.semantic_graph.model_schema(model)?`, so **every plan-level escalation came back with an unknown blast radius** and ranked on change class and age alone. Two pending backfills, or two pending gc plans, were separated only by which arrived first — including one that would delete a leaf and one that would delete a model three others read from.

A second consequence nobody had reported: `rocky audit --for <model>` compares the same field (`audit.rs:331`), so **none of these rows was findable by any model name a user would type**. A backfill of `dim_customer` sits pending review while `rocky audit --for dim_customer` says "no policy decisions recorded for this subject".

`restore.rs` is what settles that this is a class, not a backfill quirk: it has a real model name available and still fails, because the name is wrapped in display text. "Match the siblings" was not the fix — the siblings have the same bug. Scope widened deliberately to all three sites, [recorded on the issue](https://github.com/rocky-data/rocky/issues/1766#issuecomment-5576527541) before this branch was opened.

## How

```rust
PolicyDecisionRecord {
    model:  "backfill: 3 model(s)",                       // human label, UNCHANGED
    models: ["dim_customer", "fct_orders", "agg_daily"],  // the graph keys
}
```

`graph_keys()` yields the set when there is one and the single `model` when there is not, so an ordinary evaluation row — where `model` **is** the key — is untouched. The queue ranks a plan by its **widest** member: a plan is as risky as its riskiest model.

`max()` over no resolvable key stays `None` — the honest "unknown" the ranking already degrades on — and never collapses to a measured `Some(0)`.

## What deliberately does NOT change

- **The label.** Display text and the `decision_ref` drill key are byte-identical, so nothing in the SPA, `brief`, or the governor MCP path moves. (`rocky-mcp` passes `pending` through as a `serde_json::Value`; nothing parses `decision_ref`.)
- **One ledger row per plan**, so the approve path's queue cardinality is untouched.

## What DOES change

`rocky audit --for <model>` now finds these rows by the models they touch, and stops matching the summary string itself. The summary was never a model name, so no query anyone would type is lost — but it is a shipped behaviour change, so it is asserted in a test rather than left to be discovered.

## State schema 27 -> 28

`models` is serde-additive. A v27 row reads back with the set empty and falls back to `model`, which is exactly today's behaviour, so a pre-v28 plan-level row still ranks unknown rather than wrong. The blob is backward-safe in meaning too: a v27 binary reads only `model`, unchanged.

The pin gate (`on_disk_state_schema_format_is_pinned`) caught the bump and its message asks for the migration path to be confirmed in the same PR, so `a_v27_store_upgrades_in_place_and_keeps_its_policy_decisions` does that: a v27 store is stamped forward **in place**, the decision row survives, and reads back empty. "Kept" is the load-bearing half — a bump that quietly recreated the store would discard every pending review escalation.

## Evidence

`two_pending_gc_plans_rank_by_what_they_would_delete_not_by_age` is the fixture this was missing. Backfill's collision was documented; gc's was assumed to be "the same code shape" without anyone running it. It is the same user-visible collision. The discriminator is deliberate: the harmless plan sorts **first** on the plan-id tie-break, so it leads the queue unless the ranking really resolved the other plan's models.

Mutation-checked with `scripts/mutation-check.sh` — each of these fails when the thing it guards is reverted:

|---|---|
| `graph_keys()` -> `&d.model` in the ranking | `two_pending_gc_plans_rank_by_what_they_would_delete_not_by_age`, `a_backfill_ranks_on_its_widest_model_whatever_the_order` |
| `.max()` -> `.max().or(Some(0))` | `a_plan_level_row_with_no_resolvable_model_ranks_unknown_not_zero` |
| `graph_keys()` -> `d.model ==` in the audit matcher | `a_plan_level_decision_is_found_by_its_models_not_by_its_label` |
| drop `graph_keys`' fallback to `model` | `graph_keys_prefers_the_set_and_always_yields_at_least_one_key`, `test_v27_policy_decision_forward_deserializes_models_empty` |

Also pinned: an ordinary row is still found by its own model name (the regression half — without it, dropping the fallback would break almost every `audit --for` and no plan-level test would notice); a backfill ranks on its widest model whichever order the set arrives in; and the gc label's model count now comes from the same derivation as the recorded set, so the text cannot say "4 model(s)" while the row carries 3.

## Rejected shapes, recorded so they are not re-invented

- **One representative name.** Picks a winner with no rule and understates a multi-model plan.
- **One row per model.** Changes queue cardinality, and an approval marker empties the queue via `compute_review_queue` / `is_reviewed` — whether one approval correctly clears all N is unverified.

No `*Output` struct changed, so no `just codegen` cascade.

Refs #1766


hugocorreia90:
## Independent red-team disposition

**Reviewer: Codex (GPT-5), 92 tool calls, read-only at head `97ceeb5d`.** Its own second reviewer was `gpt-5.6-terra`. **Verdict: BLOCK** — three correctness defects and one test gap. Three are fixed, one is rebutted with the owner's decision on the record.

It also ran mid-review at an earlier head and caught two defects there, both fixed in `97ceeb5d` before it finished: the exclusive `graph_keys()` breaking the audit screen's custody link, and `audit --product` still ignoring the new keys.

---

### 1. `max` is the wrong aggregation — **ACCEPTED, fixed**

> a plan over two unrelated roots with three downstream models each reaches six, and `max` calls it three — the same number as a plan over one of them

Correct, and it means this PR had moved the #1766 defect rather than removed it: two plans that reach very different amounts still ranked identically. The reviewer grounds it in the codebase rather than in taste — the glossary (`docs/.../glossary.md:22-24`) defines a blast radius as a **set**, and `build_blast_radius_link` (`audit.rs:635-649`) already unions across subjects. I invented a second rule at the call site instead of reusing the one that was there.

Now a deduplicated union in `blast_radius_union`, beside `blast_radius_of`, so there is one derivation. `sum` is rejected too and pinned by `overlapping_closures_are_counted_once`.

### 2. A partly resolvable plan reports a measured zero — **ACCEPTED, fixed**

The sharpest finding, because a test in this PR claimed to guard exactly this property and did not.

```
models = ["c", "deleted_since"]      c is a LIVE LEAF

  blast_radius_of("c")             -> Some(empty)   a real, measured zero
  blast_radius_of("deleted_since") -> None          dropped by filter_map
  max([0])                         -> Some(0)
```

The queue rendered `0 downstream` for a plan half of which could not be looked up. `a_plan_level_row_with_no_resolvable_model_ranks_unknown_not_zero` only covered sets where *nothing* resolved, so it passed.

Resolution is now all-or-nothing. `a_partly_resolvable_plan_is_unknown_not_a_measured_zero` covers it, and its second half pins that the same leaf **alone** still measures `Some(0)` — so the test cannot pass by making every leaf unknown.

### 3. The schema bump is gratuitous — **REBUTTED on the facts, and the owner has ruled: keep v28**

The reviewer is right that the bump is **optional**, and it cites the codebase correctly. `state.rs:470-480` says of v11:

> "The version bump is deliberate visibility for the identity surface … rather than a storage-layout requirement; a leaner precedent (`ModelExecution::tenant`) added a blob field without a bump."

Verified independently, along with its cost claim: `SchemaMismatchPolicy::Recreate` **is** the default (`config.rs:647-657`), so a v27 binary opening a v28-stamped store bootstraps fresh local state and does one full-refresh run.

Two corrections to the framing:

- **"Destructive rollback path" overstates it.** `Recreate` never writes the downgraded state back to the shared tier — that is stated at its definition, and the loss is local. The doc calls it "a graceful one-time full refresh during the mixed-version window", i.e. the designed behaviour for a version bump, not an accident this PR introduces.
- **The precedent cuts both ways.** v11/`tenant` establishes that bumping is a choice. v17 and v18 — the two previous field additions to **this same struct** — both bumped, with the stated reason "the bump tracks the record-shape addition so `[state] on_schema_mismatch` engages as usual".

So it is a judgment call, not a defect. Put to the owner with both costs stated: **keep v28.** Reasons on the record — consistency with v17/v18 on the same struct; `rocky doctor` reports the version and the dagster fixtures capture it, so a silent shape change to a persisted ledger is precisely what a version number is for; and the downgrade cost is bounded and by design.

### 4. Changed plumbing is not mutation-pinned — **ACCEPTED, fixed**

> the new test at `audit.rs:1422-1457` only calls `PolicyDecisionRecord::graph_keys()`

True, and worse than it sounds: the test named after the fix did not execute the fix. Restoring `d.model == scope.output_model` left it green. It now drives `compute_audit` against a real spec fixture.

The reviewer also named two producer mutations that would have survived — `ordered.clone()` → `Vec::new()` at `backfill.rs`, and `gc_plan_models(&plan)` → `Vec::new()` at `gc.rs`. Both were real: every ranking test in `review.rs` hands `record_plan_review_escalation` a model set it invented, so nothing reached the production call sites. Pinned by extending the two existing end-to-end tests that already drive the real producers.

`restore.rs` is still pinned only at the derivation level — there is no end-to-end restore test that reads the ledger back, and building one needs a tombstone fixture. Said plainly rather than glossed.

---

## Mutation evidence, re-run after these changes

Every one verified with `scripts/mutation-check.sh` — the test fails when the thing it guards is reverted.

|---|---|
| `blast_radius_union` → `max` over per-model radii | `two_disjoint_roots_reach_further_than_one`, `a_partly_resolvable_plan_is_unknown_not_a_measured_zero` |
| union → sum (double-count an overlap) | `overlapping_closures_are_counted_once` |
| drop the all-or-nothing rule (skip an absent member) | `a_partly_resolvable_plan_is_unknown_not_a_measured_zero` |
| `graph_keys()` → exclusive (the custody-link regression) | `a_plan_level_decision_is_found_by_its_models_not_by_its_label` |
| audit `--for` matcher → `d.model ==` | same |
| `audit --product` filter → `d.model == scope.output_model` | `a_products_custody_chain_includes_a_plan_level_row_touching_its_output_model` |
| `backfill.rs` call site → `Vec::new()` | `backfill_plan_records_review_escalation_and_enters_queue` |
| `gc.rs` call site → `Vec::new()` | `gc_plan_enters_review_queue_and_clears_on_approve` |
| drop `graph_keys`' fallback to `model` | `graph_keys_prefers_the_set_and_always_yields_at_least_one_key` |

## Reviewer caveats, carried rather than dropped

- **Read-only, no tests run.** All findings are static control-flow tracing.
- **GitHub was unreachable** from the review sandbox, so it worked from local Git history rather than this PR's prose. Its own "most important thing missing" is that it could not read this description or the mutation transcript.
- Its "least confident" is the product question behind finding 1: *what does blast radius mean for a multi-model plan?* Answered here as the union of what the plan reaches, with the plan's own models **kept in** — because a backfill's list is transitively closed by construction, so removing them would make every backfill measure zero and re-open the collapse. That reasoning is in `blast_radius_union`'s doc comment rather than only here.
- **ModelIr / contract semantics preserved: yes** — this changes decision-ledger metadata and the queue/audit projection only.

Issue #1817, opened 2026-09-08
The path-probe class is closed at the leaf and open one directory up — a dangling ancestor still reads as absence


hugocorreia90:
Found by an independent red-team pass over the whole path-probe class (#1739, #1757, #1758, #1764, #1787, #1789, #1791, #1792). It ran 86 tool calls and its verdict was recovered after a network cut-off.

## The one-line finding

**Every fix in the class hardened the LEAF and none of them walks the ANCESTORS.** A dangling or unreadable directory *above* the file makes the leaf probe return `NotFound`, which every one of those fixes still classifies as honest absence.

```
  /data/proj -> /mnt/gone     (dangling symlink: the ANCESTOR)
        |
        v
  /data/proj/rocky.toml
        |
        +-- read()               -> NotFound
        +-- symlink_metadata()   -> NotFound     <-- the leaf-only fix agrees
        |
        v
  "the file is absent"   <-- WRONG. It is unreachable, not absent.
```

`symlink_metadata` on the leaf is what these fixes added, and it is exactly as blind to a broken ancestor as `exists()` was. The class is closed at the leaf and open one directory up.

### 1. Retention sweep picks destructive defaults — #1739

`engine/crates/rocky-core/src/path_presence.rs:43` returns `Absent` on a leaf `NotFound`. That flows to `engine/crates/rocky-core/src/config.rs:6247` as `ConfigError::FileNotFound`, which selects **default retention** at `engine/crates/rocky-cli/src/commands/state.rs:162` and sweeps at `:187`.

A dangling ancestor above `rocky.toml` therefore deletes history that the project's real, reachable policy would have retained. This is the same user-visible harm #1739 was opened for, reached through the ancestor instead of the leaf.

### 2. Approval verification is skipped — #1792

`engine/crates/rocky-cli/src/commands/product.rs:1561` (`open_store_if_present`) returns `None` for a broken store ancestor. Compile at `:1033` then concludes there is **no approval** and never verifies the snapshot at `:1046`.

An approval gate that cannot read its store does not refuse — it decides there was nothing to check.

### 3. Collision verification returns "no collision" — `product verify`

`engine/crates/rocky-cli/src/commands/product.rs:539`, `:543`, `:558`, `:911`. Four separate folds: `read_dir` returning `EACCES` hits `return Ok(())`; a failing entry read is dropped by `.flatten()`; an unstatable peer directory fails `is_dir()`; a dangling manifest fails `is_file()`.

[...]

Issue #1502, opened 2026-08-20
product commit: restore trusts a .ff-prev found on disk instead of the journal's has_prev


hugocorreia90:
Split out of PR #1498 after four red-team rounds. This is the consumer-side half of the orphan-backup problem; the producer side deliberately stops short of fixing it.

## The gap

A failed backup write can leave a partial `<final>.ff-prev` on disk before any journal entry exists (`commit.rs`, the `copy_no_follow` path). Recovery then restores whatever backup it finds:

```
commit.rs:891 (approx)  restore renames <final>.ff-prev over <final>
                        WITHOUT consulting entry.has_prev
```

So a retry that records `has_prev = false` can still be undone by a stale orphan from an earlier failed attempt.

## Why the producer will not fix it

PR #1498 tried to, twice, and both attempts were rejected by review — correctly:

```
remove_file(dst)                unconditional: deletes a file the copy
                                never created if a racer swapped the name
inode-checked remove_file       narrows the window, cannot close it;
                                check-then-unlink is non-atomic
doing nothing                   no deletion window at all
```

`O_EXCL` grants exclusivity at open time and nothing after it, so any path-based unlink in an attacker-writable directory risks destroying a bystander's file. Between a stale backup and data loss, the stale backup is the recoverable failure — so the orphan stays and is handled here instead.

## Fix shape

Restore only when `entry.has_prev` is true, and refuse (or report) an unexpected backup found on disk when it is false. That makes the journal the authority, which is what it is for, and removes any need for the producer to delete under a race.

Related: #1500 (the pathname-based races that make the delete unsafe in the first place).

PR #1819, opened 2026-09-08
fix(engine/rocky-core): an unreadable commit marker refuses recovery, instead of rolling a committed generation back

### LINKED ISSUE #1813: An unreadable commit manifest reads as uncommitted, and recovery then restores every backup and deletes the journal
Found by an independent red-team pass over the path-probe fix class (#1739, #1757, #1758, #1764, #1787, #1789, #1791, #1792). This is the most serious thing that pass found, and it is **destructive** rather than diagnostic.

> ⚠️ **Reported from an incomplete review.** The pass ran 86 tool calls and was cut off by a network error before it wrote its final report, so this is its confirmed interim finding rather than a fully written-up one. Its own second reviewer confirmed it independently before the cut-off. The file:line trace still needs to be re-derived by whoever picks this up — treat the mechanism as the claim and the exact lines as unverified.

## The mechanism

A commit manifest that is **readable as metadata but unreadable as content** — permission denied, an I/O error, a dangling symlink whose parent resolves — is folded into `committed = false`.

That is the same absent-versus-unreadable conflation the whole #1739→#1792 class exists to close, but here it does not merely mislabel a status. `committed = false` is a **decision input to recovery**: it can restore every `.ff-prev` backup and then delete the recovery journal.

```
  manifest unreadable
        |
        v  folded into
  committed = false
        |
        v  recovery reads it as "the commit never happened"
  restore every .ff-prev backup   <-- overwrites committed state
        |
        v
  delete the recovery journal     <-- removes the evidence
```

A commit that **did** happen is rolled back, and the record that would have shown it is removed in the same pass.

## Why it is worth its own issue

The other members of this class produce a wrong *report*. This one produces a wrong *mutation*, and an irreversible one. It belongs with #1228 (CAS bypass) rather than with the status-labelling fixes it shipped beside.

## What is owed

1. Re-derive the exact call chain from the manifest read to the restore-and-delete sink, and pin the file:line.
2. Fail closed: an unreadable manifest is not "uncommitted". Recovery must refuse rather than roll back on a reading it could not take.
3. A test that exhibits the condition — a manifest whose metadata reads and whose content does not — and asserts the backups are **still there** afterwards. A test that only checks the reported status would pass without the fix.

Related: #1787, #1789, #1791, #1792 (same class, non-destructive), #1769.

hugocorreia90:
## What

`recover_generation` refuses when the commit marker exists but cannot be read, instead of treating it as uncommitted and rolling the generation back.

## The defect

```rust
let committed = manifest_final.is_file()
    && std::fs::read(&manifest_final)
        .map(|bytes| content_digest(&bytes) == manifest_entry.staged_sha)
        .unwrap_or(false);
```

`unwrap_or(false)` folds a manifest the process **cannot open** — EACCES, EIO — into "uncommitted". The mutation loop then acts on it:

```
  manifest unreadable
        |
        v  folded into
  committed = false
        |
        v
  rename every .ff-prev over the committed generation
  remove the residue
  delete the recovery journal
  report RolledBack
```

A commit that **did** happen is rolled back, and the one record that would have shown it is deleted in the same pass. This is the one member of the absent-vs-unreadable class (#1739, #1757, #1758, #1787, #1789, #1792) that reaches a filesystem **mutation** rather than a wrong status — which is why it has its own issue.

## The fix

A reading has three answers, not two, and the probe now matches on all of them:

| read result | verdict |
|---|---|
| succeeds, digest matches | committed — roll forward |
| succeeds, digest differs | not yet — the previous generation stands |
| nothing at the path | not yet — the rename never ran |
| **any other error** | **refuse.** The commit's fate is unknown; nothing is touched; the journal stays. |

The `NotFound` arm goes through `path_presence::classify_not_found` rather than trusting the error kind. Containment already refused a symlink or directory at this path, but the window between that check and this read is where a swapped-in link would sit, and the shared discriminator is the proven derivation for it.

The per-entry probe further down had the same fold in the non-destructive direction: an unreadable brand-new final read as "not ours", was left in place, and the journal was then deleted. It now refuses too — walking away from a file it never identified, with the only record of what it was supposed to be, is the same mistake with a smaller blast radius.

## Evidence

The tests build the exact condition the review named: a manifest that **is a regular file** (containment passes, `is_file()` is true) whose read returns EACCES, beside a real committed contract and a real `.ff-prev`. They assert the contract still holds the **committed** bytes, the backup is untouched, and the journal survives.

Both mutation-checked with `scripts/mutation-check.sh`:

|---|---|
| restore `is_file() && read(..).unwrap_or(false)` for the marker | `an_unreadable_commit_marker_refuses_recovery_and_restores_nothing` |
| restore `Err(_) => false` in the per-entry probe | `an_unreadable_final_with_no_backup_refuses_and_keeps_the_journal` |

Both skip under root (which reads through mode 000), following the convention in `schedule/spool.rs`. The 64 existing recovery tests — crash-between-renames rollback, idempotency, the #1502 orphan-backup refusals, the symlink and traversal refusals — are unchanged and pass.

## What this does not close

The **ancestor** half of the class (#1817): a dangling or unreadable directory *above* the manifest still makes the read return `NotFound`, which `classify_not_found` reports as absent because it stats the leaf. That is a separate, shared fix in `path_presence.rs`; this PR closes the leaf-level destructive path only.

Closes #1813

### DIFF
--- a/engine/crates/rocky-core/src/product/commit.rs
+++ b/engine/crates/rocky-core/src/product/commit.rs
@@ -1331,10 +1331,57 @@ pub fn recover_generation(project_root: &Path, parsed: &ParsedSpec) -> SpecResul
         .map(|(_, path)| path.clone())
         .expect("the manifest entry was proven present above");
 
-    let committed = manifest_final.is_file()
-        && std::fs::read(&manifest_final)
-            .map(|bytes| content_digest(&bytes) == manifest_entry.staged_sha)
-            .unwrap_or(false);
+    // Did the commit marker land? Only a READING of the final can say, and a
+    // reading has three answers, not two:
+    //
+    //   read succeeds, digest == staged_sha   committed   (roll forward)
+    //   read succeeds, digest differs         not yet     (the previous generation
+    //                                                       is still in place)
+    //   nothing is at the path                not yet     (the rename never ran)
+    //   ANY OTHER ERROR                       refuse      (#1813)
+    //
+    // The fourth row used to fold into the second: `read(..).unwrap_or(false)`
+    // turned a manifest the process could not open — EACCES, EIO — into
+    // "uncommitted", and the loop below then restored every `.ff-prev` over a
+    // generation that HAD committed and deleted the journal that would have
+    // shown it. An unreadable marker is not an absent one; the honest answer
+    // is that the commit's fate is unknown, and recovery mutates nothing on an
+    // unknown. Containment above already refused a symlink or a directory at
+    // this path, so a `NotFound` here is a bare leaf — but it is still put
+    // through the shared discriminator rather than trusted, because the
+    // window between that check and this read is exactly where a swapped-in
+    // link would sit.
+    let committed = match std::fs::read(&manifest_final) {
+        Ok(bytes) => content_digest(&bytes) == manifest_entry.staged_sha,
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
+            match crate::path_presence::classify_not_found(&manifest_final) {
+                crate::path_presence::PathPresence::Absent => false,
+                crate::path_presence::PathPresence::Present { detail } => {
+                    return Err(SpecRejected::new(
+                        "commit-io",
+                        format!(
+                            "reading {} failed: the commit marker cannot be read — {detail}. \
+                             Recovery cannot tell whether this generation committed, so it \
+                             changes nothing. Fix the path and re-run.",
+                            manifest_final.display()
+                        ),
+                    ));
+                }
+            }
+        }
+        Err(err) => {
+            return Err(SpecRejected::new(
+                "commit-io",
+                format!(
+                    "reading {} failed: {err}. The commit marker exists but cannot be read, \
+                     so recovery cannot tell whether this generation committed and changes \
+                     nothing — a rollback here would restore every backup over a commit that \
+                     may have succeeded. Fix the permissions and re-run.",
+                    manifest_final.display()
+                ),
+            ));
+        }
+    };
 
     for (entry, final_path) in &resolved_finals {
         let staged = staged_sibling(final_path);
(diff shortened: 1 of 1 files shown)


hugocorreia90:
## Independent red-team disposition

**Reviewer: Codex (GPT-5), 55 tool calls, read-only against the local branch (GitHub was unreachable from its sandbox, so it worked from the commit rather than this description).** Its report is kept with the programme's review record.

**Verdict: PASS.** *"The fix closes the leaf-unreadable rollback defect. No in-scope blocker found."* All seven attack vectors it was asked to run came back either clean or out of scope. It also ran both new tests itself (Darwin, non-root) and independently re-derived from the old code that they are fix-sensitive: under `origin/main` the backup **was** renamed over the committed contract and the journal **was** deleted.

### Its one finding — out of scope, and the PR body already says so

A dangling ancestor swapped in **after** containment validates the path still reads as `NotFound` → `Absent` → rollback. It rated this a #1817 residual, not a #1813 defect, and confirmed the "what this does not close" paragraph above is accurate. Agreed; #1817 is next, and its [design note](#) lifts the ancestor walk `contained_target` already has into `path_presence` so this fix inherits it.

Its "least confident" is worth carrying forward to #1817: whether every platform maps a parent that lost its search bit to `PermissionDenied` rather than `NotFound`. The Rust logic refuses the former and would misclassify the latter — a network filesystem that masks it is in the same gap.

### The test it asked for — added before merge

*"Executable evidence for the two-entry mid-loop replay case."* It traced by hand that a refusal firing after an earlier entry was already restored replays safely; it asked for the fixture, because both existing tests put the offending entry first and so refused before anything had been mutated.

`a_mid_loop_refusal_replays_without_touching_the_restored_entry` now pins the whole story: run 1 restores the contract and refuses on the sidecar; run 2 refuses again and touches nothing; run 3, once the operator fixes the file, completes the rollback and retires the journal. Producer journals carry one artifact plus the manifest, so this needs a forged-but-permissible journal — which is exactly the shape the reviewer said it would need.

### The remaining `exists()` / `is_file()` probes it listed — routed to #1817

Thirteen sites in `commit.rs`, `product.rs` and `manifest.rs` still fold a metadata error to `false`. The reviewer classified them correctly: *"metadata/ancestor-error surfaces, not the leaf-read failure corrected by #1813."* Most are residue-cleanup probes whose fold direction is "skip a cleanup"; one — `commit.rs:1405`, `entry.has_prev && prev.exists()` — can skip a restore and then retire the journal, which is the same journal-loss shape as the per-entry probe fixed here. Appended to #1817 as scope rather than widening this PR, so the destructive leaf fix lands on its own.

### Confirmed clean

- Refusal is resumable: it returns before the mutation loop, journal deletion is unreachable, and the next phase re-enters recovery.
- The per-entry refusal cannot fire on the roll-forward path — `committed == true` exits every iteration through `continue`.
- `commit-io` is the right code: no consumer (CLI, SDK, Dagster, fulfill loop, UI) branches on it; all preserve the message.
- Successful reads keep exactly the old semantics; the fix changes nothing about a digest collision or an externally written manifest.
- **ModelIr / contract semantics preserved: yes.**

Issue #1336, opened 2026-08-01
bug(engine): Path::exists() collapses a permission error into "models directory absent"


hugocorreia90:
## Description

`locate_models_dir` decides whether a pipeline's models directory exists with `Path::exists()`:

```rust
// engine/crates/rocky-cli/src/models_loader.rs
let models_dir = project_root.join(models_base(models_glob, project_root));
if !models_dir.exists() {
    return Ok(ModelsDir::Absent(models_dir));
}
```

`Path::exists()` returns `false` for **any** metadata error, not just "not there" — `PermissionDenied` included. So a models directory that exists but cannot be traversed is reported as absent.

## Why that is worse than it sounds

The `Absent` arm is a *skip*, not an error. In `run_dag_exec.rs` the loader `continue`s past that pipeline, so its models never enter `by_pipeline` and its root never enters `contributing_roots`. Downstream, an empty `contributing_roots` is treated as **"this project has no transformation models"**, which is a legitimate, complete answer.

Concretely: `models = "locked/models/**"` where `locked/` exists and denies traversal gives

- a DAG with no transformation nodes, reported as success
- `rocky dag --column-lineage` reporting *authoritative* emptiness — after #1320, explicitly so, since `column_lineage_unavailable` is left unset for the no-models case

The #1320 work distinguishes "no lineage" from "lineage unavailable". This is a third state — *"we could not tell whether there are models"* — that currently reports as the first.

## Suggested fix

Use `Path::try_exists()` and propagate the error:

```rust
if !models_dir.try_exists().with_context(|| {
    format!("could not determine whether models directory '{}' exists", models_dir.display())
})? {
    return Ok(ModelsDir::Absent(models_dir));
}
```

## Why this is filed separately rather than fixed in #1335

`locate_models_dir` is shared: `run_dag_exec`, `branch`, `list`, `compliance`, `estimate`, `optimize`, `test`, `scope` and the API all reach it. Turning a silent skip into a hard error changes behaviour for every one of them, and at least some deliberately tolerate a missing directory.

That is exactly the shape that has bitten this repo before — making a shared function stricter without checking every tolerant caller (#1266, where emptying a set un-gated a destructive apply). It needs a caller sweep, and that sweep is not what #1320 is about.

## Acceptance

- A models directory that exists but denies traversal produces an error naming it, not a silent skip.
- Every current caller's handling of the new error path is checked, not assumed — in particular any that treat "no models" as a normal outcome.
[...]

PR #1822, opened 2026-09-08
fix(engine): a dangling ancestor is a broken path, not an absent one — walk up before calling anything absent

### LINKED ISSUE #1817: The path-probe class is closed at the leaf and open one directory up — a dangling ancestor still reads as absence

hugocorreia90:
## What

`path_presence::classify_not_found` walks the ancestors before calling anything absent, `entry_is_present` becomes "anything but a proven absence", and the product collision scan stops clearing a product on a directory it could not list.

## The defect

The absent-vs-unreadable class — #1739, #1757, #1758, #1787, #1789, #1792, #1813 — was closed at the **leaf** and open one directory up. Every fix routed through the shared discriminator, and it stats only the path it is given:

```
  proj -> /gone                    (dangling symlink: the ANCESTOR)
     |
  proj/rocky.toml
     +-- read()                    -> NotFound
     +-- symlink_metadata(leaf)    -> NotFound     <-- classify_not_found: Absent   WRONG
```

`symlink_metadata` on the leaf is exactly as blind to a broken parent as `exists()` was. Four sinks reached real decisions through it:

| Sink | Decision on a dangling ancestor |
|---|---|
| retention sweep (`commands/state.rs`) | `FileNotFound` → **default retention** → irreversible deletion of history the real config said to keep. #1739's own data-loss case, one level up. |
| `open_store_if_present` (`product.rs`) | "no store" → compile concludes **no approval to verify** |
| sidecar loader (`models.rs`, `seeds.rs`, `project.rs`) | "no sidecar" → compiles against `_defaults.toml` |
| `check_product_collisions` | `read_dir` fails → **"no collision"** → `rocky product verify` PASS |

## The fix is in the helper, so every caller inherits it

The walk is the one `product::commit::contained_target` already does to refuse a dangling ancestor before a write — lifted into `path_presence`, not reinvented. After the leaf is proven `NotFound`:

| ancestor `symlink_metadata` | verdict |
|---|---|
| `Err(NotFound)` | keep walking — the never-created chain |
| `Err(other)` | **Present**: could not be inspected |
| `Ok`, symlink, `canonicalize` fails | **Present**: dangling ancestor, named in the detail |
| `Ok`, symlink, resolves | Absent — a live link (`models -> models_real`), nothing below it |
| `Ok`, directory | Absent — nothing below it was created |
| `Ok`, anything else | **Present**: a file where a directory belongs |

The promise callers depend on — a **missing** parent is honest absence — is kept: the walk passes through a `NotFound` ancestor. A **dangling** parent is not missing; an entry is there, and the leaf's `NotFound` came from following it into nothing. Ordinary absent paths pay one extra stat on the parent.

`entry_is_present` had its own leaf-level fold — `.is_ok()` turned an unstatable leaf into `false`. Now three-way. Each of its eight callers was checked: every one propagates the read error that follows a `true`, which is what makes `true` the safe direction.

## The collision scan — a different shape, same PR

Four folds that were not the helper's: `let Ok(entries) = read_dir(..) else { return Ok(()) }`, a `.flatten()` dropping an erroring entry, `is_dir()` dropping an unstatable peer, `is_file()` dropping an unreadable manifest. Each is now proven-absence-or-refuse. New code `fulfillment-dir-unreadable` for the directory; the existing `manifest-unreadable` for a peer's manifest.

## Evidence

**This is the fixture both reviewers of the 2026-09-08 pass asked for** — lanes C and D independently named "an EACCES/dangling-ancestor suite" as the missing evidence. Built at every layer:

| Test | Old answer | Now |
|---|---|---|
| `path_presence`: dangling ancestor, dangling grandparent, unsearchable ancestor (root-skip) | Absent | Present, naming the ancestor |
| `path_presence`: never-created chain; live symlinked ancestor with absent leaf | Absent | Absent — the controls |
| `state_retention_sweep` with `proj -> gone` | sweeps on defaults, drops 5 | refuses; `run_count` stays 105 |
| `product_status_in` with `state.redb` under a dangling ancestor | "never run" | refuses, names the ancestor |
| `check_product_collisions`: unreadable fulfillment dir (root-skip), dangling fulfillment dir, peer with a dangling manifest | PASS | refuses |

Mutation-checked with `scripts/mutation-check.sh`, each verified to fail when the thing it guards is reverted:

| Mutation | Tests that catch it |
|---|---|
| leaf-only `classify_not_found` (no walk) | the dangling-ancestor tests at all three layers |
| `entry_is_present` back to `.is_ok()` | `a_leaf_under_a_dangling_ancestor_is_present` |
| `read_dir` failure → `Ok(())` | both fulfillment-dir tests |
| unreadable peer manifest → `continue` | `a_peer_with_a_dangling_manifest_refuses_the_collision_check` |

## What this deliberately leaves, and why

- `commands/state.rs` `!state_path.exists()` on the sweep's **store** path: the helper doc exempts the `state.redb` gates, and this one emits an empty report and deletes nothing. Non-destructive direction.
- The thirteen residue probes the #1819 review listed in `commit.rs` — `.ff-prev`/`.ff-staged` cleanups, `has_prev` at commit time, and `commit.rs:1405`'s restore authority. Same class, per-site changes; this PR is the shared derivation. `commit.rs:1405` can skip a restore and then retire the journal, so it needs the same care #1813 got. Tracked on #1817 as its second destructive member.

Closes #1817

---

## What changed after the description above — nine review rounds

The description above is the first commit. Eight more landed under independent review, each dispositioned in the comments. The important ones:

- **The class had more members than the helper reached.** Six model-loader probes (`_defaults.toml`, contracts, the directory walks, the models-dir locator) and the LSP's project gate sat *in front of* the shared helper and decided first. All route through it now; one `sibling_contract_path` replaces three copies of the contract probe.
- **A trailing separator defeats a leaf stat.** `symlink_metadata("models/")` follows the link it was meant to inspect. Both helpers normalise before the stat.
- **A bare relative `rocky.toml` walked off the end** — `Path::parent` gives `Some("")`. The empty parent is the working directory and is probed as one; a child-process test owns its cwd to prove it.
- **The first live ancestor must be searchable, not just present.** `dir/.` is the portable probe; Windows normalises it away and that residual is documented.
- **Every link to nowhere in a models tree is refused, whatever it is called.** A name rule was tried and failed both ways (`v1.2` dropped, `.gitkeep` refused); a dangling `README.md` refusing a compile is the accepted cost, stated in the code.
- **The filtered compile synthesised `NoModels` for an unreadable subtree** — and the policy plane reads `NoModels` as a *proven* empty set, so a layer-scoped `deny` stopped matching. Pre-existing, missed by five rounds, closed in round seven with its own `ProjectError::ModelsTree` so it cannot recur by type.
- **The LSP tells the editor** when its startup compile fails; a protocol-level test reads the client socket for `window/showMessage`.

Twenty-three mutation checks; the final round PASSED with a second model concurring. Follow-ups on #1817: a filesystem fault-injection seam, and explicit configured model roots so an entry can be *proven* outside the model namespace rather than guessed.


hugocorreia90:
## Independent red-team disposition — round one

**Reviewer: Codex (GPT-5), 52 tool calls, read-only against the local branch at `90d172f7`.** Report kept with the programme's review record.

**Verdict: NOT FULLY CLOSED** — two soundness gaps, one test gap, one low regression. All four accepted and addressed in `3759c732`, which is now under its own review. No finding was rebutted.

### 1. HIGH — an unsearchable ancestor could still read as absence — **ACCEPTED, fixed**

> Statting the directory entry `/p/locked` succeeds because that requires searching `/p`, not searching inside `locked`.

Exactly right, and it exposed that my "the walk's EACCES arm covers a masking filesystem" comment described a mechanism that did not exist: the walk never saw an error, it saw a healthy-looking directory. `searchable_or_present` now probes `dir/.` on the first live ancestor — that stat needs search permission on `dir` itself, and fails however the platform spells it. The reviewer also said the test "does not prove the comment's claim"; the new test calls the walk **directly** with a leaf under a mode-000 directory, bypassing the leaf stat, so it exercises the arm the masking case would reach. Reverting the probe fails it.

Stated rather than claimed closed: Windows normalises `dir\.` back to `dir` before the check, so the masked case stays open there. It is in the doc comment.

### 2. HIGH — model-loader probes the helper never reached — **ACCEPTED, fixed; the closure claim was false**

> These are not regressions introduced by this commit, but they disprove the broad "leaf class closed / all caller flows inherit" rationale.

Correct. Six probes sat *in front of* the helper and decided first:

| Site | What a dangling link did |
|---|---|
| `models.rs` `_defaults.toml` `exists()` | dropped the whole directory's strategy → every model fell to `FullRefresh` |
| `models.rs` contract `exists()` | compiled the model uncontracted |
| `models.rs` ×4 `!dir.exists()` → empty | a dangling models dir loaded **nothing** and reported success |
| `project.rs` ×3 | the same three shapes on the Salsa path |
| `model_walk.rs` `!dir.is_dir()` | a dangling root or subdir walked past with no error |
| `models_loader.rs` `try_exists()` → `Ok(false)` | the #1336 row that said "a broken symlink stays Absent" |

Every one now goes through the shared helper. The #1336 test that pinned "a dangling symlink resolves to NotFound, which is an absence" is rewritten to assert the opposite, with the reason — that row was a measurement of `try_exists`'s blast radius, and this class later established that the measurement's answer is not the right decision.

Found while testing it, and worth its own line: `models_base` yields `models/`, and a **trailing separator** makes the OS resolve the last component as a directory — `symlink_metadata("models/")` *follows* the link it was meant to inspect. Both helpers now strip it via `components()` before the stat. Without that, the loader fix was invisible.

### 3. MEDIUM — two collision folds not mutation-bound — **ACCEPTED**

> Add a dangling peer state-directory test … That kills the `is_dir()` reversion.

Added; it is the one mutation the other tests survived, and it now fails under it. A `DirEntry` error remains untestable without an injected filesystem seam — agreed, and said plainly rather than glossed.

The reviewer also noted `an_unsearchable_ancestor_is_present` passes against the old helper. True; its doc comment already said so, and it stays as a pin of the Linux/macOS mapping. The direct-walk test above is the fix-sensitive one.

### 4. LOW — a vanished unrelated entry now fails `product verify` — **ACCEPTED, fixed**

A `.DS_Store` unlinked between the listing and the stat is a *proven* absence and conceals nothing. `NotFound` from `metadata` now goes through `classify_not_found`: `Absent` → skip, `Present` (a dangling link) → refuse.

### Corrections the review made to my own claims

- "All eight `entry_is_present` callers propagate the read error" — false for `init.rs`, which refuses outright. Safe direction, wrong sentence. Fixed in the doc comment.
- The walk's comment about covering masked errors — see finding 1.

### Confirmed clean by the review

- The stop rule is right for every stable dangling-link chain: `symlink_metadata` does not follow its final component, so the prefix ending at a dangling component is always stat-able.
- No `classify_not_found` caller changes a decision in a less-safe direction; fresh clones and ephemeral runners stay healthy; the scheduler pays two extra stats per tick only when the spool is absent.
- Collision containment is preserved — `metadata` follows a symlinked peer exactly as `is_dir()` did.
- No consumer enumerates `SpecRejected` codes, so `fulfillment-dir-unreadable` is handled generically everywhere.

**ModelIr / contract semantics preserved: yes.**

Round two — seven more mutation checks, each verified to fail when its guard is reverted — is under an independent review of its own before merge.

hugocorreia90:
## Independent red-team disposition — round two

**Reviewer: Codex (GPT-5), 59 tool calls on the round-two delta (`90d172f7..3759c732`).** Its session stalled before writing the report — the rollout sat idle 28 minutes on a zero credit balance — and was recovered with `codex exec resume` from its banked context. The recovered verdict is complete and file:line-precise; it is kept with the programme's review record.

**Verdict: NOT FULLY CLOSED** — two soundness gaps, two error-propagation defects, one test gap. Four accepted and fixed in the third commit; one routed to its own issue. Nothing rebutted.

### 1. HIGH — a bare relative leaf bypasses the probe — **ACCEPTED, fixed**

> `Path::parent("rocky.toml")` is `Some("")`. `classify_ancestors` stats `""`, receives `NotFound`, then reaches no parent. It returns `Absent` without probing `"."`.

Exactly right, and it is the CLI's *default* spelling of the config path. The empty parent is now named `.` before the stat, so the working directory gets the same `dir/.` searchability probe as any other first live ancestor. Honest gap: this has no fix-sensitive test — exercising it needs a mode-000 working directory, and a test cannot chmod its own cwd under parallel execution. The added test pins the reachable half (the walk lands on `.` and answers Absent for a leaf that is not there, instead of tripping over the empty parent).

### 2. HIGH — inline-frontmatter models still lose a dangling contract — **ACCEPTED, fixed structurally**

> `parse_model_inline_with_context` still uses `contract_file.exists()`.

A third copy of the probe my sweep missed. Rather than patch the third copy, there is now **one** — `models::sibling_contract_path` — called by the pair loader, the inline loader and the compiler's `.rocky` loader. That also answers finding 5's "the `.rocky` route is unpinned": one mutation on the helper is caught by both the pair test and the new inline test, and there is no separate route left to pin.

### 3. — LSP and serve — **LSP fixed; serve routed to #1823**

`initialize` gated its project on `models_path.exists()`, so a dangling `models` link read as "no models project" and the server sat silent. Now anything but a proven absence is a project, and the compile that follows reports the broken link. No test: `initialize` needs a tower-lsp harness the crate does not have.

`rocky serve` is a different defect: its background compile fails, the failure is **logged**, and the project route then reports zero diagnostics and `has_errors: false`. That is the did-not-run-as-passed class through serve's async seam, independent of any path probe — filed as **#1823** rather than widened into this PR.

### 4. — the wrong object blamed — **ACCEPTED, fixed**

With `models` a dangling link the walker recorded the honest root error, but the per-directory load ran first and a strict caller saw "failed to read `models/_defaults.toml`" — a leaf under the broken root. Walk errors now come first; `a_dangling_models_root_is_blamed_before_anything_under_it` pins that the first error names the root and not the leaf.

### 5. — test gaps — **ACCEPTED**

- `.rocky` contract route: closed by the shared helper (above).
- `entry_is_present`'s trailing-separator normalisation was redundant — its `NotFound` arm falls through to `classify_not_found`, which normalises — so the test that seemed to pin it was not sensitive to it. The redundant copy is removed and the reason is in the code, so nobody reads a guard into it that is not there.

### Confirmed clean by the review

- `dir/.` is a search-permission probe per POSIX pathname resolution and the Linux path-lookup doc; the reviewer's own read-only Darwin probe found no false refusal of a healthy directory, a healthy symlinked directory, or `/`.
- `components().as_path()` strips only trailing separators and non-leading `.`; it preserves `..`, a leading `./`, root, empty paths and platform prefixes; no caller passes a deliberate `/.` or `/..`.
- Refusing a dangling models link is the right outcome; a genuinely absent root stays `Absent`, and an uninitialised submodule is a gitlink, not a symlink.
- The vanished-entry passthrough is correct; no hot-path regression — the new metadata call replaces the one `is_dir()` made.

Round three — two more mutation checks, each verified to fail under its reversion — carries the disposition above. The reviewer's "least confident" (non-conforming FUSE filesystems) and "most important missing" (a deterministic filesystem fault-injection seam) both stand; the seam is the single highest-value follow-up left in this class and is recorded on #1817.

hugocorreia90:
## Independent red-team disposition — round three

**Reviewer: Codex (GPT-5), 40 tool calls on the round-three delta (`3759c732..50db7c3c`); its own second reviewer `gpt-5.6-terra` confirmed two of three.** The session stalled again and was recovered with `codex exec resume`; the recovered verdict is complete. Kept with the review record.

**Verdict: NOT FULLY CLOSED** — three findings. All three accepted and fixed in the fourth commit. Nothing rebutted.

### 1. — the LSP moved the silence instead of removing it — **ACCEPTED, fixed; my comment was false**

> `initialize` now stores the dangling path, but the editor still receives no diagnostic. The changed comment's claim that "the compile that follows reports the broken link" is false.

Correct. The compile failure reached `info!(error = %e, "LSP compilation failed")` and nothing else; no diagnostic is built when the pipeline aborts, so the editor stayed empty and clean-looking. I wrote a comment naming a mechanism the code did not have — the exact defect this whole pass keeps finding in others' comments, now in mine. The failure arm sends `show_message(ERROR, …)` to the client, and the `initialize` comment now says precisely what happens and where. No test: a client-observed notification needs a tower-lsp harness the crate does not have. Stated.

### 2. — every walk error first was too much — **ACCEPTED, fixed**

> The strict loader reports only the distant depth-ceiling error. It hides the immediately actionable root parse error.

Right, and a regression I introduced in round three. Only the **root's** walk error goes first now — the dangling-root case that motivated the change; everything else keeps traversal order. Pinned in both directions: the dangling-root test from round two, and a new one where a root parse error precedes a dangling subdirectory.

### 3. — the bare-relative test was not fix-sensitive — **ACCEPTED, built the test the reviewer designed**

> Spawn the test binary as a child … change into its own temporary directory, remove search permission, call the private `classify_ancestors(Path::new("rocky.toml"))` … assert `Present`.

Done as specified. The child owns its working directory, drops the search bit, asks the walk about a bare `rocky.toml`, and prints its verdict; the parent asserts on it and accepts a skip under root. Reverting the empty-parent mapping fails it.

### Found on the way

The walk's subdirectory **listing** had its own `is_dir()` fold — a dangling subdirectory link was never pushed onto the stack, so round two's root check never saw it and `models/staging -> gone` was walked past in silence. Fixed and pinned.

### Confirmed clean by the review

- The `.` mapping is right for the CLI default (`main.rs:83` is process-cwd-relative); `C:rocky.toml` has parent `C:`, not `""`, so the mapping does not touch it; the Windows residual stays documented.
- `sibling_contract_path` preserves the exact naming every loader used before (`with_extension("contract.toml")`), including `fct.orders.sql → fct.orders.contract.toml`.
- No partial caller branches on error order.

Round four carries this disposition. The reviewer's two questions — Windows drive-relative behaviour, and a client-observed LSP test — both stand as gaps; the second is the one worth a harness.

hugocorreia90:
## Independent red-team disposition — round four

**Reviewer: Codex (GPT-5), 36 tool calls on the round-four delta (`50db7c3c..0f281123`).** Kept with the review record.

**Verdict: NOT FULLY CLOSED** — two regressions I introduced in round four, two test defects. All four accepted and fixed in the fifth commit. Nothing rebutted.

### 1. MEDIUM — a dangling non-model file now aborted the whole compile — **ACCEPTED, fixed; a regression of mine**

> A valid project also contains `models/README.md -> /missing`. … Compilation now fails because of a non-model file. The loader explicitly promises that filtering occurs before reading.

Right. My round-four listing change escalated *every* dangling link. A dangling link's target type is unknowable, so the rule is now what it **would** have been: an entry with a file's name (an extension) is an unrelated file and is ignored; one with a directory's name (`staging`, `marts`, `.hidden`) is a subtree the walk would have entered and is reported. Dangling model files (`x.sql`, `x.toml`, a contract) are caught by the loaders' own probes, not here. Pinned at both the walk and the compiler: `README.md -> gone` compiles clean, `staging -> gone` still refuses.

### 2. MEDIUM — the ordering fix never reached `rocky compile` or the LSP — **ACCEPTED, fixed**

> `compile()` routes through `Project::load_models_with_db`, not `load_project_models_partial_with`.

Correct — the compiler's own loader aborted on the first walk error before parsing a single model. It now applies the same precedence: the root's walk error first, then every model that loads (so a parse error has its turn), then deeper walk errors — deferred, not dropped. Pinned through `Project::load_models` in both directions, plus the dangling-root case.

### 3. MEDIUM — the child-process test skipped on the Ubuntu runner — **ACCEPTED, fixed**

> `symlink_metadata(".")` stats the target directory itself. Mode `000` on that target does not deny the stat because no inaccessible directory occurs in the path prefix.

Exactly right, and a sharper point than it looks: the gate reproduced on macOS (where I ran it) and would have skipped on Linux (where CI runs), so the test was green-by-skip precisely where it mattered. The gate is now the production probe itself — `./.` puts the working directory in the path prefix — and reproduces on every Unix except under root. Re-mutation-checked.

### 4. LOW — `show_message` had no test — **ACCEPTED, built the protocol-level test**

> Add a protocol-level `LspService` test that initializes against a failing models directory and asserts the emitted `window/showMessage` notification.

Done as specified: a real `LspService`, a real `initialize` + `initialized` exchange against a dangling `models` link, and the client socket read for `window/showMessage` with `type: ERROR`. It **has** to be protocol-level — tower-lsp drops every client notification sent before the service is initialized, so calling `recompile` on a bare struct observes nothing whether or not the message is sent. Reverting the `show_message` call fails it.

### Confirmed clean by the review

- No popup on every keystroke: `recompile` runs only from `initialized` and `did_save`; `did_change` is debounced and discards compile errors.
- The child harness works under nextest; the child env var is set nowhere else; libtest accepts the module-qualified `--exact … --nocapture` form.
- Live and outside symlinks follow exactly as `is_dir()` did — no containment change.
- Root path equality holds: the walk reports the root spelled exactly as passed.

Round five carries this disposition, with four more fix-sensitive mutation checks. The reviewer's "least confident" — whether every dangling non-model entry should invalidate a tree — is now answered and documented in `looks_like_a_directory_name`: it should not, and the name is the tie-breaker the tree's own convention provides.

hugocorreia90:
## Independent red-team disposition — round five

**Reviewer: Codex (GPT-5), 41 tool calls on the round-five delta (`0f281123..d217274c`).** Kept with the review record.

**Verdict: DO NOT SHIP** — one high, two medium, one low. All four accepted and fixed in the sixth commit. Nothing rebutted — and this round reverses a call from round four, on the record below.

### 1. HIGH — dotted dangling directories silently discarded — **ACCEPTED; the heuristic is gone**

> `v1.2 -> gone`, `staging.old -> gone`, or `.archive.v1 -> gone` … the name contains a dot, so the heuristic returns false and the entry is ignored. … a dangling symlink's former target type cannot be inferred from its name.

Right, and finding 3 is the same rule failing the other way (`.gitkeep -> gone`, part of the documented scaffold, read as a directory and failing the compile). Round four asked me to spare `README.md -> gone`; round five shows a name rule cannot do that honestly. **Every link to nowhere in a models tree is now reported, whatever it is called**, with a message that names the entry and says to repair or remove it. The cost — a dangling README refuses a compile it took no part in — is accepted and written in the code, because the exception could not be made without dropping a dotted directory in silence.

**This reverses round four's finding 1.** Both reviewers read the same trade-off from opposite ends; the fail-closed end is the one this branch stands on, and it is the class rule applied without exception.

### 2. MEDIUM — `NoModels` bypassed the deferred walk error — **ACCEPTED, fixed**

A tree whose only entry was `staging -> gone` reported "no models" over an unreadable subtree — and breached `ModelWalkError`'s own contract that strict consumers fail on the first one. The deferred check now runs before the no-models answer; pinned by `compile_path_reports_a_dangling_subdir_ahead_of_no_models`, which — as the review also asked — fails under **full** reversion of the compiler's ordering change, where the dangling-root test did not.

### 3. MEDIUM — extensionless dangling files failed the compile — **ACCEPTED, subsumed by 1**

`LICENSE`, `Makefile`, `.gitkeep`: the same rule from the other side. Gone with it. The new test walks five names — two dotted directories, two extensionless files, one plain file — and expects one refusal for each, at both the walk and the compiler.

### 4. LOW — one compiler test survived full reversion — **ACCEPTED**

`compile_path_blames_a_dangling_root_before_anything_under_it` guards the root-first half, which the old code also had. It stays as that pin; the new `ahead_of_no_models` test is the one bound to the whole change.

### Also from the review

- The LSP test's comment claimed tower-lsp drops every pre-initialisation notification; `show_message` uses the unchecked path. Corrected — the real reason the test is protocol-level is that it observes what an editor receives on the path an editor drives.
- `position()` + `remove(i)` correct; `walk_error_dir` duplication acceptable (both matches exhaustive, a new variant fails both); the `./.` gate sound on Linux and macOS per `path_resolution(7)` / `lstat(2)`; the LSP test valid and the socket confirmed server→client.

Round six carries this disposition, with two more fix-sensitive mutation checks.

hugocorreia90:
## Independent red-team disposition — round six

**Reviewer: Codex (GPT-5), 36 tool calls on the round-six delta (`d217274c..b5e032fd`).** Kept with the review record.

**Verdict: BLOCK** — one high, one low. Both accepted and fixed in the seventh commit. The product rule was **endorsed**.

### The rule — **ENDORSED**

> I endorse refusing every dangling link under the implicit models tree. A warning or "abort only when zero models loaded" permits successful partial compilation when an independent subtree disappeared.

Which settles the round-four/round-five disagreement on the record: fail-closed, no name rule. The reviewer's "safe third shape" — explicit configured model roots or an ignore list, so an entry is *proven* outside the model namespace rather than guessed from its name — is the right long-term design and is noted on #1817 as a follow-up, not attempted here.

### 1. HIGH — the filtered compile still synthesised `NoModels` — **ACCEPTED, fixed; a pre-existing fold every round missed**

> `has_matching_model_source` discards the error … it finds no accessible matching source, so `load_models_matching_with_db` returns `NoModels` before calling the changed strict loader. … `model_attributes` explicitly treats `NoModels` as a proven, healthy empty set.

Exactly right, and worse than the previous rounds' findings: this one reaches a **policy decision**. With `models/gold -> gone` as the only matching subtree, the pre-check dropped the walk error, the caller answered `NoModels` from its `false`, and the maintenance gate evaluated the target with default attributes — a `deny` scoped to `layer = "gold"` stopped matching, and compact/archive SQL could be authorised. The wrapper predates #1262's shared walk and every commit on this branch; six rounds of review of the strict loader never saw it because the pre-check returns before the strict loader runs.

The pre-check now propagates the first walk error when nothing matched (a match still wins). `a_filtered_compile_reports_a_dangling_subtree_instead_of_no_models` asserts the `ModelsTree` variant and never `NoModels`; restoring the discard fails it.

### 2. LOW — the error called the entry both a directory and a file — **ACCEPTED, fixed**

> The rendered chain calls the entry both a "models directory" and a "model file," although neither is known.

Fair: the code said the kind was unknowable and then picked `ReadDir` and let it wrap into `ModelError::ReadFile`. `ModelWalkError::UnresolvedEntry { path, detail }` names neither, and `ProjectError::ModelsTree` carries a walk failure up as a walk failure — which is also what makes the `NoModels` confusion impossible by type.

### Confirmed clean by the review

- The five-name and three-name tests exercise the real `rocky compile` path (`compile → compile_with_db → load_models_with_db → walk_model_dirs`) and cannot pass through any other arm.
- No downstream consumer folds the nested `NotFound` back into absence; the only fold was the pre-check above.
- Moving the deferred check ahead of `NoModels` did not re-break round four's ordering: a root parse error still returns via `?` first.
- All three changed tests are fix-sensitive; the LSP comment fix is behaviour-neutral.

Round seven carries this disposition. The reviewer's "most important thing missing" — that the filtered wrapper had an earlier `NoModels` return with stronger downstream meaning than a compile failure — is precisely what it fixes.

hugocorreia90:
## Independent red-team disposition — round seven

**Reviewer: Codex (GPT-5), 56 tool calls on the round-seven delta (`b5e032fd..e9e6ad0c`).** Kept with the review record.

**Verdict: BLOCK — on wording, with the soundness explicitly confirmed closed:**

> The `NoModels` policy fail-open is closed. … This does not reopen the policy fail-open. All variants still fail compilation. It does make the delta's explicit typed-error and rendering guarantee false.

One blocking finding, one test gap. Accepted and fixed in the eighth commit, with one narrow rebuttal on the record.

### Blocking — a dangling MODEL file rendered as "failed to read model file: No such file" — **ACCEPTED, fixed**

> `models/orders.sql -> gone` … strict loading collects and reads it … becomes `ProjectError::ModelLoad(ModelError::ReadFile)` before the deferred `ModelsTree`. … The neutral "cannot be resolved" and "repair or remove it" guidance is lost.

Right — the walk stores the honest error, but the loader's own read reaches the entry first, and that read said the one sentence this whole branch exists to stop saying. Three readers, one rule now: every model-tree file goes through `read_model_text`, which explains a `NotFound` on an entry the probe found present; the `.rocky` route stats before the salsa read and types a dangling one as `ModelsTree(UnresolvedEntry)`; the glob pre-check's directory listing is typed `ModelsTree(ReadDir)`, so a directory is called a directory. Three tests, one per route the review named.

**The narrow rebuttal.** The review asked for `ModelsTree` on the `.sql` route too. That would need the loader to consult the walk's stored errors before its own read, or to substitute after the fact — for a distinction only a consumer that *branches on the variant* could see, and the review's own consumer table found none besides `NoModels`. The wording — which is what an operator reads — is right on every route; the variant on the `.sql` route stays `ModelLoad`. Said here so it is a decision, not an omission.

### Test gap — **ACCEPTED**

The three new tests bind the dangling `.sql`, the dangling `.rocky`, and the unlistable-directory-before-a-healthy-match ordering. The `UnresolvedEntry` construction is bound through the walk's five-name test.

### Confirmed clean by the review

- The pre-check's order is right in both directions: a healthy match wins and the dangling subtree is reported deferred; no match plus a walk error is the walk error, and `NoModels` would be unjustified.
- **No `ProjectError` consumer folds the new variant permissively** — `model_attributes` softens only `NoModels`; `resolve_policy_and_attrs` turns any other failure into `Deny`; the maintenance gate, compact, archive, plan preview, emit-sql and the GC seam all propagate or deny. The two generic catch-alls produce intermediate empties that a later compile denies.
- No unintended `#[from]` conversions; `walk_model_dirs` returns a tuple, not a `Result`.
- `has_matching_model_source` has exactly one caller.

Round eight carries this disposition, with two more fix-sensitive mutation checks.

hugocorreia90:
## Independent red-team disposition — round eight

**Reviewer: Codex (GPT-5), 20 tool calls on the round-eight delta (`e9e6ad0c..a54ae641`); it ran the new tests (6 passed) rather than only tracing them.** Kept with the review record.

**Verdict: FAIL on one Low.** Everything else passed. Accepted and fixed in the ninth commit; the rebuttal from round seven was **upheld** by this reviewer.

### Low — a dangling ROOT lost its neutral wording — **ACCEPTED, fixed; a regression of mine**

> `entry_is_present()` returns true for the dangling link … then `read_dir(models_dir)` … immediately returns `ModelsTree(ReadDir { source: NotFound })`. That prevents surfacing the stored `UnresolvedEntry`, losing "symlink … cannot be resolved; repair or remove it."

Right, and the framing correction is accepted too: the old `exists()` let the pre-check fall through to the stored `UnresolvedEntry`; my `entry_is_present` swap made it `read_dir` the dangling root first. The pre-check now returns the **root's** walk error before any per-directory listing — the reviewer's minimal fix — while keeping "a matching healthy child wins over deeper errors". A filtered dangling-root test asserts `UnresolvedEntry` and the repair/remove wording; reverting the root-first return fails it.

### The round-seven rebuttal — **UPHELD**

> PASS — no operator-visible consumer distinction found for dangling `.sql` staying `ModelLoad(ReadPath)`. … exit behavior, MCP error code, and LSP message do not differ by `ModelLoad` versus `ModelsTree` on the traced paths.

### Nits — **ACCEPTED, fixed**

- The `.rocky` comment said the stat runs "only on the NotFound path"; the stat runs on every load and only the *return* is gated. Reworded.
- `read_model_text` was inserted between `load_dir_defaults`'s doc comment and its `fn`, so the doc attached to the wrong function. Moved above.

### Confirmed by the review

- SQL post-read behaviour (trim, allocation reuse) preserved byte-for-byte.
- The extra stat is one `metadata()` per `.rocky` per compile, reached on the LSP's debounced incremental path; existing `read_source` already canonicalises and reads every `.rocky` there. Unbenchmarked on networked filesystems — stated, not claimed.
- The `.sql` and `.rocky` wording tests are fix-sensitive; the mode-000 test skips honestly where the process can list a mode-000 directory.

Round nine carries this disposition. The reviewer's "most important thing missing" — the filtered dangling-root test and the invariant that a known unresolved root outranks a later failed `read_dir` — is exactly what it adds.

hugocorreia90:
## Independent red-team disposition — round nine: PASS

**Reviewer: Codex (GPT-5), 26 tool calls on the round-nine delta (`a54ae641..9ee8fbc2`); its second reviewer `gpt-5.6-terra` independently returned PASS.**

> PASS — No High, Medium, or Low findings. Nothing in `a54ae641..9ee8fbc2` blocks merge.

Confirmed: the root-path comparison is byte-reliable through every production caller (no normalisation boundary anywhere in the flow, trailing separator included); `remove(i)` cannot double-report; both doc corrections attach to the right functions; the new test breaks under reversion. Reverting the pre-check's `entry_is_present` to `exists()` is inert for every stable state now that root-first is in place, and matters only under a walk-then-swap race that predates this branch — recorded, not fixed.

## Where this PR ended up — nine rounds, on the record

|---|---|---|
| 1 | NOT FULLY CLOSED | unsearchable ancestor could still read as absence; six loader probes the helper never reached |
| 2 | NOT FULLY CLOSED | a bare relative `rocky.toml` walked off the end; inline models' contract probe; LSP gated on `exists()`; wrong object blamed |
| 3 | NOT FULLY CLOSED | the LSP fix moved the silence instead of removing it (my own comment claimed a mechanism that did not exist); walk-error priority too wide |
| 4 | NOT FULLY CLOSED | a dangling `README.md` aborted the compile; the compiler's own loader bypassed the ordering fix; the child test skipped on Linux; `show_message` untested |
| 5 | DO NOT SHIP | the name heuristic failed both ways (`v1.2` dropped, `.gitkeep` refused); `NoModels` outranked a walk error |
| 6 | BLOCK | **the filtered compile synthesised `NoModels`, which the policy plane reads as proven-empty** — a layer-scoped `deny` stopped matching. Pre-existing; every earlier round missed it. The product rule was endorsed. |
| 7 | BLOCK (wording) | soundness confirmed closed; a dangling model file still said "No such file" |
| 8 | FAIL on one Low | a dangling root lost its wording to the pre-check's listing; two doc nits; the round-seven rebuttal upheld |
| 9 | **PASS** | — |

Twenty-three mutation checks stand on this branch, each verified to fail when the thing it guards is reverted. Four crates green throughout.

**What the reviewers could not do, stated:** every round was read-only static tracing; the one that ran tests (round eight) ran only the new ones. Both C-lane reviewers of the original pass and rounds one through nine here all named the same missing instrument — a deterministic filesystem fault-injection seam for `EACCES`/dangling-ancestor states and for the walk-then-swap race. It is the single highest-value follow-up left in this class, and it is recorded on #1817 rather than claimed done.

Merging on green.

PR #1828, opened 2026-09-09
fix(engine/ui): a plan page carries nothing of the previous plan, and the queue names its models

### LINKED ISSUE #1815: The review screen keeps one plan's status, samples and approve command under the next plan's heading — and re-samples without consent
Found by an independent red-team pass over the SPA screens (#1737, #1740, #1745, #1746, #1747, #1748, #1763, #1765, #1781, #1783).

**The good news first: the pass found no XSS sink.** It enumerated every engine-composed field on every review, plan, sample, product and timeline surface and confirmed each renders as React text or a safe `title`; `decision_ref` never enters the DOM; and every link goes through `encodeURIComponent` behind a fixed `/ui/...` prefix, so `javascript:` and `data:` values become path segments. No `dangerouslySetInnerHTML`, `innerHTML`, `outerHTML` or `insertAdjacentHTML` in scope.

The defects are all correctness.

---

## HIGH — resource state and sample consent survive an identity change

`engine/ui/src/review/ReviewScreen.tsx:21-28` renders `<PlanDetail planId={planId}>` **without a key**. `engine/ui/src/estate/useResource.ts:21-40` initialises state once and starts a new `load()` in its effect, but never resets the previous `ready` value back to `loading`. `engine/ui/src/review/SamplePanel.tsx:89-125` likewise keeps `asked = true`.

Sequence — navigating from plan A directly to plan B, with B's reads slower:

```
  heading      -> B    (updates immediately)
  status       -> A    (stale)
  product      -> A
  diff         -> A
  sample rows  -> A
  approve cmd  -> "rocky review aaaa... --approve"   <-- A's plan id, under B's heading
```

Two consequences, and the second is the worse one:

1. A reviewer can read, act on, or copy an approval command for **the wrong plan**, under the right plan's heading.
2. Because `asked` is still `true`, changing the model **automatically starts B's warehouse sample** — a query runs under consent given for a different model. D8 made sampling per-request consent; this bypasses it without a click.

The same defect puts product B's heading and custody link above product A's status and journal.

**Why #1747's test does not catch it:** `PlanDetail.test.tsx:92-115` starts directly in the final empty-queue state and never performs an identity transition. The route tests mount one identity only.

**Fix:** key the detail components by plan/product identity, or reset each resource to `loading` when its dependency changes. Reset sampling consent and sample state when `model` changes.

---

## MEDIUM — a refused queue still produces "no longer in the queue"

`engine/ui/src/review/PlanDetail.tsx:170-190` computes `entry` as `queue.kind === "ready" ? find(...) : null`. The `Escalation` component correctly renders a non-ready queue as a resource failure, but the sample fallback treats that same `null` as **absence** and prints:

> A plan that is not product-bound and no longer in the queue has no model on this screen.

With the queue rejecting `503 engine_busy`, one page shows `refused (503)`, `engine_busy`, `retry` — and then separately asserts the plan is no longer in the queue. **An unreadable queue is unknown, not empty.** This is #1783's own distinction, not carried into the fallback.

**Fix:** three states — unknown/error, known absent, known present — and render absence-dependent copy only for known absence.

---

## LOW — the plan-label discriminator is only identifier syntax

`PlanDetail.tsx:28-32` uses `MODEL_NAME = /^[a-zA-Z0-9_]+$/` to decide whether a queue entry names a samplable model. Neither `entry.capability` nor a typed target participates. These all pass:

```
capability: "backfill", model: "backfill_3_models"
capability: "restore",  model: "restore_1_tombstoned_model"
capability: "gc",       model: "gc"
```

Each gets a "Show 20 rows" button; clicking it requests `/api/v1/models/backfill_3_models/rows?limit=20` with the warehouse-consent header, and either refuses as unknown or **samples an unrelated real model that happens to have that name**.

Latent today because the shipped labels carry punctuation (`"backfill: 3 model(s)"`), which is exactly what the existing tests use — so the tests cannot reach this.

**This one has a fix already in flight.** #1808 records the real model set on the escalation's ledger row rather than leaving the label to be parsed. Surfacing that set on the queue entry would let the UI stop guessing from syntax. Worth doing as the follow-up to #1808 rather than tightening the regex.

---

## LOW — truncation splits surrogate pairs, and the marker is forgeable

`engine/ui/src/format.ts:58-59,74-77` truncates with `.length` and `.slice`, which count UTF-16 code units:

```
shortId("aaaaaaaaaaa😀z")          -> "aaaaaaaaaaa\ud83d…"    (lone high surrogate)
elideMiddle("aaaaaaaaa😀bbbbbbbbbb") -> "aaaaaaaaa\ud83d…bbbbbbbb"
```

And the marker #1781 introduced as a trust signal can be forged, because it is derived from visible text:

```
shortId("abcdefghijklMIDDLE")  and  shortId("abcdefghijkl…")   both -> "abcdefghijkl…"
```

So a visible `…` does not prove content was removed. `format.test.ts:50-87` is ASCII-only.

**Fix:** return `{ text, truncated }` and render the indicator from the flag, not from the string. Segment with `Intl.Segmenter` where available.

---

## LOW — a hostile-value test does not reach its claimed sink

`engine/ui/src/governor/ProductsScreen.test.tsx:276-297` is named *"renders a hostile event and product name as text"* but passes `name="revenue_daily"` at line 284; the hostile value appears only in `rows[0].event`. If the product-name heading at `ProductsScreen.tsx:328-334` became an unsafe sink, this test would still pass.

The PlanDetail and SamplePanel hostile tests **do** reach their real components and DOM sinks — this gap is specific to the product heading and custody link.

---

Reviewer receipt: 48 tool calls; its own second reviewer accepted all six findings and qualified three as latent under current producer shapes.

hugocorreia90:
Closes #1815.

Navigating from plan A straight to plan B left every panel under B's heading showing A: A's status, A's sample rows, and **A's approve command**. Because the sample panel also kept its consent, changing the model started B's warehouse query with no click. The same shape put product A's standing under product B's heading. Four smaller findings from the same review are fixed alongside.

## What changed

**The identity fix is in the hook, not in a `key`.** `useResource` now derives its state from the deps it was loaded for, in render: when the identity moves, the value on hand is not shown under the new one, not even for a frame. A reload of the *same* identity (interval, refresh button) keeps the old value until the new one lands, as before. Every detail panel shares this hook — plan, product, model — so fixing it there closes the class; a `key` on `PlanDetail` would have fixed one screen and hidden the defect from the other two.

**Consent is per model.** `SamplePanel` keys its consent state on the model, so a model swap under it withdraws consent and brings the button back. D8 made sampling per-request consent; this keeps it that way.

**A refused queue is unknown, not empty.** `lookupQueueEntry` gives three answers — unknown / absent / present — and only *absent* may say "no longer in the queue". The sample fallback used to read the same `null` as absence and print that beside the queue's refusal.

**The engine names the models; the UI stops guessing.** `ReviewQueueEntry` gains `models`: the set the ledger row recorded, or the single `model` on an ordinary row — the same keys the blast radius was computed over, from one derivation, so the two cannot drift. Empty means unknown (a pre-v28 plan-level row). The UI's `^[a-zA-Z0-9_]+$` check on the label is gone; `backfill_3_models` would have passed it and sampled a real model of that name. Schema, Pydantic and TypeScript bindings regenerated.

The derivation was wrong once during this PR and is worth stating: the first version treated an empty set on a `backfill`/`gc`/`restore` row as "plan-level, unknown". But the apply-time gate records one ordinary row per touched model *with the plan's capability* (`gc` over `orders`, no set), and that rule threw the name away. The rule that holds is the name's shape by the engine's own admission rule (`validate_identifier`): a value it refuses is in no graph, and every label the engine has ever written fails it. `blast_radius_union` over no subjects now returns unknown rather than a measured zero, with its own test.

**Truncation is grapheme-safe and the marker is honest.** `clipHead` / `clipMiddle` cut between user-perceived characters and return whether they cut; `<Clip>` renders the marker as its own dimmed element only when something was cut. A value that merely contains "…" renders it as text.

**The hostile-name test reaches its sink.** It passes the hostile value as the product name, so the heading and custody link are exercised, not only the journal event.

## Tests

- UI: 8 new (`useResource` identity + same-identity reload, `SamplePanel` consent withdrawal, `PlanDetail` A→B transition with a fetch spy proving one query ever ran, refused-queue fallback wording, identifier-shaped label, unrecorded set, `ProductsScreen` A→B, `Clip` marker vs literal, surrogate pairs at both ends). Existing tests updated to the new copy and fixtures.
- Engine: `the_entry_reports_the_graph_keys_the_ranking_resolved` (ordinary row, apply-time gc row, plan-level row with its set), `a_union_over_no_subjects_is_unknown_not_zero`, and the pre-v28 assertions added to `a_plan_level_row_with_no_resolvable_model_ranks_unknown_not_zero`.

## Not in this PR

- A `key` on the route-level components. Deliberate, see above.
- The `models` cell on the queue panel shows the set only where it differs from the label.

---

## What changed after the description above — eight review rounds

The description above is the first commit. Seven more landed under independent review (Codex, with a second model concurring on each round), each dispositioned in the comments. The thread running through all of them: **a string cannot say what kind of thing it is; the writer that knew must record it.**

- **`models` on the queue entry, and who decides it.** Three reader-side rules were tried and each broke — the capability, identifier syntax, graph membership — before the rule that holds: the gate says what its subjects are (`GateSubjects`: compiled models, replication targets, or the set a resolver could vouch for), the ledger row says whether its set was deliberate (`keys_recorded`, state schema v29), and the queue takes a recorded set as written, resolving only rows that never said.
- **`preview_model`, the only licence to read.** Computed by the samples route's own admission code, lifted into one function the route and the queue both call, and gated on the route's front door (strict config, adapter registry, one resolvable pipeline). The UI offers a sample from it and from nothing else; the product's output model stands in only once the plan has left the queue.
- **Every row of a plan.** The queue keeps one row per touched model; the screen now shows each, with one command that clears them all.
- **Supersession.** A later `allow` retires an older `require_review`. A `deny` does not, because the fail-closed refusal of an unreadable ledger is recorded as one and the row cannot say which kind it is.
- **The maintenance resolver.** A table resolves to the model that targets it (the physical reading first); a string that is one model's table and another's name refuses; a table with two owners refuses only when the plan names it; ephemeral models own no table; a table no model owns is gated but records no model's key. Promote's resolver likewise vouches only for the keys it mapped.

Thirty mutation checks; the final round PASSED with a second model concurring. Residuals that need plan-time or ledger-time provenance the CLI does not yet record — compact's bare name, archive's two-part identifier, the MCP draft tools' stem/name mismatch, the fail-closed `deny` bit, clock rollback — are in #1829 with the reviewers' fixtures.

### DIFF
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -17214,7 +17214,7 @@
             "type": "string"
           },
           "models": {
-            "description": "The graph keys this entry stands for — names of compiled models, and only those: the set the ledger row recorded, or its bare `model` when the compiled graph has a model of that name. It is the set `blast_radius` was computed over. **Empty means unknown**, never \"no models\": a subject that is not a compiled model (a replication target is gated by table name; a plan-level row's `model` is a label), a model since removed, a failed compile, or a row from before its producer recorded the set. A consumer that needs a model name takes it from here; `model` is display text and is not to be parsed.",
+            "description": "The graph keys this entry stands for — names of compiled models, and only those: the set the ledger row recorded, or its bare `model` when the compiled graph has a model of that name. It is the set `blast_radius` was computed over and what `rocky audit --for` will match. **Empty means unknown**, never \"no models\": a subject that is not a compiled model (a replication target is gated by table name; a plan-level row's `model` is a label), a model since removed, a failed compile, or a row from before its producer recorded the set. `model` is display text and is not to be parsed.\n\nA key here is for ranking and audit. It is **not** a licence to read rows — `preview_model` is.",
             "items": {
               "type": "string"
             },
@@ -17223,6 +17223,13 @@
           "plan_id": {
             "description": "The plan whose approval clears this escalation.",
             "type": "string"
+          },
+          "preview_model": {
+            "description": "The one model of this entry that `GET /api/v1/models/{name}/rows` would read, decided by that route's own admission rules (a valid name, in the compiled project, not time-interval, no compile errors, a single SELECT), or `null`: the entry names no model or several, or the route would refuse the one it names — a dotted name, a model a restore plan recorded that is gone from the current project, a model that no longer compiles. A consumer offers a sample from this and from nothing else.",
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "principal": {
             "allOf": [
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -17214,7 +17214,7 @@
             "type": "string"
           },
           "models": {
-            "description": "The graph keys this entry stands for: the model set the ledger row recorded, or the single `model` on an ordinary evaluation row. It is the set `blast_radius` was computed over. **Empty means unknown** — a plan-level row written before the engine kept its set — and never \"no models\". A consumer that needs a model name takes it from here; `model` is display text and is not to be parsed.",
+            "description": "The graph keys this entry stands for — names of compiled models, and only those: the set the ledger row recorded, or its bare `model` when the compiled graph has a model of that name. It is the set `blast_radius` was computed over. **Empty means unknown**, never \"no models\": a subject that is not a compiled model (a replication target is gated by table name; a plan-level row's `model` is a label), a model since removed, a failed compile, or a row from before its producer recorded the set. A consumer that needs a model name takes it from here; `model` is display text and is not to be parsed.",
             "items": {
               "type": "string"
             },
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -17225,7 +17225,7 @@
             "type": "string"
           },
           "preview_model": {
-            "description": "The one model of this entry that `GET /api/v1/models/{name}/rows` would read, decided by that route's own admission rules (a valid name, in the compiled project, not time-interval, no compile errors, a single SELECT), or `null`: the entry names no model or several, or the route would refuse the one it names — a dotted name, a model a restore plan recorded that is gone from the current project, a model that no longer compiles. A consumer offers a sample from this and from nothing else.",
+            "description": "The one model of this entry that `GET /api/v1/models/{name}/rows` would admit, decided by that route's own rules — its front door (the strict config, the adapter registry, one resolvable pipeline with no name given) and its admission of the model (a valid name, in the compiled project, not time-interval, no compile errors, a single SELECT) — or `null`: the entry names no model or several, the route's front door is shut for this project, or the route would refuse the one model it names (a dotted name, a model a restore plan recorded that is gone from the current project, a model that no longer compiles). A consumer offers a sample from this and from nothing else.\n\nWhat this cannot promise is the request itself: the adapter's reachability and credentials at query time, the masking of classified columns, the route's concurrency and time limits. Those refusals arrive as the route's own envelope and render as themselves.",
             "type": [
               "string",
               "null"
--- a/docs/public/openapi.json
+++ b/docs/public/openapi.json
@@ -17210,8 +17210,15 @@
             "type": "string"
           },
           "model": {
-            "description": "The model the escalation is about.",
+            "description": "The model the escalation is about. On a plan-level escalation (`backfill` / `gc` / `restore`) this is a human label — `\"backfill: 3 model(s)\"` — and not a model name; the names are in `models`.",
             "type": "string"
+          },
+          "models": {
+            "description": "The graph keys this entry stands for: the model set the ledger row recorded, or the single `model` on an ordinary evaluation row. It is the set `blast_radius` was computed over. **Empty means unknown** — a plan-level row written before the engine kept its set — and never \"no models\". A consumer that needs a model name takes it from here; `model` is display text and is not to be parsed.",
+            "items": {
+              "type": "string"
+            },
+            "type": "array"
           },
           "plan_id": {
             "description": "The plan whose approval clears this escalation.",
@@ -17259,6 +17266,7 @@
           "classification_weight",
           "decision_ref",
           "model",
+          "models",
           "plan_id",
           "principal",
           "reason",
--- a/editors/vscode/src/types/generated/review_queue.ts
+++ b/editors/vscode/src/types/generated/review_queue.ts
@@ -85,7 +85,7 @@ export interface ReviewQueueEntry {
    */
   model: string;
   /**
-   * The graph keys this entry stands for: the model set the ledger row recorded, or the single `model` on an ordinary evaluation row. It is the set `blast_radius` was computed over. **Empty means unknown** — a plan-level row written before the engine kept its set — and never "no models". A consumer that needs a model name takes it from here; `model` is display text and is not to be parsed.
+   * The graph keys this entry stands for — names of compiled models, and only those: the set the ledger row recorded, or its bare `model` when the compiled graph has a model of that name. It is the set `blast_radius` was computed over. **Empty means unknown**, never "no models": a subject that is not a compiled model (a replication target is gated by table name; a plan-level row's `model` is a label), a model since removed, a failed compile, or a row from before its producer recorded the set. A consumer that needs a model name takes it from here; `model` is display text and is not to be parsed.
    */
   models: string[];
   /**
--- a/editors/vscode/src/types/generated/review_queue.ts
+++ b/editors/vscode/src/types/generated/review_queue.ts
@@ -81,9 +81,13 @@ export interface ReviewQueueEntry {
    */
   decision_ref: string;
   /**
-   * The model the escalation is about.
+   * The model the escalation is about. On a plan-level escalation (`backfill` / `gc` / `restore`) this is a human label — `"backfill: 3 model(s)"` — and not a model name; the names are in `models`.
    */
   model: string;
+  /**
+   * The graph keys this entry stands for: the model set the ledger row recorded, or the single `model` on an ordinary evaluation row. It is the set `blast_radius` was computed over. **Empty means unknown** — a plan-level row written before the engine kept its set — and never "no models". A consumer that needs a model name takes it from here; `model` is display text and is not to be parsed.
+   */
+  models: string[];
   /**
    * The plan whose approval clears this escalation.
    */
--- a/editors/vscode/src/types/generated/review_queue.ts
+++ b/editors/vscode/src/types/generated/review_queue.ts
@@ -85,13 +85,19 @@ export interface ReviewQueueEntry {
    */
   model: string;
   /**
-   * The graph keys this entry stands for — names of compiled models, and only those: the set the ledger row recorded, or its bare `model` when the compiled graph has a model of that name. It is the set `blast_radius` was computed over. **Empty means unknown**, never "no models": a subject that is not a compiled model (a replication target is gated by table name; a plan-level row's `model` is a label), a model since removed, a failed compile, or a row from before its producer recorded the set. A consumer that needs a model name takes it from here; `model` is display text and is not to be parsed.
+   * The graph keys this entry stands for — names of compiled models, and only those: the set the ledger row recorded, or its bare `model` when the compiled graph has a model of that name. It is the set `blast_radius` was computed over and what `rocky audit --for` will match. **Empty means unknown**, never "no models": a subject that is not a compiled model (a replication target is gated by table name; a plan-level row's `model` is a label), a model since removed, a failed compile, or a row from before its producer recorded the set. `model` is display text and is not to be parsed.
+   *
+   * A key here is for ranking and audit. It is **not** a licence to read rows — `preview_model` is.
    */
   models: string[];
   /**
    * The plan whose approval clears this escalation.
    */
   plan_id: string;
+  /**
+   * The one model of this entry that `GET /api/v1/models/{name}/rows` would read, decided by that route's own admission rules (a valid name, in the compiled project, not time-interval, no compile errors, a single SELECT), or `null`: the entry names no model or several, or the route would refuse the one it names — a dotted name, a model a restore plan recorded that is gone from the current project, a model that no longer compiles. A consumer offers a sample from this and from nothing else.
+   */
+  preview_model?: string | null;
   /**
    * Who authored the change (`human` / `agent`).
    */
--- a/editors/vscode/src/types/generated/review_queue.ts
+++ b/editors/vscode/src/types/generated/review_queue.ts
@@ -95,7 +95,9 @@ export interface ReviewQueueEntry {
    */
   plan_id: string;
   /**
-   * The one model of this entry that `GET /api/v1/models/{name}/rows` would read, decided by that route's own admission rules (a valid name, in the compiled project, not time-interval, no compile errors, a single SELECT), or `null`: the entry names no model or several, or the route would refuse the one it names — a dotted name, a model a restore plan recorded that is gone from the current project, a model that no longer compiles. A consumer offers a sample from this and from nothing else.
+   * The one model of this entry that `GET /api/v1/models/{name}/rows` would admit, decided by that route's own rules — its front door (the strict config, the adapter registry, one resolvable pipeline with no name given) and its admission of the model (a valid name, in the compiled project, not time-interval, no compile errors, a single SELECT) — or `null`: the entry names no model or several, the route's front door is shut for this project, or the route would refuse the one model it names (a dotted name, a model a restore plan recorded that is gone from the current project, a model that no longer compiles). A consumer offers a sample from this and from nothing else.
+   *
+   * What this cannot promise is the request itself: the adapter's reachability and credentials at query time, the masking of classified columns, the route's concurrency and time limits. Those refusals arrive as the route's own envelope and render as themselves.
    */
   preview_model?: string | null;
   /**
--- a/engine/crates/rocky-cli/src/api.rs
+++ b/engine/crates/rocky-cli/src/api.rs
@@ -3622,6 +3622,7 @@ mod tests {
             .expect("run recorded");
         store
             .record_policy_decision(&PolicyDecisionRecord {
+                keys_recorded: false,
                 models: Vec::new(),
                 timestamp: now - chrono::Duration::minutes(30),
                 plan_id: "freeze:global".to_string(),
@@ -3937,6 +3938,7 @@ mod tests {
         StateStore::open(&state_path)
             .unwrap()
             .record_policy_decision(&PolicyDecisionRecord {
+                keys_recorded: false,
                 models: Vec::new(),
                 timestamp: chrono::Utc::now() - chrono::Duration::minutes(5),
                 plan_id: "plan-revenue-daily".to_string(),
--- a/engine/crates/rocky-cli/src/commands/apply.rs
+++ b/engine/crates/rocky-cli/src/commands/apply.rs
@@ -1772,6 +1772,7 @@ fn evaluate_apply_policy_with_policy_matching_dual(
                     principal,
                     touched,
                     attrs,
+                    GateSubjects::CompiledModels,
                     &prior_decisions,
                     marker_freezes,
                     snapshot_unreadable,
@@ -1798,6 +1799,7 @@ fn evaluate_apply_policy_with_policy_matching_dual(
         principal,
         touched,
         eval_attrs,
+        GateSubjects::CompiledModels,
         &prior_decisions,
         marker_freezes,
         snapshot_unreadable,
@@ -1951,6 +1953,7 @@ pub(crate) fn evaluate_apply_policy_with_store(
     models_glob: Option<&str>,
     ledger: &StateStore,
     marker_freezes: &[rocky_core::freeze_marker::ActiveMarkerFreeze],
+    subjects: GateSubjects,
 ) -> PolicyGate {
     // Finding 1: takes the SAME `[policy]` snapshot `run` already holds (its L1212
     // `rocky_cfg`), not a reload — the in-run replication gate must evaluate the
@@ -1973,6 +1976,7 @@ pub(crate) fn evaluate_apply_policy_with_store(
         principal,
         touched,
         &attrs_map,
+        subjects,
         &prior_decisions,
         marker_freezes,
         snapshot_unreadable,
@@ -2027,6 +2031,22 @@ pub(crate) fn resolve_policy_and_attrs(
     Ok((policy.clone(), attrs))
 }
 
+/// What the keys of a gate's `touched` map name. The map holds strings; only
+/// the caller knows whether they are the compiled project's models or a
+/// replication's target tables, and the row the gate records must not call a
+/// target a model. A target that happens to share a compiled model's name
+/// still matches that model's policy attributes (the plane's standing rule),
+/// but it is not that model, and a consumer that read the row's `models` as
+/// a licence to sample would read the wrong thing (#1815).
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum GateSubjects {
+    /// Keys are models of the compiled project (apply, promote, gc, restore,
+    /// the MCP propose gate).
+    CompiledModels,
+    /// Keys are replication target table names, gated under bare `apply`.
+    ReplicationTargets,
+}
+
 /// The per-model evaluation loop shared by [`evaluate_apply_policy`] and
 /// [`evaluate_apply_policy_with_store`]. The snapshot and the record sink are
 /// supplied by the caller so the two variants differ only in how they reach the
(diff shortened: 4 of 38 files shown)


hugocorreia90:
## Independent red-team disposition — round one: NOT FULLY CLOSED, all three findings accepted

**Reviewer: Codex (GPT-5), 89 tool calls on the frozen range `618c7cc4..d78031d2`, read-only, 50 files opened, the targeted engine tests run.** It confirmed the identity and refusal fixes (all 19 `useResource` call sites inspected; the fallback matrix enumerated) and found three real defects in the new `models` field and around it. Each is fixed in the second commit.

### MEDIUM — `queue_graph_keys` decided by identifier syntax, which is neither necessary nor sufficient

Accepted. Two concrete failures, both source-proven:

- **False positive.** A replication target is gated under bare `apply` by *table* name (`raw_orders`), recorded with an empty set and a name that passes the identifier check. The queue handed it over as a model; the UI offered a sample; the preview route answers `model_not_found` — or, with a same-named compiled model, samples that one.
- **False negative.** `v2.fct_orders` is a real model name (a dotted file stem loads as one) that the identifier check rejects. The queue reported no keys and the UI said the row "was escalated before the engine kept one", which was false.

**Fix, in two halves.** The *producer* now binds the contract: the apply-time gate records `models: [model]` when its subject is a model of the compiled project, which it alone knows for certain (the same `model` field carries a target's table name). For rows with no recorded set, the queue asks the *compiled graph* whether the bare name is a model; nothing about the string can say. No keys now means exactly "no compiled model the engine can name": a non-model subject, a removed model, a failed compile, or a pre-recording row. The UI copy no longer claims which.

Three rules are now on the record as tried and rejected: capability (the gate writes ordinary rows with the plan's capability), identifier syntax (this finding), and the label itself. A mutation check reinstates each and the tests fail.

**Test evidence, as asked for:** `the_apply_gate_records_a_compiled_models_key_and_no_key_for_a_target` drives the real gate — a compiled model and a replication target under one plan — and reads back both the ledger rows and the queue. `the_entry_reports_the_graph_keys_the_ranking_resolved` gained a dotted model and a target-shaped row. `a_failed_compile_keeps_recorded_keys_and_vouches_for_no_bare_name` pins the degraded case.

### MEDIUM — `PlanDetail` kept one arbitrary row of a plan that pends several times

Accepted. The gate records one row per touched model and the queue keeps one per (plan, model); `.find()` showed one reason and one model beside a command that clears all of them. `lookupQueueEntry` now returns every row that names the plan; the escalation panel renders each (model, capability, principal, rule, blast radius) with a lead line saying how many and that approval clears them all; the sample is offered only when the rows together name exactly one model, and the approve section says what it clears. Test: two rows, two rules, two reasons, one command.

### LOW — the grapheme guarantee did not hold on the `Intl.Segmenter` fallback

Accepted as stated: the fallback uses code points, which keeps surrogate pairs whole but can part a combining mark from its base. The fallback stays (no browser floor is declared), and the doc now says exactly what it guarantees. Two tests: a combining mark stays with its base where the segmenter exists; surrogate pairs stay whole with the segmenter stubbed out.

### Also from the review

- The same-identity reload test passes on the pre-fix hook. Correct, and intended: it guards the *other* half of the contract (a refresh must not blink) against an over-eager fix. The identity test is the one that fails on the old hook.
- The reviewer could not run the UI tests (no `node_modules` in its checkout). Round two's checkout will have them.

Round two dispatched on the second commit.

hugocorreia90:
## Independent red-team disposition — round two: NOT FULLY CLOSED, all three findings accepted

**Reviewer: Codex (GPT-5), 70 tool calls on the frozen range `d78031d2..afe6db15` and the whole PR; its second reviewer (`gpt-5.6-terra`) independently found the same three and concurred.** It confirmed the multi-row and grapheme fixes, all 19 `useResource` call sites, the `decision_ref` key, and that attribute-load failure does not silently empty a compiled model's set. Each finding below is fixed in the third commit.

### MEDIUM — a replication target named like a compiled model recorded that model as its key

Accepted. Round two made the gate record `models: [model]` when `attrs_map.contains_key(model)`; the replication gate resolves the real models dir (so a target's policy attributes match by name, the plane's standing rule), so a target called `orders` beside a compiled model `orders` recorded `["orders"]`, and the UI would have sampled the unrelated compiled model.

**Fix:** the evaluator now takes `GateSubjects` from its caller — `CompiledModels` or `ReplicationTargets` — and records a key only for the first. Being in the attribute map was not enough to know; only the caller knows what its keys name. Test drives the real replication gate with a same-named compiled model and reads the ledger row back. The policy *verdict* for the collision is unchanged and remains the plane's pre-existing rule; only the row's claim about identity changed.

### MEDIUM — `models` served two roles: ranking membership and permission to read

Accepted, and this is the sharper finding. Graph membership is neither the samples route's rule (it refuses a dotted name, a time-interval model, a model with compile errors) nor stable (a restore plan's recorded tombstone name is gone from the current graph).

**Fix:** the route's admission rules are lifted into one function, `preview_rows::admit_model`, that the route itself now calls; the queue calls it too and reports `ReviewQueueEntry.preview_model` — the one model the route would read for that row, or null. `models` is documented as ranking and audit, explicitly *not* a licence to read; the UI offers a sample from `preview_model` and from nothing else. Tests: a dotted model (a graph key, refused), a restore's gone model, a model that no longer compiles, a target, a two-model set — none offered; an ordinary model — offered.

### MEDIUM — a later `allow` or `deny` did not supersede an older `require_review`

Accepted; pre-existing, and the doc comment already promised "the latest decision per (plan, model)". `select_outstanding` filtered to `require_review` before picking the latest, so the superseded escalation stayed approvable. Now the latest row per key is picked across all effects first. One consequence, stated: the fail-closed transient `deny` (ledger unreadable) also supersedes, so an escalation drops out of the queue until the next evaluation re-records it — which is the current state of that plan, and the row is in the audit ledger. Test covers both orders.

### The reviewer's uncertainty

One samplable model beside other, non-model escalations on the same plan: the screen keeps every escalation visible and offers the one readable model. Kept as is; it is internally consistent, and the reviewer said the same.

Fourteen mutation checks stood on rounds one and two; four more on this round (the collision, the offer from a graph key, the filter-before-latest, the UI offering from `models`). Round three dispatched on the third commit.

hugocorreia90:
## Independent red-team disposition — round three: NOT FULLY CLOSED; two fixed, one narrowed, one rebutted with a follow-up

**Reviewer: Codex (GPT-5), 81 tool calls on `afe6db15..d7ed50aa` and the whole PR; its second reviewer (`gpt-5.6-terra`) confirmed all four.** It verified the route refactor kept every check's order and envelope, enumerated every `evaluate_apply_policy_core` caller, and found none of the new tests vacuous.

### MEDIUM 2 — a product-bound plan still in the queue fell back to the product's output model. Fixed.

Accepted. The product fallback exists for a plan that has *left* the queue (approved, its table now real). While the plan is in the queue the engine has already said which model, if any, the route would read, and a null there is an answer. The fallback now applies only to `lookup.kind === "absent"`. Test: a product-bound pending plan whose row is `preview_model: null` shows the spec-drift card (so the product was read) and no offer.

### MEDIUM 3 — `preview_model` ignored the route's front door. Fixed for what is decidable; the rest is stated.

Accepted. The route refuses before compiling when the strict config does not load, the adapter registry cannot be built, or no single pipeline resolves without a name — and the UI sends none, so two pipelines meant a guaranteed `pipeline_error` after an offer. The queue now decides that front door once per project, with the route's own loaders, and offers nothing when it is shut. The queue's own tolerance of an unset credential and an absent config is unchanged and still pinned; it ranks, it just does not offer.

What remains undecidable in advance is the request itself — adapter reachability and credentials at query time, masking of classified columns, the route's limits — and the field's documentation now says exactly that instead of "would read". Compile-context differences (cached source schemas, mask settings) are fail-closed refusals that render as themselves; not promised. Tests: no config → no offer; two pipelines → no offer; one pipeline → `a` offered.

### MEDIUM 4 — a transient fail-closed `deny` hid a still-required review. Narrowed.

Accepted, and it was a regression of round two's own fix, so it is closed here rather than deferred: only `allow` now supersedes an older `require_review`. A `deny` is left out of the latest-row pick entirely — it may be the fail-closed refusal of an unreadable ledger, which says nothing about the plan, and the row cannot say which kind it is. A superseded-by-policy-deny escalation therefore stays approvable, which is what it was before this PR; approval then meets the deny at apply, loudly. The bit that would let a policy deny supersede belongs to the writer (`fail_closed` on the ledger row, a v29 change) and is filed as #1829. Clock rollback is a ledger-wide property and is filed with it.

### MEDIUM 1 — maintenance and promote keys. Rebutted in part, filed in full.

The reviewer's concrete input — `rocky compact orders` beside a compiled model `orders` — does not reach `GateSubjects` at all: the maintenance resolver deliberately takes a bare name that matches a model **as that model** (`--model X` semantics), so the key is *resolved*, the policy plane already evaluates the model's attributes for it, and the row's `models: ["orders"]` is the resolver's own claim, not the gate's string match. The defect is the compact CLI accepting one string for two things and regenerating `OPTIMIZE orders` from it; it predates this PR and cannot be fixed in the gate. A verbatim maintenance key equal to a model name cannot exist (the resolver would have resolved it). Promote's case is real but narrow — two loaders disagreeing on a partially broken project — and needs per-key provenance from the resolver, the same shape as `GateSubjects` but per key. Both are in #1829 with the reviewer's fixture.

Eighteen mutation checks stood before this round; three more on it (the front door removed, a deny allowed to supersede, the product fallback applied while the plan is queued). Round four dispatched on the fourth commit.

hugocorreia90:
## Independent red-team disposition — round four: NOT FULLY CLOSED on one finding, now fixed

**Reviewer: Codex (GPT-5), 43 tool calls on `d7ed50aa..0c74ecf1` and the whole PR; its second reviewer (`gpt-5.6-terra`) ran independently.** It confirmed the product fallback, the route front door, and the allow-only supersession closed, confirmed maintenance is closed (every key its resolver leaves is either a model it resolved or a string no model is named), and attacked the promote half of my round-three rebuttal successfully.

### MAJOR — a promote target kept verbatim can spell a dotted model's name. Accepted, fixed.

I had called promote's case "narrow — two loaders disagreeing on a partially broken project". The reviewer's sequence needs no broken project: a model is retargeted after a promote plan is persisted, so the old FQN no longer maps and is kept verbatim (the fail-closed D5 rule); a dotted model *name* equal to that FQN — `corp.prod.orders` is a legal name, as this PR's own dotted-key test proves — puts it in the attribute map, and the gate's string test recorded the target as that model's key. The queue then offered rows of an unrelated model beside a promote that overwrites a different table.

**Fix:** the resolver says which keys it resolved. `touched_models_for_promote` returns the resolved set beside the map; `GateSubjects::Resolved(&set)` carries it, and the gate records a key only for a key that is both resolved and in the attribute map. Every other gate now names `CompiledModels` explicitly (run, AI-authored, maintenance, gc, restore, replication says `ReplicationTargets`). Test: a promote over a mappable FQN and a verbatim FQN that spells a model's name — the first row records its key, the second records none; a mutation check reinstating the string test fails it.

The policy-attribute half (the verdict is computed against the same-named model's attributes) is the plane's pre-existing string matching and is unchanged; it is in #1829 with the fixture the reviewer asked for.

Twenty-one mutation checks stood before this round; one more on it. Round five dispatched on the fifth commit.

hugocorreia90:
## Independent red-team disposition — round five: NOT FULLY CLOSED; the maintenance half fixed, the MCP finding filed

**Reviewer: Codex (GPT-5), 77 tool calls on `0c74ecf1..90571364` and the whole PR; its second reviewer (`gpt-5.6-terra`) reproduced both findings.** It confirmed the promote fix and its test's reversion sensitivity, the dual-probe path, the `Resolved` coalescing case, and no regression from earlier rounds.

### MAJOR — the maintenance resolver's names-first order picked the wrong model. Accepted, fixed; my round-three rebuttal was wrong.

I had said a verbatim maintenance key can never be a model name. True and beside the point: the failure is not a verbatim key, it is a *mis-resolved* one. `rocky compact --catalog corp` discovers `corp.prod.orders` — model B's table; a model A *named* `corp.prod.orders` (a legal dotted name) won the names-first lookup, so the gate evaluated A, recorded A's key, and the review screen offered A's rows beside a plan whose SQL compacts B's table.

**Fix, in the resolver:** the physical reading wins — the plan's SQL mutates exactly that table, so the model whose target it is, is the model touched; a string that is one model's table *and* another model's name is refused outright, naming both readings (the plan cannot say which it meant, and the apply is destructive); and the resolver returns which subjects it could vouch for as models, so the maintenance gate now passes `GateSubjects::Resolved` and a table no model targets is gated under its own string but records no key. Tests: the reviewer's exact fixture refuses; a table resolves to its model and a stranger records nothing, read back from the ledger. Every gate now names its subjects explicitly; none says `CompiledModels` for a map it did not build from compiled models.

### MEDIUM — MCP `draft_contract` / `draft_check` / `draft_metadata` resolve by filename stem, gate by logical name. Filed as #1829 (item 5).

Accepted as a defect, not fixed here. It predates this PR (the row's `model` was already the wrong identity; `models` now repeats it), it lives in the MCP draft tools' own subject semantics — the reviewer's own open question is whether the stem was ever meant to be the canonical draft subject — and those rows are decision-only (`draft:*` ids, no plan file), so they never enter the review queue and `preview_model` is not reachable from them. The audit-identity harm is real and is in #1829 with the reviewer's fixture; the fix belongs to the tools, which have just read the sidecar and know the logical name.

Twenty-two mutation checks stood before this round; two more on it (the FQN-first order and the maintenance `Resolved`). Round six dispatched on the sixth commit.

hugocorreia90:
## Independent red-team disposition — round six: NOT FULLY CLOSED; one fixed, one test gap closed, one filed

**Reviewer: Codex (GPT-5), 85 tool calls on `90571364..9c7c2301` and the whole PR; its second reviewer (`gpt-5.6-terra`) found the same three.** It confirmed the ambiguity refusal is fail-closed and names both readings, that case folding opens no separate path, that coalescing two targets onto one model loses no SQL, that the archive wildcard is refused before gating, and that the MCP filing is accurate.

### MAJOR 2 — two models declaring one physical table collapsed to one policy subject. Accepted, fixed.

The resolver kept one owner per lowercase FQN; a denied co-owner vanished from the verdict and the ledger. The compiler's E036 exists for this and the maintenance loader (partial, deliberately) never sees it. The resolver now refuses when a second model declares a table already owned, naming both and the diagnostic — a destructive apply against a table with two owners cannot say whose policy governs it. Test: the reviewer's fixture (`pii_orders` and `zzz_public` on one table) refuses.

### MEDIUM — the stranger test did not bind production's `Resolved` handoff. Accepted, closed.

Correct: my test handed `Resolved` to the evaluator by hand, and my mutation check for it only passed because the mutation script also changed the resolver. A new test drives `gate_maintenance_apply` itself with a real config, a model's table and a stranger, and reads both ledger rows back; the mutation that makes production say `CompiledModels` now fails it.

### MAJOR 1 — `rocky archive --model` accepts one- and two-part physical identifiers. Filed as #1829 item 6; not fixed here.

Accepted as a defect, and the same class as compact's bare name: the archive CLI keeps `--model prod.orders` literally, `DELETE FROM prod.orders` resolves on the warehouse to the session's current catalog, and the resolver — which recognises only a three-part FQN as physical — matches it as a logical model of that name. The plan records no provenance, so nothing downstream can tell the two apart; this PR's refusal of a string that is both a known table and a known name cannot fire because the two-part form is never a known table. It predates this PR (names-first resolved it the same way), the fix is at planning time (refuse partial physical identifiers, or persist which kind the argument was and canonicalise), and the reviewer's own open question is whether those forms were ever meant as deferred warehouse identifiers — a product decision. Filed with the fixture.

Twenty-four mutation checks stood before this round; two more on it. Round seven dispatched on the seventh commit.

hugocorreia90:
## Independent red-team disposition — round seven: NOT FULLY CLOSED, all four accepted and fixed

**Reviewer: Codex (GPT-5), 63 tool calls on `9c7c2301..e19c0eec` and the whole PR; its second reviewer (`gpt-5.6-terra`) confirmed all four.** It confirmed the production-gate test reaches the recording sink, that case folding opens nothing new, and that two files with one logical name are refused downstream by the attribute compile.

### MAJOR 1 — the queue undid the gate's "not a model". Accepted; this was the tri-state I dodged in round two.

An empty `models` could not say "the gate wrote this on purpose" versus "the row predates the set", so the queue's graph fallback (round two) re-resolved a replication target's bare name and — beside a compiled model of the same name — handed that model over and offered its rows. The reviewer asked for a backward-compatible provenance field, and that is the fix: `PolicyDecisionRecord::keys_recorded` (state schema v29, serde-additive, forward-deserialises `false`). The gate and the plan-level writer set it; the queue takes a recorded set as written, an empty one included, and resolves only rows that never said. Test drives the real held-store evaluator as `ReplicationTargets` with a same-named compiled model present and reads the ledger row and the queue entry back: `models: []`, `preview_model: null`.

### MAJOR 2 — an ephemeral model's phantom target counted as ownership. Accepted, fixed.

The resolver now skips ephemeral models when indexing table ownership, as the compiler does for E036. Tests: a lone ephemeral target stays verbatim and is no model's key; beside a real owner it is that owner, with no two-owner refusal.

### MEDIUM 1 — the two-owner refusal fired for tables the plan never named. Accepted, fixed.

Owners are indexed per table and the refusal is raised only when a plan target resolves to a table with two; a duplicate elsewhere in the project is the compiler's to report. Test: `a`/`b` on one table, the plan naming another, resolves.

### MEDIUM 2 — the production-gate test could not tell `Resolved` from `CompiledModels`. Accepted; the reviewer's recipe is now the test.

A model spelled like the stranger is added between resolution and the gate, so it is in the attribute map but not the resolved set — the case the two variants differ on, and the case the reviewer's "observational equivalence is false" argument rests on (the resolver and the attribute compile read the tree separately). The mutation to `CompiledModels` in production now fails it; the comment that claimed so before was wrong and is corrected.

Twenty-six mutation checks stood before this round; four more on it. Round eight dispatched on the eighth commit. The dagster fixtures that print the schema version are regenerated in the same commit.

hugocorreia90:
## Independent red-team disposition — round eight: PASS

**Reviewer: Codex (GPT-5), 52 tool calls on `e19c0eec..2792e596` and the whole PR; its second reviewer (`gpt-5.6-terra`) independently returned PASS.** It ran the full `rocky-core` / `rocky-cli` / `rocky-mcp` suites and the UI typecheck itself.

> PASS. No in-scope correctness, safety, or regression defect at the frozen head.

What it verified to get there: the gate records `keys_recorded` on every evaluation and the queue takes a recorded set as written (an empty one included), resolving only legacy rows; the four production `keys_recorded: false` sites (drift custody, freeze/unfreeze, post-apply verification) never become approvable queue rows; the v29 bump's forward-deserialise, older-store stamping, newer-store refusal and doctor wording; ephemeral models stay in `names` (`--model scratch` remains the plan's claim) and only leave table ownership; the duplicate refusal fires only for a table the plan names; the production-gate test now distinguishes `Resolved` from `CompiledModels`; and the whole-PR properties from rounds one to seven (allow-only supersession, `preview_model` as the sole licence to read, identity-bound resource state, per-model consent, grapheme-safe clipping).

Two stale comments it noted — the pin test's historical stanza stopping at v28 and the v27 preservation test's prose — are corrected in a final comment-only commit.

**What it could not do, stated:** open a redb file written by an actual v28 binary (the tests synthesise v28 by dropping the key), and drive a credentialed replication run with a browser session through the collision. Both are recorded, not claimed.

## Where this PR ended up — eight rounds, on the record

|---|---|---|
| 1 | NOT FULLY CLOSED | identifier syntax as graph-key authority (both directions); one row per plan hidden; grapheme fallback overstated |
| 2 | NOT FULLY CLOSED | a target named like a model recorded as it; `models` doubling as a licence to read; allow/deny not superseding |
| 3 | NOT FULLY CLOSED | maintenance/promote heterogeneous keys; product fallback bypassing `preview_model`; the route's front door; the fail-closed deny hiding a review (a round-two regression, narrowed) |
| 4 | NOT FULLY CLOSED | promote's verbatim FQN spelling a dotted model's name |
| 5 | NOT FULLY CLOSED | the maintenance resolver's names-first order picking the wrong model; the MCP draft tools' stem/name mismatch (filed) |
| 6 | NOT FULLY CLOSED | two owners of one table collapsing to one; the production handoff untested; archive's two-part identifiers (filed) |
| 7 | NOT FULLY CLOSED | **the queue undoing the gate's "not a model"** — the tri-state I had dodged in round two; ephemeral phantom ownership; the global duplicate refusal; the vacuous production test |
| 8 | **PASS** | — |

Thirty mutation checks stand on this branch, each verified to fail when the thing it guards is reverted. Residuals with the reviewers' fixtures are in #1829 (six items). Merging on green.

PR #1830, opened 2026-09-09
fix(engine/rocky-cli): the scheduler metrics say when the webhook spool cannot be read

### LINKED ISSUE #1812: The scheduler metrics still call an unreadable webhook spool a healthy tick — #1752 wired the JSON, not the metrics
Found by an independent red-team pass over the #1733→#1795 merge window.

#1752's title is *"carry the spool refusal onto the tick report, so the JSON and the metrics say what the log says"*. The JSON half landed. **The metrics half did not**, so the commit message states something about the code that is not true.

## Repro

`rocky serve --scheduler` with `.rocky/pending-demands` present but unreadable — a dangling symlink is the easiest way to produce it.

## Trace

| Step | Location | What happens |
|---|---|---|
| Producer | `engine/crates/rocky-core/src/schedule/reconcile.rs:685` | the reconciler correctly records `report.spool_unreadable` |
| Counting | `engine/crates/rocky-cli/src/commands/scheduler/mod.rs:381` | scheduler counts include only `report.skipped` |
| Classification | `scheduler/mod.rs:408` | the tick is classified `Completed` **unconditionally** |
| OTLP | `scheduler/mod.rs:461` | metric recording iterates only `report.skipped` |
| Vocabulary | `engine/crates/rocky-observe/src/span_attrs.rs:179` | the bounded reason vocabulary has no `spool_unreadable` |

## Wrong result

```
rocky.scheduler.ticks{outcome="completed"}   ++
rocky.scheduler.skipped{reason="spool_unreadable"}   never emitted
```

In the actual metrics, a scheduler that cannot read its webhook spool is **indistinguishable from a healthy idle tick**. Only the trace span carries a boolean, and a boolean on a span is not what an alert reads.

This matters more than the JSON half it shipped with: the JSON is read by a person who already suspects something. The metric is what would have told them.

## Why the existing tests do not catch it

#1752's JSON and report tests never invoke `record_metrics`. There is no OTLP test for this state at all. They pass, and the merge message's claim about "the metrics" stays false.

## Suggested shape

Add `spool_unreadable` to the bounded reason vocabulary at `span_attrs.rs:179`, count it alongside `skipped`, and stop classifying the tick `Completed` when the spool could not be read.

The test has to drive `record_metrics` and assert on the emitted counter — a report-level assertion would pass without the wiring, which is exactly how this got through.

hugocorreia90:
Closes #1812.

`rocky serve --scheduler` with an unreadable webhook spool (a dangling symlink at `.rocky/pending-demands` is the easiest way) reported a healthy tick in its metrics. #1752 carried the refusal onto the tick's JSON and its trace span; the OTLP counters did not follow, so `rocky.scheduler.ticks{outcome="completed"}` kept climbing and no `rocky.scheduler.skipped` series ever named the spool. The metric is what an alert reads; the JSON is what a person reads once they already suspect something.

## What changed

- **Its own outcome.** Such a tick now counts under `outcome="spool_unreadable"`, a new value in the bounded vocabulary, instead of `completed`. It is not `fault` either: the cron half of the tick ran, and the counts still describe it.
- **Its own skip.** One `reason="spool_unreadable"` skip per such tick, the entry `rocky tick --output json` already lists for the whole webhook source. The span's and the log's `skipped` count it too, so the sum of the metric's reasons equals the JSON's list, as the existing invariant says it must.
- **One string.** `span_attrs::SPOOL_UNREADABLE_LABEL` is the constant behind the outcome, the reason and the JSON entry, so the three surfaces cannot drift again.
- The observability guide lists both new values (and the two skip reasons it had fallen behind on: `paused`, `config_error`).

## Tests

- `an_unreadable_spool_counts_under_its_own_outcome_and_reason` drives the resident scheduler's `record_metrics` with a dangling symlink planted before the first tick and asserts on the exported points: every `.ticks` point carries the new outcome, none `completed`; a `spool_unreadable` skip series with one increment per tick; the cron occurrence still executed. A report-level assertion would have passed without the wiring, which is exactly how #1752 shipped.
- The enumeration guard and the vocabulary tests pin the new value in both lists.


hugocorreia90:
## Independent red-team disposition — round one: NOT FULLY CLOSED, all three accepted and fixed

**Reviewer: Codex (GPT-5), 55 tool calls on `ed76f7c4..f5f67fa5`, read-only; it ran the observe and scheduler/tick suites and the core state-busy test.** It confirmed `spool_unreadable` is the right outcome (not `fault`: the tick's schedules still ran), that the test reaches the real spool scan with `dry_run=false`, that dry-run cannot synthesise the state, and that the one-way pinning test's intent survives.

### HIGH — the shipped Grafana alert selected only `config_error|fault`. Accepted, fixed.

The new outcome would have raised nothing: the provisioned rule's expression is exhaustive over the old vocabulary and `noDataState: OK` keeps it quiet. A dedicated rule, `rocky-scheduler-spool-unreadable`, now fires on the new outcome with the same window and `for` as tick-errors — separate on purpose, because cron/after/freshness work may still be running and the tick-errors message says the loop is doing nothing. Both dashboard descriptions list the current vocabularies (the reason list had also fallen behind by `paused` and `config_error`).

### MEDIUM — the "JSON, span, log and metric sums agree" claim was false for the whole-pass `state_busy`. Accepted, fixed.

Pre-existing, and my comment made the false claim explicit. `synthesized_skips` is now the one derivation of the skips a tick synthesises for a whole source or the whole pass; the JSON builder, `tick_counts` and `record_metrics` all read it. A metrics test holds the store across a tick and finds one `state_busy` skip series; a unit test pins each case, including that a pipeline-scoped `state_busy` already recorded is not doubled.

### LOW — reverting the span/log count left every test green. Accepted, closed.

A span test plants the dangling symlink and reads the closed `scheduler.tick` span: outcome `spool_unreadable`, `rocky.scheduler.skipped = 1`, the boolean `true`. The mutation that drops the count from `tick_counts` now fails it.

Also taken: the observability guide's "cron schedules" → "configured schedules (cron, `after`, freshness)", in the doc, the vocabulary comment and the changelog.

Two mutation checks stood before this round; two more on it. Round two dispatched on the second commit.

hugocorreia90:
## Independent red-team disposition — round two: NOT FULLY CLOSED, both accepted and fixed

**Reviewer: Codex (GPT-5), 46 tool calls on `f5f67fa5..4e2925c3` and the whole PR; its second reviewer (`gpt-5.6-terra`) confirmed both.** It verified the alert's UID, nesting, window and series name against the shipped Alloy config, the JSON's byte-for-byte equivalence across the four cases, no double count, and no first-party consumer keying on `source == "webhook"`.

### MEDIUM — the alert promised that deliveries were spooled while the spool was unreadable. Accepted, fixed.

They are not: `accept_journaled` fails at `create_dir_all` and the webhook endpoint answers HTTP 500, so a delivery during the fault is refused, not spooled. The rule's summary and description now say so, and that a refused delivery has to be retried by its sender.

### LOW — the whole-pass `state_busy` was tested only at the metric sink. Accepted, closed.

Two consumer-bound tests: `build_tick_output` with a held-store report yields one pipeline-less `state_busy` entry and `counts.skipped == 1`, and is not doubled beside a pipeline-scoped one; a tick over a store held across it closes a `scheduler.tick` span with `state_busy = true` and `skipped = "1"`. Restoring the old count formula fails the second; dropping the `state_busy` arm of the shared derivation fails both.

Four mutation checks stood before this round; two more on it. Round three dispatched on the third commit.

hugocorreia90:
## Independent red-team disposition — round three: NOT FULLY CLOSED on one wording, now fixed

**Reviewer: Codex (GPT-5), 38 tool calls on `4e2925c3..883ece00` and the whole PR; it ran both suites and hand-traced four mutations against the new tests.** It confirmed the round-two test gap closed at both consumers (the JSON and the closed span), the alert YAML's six unique UIDs, and no regression.

### MEDIUM — "HTTP 500 means not spooled, so retry" was unsafe. Accepted, fixed.

The reviewer's trace: acceptance links the pending file before the directory fsync, so a 500 can follow a demand that IS on disk; after repair the tick runs it, and a body-keyed retry (no `X-Rocky-Delivery`) runs it again. The rule now says a 500 leaves persistence uncertain, that the retry must carry the same delivery id, and that eligible pending demands resume after repair (a paused pipeline's stay pending). Nothing else in the PR claims otherwise.

Six mutation checks stand on this PR. Round four dispatched on the fourth (wording-only) commit.

hugocorreia90:
## Independent red-team disposition — round four: NOT FULLY CLOSED on the alert's prose; reduced to what the reconciler keeps

**Reviewer: Codex (GPT-5), 41 tool calls on `883ece00..7cdad01d` and the whole PR; sentence-by-sentence against the accept path, the status mapping, the tombstones, the pause gate and the initial state contention.** It confirmed the persistence correction, the alert condition, the provisioning structure, and the round-three test closures.

### MEDIUM 1 — "a paused pipeline's demands stay pending until resumed" overstated the hold. Accepted; wording reduced, the code defect filed.

True only when the pause cursor is readable: the reconciler reads it as `.ok().flatten()`, so a read error is "not paused" and a transient fault lets a paused pipeline's demand run. That is a fail-open in the reconciler's own gate, out of this PR's scope and filed as #1831 with the reviewer's trace and the test it needs. The alert now says only "subject to each pipeline's pause state".

### MEDIUM 2 — "eligible pending demands resume on the next tick" missed the initial state contention. Accepted, fixed.

A tick that finds the store held returns before the spool scan and counts as `completed`, so the alert can clear while nothing has inspected the spool. The wording is now the reviewer's: consumed by the next tick that opens the state store and reaches webhook consumption.

### LOW — "retry with the same id" omitted its prerequisite. Accepted, fixed.

A request that carried no `X-Rocky-Delivery` cannot be retried by id; adding one changes the dedup key. The rule now says: an original that carried the header retries with that exact id (deduplicated for 24 hours); one that carried none is at-least-once, and a retry can run the demand twice.

Six mutation checks stand on this PR; this round is wording only. Round five dispatched on the fifth commit for a final sentence-by-sentence pass.

hugocorreia90:
## Independent red-team disposition — round five: two wording reopens, both accepted

**Reviewer: Codex (GPT-5), 41 tool calls on `7cdad01d..e49cb1b5` and the whole PR.** The retry sentence PASSED. The recovery sentence was reopened twice over: "subject to each pipeline's pause state" still implied an enforced gate the reconciler does not guarantee (#1831), and "the next tick that … reaches webhook consumption" was still absolute — a store taken by a third process between two demands' children defers the rest to a later tick.

Both accepted. The sentence now promises nothing the code does not keep: *webhook consumption resumes on a later tick, and any one spooled demand can still be deferred by the reconciler's own gates.* Round six dispatched for the final sentence-by-sentence pass on the sixth (wording-only) commit.

hugocorreia90:
## Independent red-team disposition — round six: the recovery sentence PASSED; three more prose claims, all accepted

**Reviewer: Codex (GPT-5), 79 tool calls on `e49cb1b5..05273e4a` and the whole PR; its second reviewer (`gpt-5.6-terra`) found the same three.** Every code-level closure from rounds one to five was re-verified at this head; the six UIDs, the datasource and the provisioning mount were checked.

- **MEDIUM — the rule fires on a rate, so "are not being consumed" was absolute where the signal is not.** A scan failing once a minute with healthy ticks between keeps `rate[2m] > 0` for five minutes while those ticks consume demands. Accepted: "has repeatedly failed", "may be going unconsumed", and "a tick in between whose scan succeeds may".
- **LOW — `spool_unreadable` does not prove something is at the path.** Only the initial `NotFound` goes through the presence discriminator; any other scan error (descriptor exhaustion) sets the same flag. Accepted: "usually something is at that path … a scan can also fail for a reason unrelated to the path, so read the error the scheduler log names", and "fix the cause the log names" instead of "repair the spool path".
- **LOW — an empty or invalid `X-Rocky-Delivery` is body-keyed.** Accepted: "a non-empty, valid X-Rocky-Delivery"; empty or invalid values grouped with the headerless case.

This is the seventh commit and the fourth wording-only one. Round seven dispatched for a final sentence-by-sentence pass; the code and its six mutation checks have not changed since round three.

hugocorreia90:
## Independent red-team disposition — round seven: PASS (with one nit, taken)

**Reviewer: Codex (GPT-5), 39 tool calls on `05273e4a..14e8f474` and the whole PR.** Every sentence of the rule's summary and description now passes against the code: the rate-over-window semantics, the scan errors that do not prove path presence, the exact scheduler log line that names the error, the HTTP 500 uncertainty, the non-empty valid delivery id, the scheduled work that still runs, and the recovery sentence. All prior closures hold. Six unique UIDs.

The one nit — the YAML comment above the rule still said "every tick" — is taken in a comment-only eighth commit.

## Where this PR ended up — seven rounds, on the record

|---|---|---|
| 1 | NOT FULLY CLOSED | the shipped alert selected only `config_error\|fault`; the JSON/span/metric count invariant was false for the whole-pass `state_busy`; the span count untested |
| 2 | NOT FULLY CLOSED | the alert claimed deliveries were spooled during the fault; whole-pass `state_busy` tested only at the metric sink |
| 3 | NOT FULLY CLOSED | "not spooled, retry" was unsafe: a 500 can follow the link, and a body-keyed retry runs twice |
| 4 | NOT FULLY CLOSED | the pause hold overstated (the reconciler's pause gate fails open — filed #1831); "next tick" missed store contention; the retry id's prerequisite |
| 5 | NOT FULLY CLOSED | the recovery sentence still promised a gate and a tick |
| 6 | MIXED | a rate-based rule can hold while healthy ticks consume; `spool_unreadable` does not prove path presence; an empty id is body-keyed |
| 7 | **PASS** | one stale comment |

The code (rounds one to three) has not changed since the third commit; six mutation checks stand on it. Rounds four to seven were the alert's prose, sentence by sentence, until it claimed nothing the reconciler does not keep. Merging on green.

PR #1832, opened 2026-09-09
fix(engine): quality gate rides check_gate_failed, NaN threshold is refused, the sample deadline fires mid-compile

### LINKED ISSUE #1816: Three from the red-team pass: quality JSON corrupts tables_failed, NaN disables anomaly detection, the sample timeout cannot fire
Three smaller findings from the same independent red-team pass, kept together because none of them warrants its own thread and none should be lost.

## 1. `rocky quality --output json` corrupts two public fields to carry a failed gate

`finalize_quality_status` at `engine/crates/rocky-cli/src/commands/run_local.rs:499` writes the **failed-check count** into `tables_failed`. The public contract for that field says it counts tables/models and *"never counts checks"* — `engine/crates/rocky-cli/src/output.rs:243` and `output.rs:261`.

`check_gate_failed` is the field built for this, and `derive_run_status` consumes it at `output.rs:5379` — but the quality path never sets it. The persisted `RunRecord` receives the false gate value at `output.rs:5325`.

One table with three failing error-severity checks reports:

```
tables_failed:     3     <- one table
check_gate_failed: false <- the gate DID fail
status:            failure   (correct, by accident)
```

The overall status is right; the two fields a consumer would read to understand *why* are both wrong, and the persisted record carries the wrong gate flag forward.

**Why the test misses it:** `a_failed_quality_gate_is_stamped_onto_the_payload_not_just_the_record` uses exactly one table and one check, so the two counts coincidentally equal 1, and it asserts the wrong carrier at `run_local.rs:2597`. It also calls the helper directly, so moving the production call back below JSON emission would not fail it.

Related: #1795, #1788.

## 2. `anomaly_threshold_pct = nan` silently disables detection and calls the result normal

Config accepts an unrestricted `f64` at `engine/crates/rocky-core/src/config.rs:1212`; nothing validates finiteness, and TOML accepts `nan`.

At `engine/crates/rocky-core/src/state.rs:6418` both `threshold_pct > 0.0` and `threshold_pct <= 0.0` are **false** for NaN. The result is `is_anomaly = false` and the message *"within normal range"* at `state.rs:6432`. The caller only emits metrics, hooks and output when `is_anomaly` is true (`engine/crates/rocky-cli/src/commands/run.rs:6517`).

A 400% row-count spike against a flat history produces no anomaly, no warning, no metric, no hook — and is described as normal.

#1793 made `0` mean "disabled" deliberately. NaN reaches the same disabled state through the *else* of both branches, and unlike `0` it was never a choice anyone made. The tests at `state.rs:7203` cover zero, negative and finite-positive only.

**Suggested shape:** reject a non-finite `anomaly_threshold_pct` at config load, the way an out-of-range value would be. Silently disabling detection is the one outcome it must not have.

Related: #1793.

## 3. The 30-second sample timeout cannot pre-empt the compile it wraps

`engine/crates/rocky-cli/src/api.rs:1702` wraps `compute_preview_rows` in a 30s timeout. Inside that future, config loading and the **full compiler run** are synchronous — `engine/crates/rocky-cli/src/commands/preview_rows.rs:133` and `:175`. The first `await` on warehouse work is not until `preview_rows.rs:330`.

A Tokio deadline cannot fire while an inner poll is blocked on synchronous work. On a model tree whose compile exceeds 30 seconds, `GET /api/v1/models/{name}/rows` returns **no `504 sample_timeout`** at 30s; the request and a runtime worker stay occupied until compilation returns.

**Why the tests miss it:** `a_second_concurrent_sample_is_refused_immediately_with_retry_after` holds the semaphore by hand at `api.rs:3319` and never drives the timed core. The end-to-end parity test at `engine/rocky/tests/serve_samples_parity.rs:191` claims to pass a consent header but repeats the headerless helper call.

**Suggested shape:** move the compile onto `spawn_blocking` so the deadline can be observed, or apply the timeout to a stage the runtime can actually interrupt.

Related: #1733.

---

## Clean in this lane

No product defect found after producer-to-sink reading of #1742, #1754, #1755, #1762, #1768, #1780, #1794.

One evidence caveat: `an_unreadable_spool_is_a_non_zero_tick` at `engine/crates/rocky-cli/src/commands/tick.rs:743` tests `tick_verdict` only. It would still pass if `run_tick` stopped mapping that verdict to `Err` at `tick.rs:163`.

hugocorreia90:
Closes #1816. Three findings from the independent red-team pass, one PR, each fix pinned by a test that fails when the fix is reverted.

## 1. A quality run's failed gate rides `check_gate_failed`, and `tables_failed` stays a count of tables

`finalize_quality_status` wrote the failed-**check** count into `tables_failed`, whose contract says it never counts checks, and left `check_gate_failed` false. One table with three failing checks reported `tables_failed: 3` and no gate. `dagster-rocky` compares `tables_failed` with the itemised `errors` and read three unattributable failed tables. `status` came out right by accident.

```text
before                          after
  tables_failed:     3            tables_failed:     0
  check_gate_failed: false        check_gate_failed: true
  status:            failure      status:            failure
```

The helper now sets the gate and leaves the count alone. The persisted record inherits the gate. Two tests:

- a unit test with one table and three failing checks (the old test used one of each, so the two numbers coincided at 1);
- `rocky/tests/quality_check_gate.rs`, which drives the real binary with two failing checks on one table and reads the emitted payload, so the carrier **and the emit order** (#1788) are pinned from outside. Moving the call below the emit fails it (mutation m2).

**Remote state (rounds two and three).** The gate failure is a typed `QualityGateFailure` (same message, same exit 1) and the dispatcher's quality arm finalizes the remote-state session on it, as the transformation arm does for `RunFailed`; before, every error abandoned the session, so under S3 / GCS / Valkey / tiered state the record never left the pod-local file. The typed error is returned only when `persist_run_record` reports the record landed; on a failed write the run fails untyped, the message says the run is not in the history, and the dispatcher abandons, so no ledger without the run is published. Two-pod in-memory S3 tests drive the real dispatcher through both paths (mutations m7, m10).

## 2. `anomaly_threshold_pct = nan` is refused at load

TOML accepts `nan` and `inf`. Every comparison with NaN is false, so a NaN threshold flagged nothing and described a 400% spike as "within normal range". `+inf` reached the same place because no deviation exceeds it.

- `validate_checks`, a new entry in `CONFIG_VALIDATORS`: a non-finite threshold on any pipeline is `ConfigError::ChecksAnomalyThresholdNotFinite { pipeline, value }`, with a message that names the `0` off switch. `rocky validate` maps it to `pipeline.<name>.checks.anomaly_threshold_pct` (V046).
- `detect_anomaly` fails closed on a non-finite threshold for a caller that bypasses loading: `is_anomaly: true`, reason names the threshold, never "within normal range". Config loading makes this unreachable from `rocky.toml`; it is the defence for programmatic callers.
- `0`, negative, the default, and a `[checks]` table that does not name the threshold all still load (pinned).

## 3. The sample deadline fires during the compile

`GET /api/v1/models/{name}/rows` wrapped `compute_preview_rows` in a 30s timeout, but config loading and the compiler run inside it are synchronous. A Tokio deadline is checked between polls, so on a slow compile the route never answered `504 sample_timeout` at 30s: the request, a worker and the one-sample permit stayed held until the compile returned.

- The blocking stage (`prepare_preview`: config, gate, compile, admit, masking, SQL, adapter) runs on `spawn_blocking`. The warehouse stage (`ping`, `execute_query`) stays async and is dropped with the future.
- **The permit travels into the blocking task.** A compile cannot be cancelled; it finishes on its own. If the permit were released when the route stopped waiting, a retry would compile beside the orphaned compile. Now a second sample is `503 engine_busy` until the orphan returns (pinned: mutation m6 releases it early and the test fails on the 503).
- The deadline is per `ServerState` (`DEFAULT_SAMPLE_TIMEOUT` 30s, `set_sample_timeout` for tests). A `#[cfg(test)]` hook keyed by config path holds the blocking stage; the clock discriminates (1500ms hold, 100ms deadline, answer must arrive before the hold ends). Inlining the stage again (mutation m5) fails it.
- `compute_preview_rows` takes the permit as `Option<OwnedSemaphorePermit>`; the CLI passes `None`. Same core, same gate, same masking, same SQL for both callers.
- Panic on the blocking pool is `internal_error` → 500, not the 502 the fallback arm gives an adapter.
- The parity test's "with consent" check really sends `X-Rocky-Allow-Warehouse: true` now, and asserts the answer equals the headerless one.

## Evidence

- `cargo test -p rocky-core --lib`: 2176 passed. `-p rocky-cli --lib`: 1900 passed. `-p rocky-server --lib`: 103 passed. `-p rocky --test quality_check_gate --test serve_samples_parity --test serve_review_parity --test serve_governor_parity --test replication_check_gate`: all pass.
- `cargo clippy --all-targets -D warnings` on the four crates and `cargo fmt --check`: clean.
- Eleven mutation checks via `scripts/mutation-check.sh`, all exit 0 (the test fails under mutation):

| mutation | reverts | failing test |
|---|---|---|
| m1-unit | count into `tables_failed` instead of the gate | `a_failed_quality_gate_rides_check_gate_failed_not_the_table_count` |
| m1-e2e | same | `quality_check_gate::a_failed_quality_gate_rides_check_gate_failed_on_the_wire` |
| m2-order | finalize after the JSON emit | same end-to-end test |
| m3-validator | drop `validate_checks` from the chain | `a_non_finite_anomaly_threshold_is_rejected_at_load` |
| m4-detect | drop the non-finite branch in `detect_anomaly` | `a_non_finite_threshold_is_never_called_normal` |
| m5-inline | blocking stage inline in the async task | `the_sample_deadline_fires_during_a_slow_compile_and_the_permit_outlives_it` (answered after the hold) |
| m6-permit | permit held by the future, not the blocking task | same test (second sample not refused) |
| m7-abandon | quality arm abandons on the typed gate failure | `a_failed_quality_gate_is_uploaded_to_remote_state_not_abandoned` (pod B finds no record) |
| m8-field | drop the `rocky validate` field mapping | `a_non_finite_anomaly_threshold_is_rejected_v046_at_its_key` |
| m9-consent | send the real-server consent request headerless | `consent_header_reaches_the_gate_on_a_remote_adapter` |
| m10-unrecorded | typed gate failure returned when the record did not land | `a_failed_quality_gate_whose_record_did_not_land_is_not_uploaded_as_if_it_had` |

## Not in this PR

- `rocky history` does not project `check_gate_failed` on its run records; the end-to-end test reads `status` from history and the unit test pins the record's gate.
- The evidence caveat in #1816 about `an_unreadable_spool_is_a_non_zero_tick` (tests `tick_verdict` only) is unchanged.
- No output schema changed: `check_gate_failed` already existed on `RunOutput`, so no fixture regeneration. The project-config schema did change: the doc comment on `anomaly_threshold_pct` is its description, regenerated in `449a53c3`.
- #1836: the transformation, snapshot and load arms finalize on their typed sentinels (or on `Ok`) without checking that the record write succeeded, the same gap round two found on the quality arm. Scope decision on the record: quality only here.

## Least confident / most likely missing

- **Least confident:** failing closed in `detect_anomaly` on a non-finite threshold means a programmatic caller with NaN gets an anomaly hook, a metric and a warning for a misconfiguration rather than for data. The alternative (`is_anomaly: false` with an honest reason) is invisible, because the caller only surfaces `is_anomaly == true`. Config loading makes the question moot for every `rocky.toml`; the in-function choice is the fail-closed one.
- **Most likely missing:** a timeout that fires while the *warehouse* stage runs still drops the future and releases the permit while the query may keep running on the warehouse (pre-existing, documented). The permit-with-the-compile rule covers the stage that cannot be cancelled; it does not extend the lane to a query the adapter cannot cancel either.

Independent red team: Codex (GPT-5) with a second reviewer each round; dispositions are the comments below (round one: one MEDIUM accepted, two LOWs accepted, one LOW rebutted; round two: one MEDIUM and one LOW accepted).


hugocorreia90:
## Independent red-team disposition — round one: NOT FULLY CLOSED, one MEDIUM and two LOWs accepted, one LOW rebutted

**Reviewer: Codex (GPT-5), 103 tool calls on `46c53ff7..b599cfa7`; its second reviewer (`gpt-5.6-terra`) concurred on every item.** The three fixes were verified on their exercised paths; the finding was in what the fixes did not reach.

- **MEDIUM — a failed quality gate was abandoned, not uploaded, under remote state. Accepted; fixed in `41b1b1fc`.** `run_quality` persisted its `Failure` record and then `bail!`ed with an untyped error, and the dispatcher's quality arm abandoned the remote-state session on every error, which uploads nothing. Under an S3 / GCS / Valkey / tiered `[state]` backend the record never left the pod-local file; a fresh pod downloading the shared state saw the history from before the run. The gate failure is now a typed `QualityGateFailure` (same message, same exit 1), and the quality arm finalizes on it exactly as the transformation arm does for `RunFailed`, abandoning only on a pre-terminal error. The test the reviewer asked for exists: two pods over one in-memory S3, pod A runs the real dispatcher on a failing quality pipeline, pod B downloads into its own empty store and finds the `Failure` record with `check_gate_failed`. Mutation m7 (revert the branch to abandon) fails it on the pod-B read.
- **LOW — no test pinned the `rocky validate` field mapping. Accepted.** `a_non_finite_anomaly_threshold_is_rejected_v046_at_its_key` runs `validate_inner` on a `nan` config and asserts one V046 at `pipeline.dq.checks.anomaly_threshold_pct`. Mutation m8 (drop the mapping arm) fails it.
- **LOW — the parity test's consent check proved nothing on DuckDB. Accepted.** A second real-server test on a Databricks-configured project: headerless is `403 warehouse_gated`; with the header the request is past the gate, at `422 compile_error` (the project has no models). Mutation m9 (send it headerless) fails it.
- **LOW — the ninth `compute_preview_rows` argument breaks the crate's public Rust signature. Rebutted.** `rocky-cli` is the binary's internal crate, and the repository's standing rule before the 2.0 relaunch is no backward-compatibility wrappers, shims or aliases: contracts change in place. No eight-argument wrapper is added. The reviewer's own framing allowed a downgrade to NIT given such a declaration; this is it.

The reviewer's two answers, taken: the hidden assumption was that every `run_quality` error is pre-terminal, and the missing thing was a remote-state test that reads the record through a fresh session. Both are now in the tree. Round two dispatched on `41b1b1fc`.

hugocorreia90:
## Independent red-team disposition — round two: NOT FULLY CLOSED, one MEDIUM and one LOW, both accepted

**Reviewer: Codex (GPT-5), 50 tool calls on `41b1b1fc`; its second reviewer (Wegener) confirmed both findings.** Every round-one closure was re-verified at this head, and the `compute_preview_rows` rebuttal was accepted.

- **MEDIUM — the typed sentinel did not prove the record had landed. Accepted; fixed in round three.** `persist_run_record` is best-effort: a failed `record_run` (full disk, failed commit) is logged and the run goes on. So a run whose gate failed AND whose record write failed still returned `QualityGateFailure`, the dispatcher finalized on the type, and the upload published a ledger without the run — a fresh pod read an authoritative history with the failure missing. Now `persist_run_record` reports whether the record landed, and the quality path returns the typed error only when it did. Otherwise the run still fails, untyped, with a message that says the run is not in the history, and the dispatcher abandons: nothing is uploaded, nothing claims the record exists. The test the reviewer asked for exists: a test-only hook fails the write for one run id, the two-pod test drives the real dispatcher through it and asserts no typed error, no local record, and no remote object (pod B's download is `FreshStart`, not `Authoritative`). Mutation m10 (return the typed error regardless) fails it on both the type and the authority.
  - **Scope decision, explicit as asked:** quality only. The transformation arm's `RunFailed` has the same latent gap — its status-derived sentinel does not check the write either — and closing it means making the shared helper's every caller decide what a lost record means for exit code and upload. Filed as a residual rather than folded in here.
- **LOW — the remote-state test raced on the process working directory. Accepted.** The DuckDB path in the test config is absolute now and the test never calls `set_current_dir`; the shared driver says why.

The reviewer's two answers, taken: least confident was the repair boundary (decided above, on the record), and the missing thing was proof that `QualityGateFailure` is impossible without a committed record (the new test). Round three dispatched.

hugocorreia90:
## Independent red-team disposition — round three: PASS WITH NITS (four nits, three taken)

**Reviewer: Codex (GPT-5), 43 tool calls on `7ccd3d62`; its second reviewer (Pasteur, `gpt-5.6-terra`) found no missed blocker.** The round-three invariant was verified end to end: the typed gate failure only when the record landed, the untyped fallback otherwise, finalize only on the type, abandon on everything else, fail-closed under governed / suppressed / CAS postures, exit 1 on both paths, JSON already emitted with the right fields, and no periodic uploader on the quality arm. Every earlier closure was re-verified at this head, and the scope decision (#1836 for the other arms) was accepted.

- **NIT — the write-fault test injects before `record_run`, so a real redb commit failure is not exercised. Not taken here.** The injection returns the same `StateError` the helper's `Err` arm handles, and the reviewer found no control flow that would treat a commit failure differently. A commit-failure seam in the state store is #1836's test to build, since that is where the other arms' fix will need it too.
- **NIT — the test config interpolated a path into a TOML basic string, which a Windows path would break. Taken.** A literal string now.
- **NIT — the helper doc said callers "must check" the returned `bool`, which the transformation, snapshot and load arms do not. Taken.** The doc now says who checks it, who does not (#1836), and that nothing forces the check.
- **NIT — the sentinel doc said the message was unchanged; it gained the run id. Taken.** No first-party reader matches the message text; the SDK and Dagster read the JSON fields.

The reviewer's two answers, on the record: least confident is the synthetic-versus-real write failure (above, #1836); most important missing is an API that forces callers to consume the persistence outcome (an outcome enum or `Result`), which is also #1836's shape question. Nits committed as a doc-and-test-only change; merging on green.

hugocorreia90:
Main moved under this PR by four sibling merges (#1834, #1835, #1837, #1838) that touch `api.rs`, `commands/mod.rs`, `run.rs` and `rocky-server/src/state.rs`. Merged `origin/main` in (`68cd0cf9`, no conflicts) and re-gated the merged tree locally: clippy `-D warnings` and fmt clean; `rocky-server` 107 and `rocky-cli` 1905 unit tests pass; the quality-gate, replication-gate, serve parity, shadow-ownership and serve-ui end-to-end tests pass. No code decision in the merge commit; merging on green.

Issue #1632, opened 2026-09-03
content reuse: a bare read is bound to one model even when two models write a table of that name


hugocorreia90:
Found while grounding #1354 (PR #1628).

`build_reuse_target_by_model` (`rocky-cli/src/commands/run.rs`) gives each model a bare-name key. PR #1628 narrowed that to models whose configured target table is spelled like the model, and made the others poison the key. One gap stays open.

Shape:

```
orders  [target] cat.s1.orders      -> contributes the bare key "orders"
other   [target] cat.s2.orders      -> contributes "other" only
consumer (content-addressed): SELECT id FROM orders
```

Both models write a table called `orders`, in different schemas. Which one a bare `FROM orders` reads depends on the connection's search path, which Rocky does not observe. The two never claim the same key, so the poisoning guard never fires, and the consumer is bound to `cat.s1.orders`'s column hashes.

If `cat.s2.orders` (or an external `orders`) changes while `cat.s1.orders` does not, the consumer can SKIP with stale input.

Suggested direction: poison the bare key whenever more than one materialized target shares its folded table component — the same rule the existing case-collision poisoning uses, applied to the table component instead of the model name. Fail-closed: the read then records no baseline and the gate builds.

Pre-existing; #1628 neither introduced nor worsened it (verified by the red team across two rounds).

hugocorreia90:
Kept as its own decision, not folded into #1354: content-reuse keys are built independently of the compiler edge (`build_reuse_target_by_model` in `run.rs`), so resolving #1354 leaves this producer unchanged.

Issue #1631, opened 2026-09-03
typecheck: SELECT * expansion binds a bare name to a model independently of the DAG edge


hugocorreia90:
Found while grounding #1354 (PR #1628).

`semantic.rs` expands a `SELECT *` by looking the read's table name up in the model map directly (`models.get(table_name)`), not through the model's `upstream` set. So the star path binds by NAME even where the edge does not, and it would keep binding if the edge rule ever changed.

Two consequences today:

1. A `SELECT * FROM customers` reader takes its columns from model `customers` even when that model's configured target is `prod.customers_v2` and the warehouse read lands on a different physical table.
2. It checks the model map before `source_schemas`, so a genuinely external table can be typed with a model's columns.

Not fixed in #1628 on purpose. Making the read external would empty the reader's output schema, and a contract on that reader would then fail with **E010** — a compile that passes today would refuse. That is a behaviour change needing its own decision, and it is downstream of the #1354 question about which execution semantics the graph encodes.

Related: dropping an edge also reorders the `project.execution_order` walk this expansion depends on, which is one of the two reasons #1628 did not drop it.

hugocorreia90:
Tracked as a child of #1354's resolver decision: the same choice, but this issue has its own producer that must change and be tested independently — a decision on #1354 does not close it. See the sequencing note on #1354 (local execution must honour configured targets first; `executor.rs:54`).

hugocorreia90:
## Unblocked: #1354's resolver decision is made

**Decided: local execution honours configured targets.** `execute_locally` will materialize at the model's configured `[target]` instead of `CREATE OR REPLACE TABLE {model_name}` (`executor.rs:54`), and bare reads then resolve by physical target — #1354 option A.

The sequencing recorded on #1354 stands and is not reversible:

```
1.  local execution honours configured targets   (own behaviour change, owner-merge)
2.  bare reads resolve by physical target        (#1354 option A)
3.  this issue's own producer                    (needs its own change and test)
```

Per the 09-04 note on this issue, the decision does not close it: this issue has its own producer that must change and be tested independently. Dropping `blocked-decision`; the remaining work is implementation, gated on steps 1 and 2 landing first.

hugocorreia90:
## Not gated on #1354 step 2 — step 2 is gated on **this**.

Correcting a gate I recorded and repeated.

PR #1840's `resolve.rs:193-196` says it outright:

[...]

Issue #1630, opened 2026-09-03
compiler: an ephemeral model's phantom edge can suppress a real physical-read edge


hugocorreia90:
Found while grounding #1354 (PR #1628).

An ephemeral model materializes nothing. `run.rs` says so outright where shadow mode refuses it: nothing in `rocky-compiler` or `rocky-sql` rewrites a consumer's `FROM eph` into a CTE, so a consumer reads whatever physical table carries that name.

`rocky_core::physical_edges` already acts on this — `PhysicalEdgeModel::materializes` keeps a non-materializing model out of the producer index, with a comment explaining that a phantom producer could suppress a reader's true edges through the cycle guard.

`resolve_dependencies` does not. A bare read of an ephemeral model's name still becomes a compile edge, and that edge enters `existing` for the run-time derivation.

Repro shape (consumer -> producer):

```
a   materialized, reads db.s.c   (a true physical read of c's target)
e   EPHEMERAL,    reads bare a   => compile edge e -> a
c   materialized, reads bare e   => phantom compile edge c -> e
```

The true `a -> c` edge now closes a cycle through `c -> e -> a`, so the derivation skips it. Under concurrency `a` can read stale `c`.

Note the interaction with #1354: `rocky test`/`rocky ci` go through `execute_locally`, which materializes every model — ephemeral included — as `CREATE OR REPLACE TABLE <model name>`. So the edge is real there and phantom on a warehouse run, the same split #1354 is about.

hugocorreia90:
Tracked as a child of #1354's resolver decision: the same choice, but this issue has its own producer that must change and be tested independently — a decision on #1354 does not close it. See the sequencing note on #1354 (local execution must honour configured targets first; `executor.rs:54`).

hugocorreia90:
## Unblocked: #1354's resolver decision is made

**Decided: local execution honours configured targets.** `execute_locally` will materialize at the model's configured `[target]` instead of `CREATE OR REPLACE TABLE {model_name}` (`executor.rs:54`), and bare reads then resolve by physical target — #1354 option A.

The sequencing recorded on #1354 stands and is not reversible:

```
1.  local execution honours configured targets   (own behaviour change, owner-merge)
2.  bare reads resolve by physical target        (#1354 option A)
3.  this issue's own producer                    (needs its own change and test)
```

[...]

Issue #1629, opened 2026-09-03
run --dag: label inference re-creates the bare-name edge the compiler derives, independently of the compiler


hugocorreia90:
Found while grounding #1354 (PR #1628).

`unified_dag::infer_runtime_dependencies` (`rocky-core/src/unified_dag.rs`) matches each model's reads against node **labels**. For a transformation node the label is the model name, and the match never looks at the model's `[target]`.

So `run --dag` builds the same bare-name edge `classify_table_ref` builds, from a second matcher. Two consequences:

1. Any fix to the compile-time rule has to be applied here too, or the two schedulers disagree — which is why #1628 did not attempt the narrow run-path fix.
2. The suppression #1354 reports happens here as well: the label edge makes a true physical-read edge look like a cycle-closer, and `derive_physical_edges` skips it.

Repro shape (no fix shipped, so this is current behaviour):

```
customers  [target] table = customers_v2, SQL: SELECT y FROM warehouse.silver.rollup
rollup     [target] table = rollup,       SQL: SELECT x FROM customers
```

Label inference adds `rollup -> customers`. Physical inference then finds the true `customers -> rollup` and skips it as cycle-closing. `customers` can run first and read stale `rollup`.

Blocked on the same decision as #1354: which execution semantics the graph should encode.

hugocorreia90:
Tracked as a child of #1354's resolver decision: the same choice, but this issue has its own producer that must change and be tested independently — a decision on #1354 does not close it. See the sequencing note on #1354 (local execution must honour configured targets first; `executor.rs:54`).

hugocorreia90:
## Unblocked: #1354's resolver decision is made

**Decided: local execution honours configured targets.** `execute_locally` will materialize at the model's configured `[target]` instead of `CREATE OR REPLACE TABLE {model_name}` (`executor.rs:54`), and bare reads then resolve by physical target — #1354 option A.

The sequencing recorded on #1354 stands and is not reversible:

```
1.  local execution honours configured targets   (own behaviour change, owner-merge)
2.  bare reads resolve by physical target        (#1354 option A)
3.  this issue's own producer                    (needs its own change and test)
```

Per the 09-04 note on this issue, the decision does not close it: this issue has its own producer that must change and be tested independently. Dropping `blocked-decision`; the remaining work is implementation, gated on steps 1 and 2 landing first.

hugocorreia90:
## Gating correction: most of this does not wait on step 1

[...]

PR #1840, opened 2026-09-09
feat(engine): local execution materializes at the configured target (#1354 step 1)

hugocorreia90:
Step 1 of the #1354 sequence. #1629, #1630 and #1631 are gated on it, and the decision comment marks it **owner-merge** — it is a real `rocky test` behaviour change, so it goes up on its own.

## The problem it removes

`execute_locally` built every model as `CREATE OR REPLACE TABLE <model name>`, ignoring `[target]` entirely. So `rocky test` and `rocky ci` agreed with a warehouse run about a model's *output* and disagreed about *which object holds it*.

That disagreement is load-bearing. Three sites keep a false compile-time edge specifically because of it — a bare `FROM customers` matching a model NAMED `customers` that writes `prod.customers_v2`. `resolve.rs` says so in as many words: *"the edge is kept: dropping it breaks `rocky test`, which materializes by name."* Step 2 cannot drop that edge while local execution props it up.

## Two changes, and neither works alone

```
  target schema + table    a model lands at its configured schema.table
  search_path              every target schema is on it, so a bare read
                           still reaches a producer in another schema
```

I measured both against the 155 models in the in-repo examples before picking:

**Schema-qualified alone breaks the dominant idiom.** 149 of 155 models declare a target schema, and **24 projects layer across two or more** (`staging` → `marts`/`analytics`/`gold`). They read each other by **bare name** — 43 distinct such reads. `FROM stg_orders` searches only the current schema, so a consumer in `marts` stops seeing a producer in `staging`. Without the search path, those models do not compile at all.

**Table-component-only silently clobbers.** `10-route-by-tenant` has three models — `events_acme`, `events_beta`, `events_ceres` — each targeting `events` in its own schema. Materializing at the table component alone makes them one table, last writer wins, no error. **`E036` does not catch this**: it keys on the full `catalog.schema.table`, and those three are genuinely distinct objects. `three_tenants_writing_one_table_name_land_in_three_schemas` is that fixture.

Together they give local execution the warehouse's own resolution rule, which is what #1354 asks for: a bare read resolves to a **physical table**, not to a model name.

## What this deliberately does not do

- **The catalog component is dropped.** DuckDB resolves a three-part name only against an `ATTACH`-ed database and nothing attaches one, so including it would fail every model in a project that names a catalog — most of them. A model whose SQL *reads* a three-part name (`FROM poc.prod__orders.orders`) is unaffected and still does not resolve locally, exactly as before. "Honours configured targets" here means schema and table, not full target fidelity.
- **Cross-schema ambiguity stays silent.** With two schemas on the path holding the same table name, DuckDB takes the first and says nothing. That is not papered over — it is what a warehouse does with an unqualified read, so local execution now mirrors it instead of hiding it behind one flat namespace. Which model a bare read binds to when two could answer is **#1632**. Path order is sorted so the answer is at least stable across runs.
- **The false edge is still there.** Its justification is gone, but dropping it reorders `SELECT *` expansion, which is #1631 and needs its own change and test. The three comments that cited local execution as the reason are updated to say what actually holds the edge now.

## A bug found by writing the tests

Setting `search_path` changes where an **unqualified `CREATE TABLE`** lands, not just where reads resolve. A model with no configured schema was being created inside whichever target schema sorted first — another model's. `local_relation` therefore always qualifies, naming `main` explicitly, and `a_model_with_no_schema_is_not_captured_by_another_models_schema` pins it with a schema (`aaa_first_on_path`) that sorts ahead of the default.

A second test I wrote turned out to be unreachable: model loading already refuses an unsubstituted `${VAR}` in a sidecar (`ModelLoadError::EnvSubstitution`), so a placeholder target never reaches a compile result. The identifier validation stays as defence in depth on a function that interpolates into SQL, and the test now drives `local_relation` directly and says it is not a live path — rather than implying a guard that fires.

## Evidence

Six new tests, all asserting against the **database** rather than the generated SQL string. Both halves are independently **mutation-checked** with `scripts/mutation-check.sh`:

| mutation | result |
|---|---|
| revert to `CREATE OR REPLACE TABLE <model name>` | 5 tests fail |
| drop the `SET search_path` half only | exactly 1 fails — `a_bare_read_still_reaches_a_producer_in_another_schema` |

End-to-end on a real POC (`01-quality/04-local-test-with-duckdb`, which layers `staging` → `demo` and reads across them bare):

```
model executed locally  model=raw_orders  relation=staging.raw_orders
model executed locally  model=good_mart   relation=demo.good_mart
"status": "pass" ×2
```

`examples/quickstart` fails `rocky test` on this branch — and fails **identically on `main`**, for an unrelated missing `source` catalog. Checked by running the baseline, not assumed.

Green: `rocky-engine`, `rocky-compiler`, `rocky-core`, `rocky-duckdb`, 1900 `rocky-cli` lib tests, clippy `--all-targets --all-features`, fmt.

Refs #1354


hugocorreia90:
## Analysis: seven defects, six of them regressions against `main`. Recommend not merging as it stands.

The direction is right and the decision behind it stands. The implementation loses two things it has to keep: **physical identity** (the catalog is dropped, so two distinct warehouse objects become one local table) and **the compiler's own dependency binding** (one sorted global `search_path` decides which model a bare read reaches, and it can disagree with the edge the compiler derived).

Findings 1 to 5 are **differential probes**: the same project run under a binary built from this branch (`fb78012f`) and one built from `main` (`809d8f41`), both `cargo build --bin rocky`, the baseline verified to lack this branch's strings.

First, what holds up. The intended change works exactly as specified — a bare read of a renamed model's name resolves on `main` and misses on this branch. And the change is **invisible to the current corpus**: all 66 in-repo example projects with a `models/` directory produce byte-identical per-model pass/fail under both binaries, and all 16 POC `run.sh` scripts that call `rocky test` pass under both. Nothing below was found by running the suite. Each needed a project built to exhibit the shape.

---

### 1. Dropping the catalog collapses two distinct objects into one, silently

```
   a  ->  cat1.s.t   SELECT 1        b  ->  cat2.s.t   SELECT 2
                    \                      /
                     ----->  s.t  <-------      last writer wins
```

Both models report `pass`. A third model reading `s.t` sees `2`. On `main` they were `main.a` and `main.b`, two tables.

`E036` cannot catch this, and the PR body is right about why: `CollisionIdentity::of(&t.catalog, &t.schema, &t.table)` at `project.rs:102` keys on all three parts, and these are genuinely different warehouse objects. But that is an argument for the local projection being **injective**, not for dropping the component that makes them distinct.

The stated justification is too strong. `executor.rs:109` says DuckDB "resolves a three-part name only against an `ATTACH`-ed database and nothing attaches one". True — but `ATTACH` is available and the executor is the thing that would call it:

```sql
ATTACH ':memory:' AS poc;
CREATE SCHEMA poc.staging;
SET search_path='poc.staging,memory.main';   -- works, bare reads resolve
```

Two caveats if that route is taken: a catalog literally named `main` is refused as reserved, and a hyphenated catalog needs quoting. Neither is a blocker. The alternative — refuse a multi-catalog project locally with a real message — is also defensible. Silently collapsing is the one option that isn't. Three in-repo example projects already declare more than one catalog; none currently collide.

### 2. An ephemeral model materializes, and can overwrite a real model's table

`executor.rs:60` executes every model in `project.layers` without consulting its strategy. `project.rs:86` excludes ephemeral models from `E036` in as many words, because "an ephemeral model materializes nothing, so it cannot be the second writer".

```
   real  (full_refresh)  ->  s.t   SELECT 1
   eph   (ephemeral)     ->  s.t   SELECT 2      both "pass"; s.t holds 2
```

The real model's output is destroyed and both models report success. On `main` this was harmless: the two landed at `main.real` and `main.eph`, so the phantom target never collided with anything. This PR turns a quirk into data loss, and `E036` is exempted from catching it by design.

### 3. The sorted global search path can bind a bare read to a different model than the compiler chose

This is the one I'd fix first, because it produces a **wrong answer that reports success**.

```
   events   -> z.events   SELECT 2
   other    -> a.events   SELECT 1
   summary  -> marts.summary   SELECT id FROM events
```

`classify_table_ref` (`resolve.rs:96`) resolves the bare `events` to model `events` and orders `summary` after it. No `D012` fires — model `events` writes a table called `events`, so the name and the object agree. But `prepare_target_schemas` sorts the schemas, installs `a,marts,z,main`, and DuckDB resolves `a.events` first.

`summary` reads `1`, from `other`. On `main` it read `2`, from the model the compiler picked.

The PR frames cross-schema ambiguity as warehouse fidelity, deferred to #1632. That framing does not cover this case. Here the compiler has an opinion, has built the graph on it, and local execution silently contradicts it. Alphabetical precedence is not a resolution rule anyone chose.

### 4. A target schema that fails validation is skipped from setup but not from execution, so a consumer reads stale seed data

`prepare_target_schemas` filters silently at `executor.rs:187` (`validate_identifier(s).is_ok()`), while `local_relation` fails that model at `executor.rs:135`. Execution continues.

```
   data/seed.sql   CREATE TABLE main.source AS SELECT 7
   source    -> bad-name.source   SELECT 42     (fails: invalid identifier)
   consumer  -> good.consumer     depends_on = ["source"],  SELECT ... FROM source
```

`consumer` reads **7**, the seed's stale value, after its producer failed. On `main` it read 42. Run as `rocky test --model consumer` the producer's failure is filtered out of the report entirely, so the surviving signal is a consumer that ran on the wrong data.

A raw sidecar schema reaches `ModelConfig` without this validation, so the shape is reachable from a plain project.

### 5. An identifier Rocky accepts but DuckDB cannot parse unquoted fails the entire project

`validate_identifier` is `^[a-zA-Z0-9_]+$` (`validation.rs:7`). That admits digit-leading names and reserved words, which `executor.rs:193` interpolates unquoted into `CREATE SCHEMA`.

| target schema | this branch | `main` |
|---|---|---|
| `123stage` | **every model fails**, including models in valid schemas | all pass |
| `select` | **every model fails** | all pass |
| `temp` | that model fails, ambiguous catalog reference | passes |
| `information_schema` | that model fails, system-catalog write refused | passes |

The first two are the serious ones: one bad schema name takes down models that have nothing to do with it, because `prepare_target_schemas` failing marks every model in the project failed. And the trigger is realistic — `schema_template = "{region}"` with a numeric region passes the same insufficient check at `models.rs:966`.

### 6. `D012` now tells users the opposite of what the binary does

The warning at `resolve.rs:210` still reads: "`rocky test` and `rocky ci` materialize each model under its own name." On this branch they do not. A probe project printed that sentence and the failure disproving it in the same JSON document.

Same stale claim at `resolve.rs:16` (module doc), `resolve.rs:419` (test prose) and `resolve.rs:582` (assertion string). The PR updated three comments; these four are the ones that were left, and the first is user-facing.

### 7. A target schema spelled `Main` inverts the path precedence for the whole project

The filter at `executor.rs:208` is `*s != DEFAULT_LOCAL_SCHEMA`, a case-sensitive comparison, and the comment above it promises the default schema "goes LAST: a configured target outranks whatever happens to share its name in the default schema."

DuckDB folds schema names, so `Main` **is** `main` — but the filter does not, and `M` sorts ahead of every lowercase letter. So one model targeting `Main` puts the default schema first:

```
   path 'zeta,main'        bare `t` -> 222   (zeta.t, the configured target — as documented)
   path 'Main,zeta,main'   bare `t` -> 111   (main.t, the default schema — precedence inverted)
```

One model's capitalisation silently reverses the resolution order for every bare read in the project. Cheap to fix: fold the comparison, the way `bare_name_binds` and `E036` already fold identifiers.

---

### Two smaller things

**The mutation claim is off by one.** The PR body says reverting to `CREATE OR REPLACE TABLE <model name>` fails 5 tests. It fails **4**. Verified with `scripts/mutation-check.sh` on this branch, which reported OK:

```
a_model_with_no_schema_is_not_captured_by_another_models_schema     FAILED
a_model_materializes_at_its_configured_schema_and_table             FAILED
a_model_without_a_schema_still_lands_in_the_default_one             FAILED
three_tenants_writing_one_table_name_land_in_three_schemas          FAILED
a_bare_read_still_reaches_a_producer_in_another_schema              ok
local_relation_refuses_a_target_it_cannot_interpolate               ok
test result: FAILED. 2 passed; 4 failed
```

The bare-read test survives because the search path is untouched by that mutation; the last test calls `local_relation` directly. The `SET search_path` half of the claim is exact — removing it fails that one test and no other.

**A latent one.** `data/seed.sql` runs on the same connection at `test_runner.rs:137`, before `executor.rs:212` sets `search_path` unconditionally. A seed that sets its own search path loses it. No in-repo seed does, so this is latent rather than a live regression, but it is a contract worth stating either way.

**Not in scope but adjacent:** the PR carries no `CHANGELOG.md` entry. This changes what `rocky test` and `rocky ci` do — the class the issue itself marked owner-merge — and the repo's `### Changed` section documents smaller contract moves than this one.

---

### What I am least confident about

Whether local execution should **emulate** catalogs through `ATTACH` or **refuse** a multi-catalog project outright. I verified `ATTACH` works and gives real three-part fidelity, but I have not measured what it costs on a project with many catalogs, nor whether a per-catalog in-memory database interacts badly with the seed loader, which writes to `main` unqualified. That is a product decision, not a detail. What is not in doubt is that the current projection is lossy in a way nothing downstream can detect.

I am also not confident in the PR's corpus numbers. It cites 155 models and 24 projects layering across two or more schemas; my scan of the same tree found 131 models across 100 projects, 20 of them multi-schema. The direction is identical and the argument survives either count, but the exact figures did not reproduce, so I would not quote them in the commit message.

### What I think is being missed

**There is no written contract mapping a compiler model identity to a local DuckDB object.** Without one, the executor is free to pick a lossy projection (drop the catalog) and an unrelated resolution rule (sort the schemas), and nothing anywhere can notice the two disagree with the compiler. Findings 1, 3 and 4 are all instances of that single absence. Writing the mapping down first — injective, and binding a bare read to the producer the graph selected — would make each of them either impossible or a test.

The narrower version: none of the six new tests asserts a `rocky test` **result** for a shape with more than one candidate producer. They assert placement, which is why a complete-looking suite is green beside all six findings.

---

### Independent red team

**Codex (`gpt-5.6-terra`), 74 tool calls, verdict BLOCK.** It ran against the same worktree and reached findings 1 through 6 independently, plus the seed-path regression and the mutation-count correction. It also ran its own second reviewer on its analysis. Its summary: *"local target projection must preserve both physical identity and compiler-selected dependency binding; dropping catalog and using one sorted global search path preserves neither."*

Two of its claims I checked and would narrow. It reported the DuckDB three-part limitation as overstated — correct, and my `ATTACH` probe above is the evidence. It could not run `scripts/mutation-check.sh` under its read-only sandbox and inferred the count from equivalent SQL; I ran the real script and its inferred count of 4 was right.

One place it stopped a step short. It checked `Staging` versus `staging` and called case clean, which is right for that pair. It did not carry the same fold into the `!= "main"` filter, which is finding 7 above — the only one of the seven that came from neither its pass nor the suite.

Refs #1354, #1629, #1630, #1631, #1632

hugocorreia90:
## Constraint on finding 2's remedy: "skip ephemerals locally" is not available

Adding this because the obvious fix is wrong, and the first review pass recommended it in those words — "Skip ephemerals both when preparing schemas and executing, with a database-observed absence/collision test."

That would break every local consumer of an ephemeral model. `run.rs:8285` states it directly:

> Ephemeral models emit no statements … and the comments throughout this codebase describe them as "inlined as CTEs in downstream queries". **No such inlining exists**: nothing in `rocky-compiler` or `rocky-sql` rewrites a consumer's `FROM eph` into a CTE … So a consumer of an ephemeral model reads **whatever physical table happens to carry that name**.

So the phantom table is not a stray artefact — locally it is the only thing that makes an ephemeral model's consumers work at all. `rocky test` resolves `FROM eph` to the table `execute_locally` created by name.

That does not make finding 2 less real. Both statements hold:

```
ephemeral materializes locally   -> its consumers resolve (by accident, by name)
ephemeral materializes AT ITS TARGET (this PR)
                                 -> it can overwrite a real model's table, silently
```

The PR turns a load-bearing accident into data loss. Removing the accident without replacing it turns it into a broken read.

**So the remedy is one of:**

1. **Real CTE inlining** — rewrite a consumer's `FROM eph` into a `WITH` clause, which is what the codebase's comments have claimed all along. Then skipping materialization is correct and the phantom table stops existing anywhere.
2. **Refuse an ephemeral consumer on the local path**, the way `run.rs:8300` already refuses shadow/branch execution for exactly this reason: "a consumer would read the production table instead of an isolated one."

Either way the test has to execute `eph → consumer` and assert the **consumer's data**. A table-absence assertion alone would pass against a fix that silently breaks every reader.

**Provenance:** the "skip them" recommendation was mine to inherit and I nearly acted on it — I recorded it as buildable-now in a follow-up analysis. A second Codex pass rejected that, and a second-opinion analysis in another session independently confirmed the citation. Recording it here so the constraint sits with the finding rather than in a session that ends.

Refs #1630, #1354

hugocorreia90:
## Recommendation: close this PR, keep #1354 open, rebuild behind a test harness first

Re-checked today before recommending anything. The branch has not changed since the analysis above: the only code commit is `4e845a92`, followed by two merges from main (`663e1be5`, `fb78012f`). Both of my comments postdate all three, so the seven defects were written against this exact head and none has been addressed.

### Why not fix it in place

Three of the seven — catalog dropped (1), a bare read binding to a different model than the compiler chose (3), and an invalid target schema silently filtered from setup (4) — are not three bugs. They are one absence:

> there is no written contract mapping a compiler model identity to a local DuckDB object.

Patching them one at a time re-creates the class, because the next change adds a fourth case to the same undefined mapping. Finding 7 (a schema spelled `Main` inverting path precedence project-wide) is the same root reaching a fourth place, and it came from neither the test suite nor the red team.

The decision behind the PR is **not** in question. #1354's ruling — local execution should materialize at the configured target — is correct and unchanged. This implementation loses two things it has to keep: physical identity, and the compiler's own binding.

### What to settle before new code

**1. The identity contract.** One sorted global `search_path` is the mechanism behind findings 1, 3 and 7. The PR's own justification concedes the alternative: `ATTACH ':memory:' AS poc` combined with `SET search_path='poc.staging,memory.main'` works, so per-catalog attach is a real option. Two caveats found while checking it: a catalog named `main` is refused as reserved, and a hyphenated catalog needs quoting.

**2. Ephemeral models (finding 2).** The obvious remedy is not available. `run.rs:8285-8291` states that no CTE inlining exists anywhere, so a local consumer of an ephemeral model reads the phantom table by name today; skipping materialization breaks every such consumer. The choice is real inlining, or refusing ephemeral consumers on the local path. Either way the test must run `ephemeral -> consumer` and assert the consumer's **data**, not just that it ran.

### The harness has to come first

All 66 in-repo example projects with a `models/` directory produce byte-identical per-model pass/fail under both binaries, and all 16 POC `run.sh` scripts pass under both. The suite could not see any of the seven. Every finding needed a project built to exhibit the shape — six of them (crosscat, ephclobber, orderdep, wrongbind, badschema, renamed). Those should become real fixtures before the rework, or the same class ships again unseen.

Method that worked, if someone repeats it: archive **the whole repo** rather than `engine/` (`rocky-mcp` does `include_str!("../../../../.claude/skills/…")`, so an engine-only archive fails to compile and leaves the previous binary in place), build both with a shared `CARGO_TARGET_DIR`, then diff per model. Verify which binary you actually ran with `strings <bin> | grep -c "<a string only the PR adds>"` — that check caught exactly the stale-binary case.

### What this does and does not unblock

Correcting something I said earlier in shorter form:

```
  #1631   CLOSED 2026-09-10 — its recorded gate was CIRCULAR (step 2 waits on
          #1631, not the reverse). Already landed.
  #1629   NOT fully blocked. Its P1 half (derive_physical_edges loses a true
          edge permanently on the catalog-omission 3-part shape) and P2 half
          (seed/load label collision policy) are buildable now. Only the
          wiring is downstream, and that lands atomically with step 2.
  #1630   genuinely blocked — it needs the ephemeral decision above.
```

So closing this PR unblocks less than the issue list suggests, and #1629 can start regardless.

### Red team

Codex `gpt-5.6-terra`, 74 tool calls, VERDICT **BLOCK**, reaching findings 1-6 independently. Its summary: *"dropping catalog and using one sorted global search path preserves neither physical identity nor compiler-selected dependency binding."*

hugocorreia90:
## Closed without merging — recommendation accepted (2026-09-16)

The recommendation above is accepted: close this PR, keep #1354 open, and rebuild step 1 behind regression fixtures.

The branch `feat/1354-step1-local-targets` is **kept on purpose**. It is the record of what was attempted, and the rebuild will want to diff against it.

What happens next:

```
  #2045   regression fixtures for the seven shapes; needs no decision; lands FIRST
  #2044   two rulings: catalog identity mapping, and ephemeral models locally
  #1354   stays open; step 1 is rebuilt after both of the above
```

The child issues still read "gated on steps 1 and 2 landing first", which now implies they have no path. Each one is corrected on the issue itself:

- **#1629** — its P1 half is buildable now and does not wait on this PR or a replacement.
- **#1630** — genuinely waits, on decision 2 in #2044.

Issue #1501, opened 2026-08-20
security: Phase B reads the model sidecar through a raced symlink and persists external keys


hugocorreia90:
Found by an independent red team on PR #1498. Separate from #1482 and from #1500 — this one has a concrete data-exfiltration path into a committed artifact.

## The path

```
commit.rs:1030   Phase B reads models/<model>.toml with read_to_string
                 -> FOLLOWS a symlink
lowering.rs:237  the merge preserves every unowned TOML key
                 -> whatever it read is kept
the commit pre-check runs only AFTER this read
```

A process that can write in the models directory points the sidecar at any Rocky-readable external TOML file, lets Phase B read it, then swaps a regular file back before the pre-check. External keys — credentials, for instance — are merged into the project sidecar and committed as project content.

A **hardlink** needs no timing at all where the filesystem permits one.

## Fix shape

Open the sidecar no-follow, require a regular file, and bind the later commit to that same descriptor/inode rather than re-resolving the path. Binding to the descriptor is the part that matters: re-reading by path after the check reintroduces the race.

Reported as reachable by the reviewer; I have **not** independently reproduced the exfiltration end to end, so treat the mechanism as traced-from-source rather than demonstrated.

hugocorreia90:
**Chain link: #1500 is the fix-carrier.** These two share one root — validate a path, then re-resolve it — so one PR should close both.

```
#1500  pathname-based containment (hardlink / Windows / parent-dir race)
   |   dirfd-relative operations are the shared fix
   v
#1501  Phase B sidecar re-read  <- this issue: the only DATA-EXFILTRATION outcome
```

Keeping this open separately because its regression test belongs in that PR: the others are integrity failures, this one commits external TOML keys into the project sidecar. That test is the thing that proves the dirfd approach actually closed the exfiltration path, not just the overwrite paths.

PR #1841, opened 2026-09-09
fix(engine): the commit protocol publishes through captured directory descriptors

hugocorreia90:
The last of #1500's three decided items. The other two are already on `main`, so this PR is narrower than the issue:

| decided | state |
|---|---|
| refuse a source with `nlink > 1` | **shipped** — checked on the opened descriptor in `write_no_follow` / `read_no_follow`, documented as a stated compatibility break |
| Windows documented as not guaranteed | **shipped** — the header's closing paragraph already says `O_NOFOLLOW` is unix, `number_of_links` is unstable, and neither leaf guard is present there |
| **dirfd-relative `openat` / `renameat`** | **this PR** |

#1501, filed alongside it, is closed with its own test at `step.rs:2046`.

## Where the redirect actually bites

I built the obvious exploit first and it did not work, which is worth stating because it is the intuitive attack:

```
  swap between staging and publishing   the staged SOURCE name stops
                                        resolving -> the rename fails
                                        closed. No redirect.

  swap before cleanup                   remove_file("models/x.ff-prev")
                                        follows the replacement and deletes
                                        a file in a tree the project never
                                        owned. THIS one works.
```

So the primitive #1500 hands an attacker is **deletion outside the project**, not a redirected write. The fix is the same either way — `DirCustody` opens every parent ONCE, before the first write, with `O_DIRECTORY | O_NOFOLLOW | O_CLOEXEC`, and the publishing mutations become `renameat` / `unlinkat` against those descriptors, which do not re-traverse.

## Evidence: a pair, because one test would not be evidence

| test | asserts |
|---|---|
| `a_swapped_directory_redirects_a_pathname_based_cleanup` | the attack **works** — a file outside the project is deleted |
| `captured_directories_survive_a_swap_that_defeats_pathnames` | the same swap against the captured descriptors deletes **nothing** |

A single green containment test could be green because the swap never fired. The harness half is there to fail loudly if that ever becomes true — its assertion is *"if every decoy survives here, the sibling test proves nothing."* Both assert on **where the bytes went**, never on a message.

**Mutation-checked**: reverting `DirCustody::remove` to `std::fs::remove_file` fails the containment test.

One decoy is planted per published leaf rather than per the leaf that happens to be removed first — pinning the removal order would make the test fail for a reason unrelated to containment the day the artifact set changes.

## Not closed, and the header now says so

A symlink swapped in at the staged **leaf name** between the close and the rename is renamed into place by `renameat` exactly as by `rename`. The descriptor protects the directory, not the leaf. The old header said this residual was *"closed by the same fix (a dirfd plus `renameat`, **or** holding the staged descriptor through publication)"* — only the second remedy reaches it, and `linkat` with `AT_EMPTY_PATH` is Linux-only, wants `CAP_DAC_READ_SEARCH`, and **`libc` does not define `AT_EMPTY_PATH` on Apple targets at all** (checked, not assumed). There is no portable spelling, so v0 keeps the residual and the header states it as a residual rather than as closed.

## Structure

Capture lives in `commit_generation_with_ops`, which both `commit_generation` and `run_phase_b` reach with `ops: None` — so neither can quietly keep the path-based calls, and Phase B is covered without a second capture site. `Some` still belongs to the four crash drills that freeze the protocol mid-flight; `CommitOps` now owns its closures (`Box<dyn FnMut>`) so the production default and an injected one are the same type.

A mutation whose parent was not captured is **refused**, not silently fallen back to `std::fs::rename` — a fallback would reintroduce the vector on exactly the paths nobody enumerated. The two `create_dir_all` calls inside the protocol are gone; capture creates every parent, deliberately, so nothing re-resolves a path after custody is taken.

On non-unix the operations stay path-based behind the same type, so the protocol has one shape rather than a second code path nobody runs.

Green: 2175 `rocky-core` lib tests plus every integration suite, clippy `--all-targets --all-features`, fmt.

Refs #1500, #1501


hugocorreia90:
Blocked by #1844 (CI infrastructure), fixed in #1845 — not by anything in this diff.

hugocorreia90:
The three red checks are #1844, not this diff: every engine job is dying at `sudo apt-get update` because Google's Chrome apt repo is serving a stale index, and `apt-get update` fails when any configured source does. Reproduced on two runs of this same commit 22 minutes apart, and on #1842 whose diff shares no files with this one.

Fixed in #1845. This PR needs a rebase once that lands.

Issue #1583, opened 2026-08-31
resume scope: decide the source-side identity (endpoint field vs config fingerprint)


hugocorreia90:
The resume scope (#1573, #1582) covers pipeline name, `--filter`, and TARGET routing. It records nothing about the source. A `[source]` edit between runs still skips silently.

## Reproduce

1. Run a replication pipeline to success.
2. Edit its `[source]` block — point `source.discovery.adapter` at another adapter, or change `schema_pattern` so discovery maps different source tables onto the same target names.
3. `rocky run --resume-latest` after a failure. `completed_keys` matches the checkpoint's `Success` entries by TARGET key, so tables never copied from the new source are skipped and the run exits 0.

Same code path as the endpoint hole #1582 closes, one field short.

## The decision

What defines "the same run"? Two candidate shapes:

- Add the discovery adapter's `endpoint_identity()` plus the schema pattern to `ResumeTarget`. Narrow, matches #1582's pattern.
- Stamp the pipeline-block fingerprint that `load_rocky_config_fingerprinted` already computes, and refuse (or warn) on mismatch. Total, but refuses resume on ANY pipeline edit, harmless ones included.

Refusing more is fail-closed; refusing everything makes `--resume-latest` useless after any tweak. This is a contract call, not a patch.

Found by independent adversarial review (Codex) of #1582's branch; the reproduction was verified against the real code path before filing.

hugocorreia90:
Related, from #1591 (which closes #1586).

A concrete trigger for this gap, found by independent red-team review:

- `[source.schema_pattern] separator = "__"`, no `[target] separator` (so the target inherits it)
- `catalog_template = "wh"`, `schema_template = "raw__{regions}"`
- the run uses `--branch feature`, so the target schema is always `branch__feature`

`SchemaPattern::parse` splits every source schema name on the pattern separator. Change it to `"---"` and `src__acme__us_west__shopify` stops parsing while `src__acme---us_west---shopify` starts. The catalog is static and the schema is the override, so the new source resolves to the **same** `wh.branch__feature.orders` key. A resume then skips a table it never copied.

#1591 blocks that specific path by never eliding a separator the target inherits from the pattern. That is a patch over one symptom, not the class:

- with an explicit `[target] separator`, editing `[source.schema_pattern] separator` is recorded nowhere and the resume still crosses;
- the pattern `prefix` and `components` are not in the scope at all.

[...]

PR #1842, opened 2026-09-09
fix(engine): the resume scope records where the run read from

hugocorreia90:
The resume scope covered pipeline, filter and the whole TARGET block, and recorded **nothing** about the source. This issue's own reproduction still worked on `main`.

## The hole

```
  edit source.discovery.adapter  ─▶  no field in ResumeScope moves
        │
        └─▶ completed_keys matches the checkpoint's Success entries
            by TARGET key
              └─▶ tables never copied from the NEW source are skipped
                    └─▶ the run exits 0
```

A resume that skips work it never did is the same class of exit-0 lie #1582 closed on the endpoint, one field short.

## What the scope now records

`ResumeSource`, alongside `ResumeTarget`:

| field | why it is not redundant |
|---|---|
| `discovery_adapter` | the issue's step 2. `None` is not "unrecorded" — it is a pipeline with no `[source.discovery]` block, itself a difference worth refusing over |
| `endpoint` | an alias can be re-pointed without being renamed; the same argument `ResumeTarget::endpoint` answers on the write side |
| `catalog` | `[source] catalog`, verbatim |
| `pattern_prefix` / `pattern_separator` / `pattern_components` | the shape that maps source schema names onto target names |

The pattern goes in as a **shape**, not as itself. Discovery parses each source schema name with it, and the captured values render the target names the checkpoint is keyed by — so two patterns that differ can map different source tables onto one key. Components render to stable tags (`fixed:`, `var:`, `varlen:`, `term:`) so the recorded blob does not move the day `PatternComponent` gains a variant or a field.

The pattern's own separator is recorded separately from `ResumeTarget::separator_role`: that one is the **target** separator, which `[target] separator` may override, while the pattern's splits every source schema name regardless.

## Option A, and why not B

Option B — stamp the config fingerprint — stays rejected for the reason recorded on the issue: it refuses resume on every credential rotation and cannot name what changed. This is the minimal form of option A: the discovery adapter's `endpoint_identity()`, `source.catalog`, and the pattern's shape.

## Old checkpoints

A checkpoint written before this field deserializes with `source: None`; a current scope always records `Some`, so it never compares equal and the resume refuses as a scope mismatch. That is the same fail-closed treatment `target: None` already gets, and it is deliberate — a checkpoint whose source is unknown is not evidence any table was copied from *this* source.

## Tests

| test | asserts |
|---|---|
| `re_pointing_the_discovery_adapter_changes_the_resume_scope` | the issue's step 2 |
| `re_pointing_the_adapter_behind_the_alias_changes_the_resume_scope` | same alias, different database |
| `a_different_source_pattern_changes_the_resume_scope` | prefix and components each move it |
| `an_unchanged_source_leaves_the_source_scope_equal` | the refusal is not simply "always" |

`resume_scope_blob_shape_is_pinned` is updated deliberately — the pinned blob gaining a `source` object *is* the fix, since its absence was the defect.

## Two test fixes worth naming, because neither was mechanical

**`resume_refuses_a_checkpoint_from_another_gateway_path`** asserted `!rendered.contains("tenant")` to prove a Trino gateway path stays out of the message. The source half legitimately renders a schema-pattern component named `tenant`, so that check would have failed for a reason unrelated to gateway routing. It now asserts `!rendered.contains("/tenant/")` — the path, which is what it means.

**Three seed sites** built their checkpoint scope with a test fixture and then drove the real run path against it. With a source half in the scope, a fixture seeds something the run never builds and the lookup misses. They now derive `&replication.source` plus the discovery adapter from the same adapter map the run uses.

Green: 1904 `rocky-cli` lib tests, clippy `--all-targets --all-features`, fmt.

Refs #1583, #1582


hugocorreia90:
The three red checks are #1844, not this diff — every engine job is dying at `sudo apt-get update` on a stale Google Chrome apt index. Fixed in #1845; this PR needs a rebase once that lands.

PR #1843, opened 2026-09-09
fix(engine): a schema-wide quality target that lists no tables is not a clean run

### LINKED ISSUE #1811: A quality target whose schema lists as empty still reports a clean run — the zero-rows door #1786 left open
Found by an independent red-team pass over the #1733→#1795 merge window. #1786 closed the case where listing a schema **errors**; the case where it succeeds with **zero rows** is still open, and it is the case a renamed schema produces.

## Repro

A schema-wide quality target on DuckDB: valid catalog, `table` omitted, the schema renamed or never created, checks enabled, `fail_on_error = true`.

## Trace

`engine/crates/rocky-duckdb/src/dialect.rs:235` builds a filtered `information_schema.tables` query. A missing schema is not an error there — it returns `Ok` with zero rows.

The success arm collects that empty vector at `engine/crates/rocky-cli/src/commands/run_local.rs:579`, so the check loop at `run_local.rs:603` iterates nothing and emits nothing. `error_failures` is still zero at `run_local.rs:705`.

## Wrong result

```
status:        success
tables_failed: 0
exit code:     0
```

for a run that checked **no tables at all**. A scheduled quality pipeline pointed at a schema someone renamed reports green forever.

This is the same class as #1780 and #1786 — "did not run" reported as "passed" — reached through a different door: an empty successful listing rather than a failed one.

## Why the existing test does not catch it

`a_schema_target_that_cannot_be_expanded_reaches_the_error_bucket` at `run_local.rs:2707` calls the expansion helper directly and supplies an explicit SQL/query error. It never drives the live `Ok(empty)` path, so it passes with this defect present.

## Suggested shape

An expansion that resolves to zero tables is not a clean run. Either refuse it the way an unreadable target is refused, or record a `not_evaluated` reason so the gate can tell "nothing matched" from "everything passed" — the distinction v27 already introduced for checks.

The fix needs a test that goes through the real expansion against a real missing schema, not through the helper.

hugocorreia90:
Closes #1811. The zero-rows door #1786 left open.

## What was wrong

A `[[pipeline.<name>.tables]]` entry with no `table` names a whole schema, and `run_quality` lists it with the dialect's `list_tables_sql`. #1786 made a listing that **errors** an error-severity check the engine could not evaluate, so the gate fails. A listing that **succeeds with zero rows** was still collected as an empty vector; the check loop iterated nothing, emitted nothing, `error_failures` stayed 0, and the run reported:

```text
status:            success
check_gate_failed: absent
tables_failed:     0
exit code:         0
```

for a run that checked no table at all. That is exactly what a renamed or never-created schema produces: to `information_schema` it is not an error, it is an empty answer. A scheduled quality pipeline pointed at a renamed schema stayed green forever.

## The fix

In the schema-wide branch of `run_quality` (`engine/crates/rocky-cli/src/commands/run_local.rs`), an `Ok` listing with zero rows is recorded through the same `checks_for_unexpandable_target` helper as an unlistable one: one `schema_expansion` check, error severity, `not_evaluated` with a reason that says nothing was checked and names `catalog.schema` as missing, renamed, or empty; then `continue`. Nothing else changes: an explicit `table` target is untouched, and a schema that lists tables runs exactly as before.

```text
list_tables_sql
   |-- Err            -> schema_expansion, not evaluated   (#1786)
   |-- Ok, 0 rows     -> schema_expansion, not evaluated   (this PR)
   '-- Ok, N rows     -> N tables checked                  (unchanged)
```

**Decision, on the record:** a schema that exists but is empty fails the gate too. The pipeline's job is to check the tables in that schema; checking none is not evidence that they are fine, and the reason says "may be missing, renamed, or empty" so the operator knows which to look for. Telling the three apart precisely would need a schema-existence query per dialect; that is not in this PR. `fail_on_error = false` keeps the run green and still reports the unevaluated target.

## Tests

The issue asked for a test through the real expansion against a real missing schema, not through the helper. Three, in `engine/rocky/tests/quality_check_gate.rs`, each through the real binary on DuckDB:

- `a_schema_wide_target_that_lists_no_tables_is_not_a_clean_run`: schema `ghost`, gate on → exit 1, `status: Failure`, `check_gate_failed: true`, `tables_failed: 0`, one `schema_expansion` check keyed `["fixture", "ghost"]` whose reason names the schema.
- `an_empty_schema_expansion_is_reported_even_when_the_gate_is_off`: same target, `fail_on_error = false` → exit 0, `Success`, and the unevaluated check is still in the payload.
- `a_schema_wide_target_with_a_table_still_checks_it_and_passes`: the discriminator — schema `main` with the seeded table → the table is checked and the run is green, so the fix keys on the empty listing, not on schema-wide targets as such.

## Evidence

- `cargo test -p rocky --test quality_check_gate`: 5 passed. `cargo test -p rocky-cli --lib -- run_local`: 28 passed. `cargo clippy --all-targets -D warnings` and `cargo fmt --check`: clean.
- Two mutation checks via `scripts/mutation-check.sh`, both exit 0: reopening the door (`if false && listed.is_empty()`) fails the gate-on test on the exit code and the gate-off test on the missing `schema_expansion` result.

## Not in this PR

- Distinguishing "schema missing" from "schema exists but empty" (a per-dialect existence query).
- The other dialects' `list_tables_sql` are unchanged; the fix is at the consumer, so every adapter that answers an empty listing gets it.

## Least confident / most likely missing

- **Least confident:** whether an existing-but-empty schema should fail the gate. Argued above; the alternative (pass with a warning) is the shape this issue is about.
- **Most likely missing:** a dialect whose listing query answers a missing schema with an *error* is already covered by #1786; one that answers with a row for the schema itself and no tables would slip through neither path — none of the in-tree dialects does that, as far as the `list_tables_sql` implementations show.

Independent red team: Codex review dispatched on this PR; disposition follows in a comment.


hugocorreia90:
## Independent red-team disposition — round one: PASS WITH NITS (one nit, taken)

**Reviewer: Codex (GPT-5), 30 tool calls on `0c712cc1`; its second reviewer (`gpt-5.6-terra`) reached the same verdict and found no additional blocker.** The guard's placement, the gate path on and off, the persisted record, the Pipes emitter, the SDK and Dagster readers of a two-element `asset_key`, the absence of a hook seam in `run_quality`, and every in-tree dialect's listing query were verified; the existing-but-empty decision was judged appropriate, with `fail_on_error = false` as the explicit escape hatch; the three tests were judged non-vacuous; the changelog true.

- **NIT — the text summary said "across 1 table(s)" for a run that found zero tables. Taken.** `check_results.len()` counts the schema-level entry as a table. The summary now counts tables actually checked and says separately how many schema targets could not be expanded: `0 check(s) across 0 table(s), 1 schema target(s) could not be expanded, 1 error / 0 warning failed`. JSON, the gate, the record and the exit code were already right; the JSON-only test helper could not see the text line, which the reviewer noted.
- **Residual the reviewer named (not a defect): a listed row whose first cell is not a string is dropped by the `filter_map`.** Every in-tree dialect projects a string `table_name`, so an all-malformed answer would fail closed through the new guard and a mixed one could silently check fewer tables. Taken as a warning in the same commit: when usable names are fewer than rows listed, the run logs how many were skipped. No behaviour change otherwise.

The reviewer's two answers, on the record: least confident is live warehouse metadata behaviour (visibility, pagination, cell typing), which static tracing and vendor documentation support and which no in-tree test exercises live; most important missing is a non-JSON end-to-end assertion, which is the nit above. Both changes are in the summary line and a log line; merging on green.

PR #1849, opened 2026-09-09
fix(engine/rocky-server): rocky serve reports a compile that produced no result, not a clean project

### LINKED ISSUE #1823: rocky serve reports a clean project when its background compile failed — no diagnostics, has_errors false
Found by the second independent review of #1822 (finding 3). Routed here because it is about how `rocky serve` surfaces a compile failure, not about the path probes that PR closes.

## The state

The project's `models` entry cannot be read — a dangling symlink is the simplest way to make one.

## What happens

`rocky serve` reaches the stricter model walker (which now reports the broken root, #1817), but the compile runs **asynchronously**, its failure is **logged**, and the server continues. With no compile result to show, the project route reports:

```
diagnostics: []
has_errors:  false
```

A project whose models cannot be read at all presents to the SPA and to `/api/v1/project` as a clean, healthy project with nothing to say.

Sites: `engine/crates/rocky-cli/src/api.rs:916` (the async compile whose error is logged and dropped) and `engine/crates/rocky-server/src/state.rs:255` (the project state that reads "no result" as "no problems").

## Why it is its own issue

"Did not run" reported as "passed" is the #1780 / #1782 / #1811 class, reached through the serve process's background compile. It is independent of the probe fix: even with every path probe honest, a compile that fails for any reason after the server is up produces the same clean-looking state.

## Suggested shape

The project state needs a third value beside "compiled clean" and "compiled with errors": **"no compile result"** — with the last failure's message. `has_errors` must not be `false` while that state stands, and the route should say the compile failed rather than listing zero diagnostics.

Related: #1817, #1811, #1780.

hugocorreia90:
Closes #1823.

## The defect

`rocky serve` compiles in the background. When that compile failed, the server wrote a log line and continued. `GET /api/v1/project` read "no result" as "nothing to report".

A project whose `models` entry cannot be read looked healthy:

```text
                     before        after
compile_error        absent        "models ... cannot be resolved"
models_compiled      absent        absent
has_errors           false         true
```

## The change

The server now keeps a third compile state: the compile made no result, and why.

```text
recompile (one at a time)
  compile ok       ->  result kept,     failure cleared
  compile failed   ->  result dropped,  failure kept
  task panicked    ->  result dropped,  failure kept
```

- A failed compile drops the previous result. The model, lineage and DAG routes answer `503 engine_not_ready`. They no longer serve models the project does not have.
- Only one compile runs at a time. Each caller gets the outcome of the compile it asked for.
- `GET /api/v1/project` sends `compile_error`. `models_compiled` is absent. `has_errors` is true.
- `POST /api/v1/compile` answers `status: "compile_failed"`. It answered `"recompiled"` before.
- The project card shows the reason. Before the first compile it shows "compile pending".

## Decisions

- A failed compile drops the old result. Counts from an older compile do not describe the project.
- The state before the first compile does not change on the wire. `models_compiled` absent with no `compile_error` means pending, and the card shows it as pending.

## Tests

| Test | What it proves |
|---|---|
| `a_failed_compile_is_recorded_and_a_later_success_clears_it` | The failure is recorded. A good compile clears it. A second failure drops the result. |
| `overlapping_recompiles_are_serialised_and_each_reports_its_own_outcome` | One compile at a time. Each caller gets its own outcome. |
| `the_project_route_reports_a_failed_compile_rather_than_a_clean_project` | The route and `POST /api/v1/compile` report the failure. `GET /api/v1/models` answers 503. |
| `EstateScreen.test.tsx` | The card shows the reason, and shows "compile pending". |

Seven mutation checks pass with `scripts/mutation-check.sh`. Each test fails when its own fix is removed.

Clean: `cargo clippy --all-targets -D warnings`, `cargo fmt --check`. Passing: `rocky-server` 109, `rocky-cli` `api` 117, `engine/ui` 121.

## Not in this PR

The LSP keeps its own compile state. How it tells the editor about a failed compile does not change.

Red team: Codex, four rounds. Verdicts are in the comments.


hugocorreia90:
Merged `origin/main` in (`ad09d0ec`) after #1843 landed: one conflict, in `engine/CHANGELOG.md`, where both PRs added a bullet at the top of Unreleased → Fixed; resolved by keeping both. No code conflict. Re-gated the merged tree: `cargo check --tests` on `rocky-cli`, `rocky-server`, `rocky` clean; the merged-in `quality_check_gate` end-to-end suite passes (6).

hugocorreia90:
## Red team round 1 — Codex. NOT FULLY CLOSED

| | Finding | Done |
|---|---|---|
| HIGH | A failed compile kept the old result. Five routes served old models as current. | Drop the result with the failure. Those routes answer `503 engine_not_ready`. |
| HIGH | Two compiles could run at once. `POST /api/v1/compile` could report the other compile's outcome. | One compile at a time. Each caller gets its own outcome. |
| LOW | Before the first compile, the card looked healthy. | The card says "compile pending". |
| NIT | The changelog gave two wrong field shapes. | Corrected. |

My reason for keeping the old result was wrong. I wrote "the LSP needs it". The LSP has its own compile state.

hugocorreia90:
## Red team round 2 — Codex. NOT FULLY CLOSED

The reviewer found no defect in the code. Two items:

| | Finding | Done |
|---|---|---|
| LOW | No test proved that two compiles cannot run at once. | Added `overlapping_recompiles_are_serialised_and_each_reports_its_own_outcome`. |
| NIT | Two comments were still wrong. `models_compiled` is absent, not null. The old result is dropped, not kept. | Corrected. |

hugocorreia90:
## Red team round 3 — Codex. NOT FULLY CLOSED

| | Finding | Done |
|---|---|---|
| LOW | The new test waited 300 ms, then checked that nothing had changed. A slow second compile gives the same picture, so the test could pass with no lock at all. | The test reads the lock instead. While compile A waits, `compile_gate.try_lock()` must fail. This is true at any speed. |
| NIT | The route doc said `models_compiled` is `null`. It is absent. | Corrected. |

hugocorreia90:
## Red team round 4 — Codex. PASS

No findings. The reviewer agrees that the lock check fails at any speed when the lock is removed. Merging on green.

hugocorreia90:
Merged `origin/main` in again after four sibling merges (#1842, #1847, #1848, #1850), no conflicts. #1848 guarded the repository's dangling-symlink tests with `cfg(unix)`; the three this PR adds use the same Unix symlink call, so they carry the same guard now (one mechanical commit). Re-gated the merged tree: clippy and fmt clean, `rocky-server` 109 and the `api` module's 117 tests pass. No code decision in either commit; the merge chain is re-armed on the new head.

PR #1858, opened 2026-09-10
fix(engine/ui): a DAG node click opens the model, not a 404

### LINKED ISSUE #1857: Browser UI: a DAG node click opens a 404, a keyboard cannot open one, and the canvas ignores dark mode
Three defects in the browser UI's DAG panel. All confirmed against a live `rocky serve`.

## 1. A node click opens a 404

The panel sends the node's `id`. The route wants the `label`.

```
GET /api/v1/dag  ->  { id: "transformation:customer_orders",
                       kind: "transformation",
                       label: "customer_orders" }
                                |
      DagPanel sends the id ----+
                                v
GET /api/v1/models/transformation:customer_orders  ->  404
GET /api/v1/models/customer_orders                 ->  200
```

The route searches the compiled model set. It never strips the `kind:` prefix.

### Only one kind of node has a model to show

A prefix strip is not the fix. Seven of the eight kinds label something else.

| kind | its label | `/models/{label}` |
|---|---|---|
| transformation | the model name | 200 |
| source | `ecommerce (source)` | 404 |
| load | `ecommerce (load)` | 404 |
| quality | the pipeline name | 404 |
| snapshot | the pipeline name | 404 |
| seed | the seed name | 404 |
| test | the test label | 404 |
| replication | the pipeline name | 404 |

Two labels have a space in them. They are not identifiers at all.

## 2. A keyboard cannot open a node

React Flow sends Enter and Space to selection. It never sends them to a click.
So a keyboard user can highlight a node and can never open it.

## 3. The minimap and the controls paint light on a dark page

`<ReactFlow>` gets no `colorMode`. Its default is a fixed `light`.

## Why no test caught the first one

The only captured DAG fixture is replication-only. It holds two nodes, both of
kinds the model route cannot serve. No transformation node has ever crossed the
real route in a test.

## Expected

- A transformation node opens its pane, under the bare label.
- Every other kind is not clickable: no pointer cursor, no tab stop, `aria-disabled`.
- Enter and Space open a focused transformation node.
- The canvas follows the viewer's colour scheme.

hugocorreia90:
Fixes the three defects in #1857.

## What changed

The DAG panel sent the node's `id`. The model route wants the `label`.

```
                       id     "transformation:customer_orders"  ->  404
GET /api/v1/dag  --->
                       label  "customer_orders"                 ->  200
```

A prefix strip would not be enough. Only a `transformation` node labels a
compiled model. `nodeRoute` holds that table, and the panel opens a pane
only where the table says there is one.

| kind | clickable |
|---|---|
| transformation | yes, under the bare label |
| source, load, quality, snapshot, seed, test, replication | no |
| a kind the table does not name | no, and a test fails |

The other seven kinds now have no pointer cursor, no tab stop, and carry
`aria-disabled`.

**Keyboard.** React Flow sends Enter and Space to selection, never to a
click. So the panel handles those two keys itself.

**Dark mode.** `<ReactFlow colorMode="system">`. Its default is a fixed
`light`, which is why the minimap and the controls painted light.

## Tests

```
cargo test --features ui --test serve_ui
```

**A new fixture.** `engine/ui/src/test/fixtures/dag-mixed-kinds.json`, captured
from a real `rocky serve`. It holds seven of the engine's eight node kinds. The
existing corpus (`@rocky-fixtures`) is replication-only with two nodes, which is
why no transformation node had ever crossed the route in a test. Its capture
recipe is in the README beside it.

**How the tests reach it.** A relative import, not a new alias. The fixture
lives inside `engine/ui/src`, so `vite.config.ts` is unchanged.

**Against a live server.** `serve_ui.rs` starts `rocky serve` on a project built
to emit those same seven kinds, then asks the route about every node: the label
answers `200` only for a transformation, and the id answers `404` every time. It
classifies through the engine's own `NodeKind` enum over an exhaustive match, so
a new variant fails this file to compile. It asks only the API, so unlike the
page test beside it, it does not skip when the embed is empty.

**Where CI runs it.** The `Test` job, through `cargo nextest run
--all-features`, which turns on the `ui` feature. Read out of that job's own
log on this branch:

```
PASS [ 0.305s] (111/7399) rocky::serve_ui only_a_transformation_node_is_servable_and_only_under_its_label
PASS [ 0.006s] (112/7399) rocky::serve_ui real_server_prints_the_token_address_and_serves_the_public_page
```

The two timings are the proof that the new test does not skip. It spends
0.305s starting a server and making 18 requests. The older page test returns
in 0.006s because that job has no `engine/ui/dist`, so it takes its early
return, which is exactly the condition the new test had to survive.

The `Browser UI` job is npm only, with no Rust toolchain. The release-build
smoke does build `--features ui`, but it runs curl, not the test binary.

**Mutation checks.** Eleven guards, each proven to fail when its fix is
reverted (`scripts/mutation-check.sh`, clean tree): the click passes the label;
an unknown kind is its own state; a transformation is servable; only a servable
node is focusable; the rest are `aria-disabled`; Enter and Space open;
`colorMode` follows the system; the pointer cursor is only where something
opens; the live-route classification; the estate test's loader refuses an id;
and the fixture-drift comparison.

The one guard with no mutation check is the compile-readiness latch. Removing
it restores a race that usually passes, so a mutation check would prove
nothing.

**One test was still teaching the defect.** `EstateScreen.test.tsx` is the test
this issue names as the reason nothing caught the bug. It drove the detail pane
with `capturedDag.nodes[0].id` through a fake loader that answered any string,
and its comment claimed that loader was the one the click reaches. Both were
false after this fix, and node zero of the replication corpus is a source node,
which has no model at all. It now uses a transformation node's label and a
loader that refuses any name the real route would refuse.

## What this does not close

Rows in TypeScript and rows in Rust both fail when a kind is unclassified. That
is two tests, not one contract. `DagNodeOutput.kind` is a plain `string` in the
generated types, so a kind added in Rust cannot break the TypeScript build.
Closing that properly means an enum on `DagNodeOutput`, which drags in the full
`just codegen` and `just regen-fixtures` cascade. It is not in this PR. The Rust
test does not protect the click path; it protects the route's side of it.

## Red team

The Codex plugin, with `gpt-5.6-terra` as a second model. Round one returned
**NOT FULLY CLOSED** and it was right: the new Rust test could pass on timing
alone, because `/health` and `/dag` both answer before the project has
compiled while the model routes answer `503` until it has. Fixed, and
dispositioned in a comment on this PR. Round two is reviewing that fix.

Closes #1857


hugocorreia90:
## Red team, round one: NOT FULLY CLOSED

Reviewer: the Codex plugin, with `gpt-5.6-terra` as a second model. Range
`8e26660e..4c92824c`. It found one blocking defect. **It was right, and it is
fixed in `2c3a703d`.**

### The block: my new test could pass on timing alone

```
GET /api/v1/health   no gate                 -> 200 at once
GET /api/v1/dag      gates on config_path    -> 200 at once
GET /api/v1/models   gates on compile_result -> 503 until compiled
serve.rs             sleeps a fixed 100 ms for the compile it spawned
```

The test waited for health, asserted the DAG, then asked the model routes.
A compile slower than 100 ms answers the first model request
`503 engine_not_ready`. That is neither the `200` nor the `404` asserted, so
**every** assertion in the loop fails on timing, not on behaviour.

I checked all four call sites before accepting it. `full_dag` never reads
`compile_result`, and `get_model` returns `engine_not_ready` before it can
return `model_not_found`. So the race is real. One passing run proved nothing.

The test now waits for `GET /api/v1/models` to answer `200` first. That route
reads the same `compile_result` the per-model route reads, so it is the
readiness the assertions need.

### The nit: the fixture could drift silently

Nothing regenerates the captured fixture. Taken: the expected kind set is now
read from that fixture instead of written out a second time in Rust. A live
server that stops emitting a kind the fixture carries, or starts emitting one
it does not, now fails the test. Mutation-checked: dropping one kind from the
comparison fails it.

### Points 1 to 4: HOLDS

The keyboard path resolves the nearest wrapper and cannot open the wrong node.
An unknown kind is refused at all three call sites, fail-closed. Per-node
`focusable` and `selectable` are honoured, and `onNodeClick` still fires for an
unselectable node, so both gates are needed. The Rust test fails loudly, not
silently, if the scaffold changes.

### One thing I did not do

The readiness latch has no mutation check. Removing it restores a race that
usually passes, so a mutation check would prove nothing. Eleven other guards
are checked.

Also corrected here: the file's own doc named the `Browser UI` job as the one
that runs it. That job is node only. The job is `Test`.

hugocorreia90:
## Red team, round two on the fix: PASS WITH NITS

Same reviewer, on the delta only. The readiness latch, the CI doc correction
and the merge all **HOLD**. It confirmed what I had also checked: the list and
detail routes gate on the same `compile_result`, this test passes no `--watch`,
and `/dag` never recompiles, so nothing here can drop the result after the
latch observes it.

### The nit, taken

My comparison reduced both DAGs to a set of **kinds**. A live server could add
a node, drop one, or rename one and still match, while my assertion text said
the two agreed about their nodes. The text claimed more than the check did.

```
before   {seed, snapshot, source, load, quality, transformation, test}
after    {(transformation, transformation:raw_orders, raw_orders), ...}
```

Now compared per node as `(kind, id, label)`. Mutation-checked: renaming one
node fails it, which the kind-set version passed.

Deliberately still not compared: edges, targets and the engine version string.
That is release churn, not drift the SPA can see, and those three fields are
what the SPA actually reads from this fixture. The reviewer's open question was
whether full-payload parity was the intended contract. It is not, and the code
now says so.

PR #1861, opened 2026-09-10
feat(engine/rocky-cli): rocky history --run <id>, and the second batch of producer seams

hugocorreia90:
Second and last batch of the C1-P1a producer seams (`rocky-plans/plans/ui-control-plane-2026-09/C1-producers.md`), after #1860. No route is added; every read verb the plan lists now has one function a route can call for the document the CLI prints.

**Seams.** Each `run_*` is now "call the seam, render, decide the exit code"; the JSON branch is the seam plus one print, so the CLI bytes are unchanged.

| Verb | Seam | Returns |
|---|---|---|
| `replay --check` | `compute_replay_check(state_path, target, model_filter)` | `ReplayCheckOutput` |
| `policy check` | `compute_policy_check(config_path, models_dir, principal, capability, model)` | `PolicyCheckOutput` |
| `policy test` | `compute_policy_test(config_path)` | `PolicyTestOutput` (`run_policy_test` still exits 1 after printing when a scenario fails) |
| `branch list` | `compute_branch_list(state_path)` | `BranchListOutput` |
| `branch show` | `compute_branch_show(state_path, name)` | `BranchOutput` |
| `estimate` | `compute_estimate(config_path, models_dir, pipeline, model_filter)` (async) | `EstimateReport { output: EstimateOutput, skipped, explain_failed, pricing_source, pricing, matched }` |

`estimate` returns a report, not the bare output, because its text shows what the JSON deliberately omits: SQL-generation skips, EXPLAIN failures and the pricing table. Its text is unchanged too: the `! <model> — explain failed` lines were printed inside the loop and the table after it, and the renderer prints them in that order. All seams are exported from `rocky_cli::commands`.

**One new selector: `rocky history --run <id>`.** No CLI command emitted a single run, so a route for one run had no oracle (ROADMAP: "parity needs an oracle"). `--run` prints one run in the list's own envelope, `count` 1, `--audit` composing; an id the store does not hold is an error naming it, with nothing on stdout. Conflicts with `--model`, `--recipe`, `--since`. Same shape as the list on purpose: a reader of `history --output json` reads one shape either way, and nothing in `output.rs` changes, so no schema, binding or fixture regenerates. The plan wrote "one run → `RunHistoryRecord`"; this is that record inside the existing envelope rather than a new top-level output. If C1-P1b wants a bare record for `/runs/{id}`, that is one new struct and a codegen cascade, deferred until a route asks for it.

**Tests.** One seam test per module: replay check serves the recorded run, filters, refuses an unexecuted model; policy check serves the decision for an uncontracted agent apply (default posture, no rule matched); policy test reports a passing and a failing scenario without exiting; branch list and show serve a created branch and refuse an unknown name; estimate (behind `duckdb`) reports no models in both shapes and refuses an unknown filter; history serves exactly one run with and without `--audit` and refuses an unknown id. The binary-stdout parity test (`engine/rocky/tests/producer_seams_stdout.rs`) now covers all ten commands: trace, cost, compliance, replay check, branch list/show, policy check/test, estimate, `history --run` (plus the unknown-id refusal: exit 1, id on stderr, empty stdout). Where the text or exit path has structure of its own it is pinned too: `policy test` with a failing scenario prints the full report, then exits 1 with the unchanged reason; `estimate` on a DuckDB project with three models (`bad` fails EXPLAIN, `skip` has no DuckDB SQL, `good` is estimated) prints the failure line before the table and the skip only under `--verbose`, with the JSON holding only `good`; `history --run` in text mode prints a run table byte-equal to the list's and, with `--audit`, the governance table after it.

**Mutation checks** (`scripts/mutation-check.sh`): ignoring `audit` in `history_run_output` fails the history seam test; making every policy scenario pass fails the policy test seam test; printing compact JSON from `run_branch_list` fails the stdout parity test; dropping the empty-report message from `compute_estimate` fails the estimate seam test; exiting before printing in `run_policy_test` fails the failing-scenario binary test; printing EXPLAIN failures after the table fails the mixed-outcomes binary test. All six: the test failed under mutation.

**Gates run locally.** `cargo fmt --check`, `cargo clippy --all-targets --all-features -D warnings`, `cargo test -p rocky-cli`, `cargo test -p rocky --test producer_seams_stdout`.

**Codegen.** No `*Output` shape changes; no schema, binding or fixture regeneration. `EstimateReport` is not serialised.

**Docs.** `rocky history` flag table and an example. Changelog: one Added entry, placed at the bottom of the section so a parallel PR adding at the top never shares the hunk.

## Red team

Codex (GPT-5), two passes.

**Pass 1** (marker `ROCKY-C1-P1A-SEAMS-2-2026-09-10`): **REJECT** on coverage and docs; every seam was CONFIRMED behaviour-preserving (JSON, text ordering in all four estimate cases, exit codes, clap conflicts probed live, exports, no codegen). Six adopted, each verified before the fix:

| # | Finding | Verified | Fix |
|---|---|---|---|
| 1 | The policy binary test used only passing scenarios, so exiting before printing would survive | true | a failing-scenario run: full report on stdout, exit 1, unchanged stderr |
| 2 | Estimate was exercised only with no models | true | a three-model DuckDB fixture: EXPLAIN failure, SQL-generation skip, one estimate; text order and `--verbose` pinned |
| 3 | `history --run` was checked as JSON only | true | text test: run table byte-equal to the list's, audit table after it |
| 4 | The replay test derived `replayable` from `replayable_count` | true | asserts the seeded truth: no artifacts, count 0, not replayable |
| 5 | The `--run` doc example sat before a two-run JSON document | true | moved after the list output |
| 6 | Stale rustdocs on `compute_policy_check`/`compute_policy_test`, `model_history_output`'s doc attached to the new fn, a malformed changelog sentence | true | corrected |

**Pass 2** (marker `ROCKY-C1-P1A-SEAMS-2-FIX-2026-09-10`) on the fix delta: **REJECT** on one point, five CONFIRMED (the mixed estimate fixture and all its named mutations; the independently built run table and audit order; the replay seeded truth; the doc placement; the rustdocs and changelog). The point: the failing-policy binary test accepted any non-zero exit and any stderr containing the old substring. Verified (the binary exits 1 and stderr is exactly `Error: 1 of 2 policy scenario(s) failed`, captured from a probe run) and fixed in a third commit: the test asserts exit code 1 and the whole of stderr. Not re-reviewed: a test-assertion tightening Codex itself specified, no behaviour.


hugocorreia90:
The one Test failure on this PR (`commands::review::tests::the_entry_offers_no_read_where_the_routes_front_door_is_shut`, green on rerun) is the same flake #1855 hit. Cause and fix: #1869.

Issue #1859, opened 2026-09-10
Browser UI: the DAG node glyph and accent key on kinds the engine never sends


hugocorreia90:
The DAG node card colours and labels itself by kind. Its table names kinds the engine does not emit.

`engine/ui/src/estate/ModelNode.tsx`, `kindClass` and `kindGlyph`:

```
the card asks for      the engine sends
  "source"       <-->    "source"        match
  "load"         <-->    "load"          match
  "model"                 —              never sent
  "view"                  —              never sent
  "materializedview"      —              never sent
   —                     "transformation"  falls to the default
   —                     "quality"         falls to the default
   —                     "snapshot"        falls to the default
   —                     "seed"            falls to the default
   —                     "test"            falls to the default
```

`NodeKind` in `engine/crates/rocky-core/src/unified_dag.rs` has eight variants:
`source`, `replication`, `transformation`, `quality`, `snapshot`, `load`,
`seed`, `test`.

## Effect

A transformation node shows the default dot glyph and a grey accent. That is
the one kind a user can click, after #1857. Five of the seven remaining kinds
look identical to each other.

## Expected

One glyph and one accent per engine kind. The same table that
`engine/ui/src/estate/nodeRoute.ts` already holds, or built from it.

## Notes

Found while fixing #1857 (PR #1858). Cosmetic, so it was left out of that PR
on purpose. `nodeRoute.ts` now gives a single place the kind list is written
down, which is the natural thing for this to reuse.

hugocorreia90:
Grounded from both ends now, and visible on screen.

**The producer.** The eight kinds the engine can send are `source`,
`replication`, `transformation`, `quality`, `snapshot`, `load`, `seed`, `test`
— the `Display` impl for `NodeKind` in `rocky-core`'s `unified_dag.rs`. The DAG
route copies the node's kind straight through, so there is no translation layer
between those strings and the page.

**The consumer.** `kindClass` and `kindGlyph` key on `model`, `view` and
`materializedview`. None of the three is a value the engine emits.

**What that means in practice.** A live estate realistically shows `source`,
`transformation`, `load`, `seed`, `test`, and sometimes `quality` or
`snapshot`. Of those, only `source` and `load` match an arm. Everything else
falls to the default dot and grey accent — including `transformation`, which
after #1858 is the one kind a user can click.

Confirmed on a rendered page rather than inferred: on a fresh `rocky playground`
served with `rocky serve --ui`, all three transformation nodes draw with the
[...]

PR #1863, opened 2026-09-10
fix(engine/ui): a tab with no token asks the engine for nothing

### LINKED ISSUE #1862: Browser UI: a tab with no token fires four doomed requests, and the empty states do not say what fills them
Two problems on a project that has nothing in it yet. Both are what a new user sees first.

## 1. A tab with no token fires four requests it cannot make

Open `/ui/` without the token `rocky serve --ui` prints:

```
  No token for this tab                     <- the true thing, said once
  REFUSED (401) unauthorized ...            <- Project
  REFUSED (401) unauthorized ...            <- DAG
  REFUSED (401) unauthorized ...            <- Runs
  REFUSED (401) unauthorized ...            <- Schedule
```

`App.tsx` gates only the engine panel. Every lane still mounts its panels, and
every panel still fetches. Four refusals describe the same single fact, and
together they read as a broken install.

It is not only the estate lane. The review and governor lanes mount
resource-backed screens the same way, so a gate on one lane leaves the other
two reachable from the same page.

## 2. The empty states do not say what fills them

Honest, but a dead end. And the obvious copy is false:

- **Review** — the queue is not AI-only. `rocky backfill --model <name>` puts a
  plan in it, and backfills are always review-gated. The prerequisite is that
  the model compiles, not that anything has run.
- **Products** — no `rocky product` subcommand creates a spec file. But saying
  "nothing creates one" is also wrong: `rocky fulfill` stages one.

## 3. Smaller things on the same page

- The engine block spends the top of every screen on three cards, including all
  19 capability tokens joined into a paragraph.
- Estate prints `GET /api/v1/project` beside each heading. The governor tabs
  already solved this with a tooltip.

## Expected

- A tab with no token issues zero requests, on every lane, and says one thing.
- The engine block is one line and still carries version, schema version and
  capability count.
- The review and products copy name a real command and its real prerequisite.
- Estate's route labels move into a tooltip.

hugocorreia90:
Fixes #1862.

## The page with no token

```
  before                                after
  ──────────────────────────────        ──────────────────────────────
  No token for this tab                 No token for this tab
  REFUSED (401) unauthorized            (nothing else, and no request)
  REFUSED (401) unauthorized
  REFUSED (401) unauthorized
  REFUSED (401) unauthorized
```

Four refusals described one fact. Only the engine panel was gated; every lane
still mounted its panels and every panel still fetched.

**Where the gate went, and why.** One boundary in `App`, above the lane. This
is the answer to row 2. A lane cannot gate itself: returning after its
`useResource` calls is too late, the loads have started, and returning before
them makes the hooks conditional. One boundary is also the only shape that
stays true when a fourth lane is added — three per-lane gates would let that
lane fire requests nobody notices. The other two designs the brief offered
make row 1 true three times over instead of once by construction.

## The rest of the page

| | before | after |
|---|---|---|
| engine block | three cards, 19 capability tokens as a paragraph, above every screen | one line: `rocky 1.74.0 · state schema v23 · 2 capabilities`, list in a tooltip |
| Estate headings | `GET /api/v1/project` printed beside each | in the heading's `title`, the idiom the governor tabs already used |

## The copy, probed before it was written

The brief warned the obvious copy is false. It is, in both places, so both were
reproduced against a real `rocky serve` first.

**Review.** `rocky backfill --model customer_orders` on a fresh playground:

```
GET /api/v1/review/queue  ->  "total": 1
                              "capability": "backfill"
                              "principal":  "agent"      <- typed by a person
```

Two things that would have been wrong. The prerequisite is **not** "run it
first": the same command works on a project that has never run, only the cost
estimate changes basis from `historical_observed` to `unavailable`. The real
prerequisite is that the model compiles — an unknown one exits with
`unknown model(s) requested for backfill`. And the queue is **not** AI-only:
the row above says `principal: "agent"` for a plan a person asked for, so
copying that field into the page would have said something false.

**Products.** `rocky product --help` lists verify, compile, status, list,
journal, approve. None creates a spec. But "nothing creates one" is also wrong:
`rocky-fulfill` writes `candidate spec products/<name>.toml written (<digest>);
awaiting approval` (`machine.rs:912`). The copy names both routes.

Every `rocky` command the UI names was checked to exist: review, product,
backfill, serve, fulfill, compile.

## Tests

```
```

Row 1 renders the **real** lanes, not slots — a slot would prove nothing about
the screen that actually reads the API — with a counting `fetch` stub, once per
lane. It also asserts the count is non-zero **with** a token, so the zero
assertion cannot pass vacuously.

**Fourteen** mutation-checked guards (`scripts/mutation-check.sh`, clean
tree): the boundary, the production token default, an empty stored token in
both readers, the one-line engine block, the review prerequisite, both product
routes, the fulfill prerequisite, the "still registers the product" clause, the
hidden descriptions, and the Estate tooltip.

Two of those guards exist because a mutation check failed first: the fulfill
clause and the bootstrap fallback were both fixes with no test behind them, and
deleting either left the suite green.

## Row 3's screenshot

Taken. I first reported this as undeliverable — the Chrome extension is not
connected — then found a headless Chrome already cached on the machine from
some earlier playwright work, which runs standalone with no npm dependency and
no `package.json` change.

Against a real `rocky serve --ui` on a fresh playground:

- **No token.** One message. The four `REFUSED (401)` cards are gone.
- **With a token.** `rocky 1.73.0 · state schema v29 · 19 capabilities` on one
  line where three cards used to sit, and the panel headings carry no route
  text.

The mechanical assertion stays as well, since it is the part CI can run: the
engine block is a single `<p>` whose visible text, with hidden content
removed, is exactly that line, and no grid survives.

The screenshot also shows #1859 rather than describing it: every transformation
node renders with the generic dot glyph, because the node card's colour table
keys on kinds the engine never sends. Not in this PR, but it is the first
screen a user sees.

## Red team

Codex plugin with `gpt-5.6-terra` as a second model. **Two rounds so far, both
NOT FULLY CLOSED, eight findings, all eight real and all eight fixed.** A third
round is running on the latest delta. Each round is dispositioned in a comment
on this PR. The two that mattered most:

- **The gate had a hole.** A stored empty-string token passed my check but is
  falsy to `apiGet`, so every lane would mount and fetch with no credentials —
  rebuilding the exact wall of refusals this PR removes.
- **The copy contradicted the screen it sits on.** "Does nothing without one"
  was false: driverless `rocky fulfill` stages no file but registers the
  product, so the name appears in that very list. Verified at the route the
  page reads, not just at the CLI.


hugocorreia90:
## Red team, round one: NOT FULLY CLOSED

Codex plugin with `gpt-5.6-terra` as a second model. **Five findings, all five
real, all five fixed** in `51daed65` and `d20353f0`.

### 1. The gate and the request sender disagreed about `""`

```
sessionStorage["rocky.ui.token"] = ""
  App          "" !== null   ->  mounts every lane
  apiGet       "" is falsy   ->  sends NO Authorization header
  result                     ->  the wall of 401s, back again
```

The one thing this gate exists to prevent, reachable through a stored empty
string. `currentToken` now applies the same zero-length rule
`tokenFromFragment` already applied on the way in.

### 2. Nothing covered the production token source

Sharp and correct. Every boundary test passed `token` explicitly, so replacing
the default parameter with a sentinel left **all six green** while the real
page mounted lanes with empty storage. I ran the reviewer's own mutation: it
now fails. Three tests render a bare `<App />` against real `sessionStorage` —
a stored token, cleared storage, and the empty string.

### 3. The products copy overstated `rocky fulfill`

Reproduced on a default project:

```
$ rocky fulfill demo
blocked: ... no [fulfill.driver] is configured in rocky.toml
```

Following the copy created nothing. It now names that prerequisite. This is
the second time in this package the obvious copy was false, which is what the
brief warned about.

### 4 and 5. The tooltips had become hover-only

Fair. A `title` is not in the tab order and screen readers treat it
inconsistently, so moving the capability list and Estate's producer routes
into one took them away from anyone not using a mouse. Both are now also in
the accessibility tree as `sr-only` text. Estate's sits **outside** the
heading, so navigating by heading still announces "DAG" and not
"DAG GET /api/v1/dag".

### What the mutation checks caught that I had missed

Guard 11 failed on the first run: deleting the new `[fulfill.driver]` clause
left the products test green. The fix was unguarded and could have been
dropped later without a failure. Now asserted. Ten guards pass.

### Still open, and not fixable here

Row 3's screenshot. This environment has no browser. The claim is pinned
mechanically instead — the engine block is one `<p>` whose visible text, with
`sr-only` content removed, is exactly `rocky 1.74.0·state schema v23·2
capabilities`. Someone with a browser should still look at it.

hugocorreia90:
## Red team, round two: NOT FULLY CLOSED

Three findings on the fix delta. **All three real, all three fixed** in
`c2385837` and `97c8c7e1`.

### 1. The bootstrap reader still disagreed about `""`

I fixed `currentToken` in round one and left `bootstrapToken` reading storage
directly. Production discards that return so it was unreachable, but it is an
exported function and the disagreement was the whole finding. It now goes
through `currentToken`.

### 2. "Does nothing without one" was false, and false on this very screen

The sharpest of the three. Reproduced on a fresh playground:

```
rocky fulfill demo          (no [fulfill.driver] configured)
  products/                 does not exist       <- nothing staged
  rocky product list        0 -> 1, "demo"       <- but something happened
```

The command takes the product's ownership and persists a `blocked` record, and
the product list includes state-store records as well as spec files. So a
reader following my copy would run the command, refresh Products, and watch a
name appear on the screen that had just told them nothing would happen.

It now says it cannot stage a candidate without a driver, and that running it
anyway still registers the product. The old assertion was locking the false
claim in, so it asserts the opposite now.

This is the third time in this package that the obvious copy was false. The
brief warned about exactly that, and I still needed two rounds to get it right.

### 3. The route and the capability list were announced twice

An otherwise unused `title` already becomes the accessible description, so the
`sr-only` sibling I added in round one was a second node in the reading order.
Both now use `aria-describedby` pointing at a `hidden` element, which
supersedes `title` and is not read in flow. The mouse tooltip is unchanged.

### What the mutation checks caught

Guard 12 failed on the first run: reverting `bootstrapToken` left the token
tests green. The reviewer had asked for that test by name and I had fixed the
code without it. Three cases now cover it. **Fourteen guards pass.**

### Still open

Row 3's screenshot. No browser exists in this environment. The structure is
pinned mechanically; the appearance is not.

hugocorreia90:
## Red team, round three: PASS

No findings in the delta. All three round-two findings confirmed closed.

- **The token readers agree.** Reverting `bootstrapToken` fails the new test.
  The only other direct reader of the storage key is `apiGet`, whose
  truthiness check also rejects `""`.
- **The fulfill copy is accurate.** Acquisition persists a fulfillment record
  before the missing-driver path runs, and the product listing unions those
  records with spec filenames, so the driverless name lists with no spec. The
  reviewer checked the cases that would have made the new sentence wrong in
  the other direction and found none that contradict it.
- **One announcement, not two.** A directly referenced hidden node does
  contribute to `aria-describedby`, and description computation uses
  `aria-describedby` before `title`, so `title` is no longer added as a second
  description. The ids are unique and collide with nothing.
- **The merge lost nothing.** All five main-side files are byte-identical, and
  the eight UI files are unchanged from the first parent.

### Three rounds, eight findings, all mine

Round one found five, round two three, round three none. Every one was real.
The two worth remembering: a stored empty-string token would have mounted every
lane and fetched with no credentials, rebuilding the wall of refusals this PR
removes; and the products copy contradicted the very screen it sat on.

Two of the fourteen guards exist only because a mutation check failed first —
the fulfill clause and the bootstrap fallback were each a fix with no test
behind it.

### What is still unverified

The reviewer could not run a browser or a screen reader, and neither could it
run vitest (its sandbox refuses the jsdom temp directory). The accessibility
behaviour is verified against the normative computation, not against VoiceOver
or NVDA. I have since taken screenshots with a headless Chrome, which settles
the rendering claims in the PR body but not the assistive-technology ones.

PR #1876, opened 2026-09-10
fix(engine/rocky-core): an unreadable pause state defers a spooled webhook demand

### LINKED ISSUE #1831: The reconciler's pause gate fails open: a cursor read error runs a paused pipeline's spooled webhook demand
Found by the independent review of #1830 (the #1812 fix), round four, while checking what the new spool alert may promise about paused pipelines.

## The fail-open

When the reconciler consumes a spooled webhook demand it asks the state store whether the pipeline is paused, and reads the answer as `get_schedule_state(...).ok().flatten()` (`engine/crates/rocky-core/src/schedule/reconcile.rs`, the pause gate before `execute_demand`). A read *error* — the redb read transaction failing to open, the table failing to open, the JSON failing to decode — is therefore the same as "not paused". There is no second pause check before the claim transaction, so a transient read fault that clears before the claim lets the demand claim, spawn and, on success, dispose of its pending file: a paused pipeline runs.

The existing test (`reconcile.rs` ~3277) proves only the readable case: a cursor with `paused = true` records a skip and leaves the file untouched.

## What should happen

An unknown pause state must not read as "not paused". Fail closed: an error reading the cursor defers the demand (a skip with its own reason, e.g. `history_unavailable`, which already exists for the analogous read failure elsewhere) and leaves the file for the next tick. A test that injects a decode failure (a malformed cursor blob) and asserts the demand stays pending pins it.

## Why it is filed rather than fixed in #1830

#1830 is the metrics vocabulary and the alert; this is the reconciler's own gate and a behaviour change to what a tick does with a demand. The alert's wording was reduced to "subject to each pipeline's pause state" so it promises nothing this code cannot keep.

Related: #1812, #1830, #1716 (the pause/tombstone ordering).

hugocorreia90:
Closes #1831.

The reconciler's pause gate read the schedule cursor as `.ok().flatten()`, so a read **fault** was indistinguishable from a live cursor saying "not paused":

```
get_schedule_state(...)      gate saw        should be
  Ok(Some(paused: true))     paused          paused
  Ok(Some(paused: false))    not paused      not paused
  Ok(None)                   not paused      not paused
  Err(store fault)           not paused  ✗   unknown -> defer
```

There is no second pause check before the claim. So a transient fault that cleared before the claim let a **paused pipeline claim, spawn, and dispose of its pending file** — the webhook bypass #1334's red team found, reachable again through an error path.

## The justification for the old behaviour does not hold

The comment argued that failing closed would "hold every webhook hostage to a transient store error".

That conflates **deferral** with **loss**. A deferred demand leaves its pending file untouched and is retried next tick — which is exactly what `Paused` already does. An accepted (202'd) delivery is dropped in neither case.

So the trade is: one tick of latency on a store fault, against running a pipeline an operator explicitly paused.

## Reusing `HistoryUnavailable`, not adding a variant

`TickSkipReason` mirrors the frozen tick contract, is rendered as a stable string at two sites, and is asserted against `span_attrs::SCHEDULER_SKIP_REASONS` by an exhaustiveness test. Adding a variant is a user-visible contract change.

`HistoryUnavailable` already names this class — "a store read fault that makes the classification unreliable, surfaced loudly rather than silently dropped" — and the analogous history read at `reconcile.rs:841` already fails closed the same way. Its doc is widened to say it covers both reads rather than leaving the second use unexplained.

## Evidence

The decision is extracted into `webhook_pause_gate` so it can be tested directly. Injecting a store fault through the real `StateStore` would need a corrupt-row write helper that does not exist, and the function *is* the whole decision.

**Mutation-checked.** `scripts/mutation-check.sh` restoring the fail-open arm:

```
test an_unreadable_cursor_defers_a_spooled_demand_instead_of_running_it ... FAILED
```

Three tests. The third — `a_live_cursor_and_an_absent_one_both_run` — is what keeps the fix honest: without it, "always skip" would defer every webhook forever and still satisfy the fault test.

`cargo test -p rocky-core --lib` — 2185 passed, 0 failed. Clippy `--all-targets --all-features` clean, `cargo fmt --check` clean.

Whether a *persistently* unreadable cursor now strands a demand indefinitely. The file stays pending and each tick records a skip, so it is visible rather than silent — but nothing escalates it, and a permanently corrupt cursor row would defer that pipeline's webhooks forever. `StateBusy` has the same property today, so this is consistent rather than new; it is still a shape worth a bound.

Only the webhook path is fixed. The cron/standing path resolves its pause state elsewhere, and I did not audit whether that read has the same `.ok().flatten()` shape. The two should agree — an unreadable pause state is unreadable regardless of what asked — and if the other path fails open too, this PR closes half a class. Worth a follow-up grep rather than an assumption either way.

**Independent red team: not run.** The Codex plugin cannot return a verdict into an autonomous loop, so this is self-reviewed at maximum effort and disclosed as such, per `AGENTS.md`. The defect itself was found by an independent review of #1830.

Refs #1830, #1334, #1812


hugocorreia90:
Follow-up done, and it found the other half.

This PR's "what I think is being missed" said the cron/standing path should be grepped rather than assumed. It has the **same fail-open**, filed as #1877.

`ScheduleStateView::get` (`demand.rs:260`) returns `ScheduleStateRecord` with no way to express failure, and the live impl at `reconcile.rs:822` collapses a fault into `.unwrap_or_default()`. `evaluate_one` then gates on `cursor.paused` at `demand.rs:516` — so a read fault runs a paused pipeline on the primary scheduling path.

It is worse there than here: the default discards the **whole** cursor, so a fault also resets the failure-backoff ladder and the catchup anchor. A pipeline backing off is doing so precisely when the store is under stress.

Kept out of this PR deliberately — that fix changes a trait signature and threads a new skip reason through `evaluate_one`/`evaluate_demands`, which is a different blast radius from a local `Result` at one call site.

PR #1885, opened 2026-09-10
fix(engine/ui): the page on a phone, the graph beside an open pane, and a minimap that drew nothing

### LINKED ISSUE #1881: Browser UI: the graph is cut off on a phone, four tables push the page sideways, and the minimap draws no nodes
The browser UI had only ever been viewed in dark mode at 1568×724. Looking at 320px, 375px, and at the detail pane open, found five defects. All measured rather than eyeballed.

## 1. The DAG graph is cut off, not fitted

`fitView` cannot zoom out past `minZoom`, and the library default is 0.5:

```
three nodes span  3×184 + 2×96          = 744px
320px viewport    -> ~288px of canvas   -> needs zoom 0.39
                     clamped at 0.50    -> both end nodes off-canvas
```

A deeper DAG hits the same wall at wider viewports.

## 2. The graph never re-fits

`fitView` runs at mount only. Opening the detail pane takes the canvas from
990px to 620px at a 1024px viewport, and the graph keeps its old zoom, so the
last node leaves the canvas. Rotating a phone does the same.

## 3. Four tables push the whole page sideways

Runs, Schedule, and the two governor section-card tables have no scroller, so
the widest cell moves the page body — every panel above and below it included.
`SamplePanel` and `ProductsScreen` already wrap theirs; these four were missed.

## 4. The governor tabs overflow at 320px

Five tabs do not fit. Measured at 3–4px past the page body on every governor
route. The lane nav, which is the one that gets flagged for this, fits.

## 5. The minimap has never drawn a single node

The starkest one. Its SVG contains **zero** `.react-flow__minimap-node` rects,
in both themes, while its `viewBox` spans the correct 857px of node bounds. So
it knows where the nodes are and draws none of them.

`NodeComponentWrapperInner` skips any node failing `nodeHasDimensions`, which
reads `measured`/`width`/`height` off the node object the app passes. The graph
is controlled (`nodes={flow.nodes}`) with no `onNodesChange`, so React Flow has
nowhere to write a measurement back to, and those stay `undefined` forever.

## Expected

- Every lane and every governor tab: no horizontal overflow of the page body at
  320px and 375px.
- The graph fits its canvas, and re-fits when the canvas changes width.
- The minimap draws the nodes, in both themes.

## How to check it

No Chrome extension is needed. A headless Chrome is cached from playwright and
runs standalone; Node 22's global `WebSocket` makes a CDP client dependency-free.
`scrollWidth − clientWidth` is the overflow, and walking up from each too-wide
element while ignoring any with a scrolling ancestor names the real culprit.
Note that `--force-prefers-color-scheme` does nothing — it produces a
byte-identical light capture. `Emulation.setEmulatedMedia` is the one that works.

hugocorreia90:
Fixes #1881.

The page had only ever been viewed in dark mode at 1568×724, and the brief said
so. Looking at 320px, 375px, and at the detail pane open found five defects.
Four of the five were not in the brief.

## 1. The graph was cut off, not fitted

```
three nodes span  3×184 + 2×96         = 744px
320px viewport    -> ~288px of canvas  -> needs zoom 0.39
                     minZoom is 0.50   -> both end nodes off the canvas
```

`fitView` cannot zoom out past `minZoom`. The floor is a real limit, not a
preference, so it is now 0.05; a deeper DAG needs it lower still.

## 2. The graph never re-fitted

`fitView` runs at mount. Opening the detail pane takes the canvas from 990px to
620px at a 1024px viewport, and the graph kept its old zoom, so
`revenue_summary` simply left the canvas. Rotating a phone does the same. A
`ResizeObserver` now re-fits on a width change.

**This is the case the brief asked about**, and it could only be seen with the
pane actually open — which needed a real click, not a URL.

## 3. Four tables pushed the whole page sideways

Runs, Schedule and the two governor section-card tables had no scroller, so the
widest cell moved the page body, every panel above and below it included. They
scroll themselves now, the idiom `SamplePanel` and `ProductsScreen` already used.

## 4. The governor tabs overflowed

Five tabs do not fit 320px: measured at 3–4px past the page body on every
governor route. They wrap now. The brief predicted this for the **lane** nav.
That one fits. The tab nav it did not mention is the one that did not.

## 5. The minimap had never drawn a single node

The starkest one. Its SVG held **zero** `.react-flow__minimap-node` rects, in
both themes, while its `viewBox` spanned the correct 857px of node bounds — so
it knew where the nodes were and drew none of them.

`NodeComponentWrapperInner` skips any node failing `nodeHasDimensions`, which
reads `measured`/`width`/`height` off the node object the app passes. The graph
is controlled with no `onNodesChange`, so React Flow has nowhere to write a
measurement back to. `layout.ts` already knew both numbers.

```
before: nodeCount 0   viewBox spans 857px
after:  nodeCount 3   fills rgb(43,43,43) dark / rgb(226,226,226) light
```

## Measured, not eyeballed

| | 320px | 375px |
|---|---|---|
| estate, review, governor | 0px over | 0px over |
| the four governor tabs | 0px over | 0px over |

Overflow is `scrollWidth − clientWidth`, and the culprit is found by walking up
from each too-wide element and ignoring any with a scrolling ancestor — which
names the offender rather than the widest box. A viewport screenshot cannot
answer this at all: a page that scrolls sideways looks identical to one that
clips.

Row 3, with the pane open: at 1024px the pane is 360px and the canvas 620px,
side by side; at 1280px, 360 and 748. Below the breakpoint it stacks, canvas
first, then detail. **The `lg` breakpoint is confirmed, not moved.**

Row 2, as rendered rather than as a passed prop: `.react-flow` carries `dark`
or `light`, the minimap background goes `rgb(20,20,20)` / `rgb(255,255,255)`,
and the controls `rgb(43,43,43)` on `rgb(248,248,248)` / `rgb(254,254,254)` on
near-black.

## Tests

```
```

Two mutation-checked guards, for the two fixes that can silently come undone:
every flow node states its size, and the runs table sits inside a scroller. The
other three need real layout, which jsdom does not do; those stay
screenshot-and-measure evidence, and the measurement is repeatable.

## How the screenshots were taken

No Chrome extension. A headless Chrome is cached from playwright and runs
standalone, so no npm dependency and no `package.json` change. Node 22's global
`WebSocket` makes a CDP client dependency-free, which is what allowed clicking a
node to open the pane and emulating the colour scheme.

`--force-prefers-color-scheme` is inert — it produced a byte-identical light
capture. `Emulation.setEmulatedMedia` is the one that works.

## Red team

Dispatched to the Codex plugin. Findings and their disposition will be posted
here before merge.


hugocorreia90:
## Re-measured on a deeper graph

My evidence used the playground: three nodes, three layers. The brief's worry
was "a graph with many layers", so that was the easy case, not the one it asked
about.

Re-ran everything against a project with **nine nodes across four layers**
(seven of the engine's eight node kinds):

```
320px estate          over=0px, no unscrolled culprits
1024px, pane open     pane=360px  canvas=620px  side by side  over=0px
minimap               9 of 9 nodes drawn
```

Every node fits the 620px canvas beside the open pane, and the minimap draws
all nine. The three-node case would have passed even with the old `minZoom` of
0.5 at that width; this one would not have.

hugocorreia90:
## Red team on this PR: two rounds so far, both NOT FULLY CLOSED

Codex plugin with `gpt-5.6-terra` as a second model. **Six findings, all six
real, all six fixed.** Three of them were defects I introduced while fixing
the ones before.

### Round one, four findings

1. **The zoom floor answered the wrong question.** `minZoom` is the floor for
   the *hand* as well as the fit, so lowering it to fit a phone let a person
   pinch a node down to 9×2 pixels.
2. **The resize observer watched the wrong panel.** `document.querySelector`
   finds the first flow on the page; with two panels both would watch the
   first, and resizing it would reset the other.
3. **The card was not the size I declared it to be.** Measured in a browser,
   which is what the reviewer said would settle it:

   ```
   before   wrapper 184×46   card 184×34   handles 23   card centre 17
   after    wrapper 184×46   card 184×46   handles 23   card centre 23
   ```

   So the node carried a 12px invisible clickable band and its handles sat
   6px below the card's visible centre. That came from my own minimap fix
   declaring a height the card never had.
4. **The scroller I added was itself a barrier.** A scrolling div is not in
   the tab order in every browser, so at 320px a keyboard-only reader could
   not reach the right-hand columns at all. Fixing the page overflow had made
   those columns unreachable another way.

### Round two, one finding and one nit

**My round-one trade was wrong, and it was reproduced.** I split the floors and
kept the gap small, assuming the resulting snap would be rare. It is not:

| layers | fit at 320px | against a 0.1 floor |
|---|---|---|
| 10 | 0.1065 | fine |
| 11 | 0.0965 | snaps |
| 20 | 0.0523 | snaps |
| 42 | 0.0247 | snaps 5×, most of the graph lost |

An eleven-deep model chain is ordinary. **Both rounds are satisfiable at once**,
which beats choosing between them: the interaction floor now *follows* the fit,
`min(MIN_ZOOM, fitted)`. A small graph keeps its 0.1 floor, so round one's
objection still holds. A deep graph lowers the floor to exactly what the fit
needed, so there is nothing to snap to.

Verified on a real 14-layer chain, through the library's own zoom control:

```
fit 0.0680  ->  one zoom step  ->  0.0816   (1.20×)
```

`0.0816` is below the old floor, which is only possible because the floor moved.

**The nit, taken.** `role="region"` on each table scroller added a landmark
inside an already-named section — a full brief would expose nine sections plus
eleven nested regions. They are `role="group"` now, keeping the tab stop and
the name without the landmark noise.

### One thing about my own evidence

My first attempt to prove the zoom fix used a synthetic wheel event that moved
nothing. The ratio came back exactly `1.00`, which reads as "no snap" and
proves nothing at all. Re-run through a path that genuinely moves the zoom.

### Also checked, unprompted

Forcing the card height could have clipped a long label. Added a model named
`a_very_long_model_name_that_will_not_fit_the_card` and measured every node:
no card clips vertically (`44/44` throughout), and the long label truncates on
one line, which is the component's existing behaviour rather than a new
consequence.

hugocorreia90:
## Red team: five rounds, eleven findings, all real

Round five returned **PASS WITH NITS** and the nits are taken. Six of the
eleven were defects I introduced while fixing an earlier one.

### Round four, two findings

1. **The refit identity was the node set, not the layout.** A node's position
   comes from its `execution_layers` index, so adding a `depends_on` between
   two models that already exist re-layers the graph and changes its span with
   an identical id set. Keyed on position now.

   The reviewer also named why my 3-to-14-node trace passed anyway: growing a
   project changes the id sequence, so that scenario could not exhibit this.

2. **The fit control recorded the zoom from before the fit.** The library
   calls `fitView()` then `onFitView` immediately, but that call only *queues*
   the transform. My trace pressed Fit while already at the target zoom, so
   the stale read happened to equal the right answer — **the evidence was
   shaped so the bug could not appear.** The control runs its own awaited fit
   now. Verified on the path the old trace could not reach:

   ```
   fitted        0.0680
   zoomed in to  0.3508   (deliberately well above the 0.1 floor)
   fit again     0.0680
   one zoom-in   0.0816   1.20x, no snap
   ```

   Stale, that would have recorded 0.3508, raised the floor to 0.1, and
   snapped.

### Nothing had guarded any of this

Three rounds went into the zoom behaviour and its correctness rested entirely
on browser traces I ran by hand. The two decisions inside it are pure, so
`floorFor` and `layoutIdentity` are exported and tested, and **both mutation
checks fail when the fix is reverted**. The wiring — which paths call refit —
needs real layout, so it stays browser-verified. I am not claiming the suite
covers it.

### Round five: PASS WITH NITS

No behaviour-blocking defect. All three nits were comments of mine that said
things the code does not do, including one that contradicted the correct
comment directly below it, and one that was the stated justification for the
whole design.

That is the fourth time in this package my prose outran what I had verified,
and by the end it was the only category of finding still appearing. The code
converged after four rounds; the descriptions of it took one longer.

PR #1907, opened 2026-09-10
feat(engine): GET /api/v1/settings reports the server's posture

hugocorreia90:
`GET /api/v1/settings` reports how the running server is bound and what it will accept. Read it to check what a server is actually enforcing, rather than what its start command looked like.

It reports the bind host, the CORS allowlist, the `Host` values the UI guard accepts, whether the scheduler and the UI are on, whether `ROCKY_WEBHOOK_SECRET` can sign a webhook, the token's scope, and the state backend.

No secret is in the response. The token appears as its name and scope. The webhook secret appears only as whether it is usable.

## An allowlist, not a config dump

The document is built field by field, by hand, in `settings_output` — from a snapshot of primitives plus two fieldless enum labels taken out of the config file, after which the config is dropped. No `RockyConfig` is serialised or `Debug`-printed anywhere on that path; `AdapterConfig`'s `Debug` prints its `.extra` map, which is unbounded caller-supplied TOML.

To be exact about what enforces that: the **projection function** does, not the type. Rust would happily let a future field carry a secret. Three tests stand behind it:

- `settings_reports_exactly_the_allowlisted_fields` asserts the field list against a literal, so **adding** a field fails.
- `settings_never_discloses_a_configured_secret` configures three real secrets and greps this document for each.
- `no_safe_route_discloses_a_configured_secret` greps **every safe route** for the same three, error bodies included, and asserts it exercised at least 20 routes so it cannot pass vacuously.

## Three things the obvious implementation gets wrong

**1. The webhook secret cannot be probed with `env::var(..).ok()`.**

The startup gate is two reads deep. `webhook_secret_fail_closed` refuses a blank value; `env_var_fail_closed` beneath it refuses `NotUnicode`. The obvious probe reports a non-UTF-8 secret as *absent* — the display contradicting the gate it describes.

```text
  present            scheduler starts; webhooks must carry a signature
  absent             scheduler starts; loopback accepts UNSIGNED,
                     non-loopback answers 404
  set_but_unusable   scheduler REFUSES to start (blank, or not valid UTF-8)
```

The probe uses `var_os` and never bails, so presence is captured **even when the scheduler is off** — which is the point, since it says what will happen when you turn it on. A parity test asserts the probe and the gate agree across four env states; a second pins that they read the same variable.

**2. `bind_host` must not be able to name a host the server is not on.**

`run_serve` builds `ServeConfig` first and lends the snapshot that same `String`, so the reported host and the bound host are one value rather than two that happen to agree.

**3. Reading `rocky.toml` must not become a new reason to refuse to start.**

Nothing on the path to `TcpListener::bind` reads a file. (The initial compile does read the config, but it is `tokio::spawn`ed, so it never gates the listener.) An eager read for two report fields would put a blocking full-file read *on* that path, so a `rocky.toml` that is a FIFO or sits on a stalled mount would stop the server binding at all.

The labels resolve once, on the first request to this route — and the projection takes them as an argument, so there is no "not read yet" state that could report itself as `absent`.

That read is bounded, because moving a hang is not fixing it. `spawn_blocking` cannot be cancelled and `OnceLock::get_or_init` makes every concurrent caller wait on the one initializer, so unbounded it would starve the blocking pool the state-store reads share. It takes the shape the sample route already uses: its own single permit, refused immediately with `503 engine_busy` rather than queued, and a 5 second deadline, with the permit carried into the blocking stage so a slow read keeps the lane until it returns. At most one worker can ever be stuck, and once resolved every later caller touches neither the permit nor the disk.

```text
  loaded       read and parsed
  absent       no rocky.toml -- an ordinary fact, not a failure
  unreadable   present but would not parse
```

## Both lists report what is enforced

`--allowed-host` only becomes a guard under `--ui`, so `allowed_hosts` is empty without one. `--allowed-origin` values that are not valid header values are dropped when the CORS layer is built, so they are absent from `allowed_origins` too. Both come from the same derivation the enforcement uses, so neither report can drift from the thing it describes.

No token configured is `null`, not a token named `"none"` — the most exposure-relevant answer this route gives, so it is distinguishable rather than something a careless reader takes for a configured token.

## API-only, deliberately

There is no `rocky settings` verb. This describes *a server that is running*, which a one-shot CLI invocation would have to invent. `ScheduleStatusOutput` is the in-repo precedent: registered in `export_schemas.rs`, with no `types.py` bridge and no `RockyClient` method. `sdk/python/AGENTS.md`'s "one typed method per CLI command" invariant is keyed on a CLI command, and there is none here.

So there is no CLI-to-route byte-parity test, which the last two producer PRs had. Standing in for it: the field-list allowlist, the two secret greps, the probe/gate parity test, the malformed-config guard, and a route test pinning the response bytes to the projection.

## Also fixes a guard that is red on main

`vscode-ci`'s `generatedBarrelCompleteness` test fails on `main` today. The generated TypeScript barrel is hand-maintained — codegen writes `types/generated/<command>.ts` but never `index.ts` — and two outputs that already landed were never added: `PolicyRulesOutput` (#1874) and `ScheduleSpoolOutput` (#1899). `SettingsOutput` would have been the third. All three are exported here rather than leaving the guard red for the two already merged.

## Beyond the plan's list

The plan named nine fields. This adds two: `ui`, so an empty `allowed_hosts` is explainable, and `config_status`, so a `null` `state_backend` is.

## Red team

Codex, via the `codex:codex-rescue` plugin, in two passes. Every finding was checked against the code before being adopted.

**Round one** ended without issuing a verdict, so it is recorded as an incomplete review, not a clearance.

| Finding | Verified | Disposition |
|---|---|---|
| `allowed_origins` reports the raw list while CORS drops unparseable origins | Confirmed at `auth.rs:492` | **Fixed.** One derivation feeds both the layer and the report |
| The config read sits on the pre-bind path | Confirmed | **Fixed.** Resolved lazily on first request |
| The labels can go stale against the scheduler, which re-reads each tick | Confirmed; already known | Documented; `config_status` explains the nulls |
| Two "by construction" claims broader than the types enforce | Fair | Prose corrected — the projection and its tests enforce it, not the type |
| The TypeScript barrel omission | Confirmed independently | Fixed; see above |

**Round two** reviewed the fix delta and returned **BLOCK** on two counts. Both are fixed.

| Finding | Verified | Disposition |
|---|---|---|
| The lazy read is unbounded: `spawn_blocking` cannot be cancelled and `get_or_init` makes concurrent callers wait, so a stalled `rocky.toml` starves the blocking pool | Confirmed | **Fixed.** Own single-permit lane, `503` rather than queueing, 5s deadline, permit carried into the blocking stage |
| `SettingsSnapshot` lost public `Clone` on a false premise | Confirmed in std: `impl<T: Clone> Clone for OnceLock<T>` | **Fixed.** Restored |
| A panicked read reported as `config_status: unreadable` | Fair | **Fixed.** Now `500` — a task failure is not a verdict about the config |
| "Nothing reads the config before the bind" is false | **Partly.** `recompile` does read it, but it is `tokio::spawn`ed (`state.rs:404`) and does not gate the listener, so the fix stands. The *claim* was still too broad | **Fixed.** Test renamed to `build_serve_state_does_not_read_the_config`; docs say the initial compile reads it on its own task |
| The route walk proves error paths, not success paths | Fair | Now uses a real plan-id shape so review routes reach their handlers; the doc states what it proves |

**CodeQL** then failed with two high-severity alerts, both in my own test code: the containment assertions interpolated the secret, and the body containing it, into their failure messages. A test written to prove nothing leaks would have written the value into the CI log on failure. The three secrets now live in one constant paired with names, and a failure reports the name.

## Evidence

- **10 mutation checks, all fix-sensitive**: the naive probe spelling; an unreadable config reporting `loaded`; the token field carrying the Bearer secret; an unconfigured token reported as `"none"`; a hard-coded `bind_host`; `allowed_hosts` reporting a wildcard; `allowed_origins` reporting the typed list; an eager pre-bind config read; the admission lane removed; and the fast path removed.
- One check **failed first** and was right to: the CORS test's fixture held only installable origins, so raw and enforced were identical and no assertion could tell a correct projection from one that echoes what was typed. The fixture now carries an uninstallable origin and asserts that it does.
- `cargo test -p rocky-cli -p rocky-server --all-features` green; clippy `--all-targets --all-features -D warnings` and `cargo fmt --check` clean.
- `just codegen` and `just regen-fixtures` produce no drift; vscode `vitest` 319 passing; SDK `pytest` 269 passing.

Whether `state_backend` and `concurrency_control` earn their place. They are the only two fields that need a file read, the only two that can go stale, and every problem in both review rounds came from them. The plan asks for them, so I kept them and bounded the read rather than narrowing the scope on my own. If you would rather they went, the route gets simpler and loses a whole failure class.

## What I think I am missing

The route is authenticated but is a `GET`, so a read-only `--ui` token can read it. That discloses the bind host, the allowed origins, and whether a webhook secret exists — and `webhook_secret: absent` tells the holder that unsigned POSTs are accepted on a loopback bind. I believe that is moot, because the same holder can probe the webhook route directly and learn the same thing. I would rather have that challenged than assume it.

Refs #1874


hugocorreia90:
Rebased onto `main` at `0e36b921` (14 commits, force-pushed with lease). The only conflict was the generated `editors/vscode/src/types/generated/index.ts`; `just codegen` on the rebased tree produced no further diff to any generated file, so the bindings, schema and OpenAPI document already match what the engine emits.

The two CodeQL alerts on this head (186, 187, `rust/cleartext-logging`) point into `mod tests`: the `project_with_three_secrets` fixture writes deliberately fake secrets (`CONFIG_SECRET_XYZ`, `BEARER_SECRET_ABC`, `WEBHOOK_SECRET_DEF`) into a temp `rocky.toml` so the settings route's redaction can be asserted. They are the test's inputs, not a leak; "used in tests" is the honest dismissal. CodeQL is not a required check.

The session that opened this PR is no longer running; I took it over for the rebase only. No code change beyond the rebase.

Issue #1853, opened 2026-09-09
Nothing gates the two hand-maintained binding barrels, so a generated type can be inert


hugocorreia90:
#1852 closes the roster half of #1760: a CLI output whose type derives `JsonSchema` and is registered in no schema, and inlined into none, now fails CI.

Two manual steps sit *behind* the roster, and that gate cannot see either. A type can be correctly registered, have its schema exported and its class generated, and still be unreachable by anyone importing it.

```
output.rs ──▶ schemas() ──▶ schemas/*.json ──▶ generated class
                  ▲                                   │
        #1852 gates here                    ┌──────────┴──────────┐
                                            ▼                     ▼
                          types_generated/__init__.py    generated/index.ts
                             a _LAZY map + __all__        explicit re-exports
                                  hand-maintained         hand-maintained
                                      ▲                        ▲
                                 no gate here            no gate here
```

## What exists today

`sdk/python/tests/test_barrel_completeness.py` asserts every name in `__all__` resolves. So it catches a name in `__all__` with no `_LAZY` entry — but **not** a generated type that is in neither, which is the failure this describes. The TypeScript barrel has no equivalent check at all.

## The trap that makes hand-editing worse than it looks

`justfile:165` restores `types_generated/__init__.py` from HEAD after every codegen pass, because `datamodel-code-generator` overwrites it with an empty stub. So a barrel edit made *before* a codegen run is silently reverted by it. The working order is: run codegen, then edit the barrel.

## What would close it

Both barrels are derived from the same set the schemas are. Something should assert that every generated module exports through its barrel — a Python test comparing `types_generated/*.py` against `_LAZY` + `__all__`, and a TypeScript equivalent comparing `generated/*.ts` against `index.ts`.

Whether that is two tests or one generator that writes the barrels — removing the hand-maintenance and the `justfile:165` restore along with it — is the open question. The generator is the better shape and the larger change.

Refs #1760, #1852, #1699, #1810

hugocorreia90:
## Correcting this issue: its Python claim is wrong, and it understated the TypeScript gap

I filed this. Both halves need correcting, and I found out by trying to build the fix.

### The Python half was already covered — my error

This issue says:

> `sdk/python/tests/test_barrel_completeness.py` asserts every name in `__all__` resolves. So it catches a name in `__all__` with no `_LAZY` entry — but **not** a generated type that is in neither, which is the failure this describes.

[...]

PR #1916, opened 2026-09-11
fix(vscode): re-export the four generated output types missing from the barrel

hugocorreia90:
Four generated output types are unreachable from the VS Code barrel. Split out of #1907 so a release cut does not have to wait on it.

## What is missing

`editors/vscode/src/types/generated/index.ts` is hand-maintained: `just codegen` writes `types/generated/<command>.ts` but never the barrel. So an output's interface can be generated, committed, and still be unreachable by anyone importing from the barrel.

```text
  policy_show.ts     PolicyRulesOutput            #1874
                     PolicyAutonomyBudgetOutput
                     PolicyRuleScopeOutput
  schedule_spool.ts  ScheduleSpoolOutput          #1899
```

Both of those PRs were mine, and both landed their generated `.ts` without the barrel entry.

The nested and enum types from the same two modules are exported alongside them, so a consumer can name what the outputs actually contain.

## The guard exists and has never run

`src/__tests__/generatedBarrelCompleteness.test.ts` was added in #1853 for exactly this failure, and it does catch it:

```text
  before   2 failed | 108 passed
  after    110 passed
```

It has never gated anything. `vscode-ci` runs `npm test`, which is `node ./out/test/runTest.js`. The vitest suite is `npm run test:unit`, and **no workflow invokes `vitest` or `test:unit`**:

```bash
$ grep -rn 'vitest\|test:unit' .github/workflows/
  (nothing)
```

So the whole vitest suite — 318 tests — is inert in CI, and this guard has been decorative since it was written. That is the more useful finding here; wiring `test:unit` into `vscode-ci` is a follow-up, not this PR.

## Verification

- `npm run test:unit` — 318 passed.
- `npm run compile` (`tsc -p ./`) — clean.
- Before/after on the guard as above, so the change is load-bearing rather than assumed.

## Scope

Barrel only, one file, no generated output changed. `SettingsOutput` is deliberately absent: `settings.ts` does not exist on `main` yet, and its barrel entry travels with #1907.

### DIFF
--- a/editors/vscode/src/types/generated/index.ts
+++ b/editors/vscode/src/types/generated/index.ts
@@ -60,6 +60,25 @@ export type {
   ScheduleThrottleKind,
   TickLockState,
 } from "./schedule_status";
+export type {
+  ScheduleSpoolOutput,
+  SpoolCounts,
+  SpoolPendingEntry,
+  SpoolSkippedEntry,
+} from "./schedule_spool";
+export type {
+  PolicyRulesOutput,
+  PolicyAutonomyBudgetOutput,
+  PolicyRuleScopeOutput,
+  PolicyRuleEntry,
+  PolicyFreezeInForce,
+  PolicyFreezeSources,
+  PolicyLedgerSource,
+  PolicyMarkerSource,
+  PolicyEffect,
+  PolicyPrincipal,
+  PolicyCapability,
+} from "./policy_show";
 
 // serve HTTP API — the five estate routes (typed, no CLI counterpart)
 export type { HealthOutput } from "./health";


hugocorreia90:
Filed #1917 for the reason nothing caught this: `vscode-ci` never runs vitest, so `generatedBarrelCompleteness.test.ts` has never gated anything.

Worth recording there and here: #1853 asked for **both** hand-maintained barrels to be gated and is closed by #1875, but only the Python half actually gates — `sdk-ci` runs pytest, and nothing runs vitest. This PR backfills the four missing exports; #1917 is what stops the next four.

Deliberately not fixing the workflow here, so a one-file release-unblocker stays one file.

Issue #1919, opened 2026-09-11
Make a resolved ${VAR} value unprintable: a secret type that renders only as ${NAME}


hugocorreia90:
Follow-up to #1897. Refs #1886, #1878.

## Why

The 1.74.0 fix redacts resolved `${VAR}` values on the way out of `rocky serve`. That closes today's leaks, not the class. A value is still a plain `String` after substitution, so every new field, error variant or route is a new chance to print it.

## What

Resolve `${VAR}` into a secret type whose `Display`, `Debug` and `Serialize` all render `${NAME}`. Code that needs the plaintext (for example, the adapter connect path) calls an explicit `.expose()`. A leak then becomes a compile error, not a missed producer.

## Decided

Owner, 2026-09-11: do this after the 1.74.0 tag, not before. The 1.74.0 changelog states that the class is not yet closed.

hugocorreia90:
Two more call sites, added after the survey this issue rests on.

#1934 merged while #1920 was in review and added two `ConfigError` variants that interpolate config-derived values:

```rust
#[error("pipeline '{pipeline}' {field} is {separator:?}, which is outside the allowed set …")]
#[error("pipeline '{pipeline}': metadata_columns entry '{column}' has a value that cannot be used: {reason}")]
```

`{separator:?}`, `{column}` and `{reason}` can each carry a resolved `${VAR}` value.

They are covered at runtime by #1920's response filter, so nothing leaks today. They matter here because when `RedactedString` replaces the filter these are two more sites needing the typed treatment — and they are easy to miss precisely because they postdate the count of "twenty-odd variants" that everyone has been working from.

Found by ui-revamp during a re-review, not by either PR's own checks. Refs #1934.

hugocorreia90:
**Recommendation (2026-09-22): start it now, as the 1.75 item.** The gate this issue named (the `engine-v1.74.0` tag) is behind us since 2026-09-13. It does not depend on PR #1944, which scrubs the job store; this issue is the `RedactedString` type that renders `${NAME}`. Include the two `ConfigError` call sites #1934 added. It needs the independent review before merge, so the earliest merge is after 2026-09-26.

*Checked against `main` on 2026-09-22. No independent model review: Codex credits return 2026-09-26. Same-model recommendation, disclosed.*

hugocorreia90:
[...]

Issue #1897, opened 2026-09-10
Define the trust boundary for environment-expanded config, so a resolved secret cannot reach an error string


hugocorreia90:
`${VAR}` is resolved **before** any config text is parsed. From that point on, the text holds live secrets, and every error that quotes it is a potential disclosure. #1888 closed one such error. A red team then found two more classes, and there is no argument that a fourth does not exist.

This issue is the decision #1888 deliberately did not make: **where is the boundary that guarantees a resolved value cannot leave the process?**

## Why chasing producers does not terminate

```
rocky.toml ─┐
_defaults.toml ─┼─▶ substitute_env_vars ─▶ EXPANDED TEXT ─┬─▶ parse error ──┐
sidecar.toml ─┤        (report DISCARDED)                 ├─▶ validation ───┼─▶ format!("{e:#}")
frontmatter ─┘                                            └─▶ ...next one   │        │
                                                                            │        ▼
                                                        any future caller ──┘   HTTP + browser
```

`substitute_env_vars` (`config.rs:2344`) returns expanded text and throws the substitution report away. Any caller can then parse that text and build an error from it. The redactor added in #1888 needs the report, so it is reachable from exactly one producer. Every other one is open, including ones not written yet.

Two are open today.

**Class A — post-parse validation errors.** Provenance is gone once `parse_rocky_config` returns, so no validation error can redact. **20 of 28 `ConfigError` variants interpolate a `String`/`PathBuf`/`Vec<String>` field.** Worked example: `autonomy_budget = { failures = 2, window = "${ROCKY_SECRET}" }` is cloned at `config.rs:3423` and rendered as `{window:?}` by `PolicyBudgetInvalidWindow` (`config.rs:333`), reaching `GET /api/v1/policy` at `api.rs:2135`.

**Class B — model sidecars and frontmatter.** `models.rs:683` (`_defaults.toml`), `models.rs:1479` (sidecars) and `models.rs:1540` (inline frontmatter) all substitute then parse. `ModelError::ParseFrontmatter` (`models.rs:38`) embeds the raw `toml::de::Error` and interpolates it into its own `Display`, so `e.to_string()` at `rocky-server/src/state.rs:467` publishes the echoed line, which `GET /api/v1/project` returns and the SPA renders.

## The options

[...]

Issue #1886, opened 2026-09-10
A config parse error carries resolved env var values to the browser UI, and the hint prints them by design


hugocorreia90:
A config parse failure carries **resolved** environment variable values to the browser UI, through `config_error` on `GET /api/v1/project`. The estate screen renders it as text on the project card.

Related to #1878 but not the same defect. That one asks whether a resolved `${VAR}` may appear in a load-bearing config **field**. This is the **diagnostic text**, which carries values regardless of which field they came from, and one half of it does so deliberately.

## Reproduced

`${VAR}` is substituted before the TOML is parsed, so a value that breaks the parse is quoted back with its resolved contents.

```toml
[adapter]
type = "duckdb"
path = ${ROCKY_PROBE_SECRET}     # valid TOML before substitution, broken after
```

With `ROCKY_PROBE_SECRET` set, `GET /api/v1/project` answers:

```
config_error: failed to parse TOML: TOML parse error at line 3, column 8
  |
3 | path = SUPERSECRET-abc123-DO-NOT-LEAK
  |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
string values must be quoted, expected literal string

  hint: config had these env var substitutions — check one is the culprit: ROCKY_PROBE_SECRET="SUP…
```

## Two separate paths, and the second is by design

1. **The parse error echoes the offending line**, which by then holds the
   resolved value, underlined with carets.
2. **The hint prints the substitutions as `NAME="VALUE"`**, truncated to
   `MAX_VALUE_LEN` (`config.rs`, near the "env var substitutions" message).
   This is a deliberate debugging aid. It is a reasonable one in a terminal
   the operator already owns, and a disclosure over HTTP.

The realistic shape is not a probe variable. It is a warehouse credential:

```toml
[adapter]
token = "${DATABRICKS_TOKEN}"
```

Any syntax error elsewhere in the file makes the hint list that token's value.

## Why it reaches a browser

`api.rs` builds `config_error` as `format!("{e:#}")` — the whole error chain, unfiltered — and `GET /api/v1/project` returns it. `ProjectStrip` renders it as the project card's subtitle. `rocky serve --ui` is exactly the case where a person is looking at that card.

The UI token is read-only, which limits **who** can read it, not **what** is in it.

## What to decide

- The hint should probably name the variables and not their values. The name alone is what makes it a useful hint.
[...]

Issue #1878, opened 2026-09-10
Decide whether a resolved ${VAR} may appear in load-bearing config fields on read surfaces


hugocorreia90:
## What

Decide, once, whether a resolved `${VAR}` value in a **load-bearing** config field may be shown in CLI JSON and over `rocky serve` read routes. Today the answer is implicit and inconsistent.

## Why now

`load_rocky_config` substitutes environment variables **before** parsing, so by the time any output struct clones a config value, a `${DEPLOY_TOKEN}` is a live credential and not a template. There is nothing at the output layer that can tell the two apart.

PR #1874 hit the concrete version of this. A policy rule's `conditions` accepted arbitrary JSON, was documented "parsed and ignored", and was carried into `rocky policy show --output json` and `GET /api/v1/policy`. So `conditions = { token = "${DEPLOY_TOKEN}" }` was served. That field was dropped, which fully closes it, because the field decided nothing and nothing was lost.

The general case is not closed and cannot be closed the same way. These fields are carried because they decide outcomes and a reader cannot work without them:

- a policy rule's `scope` (`models`, `tags`, `layer`, `classifications`)
- a rule's `verify_after` check names
- an autonomy budget's `window`

Dropping them is not an option, and key-based redaction does not work on values whose keys are author-chosen.

## The decision

One of:

1. **Accepted disclosure.** Say so explicitly in the docs and in the schema descriptions: any config value a read surface shows may be a resolved variable, and operators must not put secrets in policy fields. Cheapest, and honest, but it is a real constraint on config authors that nothing currently states.
2. **Track provenance.** Keep the raw template alongside the resolved value at parse time, and have output structs carry the raw form (`${DEPLOY_TOKEN}`, not its value). Correct, and a change to the config layer that reaches every output.
3. **Deny-list the substitution.** Refuse `${VAR}` in a named set of fields at parse time, so a secret cannot reach them. Narrow, and it breaks any project legitimately templating a scope.

Recommendation: option 1 now, stated in `docs/STYLE`-conformant prose in the config reference, and option 2 recorded as the eventual shape. Option 3 is a behaviour break for a case nobody has reported.

## Scope note

This is deliberately NOT part of #1874. That PR closed the one field where the fix was unambiguous (unbounded, and load-bearing for nothing). Two independent reviewers agreed the general policy needs its own design rather than a redaction bolted onto one producer.

## Evidence

[...]

PR #1920, opened 2026-09-11
feat(engine): no resolved secret leaves rocky serve — the registry and the outermost filter (1/3)

hugocorreia90:
Refs #1897. **PR 1 of 3** for the 1.74 tag gate: the registry and the outermost filter. Jobs durability is PR2; the `${NAME}` placeholder in successful `/policy` fields is PR3.

`${VAR}` is expanded *before* the config is parsed, so a resolved secret is ordinary bytes by the time a producer carries it. Nothing downstream knows it is holding a credential.

```
         rocky.toml / model files
                 │  ${TOKEN} expanded here
                 ▼
   substitute_env_vars_inner ──▶ secret_registry   (name + value, monotonic)
                 │                      │
                 ▼                      │
   any producer: 28 ConfigError         │
   variants, 8 model sites, a           │
   successful /policy field …           │
                 │                      │
                 ▼                      ▼
            router() ───────▶ OUTERMOST response filter ──▶ the wire
```

## Why a boundary and not an audit

Auditing producers does not terminate. Eight substitution call sites in `models.rs`, twenty-odd `ConfigError` variants interpolating a `String` or `PathBuf`, and a successful `GET /api/v1/policy` that copies resolved scope fields with no error involved at all. Each one a future edit can re-open.

So the guarantee moves to the boundary: whatever any producer built, the bytes on the wire carry no registered value.

## Two more disclosure surfaces appeared while this PR was in review

The strongest argument for doing this at the boundary is not in the design. It is in the timeline.

This PR's case for a boundary is that auditing producers does not terminate — each known site is one a future edit can re-open. **While the PR sat in review, #1934 merged and added two more**, both interpolating config-derived values into an error:

```rust
#[error("pipeline '{pipeline}' {field} is {separator:?}, which is outside the allowed set …")]
#[error("pipeline '{pipeline}': metadata_columns entry '{column}' has a value that cannot be used: {reason}")]
```

`{separator:?}`, `{column}` and `{reason}` are all config-derived, so all three can carry a resolved `${VAR}` value. They are exactly the class this issue is about, they postdate the survey that counted twenty-odd such variants, and nobody involved in either PR noticed the interaction until a reviewer looked.

A fix that audited the known producers and shipped yesterday would already be incomplete today. The boundary covers both without knowing they exist.

Found by ui-revamp while re-reviewing after a base update — worth recording that it took a deliberate re-read to see, because the two PRs were green independently and nothing connects them mechanically.

## The registry

Fed at `substitute_env_vars_inner` — the one point the config loader, the report variant, and all eight `models.rs` call sites converge on. A future caller is covered by construction; there is no "remember to register" step to skip.

- **Monotonic within a process.** A value stays registered for the life of the process, so a config edit that stops referencing `${OLD_TOKEN}` does not un-protect records written while it was live. It does **not** survive a restart — see the stated residual below, which is the case this bullet originally over-claimed.
- **`${VAR:-default}` literals are NOT registered.** A default is written in `rocky.toml` in cleartext, so anyone who can read the config can already read it. Registering it would mangle diagnostics for no secrecy gain.
- **A poisoned lock is recovered, not propagated.** The map is owned strings with no cross-field invariant, and treating poison as "refuse every response" would let one unrelated panicking thread take the server's output down.

## Outermost, and why that word is doing work

The filter is the **last** layer `router()` applies, so on the response path it runs after everything else:

```
413 payload_too_large   (a map_response layer)
421 host_not_allowed    (require_known_host, outside the bearer layer)
401 / 403               (require_bearer_token)
404 / 405 fallbacks
a handler that builds its own Response
```

That last one is not hypothetical. `POST /api/v1/compile` hand-builds its body with axum's `Json` and inserts `config_error` verbatim. A filter installed at the `PrettyJson` responder — my first design — would have missed it **while looking correct**. Codex caught that; the test below is the receipt.

Two tests pin it, both mutation-checked:

```
remove the filter layer       -> a_hand_built_response_body_is_filtered_too FAILED
move it one layer inward      -> the_secret_filter_is_the_outermost_layer   FAILED
```

The first mutation prints the live defect verbatim, which is the clearest statement of what this PR closes:

```
{"config_error":"… failed to parse TOML: TOML parse error at line 3, column 8
  |
3 | path = OUTERMOST-PROBE-SECRET-8e26660e
  |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

## JSON only, deliberately

The filter rewrites bodies whose `content-type` is JSON and passes everything else through untouched. With `--ui` the same router serves the embedded JS, CSS and fonts, and a byte-level replacement inside a minified bundle that happened to contain a registered value would corrupt the asset silently — plausible, since the 8-byte floor admits short sequences.

Safe because the UI files carry no config text: `index_response` serves the embedded `index.html` byte for byte and `serve_asset` serves hashed files, with no templating and no state injection. Anything config-derived reaches the browser through the JSON API, which this filter covers.

## The four failure modes #1897 names

| Mode | How it is handled |
|---|---|
| escaping | each value is searched raw **and** in its `serde_json`-escaped form — serde writes `a"b` as `a\"b`, which a raw search walks past |
| line splitting | a newline is escaped to `\n` by serde, so a multi-line value is contiguous in the body and the escaped pass catches it |
| short values | the 8-byte floor, below |
| truncation | redact **before** truncating at the two marked sites, plus an anchored prefix match for records an older binary wrote — see below |

## The owner's ruling, and my dissent on the record

Hugo ruled: every `${VAR}` value is redacted, with an **8-byte floor**; shorter values are shown. I recommended fail-closed instead (refuse the response rather than show a short value). He ruled otherwise and the floor is what this implements. `SECRET_LENGTH_FLOOR` carries the trade in its doc comment.

Two consequences, both stated rather than discovered later:

**Collision.** A non-secret value at or above the floor is rewritten wherever those bytes appear. With `CATALOG=analytics`, a compile error about `analytics_orders` becomes `${CATALOG}_orders`. The `${NAME}` form is what keeps that workable — `[REDACTED]_orders` would leave an operator unable to tell which model failed, while `${CATALOG}_orders` names the variable, so anyone who can read the environment can reconstruct the identifier. The variable name is not a secret; `format_env_var_hint` already lists names.

An exemption for values matching a model/pipeline/target identifier was considered and **rejected**: it fails open, and decisively, the identifier set comes from a successful compile — `recompile()` returns from its `Err` arm with no compile result, so on a config parse failure the exemption list is empty in exactly the case that leaks.

**A length oracle.** Showing values under 8 bytes and rewriting those at or above reveals which side of 8 a value falls on. Small, real, and inherent in the ruling rather than chosen here.

## Verification

`cargo test -p rocky-cli -p rocky-core` — 2001 + 2209 and every other target, 0 failed. Clippy `--all-targets --all-features` clean on both, `cargo fmt --check` clean. The full suites were run immediately after wiring the registry, before the filter existed, specifically to catch test pollution from a process-global; there is none.

## Truncation: reachable, found, and fixed in the follow-up commit

This section said the opposite when the PR opened. I had written that truncation was probably unreachable on the strength of grepping three files. **It is reachable**, at two sites, and I had not grepped the right ones.

```
resilience.rs:258   truncate_error    format!("{err:#}") then truncate(500)
                      -> AttemptRecord.error -> MaterializationOutput.attempts
                      -> `rocky run --output json` stdout
                      -> captured as PersistedJob.result -> GET /api/v1/jobs/{id}

machine.rs:2403     truncate_detail   caps at 4000 with a `…` marker
                      -> observation_detail -> GET /api/v1/products/{name}/journal
```

A resolved value straddling the cut ships as a PREFIX, which a filter searching for whole values cannot match.

**Correction on the route**, because an earlier revision of this section named the wrong one: the attempt trail does NOT reach `GET /api/v1/runs`. `RunModelRecord` (output.rs:2487) carries `model_name`, `duration_ms`, `rows_affected`, `status` and `recipe_identity` — no `attempts`. The persisted copy has no field on the history response to come out through. Persisted is not rendered.

Two defects at that one line, not one. `String::truncate` also **panics** off a char boundary, so a multi-byte character straddling byte 500 aborts the retry path — pre-existing, unrelated to secrets, fixed here.

Fixed by redacting before truncating, plus an anchored prefix match for records an older binary already wrote. Both land in the fix delta on this branch.

**That two marked sites are all of them.** Both truncations found so far append `…`, which is what makes the anchored prefix match possible — it reads a cut we made ourselves rather than hunting for secret-shaped text. A site that caps SILENTLY, with no marker, would still ship a fragment and nothing would fail.

I have now been wrong once about having found them all, so this is stated rather than claimed. `truncate_detail`'s own doc says "Truncation is visible, never silent" — so the repo's convention is on my side, and `truncate_error` was the one that had not followed it. That is a convention, not an enforcement.

The registry is per-process, and PR2 inherits that. `execute_job_subprocess` spawns a separate `rocky` with its own registry, and the parent scrubs the child's output against the **parent's** set — so a value the child expanded and the parent never loaded cannot be redacted. The ordinary case is that both loaded the same config, and "ordinary case" is not fail-closed. It is documented in the registry's module doc and backlog is pinning it as a known gap so that if it is ever closed, the test says so.

The wider point: this PR moves the guarantee from "every producer remembered" to "the boundary caught it", which is a real improvement, but it is still a filter over a value that should never have been a plain `String`. `RedactedString` already exists in this repo (`rocky-core/src/redacted.rs`, 28 consumers) and was ruled out of scope for the tag, tracked as #1919. Until then a leak is a missed match rather than a compile error.

### A component reasoned into "redundant" is the one to mutate before removing

The rescan checks the finished body in two views: the wire bytes, and the JSON-decoded string values. **I concluded the decoded view was redundant and said so in writing** — I had deleted it, watched every test stay green, and inferred it was dead weight kept for safety.

It is load-bearing. The construction:

```
value = ABCDEFGH/
body  = {"m":"ABCDEFGH\/"}
```

JSON permits `\/` as an escape for `/`, but neither `serde_json` nor Rust `Debug` ever *generates* it. So every form `escaped_forms` produces collapses to the unescaped value, none of which is in the wire bytes. Only decoding recovers it.

The error is worth more than the fix: **I checked which escapes our SERIALIZER emits and concluded no other form could appear.** But this filter reads bodies, and a body may carry any escape the JSON *spec* allows. Generalising from the writer to the reader is the mistake, and it is not specific to escaping.

The same shape appears three times in this PR's history — a claim about reach, a claim about a failure mode being unreachable, and this. In each case a test passed and the reasoning behind it was wrong. Mutating before removing is what caught it.

### A cap that bounds bytes does not bound work

The filter refuses bodies over 64 MiB. Two ways to exhaust the process *inside* that cap:

- **Match density.** A registered value of `AAAAAAAA` against a 60 MiB run of `A` produced roughly 62.9 million overlapping spans — about **2.3 GiB** for the span vector alone, before replacement strings and sort workspace. Overlapping matches of one form now collapse as they are found; a run of any length is one span.
- **Per-string rebuild.** The rescan rebuilt the escaped-form set for every decoded string. A 1,000-entry registry against a 300 KiB body holding 100,000 strings is ~100 million `escaped_forms` calls, each allocating four owned strings. The form set is now built once per rescan.

Neither is reachable through the body-size limit, because the limit bounds the input and these are quadratic in the work derived from it. Worth stating as a class: **any limit expressed in input bytes should be checked against the work that input can cause.**

## Stated residual: persisted records can surface a rotated or foreign-process `${VAR}` value after a restart

**Owner-ruled (#1897): disclose in 1.74, close with #1919.**

The registry is process-local and in-memory. A restarted server knows only the values *its own* config load expands. So a value that was `${VAR}`-derived when a record was written, and is no longer expanded by the running process — rotated, a different config, or written by another process such as a CLI `rocky run` — is not in the registry, and the filter has nothing to match. No truncation is involved; the whole value ships.

Found by the Codex pass on this PR, which named it before it died mid-trace.

### The fields, checked by the Codex pass

My first enumeration had two entries that do **not** leak and missed three surfaces that do. Corrected:

| Route / struct | Field | Lookup key? |
|---|---|---|
| `/api/v1/runs` · `RunHistoryRecord` | `pipeline` | **yes** (filters, claims) |
| `/api/v1/runs` · `RunModelRecord` | `model_name` | **yes** (resume, history) |
| `/api/v1/jobs/{id}` · `JobStatus` | `result` — the child's stdout verbatim | no |
| `/api/v1/jobs/{id}` · `JobStatus` | `error` | no |
| `/api/v1/audit` · `AuditDecisionEntry` | `model`, `reason` — drift embeds the model in both | `model` **yes** |
| `/api/v1/brief` | the same persisted decisions | — |
| `/api/v1/review/queue` | the same persisted decisions | — |
| custody (`audit.rs:358`) | the same persisted decisions | — |
| `/api/v1/schedule/spool` · `PendingDemand` | `pipeline` — written durably, and the producer deliberately does not read current config | — |

**Withdrawn — these do NOT leak, and I claimed they did:**

- `/api/v1/runs` · `target_catalog`: the route passes `audit = false` (`api.rs:1296`), so the field is omitted entirely (`history.rs:59`).
- `/api/v1/products/{name}/journal` · `observation_detail`: the projection enumerates its fields (`product.rs:1874`) and that is not among them.

**Also checked and clean:** schedule status loads current config and attaches claims only to currently-resolved pipeline names.

### What would prove this residual, and is not here

A two-process regression test: persist runs, decisions and spool entries while a value is expanded, rotate it, start a fresh server, and read those routes back. Codex named this as the most important missing item and I agree. It is not needed to FIX the residual — it is what would demonstrate it, and its absence is why the table above rests on reading rather than on a failing test.

### Why a write-time scrub is not the fix

The leaking fields and the lookup keys are the **same class**. `target_catalog` is safe to scrub; `model_name` and `pipeline` are not — and all three are config-derived strings that may or may not be `${VAR}`-sourced. Nothing at write time carries that bit, so a scrub either corrupts resume and history matching or misses the leaking fields.

Read-time filtering is not what is broken: it rewrites the response and touches no stored key. The gap is the registry's **knowledge**, not the mechanism. Closing it properly means the value is never a plain `String` in the record — which is #1919.

**Independent red team: INCOMPLETE at time of writing.** Codex reviewed the design SKETCH and returned **UNSOUND**; the design here is the corrected one, and its three load-bearing findings were each verified against the code before being accepted. Its architectural correction — outermost middleware rather than a responder-level filter — is the reason `POST /api/v1/compile` is covered.

**The completed pass found NINE defects**, six of them in this PR's own code, each with a reproduction. All six are fixed on this branch; the two highest-value fixes are mutation-checked. Its two clean categories are worth stating too: it found no way for a response to bypass the outermost filter (it checked SSE, WebSocket, upgrades, streamed bodies and direct socket writes — none exist in production), and no soundness problem in the registry. It ran its own second reviewer, which confirmed every finding.

The six, and what each was: a PANIC in the tail search from an unchecked char boundary; HEAD requests turned into manufactured 500s; a doubly-escaped value going unmatched; a trailing backslash producing invalid JSON; two overlapping values leaking seven bytes of one; and a variable NAME that itself carried a registered value.

An earlier Codex pass over this PR **failed mid-trace** (`status: failed`, no files touched). Its partial output named the persisted-records residual above, which is why that section exists. A third pass covering this PR plus the fix delta is running; this PR should not merge on the strength of a review that did not finish.

Note for that reviewer: Codex may have read either version of `truncate_error`; its earlier pass covers the #1920 diff, not that function. Its three load-bearing findings — the jobs path is durable, eight model sites not three, and `RedactedString` already exists — were each verified against the code before being accepted. Its architectural correction (outermost middleware, not a responder-level filter) is the reason `POST /api/v1/compile` is covered. Reviewed by ui-revamp; the collision behaviour and the overlapping-pair test come from its pre-build notes, and the exhaustive route sweep is backlog's independent work on `test/serve-route-secret-sweep`.

Refs #1886, #1878, #1919


hugocorreia90:
Reviewed as the assigned reviewer for the three-PR redaction series. **The design holds up under every check I set out to make, and two of them are answered more thoroughly than I asked.** One blocking item, which is CI-visible and has a known fix.

## Blocking: CodeQL is red, and it is the same defect this PR is about

Four annotations, all in the new tests:

```text
secret_filter.rs:616  Cleartext logging of sensitive information
secret_filter.rs:618
secret_filter.rs:696
secret_filter.rs:697
```

```rust
assert!(!out.contains(secret), "the value survived: {out}");
```

`out` is the redacted body. On the failure path — the only path that prints — it is the body that still *contains* the value. So a failing containment test writes the secret into the CI log it exists to keep it out of.

Harm is low (fixture values), but this is a PR about not leaking secrets, and it leaves CodeQL red on a release gate.

I hit the identical finding on #1907 earlier today, same cause, same four-line shape. The fix that cleared it: put the values in one constant paired with a NAME, assert on the value, print only the name, and never print the body.

```rust
const SECRETS: [(&str, &str); 3] = [("the Bearer token", "BEARER_SECRET_ABC"), ...];
for (label, secret) in SECRETS {
    assert!(!body.contains(secret), "the settings document disclosed {label}");
}
```

## What I verified, rather than took from the description

**The central claim — "a future caller is covered by construction".** This is what the whole boundary rests on, so I checked that `substitute_env_vars_inner` really is the single convergence point. It is: `substitute_env_vars` → `..._with_report` → `_inner`, `config.rs:6447` calls `_inner` directly, and both `std::env::var` calls (`:2416`, `:2426`) are inside it. Registration there does cover every path.

**Longest-first ordering**, which I asked for — implemented, documented as load-bearing, and pinned by `an_overlapping_pair_reads_back_longest_first`. The doc comment states the failure it prevents (a fragment of the longer secret left behind by the redactor itself), which is the right way to record it.

**Fail closed on an error, not only on a match** — and further than I asked. Over-cap or unreadable body, non-UTF-8, replacement that broke the JSON, and then `any_value_survives` as a final independent gate that does not depend on any generator being right. That last one is the part I would have asked for if it were missing.

**`any_value_survives` checks the decoded view too**, not just raw bytes, and `walk_strings` covers object keys. A value only visible after a client parses the JSON is caught.

**The 64 MB cap** is a memory guard well above any realistic response and fails closed when met. Fine.

## A correction to something I told you

I said the truncated attempt trail reaches `GET /api/v1/runs`. **That was wrong and your correction is right.** `RunModelRecord` (`output.rs:2487`) carries `model_name`, `duration_ms`, `rows_affected`, `status`, `recipe_identity` — no `attempts`. The persisted copy has no field on the history response to surface through.

The concern was real but travels the route you found: `rocky run --output json` → `PersistedJob.result` → `GET /api/v1/jobs/{id}`. Thank you for tracing it rather than accepting my route.

## One honest limit on "covered by construction"

The claim is true for callers that go through `substitute_env_vars_inner`. It is not true for a future producer that calls `std::env::var` directly and puts the result in a response — that value is never registered, and `secret_registry::is_empty()` would additionally let the filter skip the body entirely if nothing else had registered.

Not a defect in this PR: today all expansion converges, and I verified it. But the property is an invariant maintained by convention, and the module doc reads as though the boundary is unconditional. Worth one sentence in the registry's module doc naming the invariant — "a value that does not pass through `_inner` is not protected" — so the next person adding an expansion path sees what they are opting out of.

## On the parts that are rulings, not code

The 8-byte floor, the collision consequence, and the length oracle are all Hugo's call and are stated in the PR rather than buried. The `${NAME}` form over `[REDACTED]` is the right choice for the collision case, for the reason given — an operator can still identify `${CATALOG}_orders` and cannot identify `[REDACTED]_orders`.

Rejecting the identifier-exemption is correct, and the decisive argument is the one in the PR: the identifier set comes from a successful compile, so on a config parse failure it is empty in exactly the case that leaks. That is a fail-open exemption and it belongs rejected.

## Verdict

Approve on substance. Fix the four assertion messages and I will confirm — required checks are already 7/7, so that is the only thing between this and green.

hugocorreia90:
**Approving.** Re-reviewed at head `a4309e21`, which is a different tree from my first read — sixteen-plus commits and a base update landed in between, so I re-derived rather than carrying the earlier verdict across.

```text
required checks    7/7 green, 0 pending
CodeQL             completed | success — "No new alerts in code changed by this pull request"
Analyze (rust)     completed | success
```

The four cleartext-logging annotations are cleared, and the messages went further than the fix I suggested: label **and** body length, each fixed at the call site, so a failure says which configured value got through without disclosing what came back. I have aligned #1907's containment tests to the same shape, so two secret-containment suites in one release fail the same way.

**What I verified rather than read for plausibility:**

- **The central claim.** `substitute_env_vars` → `_with_report` → `_inner`, plus the direct `config.rs:6447` call, with both `std::env::var` calls inside `_inner`. Registration there really does cover every path, which is what the whole boundary rests on — if it were wrong the PR would be decorative.
- **The extraction is a pure refactor.** Reduced to three moves: the empty check keeping its position, a reference adjustment, and the sort relocating into `rewrite_merged` with `mut spans`. Nothing dropped between the loop's end and the sort. The worry that a testability split had changed behaviour does not hold.
- **The base update did not touch the reviewed surface.** A `git diff --stat` across the five files, between my reviewed head and this one, shows a single hunk: the `<= 2` → `== 1` tightening in a test.

**What makes this more than a filter:** fail-closed on over-cap, on non-UTF-8, on a replacement that broke the JSON, and then `any_value_survives` as a final gate that trusts none of the generators and checks the decoded view as well as the raw bytes. Longest-first ordering with the failure it prevents written down beside it. And the residuals — the process-local registry across a restart or a subprocess, the length oracle, the collision — are stated in the body rather than discovered later.

Two lessons recorded in the body are worth more than the code to a future reader: *a component reasoned into "redundant" is the one to mutate before removing* — the `\/` case, where checking what our serializer emits stood in for what the format permits — and *a cap that bounds bytes does not bound work*.

Merge is release-cut's call; the branch reads `BEHIND` again as of now.

PR #1922, opened 2026-09-11
fix(engine): every user expression spliced into generated SQL is parsed and allowlisted, and a refusal trips the gate

hugocorreia90:
A `[quarantine]` `expression` is user SQL spliced **twice** into a CTAS that runs with warehouse credentials, guarded only by `reject_statement_terminator` — which stops a predicate *ending* the statement and nothing else. A subquery reads any table the run's credentials can see; a content-reading function sits in scalar position and reads whatever it is pointed at.

[#1820](https://github.com/rocky-data/rocky/pull/1820) closed that on the **checks** path. This closes it everywhere else.

## Every site, not the one that was reported

Grepping `reject_statement_terminator` for user SQL reaching a generated statement gives six. Five are in scope here and all five now run the same validator under the same dialect mapping, so no two can drift on what is allowed.

| Site | Field | Shape |
|---|---|---|
| `quarantine.rs` `lower_valid_predicate` | quarantine `expression` | boolean |
| `quarantine.rs` `wrap_filter` | quarantine `filter` | boolean |
| `tests.rs` ~445 | checks `filter` | boolean |
| `tests.rs` ~621 | `unique_expr` `key_expr` | **scalar** |
| `checks.rs` ~280 | `cross_source_overlap` `key_expr` | **scalar** |

The checks `filter` is the one nobody had named: #1820 guarded the expression beside it and left the filter on the terminator-only check, so that hole predates quarantine.

**A sixth site is out of scope and handled separately.** `metadata_columns[].value` is spliced into `CAST({value} AS {type}) AS {name}` (`trino/dialect.rs:118-130`, and the same shape in `snowflake/dialect.rs:182`, `databricks/dialect.rs:141`, with the constructor check at `rocky-ir/src/ir.rs:289`). It is terminator-guarded only and reachable from `rocky.toml`.

It is not fixed here because the fix has a different shape: `MetadataColumn::new` is the one chokepoint every adapter passes through, but `rocky-ir` is data-only and has no dialect, so the check belongs at config load rather than at the splice. That is its own PR.

## The boundary depends on where the expression is used

The independent review found the hole in the widening itself: the allowlist has carried `now`, `current_timestamp` and friends since it was written for **predicates**, and applying that same list to grouping keys made them reachable where they are wrong.

```
  key_expr = "now()"   accepted. The key comes from evaluation time, not
                       the row, so the table collapses into one group.
```

The exclusions doc already said a non-deterministic key groups nothing — it said it about `uuid` and `random` while `now` sat on the list.

So the validator takes an `ExpressionUse`:

| mode | used by | volatile | `COLLATE` |
|---|---|---|---|
| `SinglePredicate` | checks `expression`, checks `filter`, MCP `draft_check`, quarantine in `drop` and `tag` | **allowed** | allowed |
| `ScalarProjection` | `metadata_columns[].value` | **allowed** | allowed |
| `GroupingKey` | `unique_expr.key_expr`, `cross_source_overlap.key_expr` | refused | **refused** |
| `ReevaluatedPredicate` | quarantine in `split` only | refused | allowed |

Volatile is allowed wherever the expression is evaluated **once**: `created_at > now() - interval` is a legitimate freshness-shaped check, and `_loaded_at = current_timestamp()` is the reason metadata columns exist. Refusing either would break working configs to close a bug that does not exist in that position.

**The two policies are separate questions, so they are separate predicates.** The first version gated `COLLATE` on volatility tolerance, which refused it in `ReevaluatedPredicate` too. That was wrong, and the code comment beside it already said why: an explicit collation is *deterministic*, so two evaluations agree. What it changes is the meaning of equality — which matters only where values are compared against each other to form groups. So `COLLATE` is refused in `GroupingKey` and nowhere else.

**There is no `Default`.** A new call site has to choose, rather than inherit the permissive answer.

That was not a theoretical benefit. While this branch was out, #1934 landed on main with an eighth caller — `metadata_columns[].value`. The merge did not compile until that site picked a mode. A call site written on a branch this one had never seen could not silently inherit the permissive answer, which is the entire argument for the missing `Default`.

**Volatility is not restated.** The check reuses `determinism::VOLATILE_FUNCTIONS`. That reuse had a gap worth recording: `localtime` was on this module's allowlist and missing from `VOLATILE_FUNCTIONS`, so it passed both restricted modes. The old caller was safe because it refuses any function it does not recognise; this module has its own allowlist, so that fallback stopped covering it. Fixed, and `every_clock_name_on_the_allowlist_is_known_volatile` now fails if either list gains a clock name the other lacks.

### Why only quarantine `split` is `ReevaluatedPredicate`

In `split`, the predicate is spliced into the quarantine CTAS **and again** into the valid-table CTAS that follows, and those run as two statements. `created_at <= now()` can call a boundary row invalid in the first and valid in the second, putting one row in **both** outputs.

```
  split   statement 1  CREATE ... __quarantine AS ... WHERE NOT (<pred>)
          statement 2  CREATE ... __valid      AS ... WHERE     (<pred>)
                                                            ^ evaluated again
  drop    statement 1  CREATE ... __valid      AS ... WHERE     (<pred>)
  tag     statement 1  CREATE ... <source>     AS ... CASE WHEN ...
```

`drop` and `tag` emit **one** statement each, so a clock function there is as legitimate as in any ordinary check. The first version of this change refused all three modes — a false refusal that would have broken working `drop` and `tag` configs to close a bug only `split` has. Caught by the independent review.

**The engine's own generated predicates take the same rule.** `not_in_future` lowers to `col <= <current_timestamp>` and `older_than_n_days` to a `<current_date>` comparison. Neither passes through the expression validator, because that gate only ever sees user-written fragments. So the first version refused a user's `created_at <= now()` in `split` while the engine generated the identical predicate itself, one line away. Both kinds are now refused in `split` and stay available in `drop` and `tag` — the refusal is on the **mode**, not on who wrote the expression.

Filed as #1937. The refusal here closes the door; it does not fix the double evaluation, and #1937 says so.

**The two `key_expr` sites are scalar, not boolean**, which was the risk worth checking before widening — a validator sized for boolean predicates would refuse every valid key. It isn't: `validate_check_expression` parses one expression and judges its nodes, and requires nothing about the result type. Pinned from both sides.

## A placeholder is a session read

Found by the independent review, confirmed before fixing:

```
$foo => Ok(Value(ValueWithSpan { value: Placeholder("$foo"), .. }))
$foo validated => Ok(())          ← accepted
```

The same read spelled as a function was already refused (`getvariable('x')`, `identifier($foo)`). So the boundary refused two spellings and accepted a third. This is the **shared** validator, so the gap was in #1820's gate too; both close here.

Rocky splices these expressions as text and never binds a parameter, so a placeholder can only be a session variable resolved at execution time. Refusing all of them is fail-closed on purpose.

## What the walker actually judges

Not a full AST allowlist, and the PR should not imply otherwise.

**Refused**: anything that does not parse as exactly one expression under the target dialect; trailing tokens; any nested query anywhere; any qualified function name; any unqualified function not in `CHECK_EXPRESSION_FUNCTIONS`; a lambda; a bare session-identity keyword; and now any placeholder.

`to_char` is **not** on the list. Snowflake's one-argument form falls back to the session's output-format parameter, which makes it a read of session state rather than a function of its arguments. The reviewer flagged it as its own least-confidence item and there is no Snowflake here to settle it, so it stays off rather than shipping on "probably fine". The way back in is recorded beside the list.

**Still accepted**: every other parsed expression node and operator. The walker judges *node kinds*, not semantics. A future dialect construct that reads outside the row and is none of the above would pass.

**Known unanswered, filed as #1935**: the walker matches function NAMES, and a warehouse can rebind a name (`CREATE MACRO`, a UDF). Whether a shadowed `md5` or `sha2` would pass this boundary is not established on any dialect — the independent review raised it and its run ended before it could test. It is a property of #1820's validator, which this PR inherits and slightly widens rather than creates. Recorded so the gap is on the record rather than lost with that run.

`dialect_for` falls back to `GenericDialect` for an unrecognised name, including Trino. That is deliberate and already documented in the module header, with an existing test (`unknown_dialect_falls_back_to_generic_and_still_judges`): a broader dialect can only *accept* more syntax, and acceptance still has to clear the walker.

## A refused predicate trips the check gate

The validator turns a dangerous predicate into a compile error. That error was being swallowed by `warn!(... "skipping")` — execution continued, the table was never split, and every row quarantine existed to catch flowed on with nothing in `RunOutput` saying so.

**Part one alone would have been a regression**, trading a security hole for a silent data-quality one. A refusal is now a failing error-severity check named `quarantine:compile`, so the gate trips and the reason reaches the JSON.

Not a hard abort: the compile runs **before** any quarantine statement executes.

**Per table, not per run.** Nothing is half-written *for the table that failed*. Across a run, table A can split successfully before table B's predicate is refused — so a failed run may leave some tables split and others not. The gate reports the failure; it does not unwind the tables that succeeded.

## How this shows up in Dagster

The step **fails** on the exit code: a non-zero exit becomes `RockyCommandError` → `dg.Failure` (`resource.py:224-233`).

The check appears as a **warning observation**, not a failing check: the component emits only declared `(asset_key, check_name)` pairs, and `quarantine:compile` is not in any spec (#1673).

The name is `quarantine:compile` because this result is a second `TableCheckOutput` on an `asset_key` that already has one, and `emit_check_results` has no dedup: a duplicate `(asset_key, check_name)` raises "returned an output multiple times".

**It covers `checks.custom` too, not only assertions.** The first version walked assertions alone, and a custom check carries a user name that reaches the same Dagster keying — so the collision was still open through the other door. Worse than a duplicate: the component path *silently skips* the later one, so the quarantine failure would vanish behind a passing custom check.

**The colon is not what makes it safe.** `sanitize_check_name` (`contracts.py:279`) replaces every invalid character with `_`, so Dagster sees `quarantine_compile`. Names are therefore compared *after* sanitization — `quarantine:compile`, `quarantine_compile` and `quarantine.compile` are all refused, because checking the raw string would miss exactly the spellings that collide once mapped.

**Dagster owns more names than the engine does — but not unconditionally.** The first version reserved `quarantine:compile` alone, which was too narrow: a custom check called `contract.required-columns` sanitizes onto `contract_required_columns`, and on collision the **first** spec wins, the later one dropped with only a log line. The symptom is a user's check disappearing, not an error.

The second version reserved all nine names Dagster can declare, which was too **broad**. Five are conditional:

| name | Dagster declares it | so |
|---|---|---|
| `row_count`, `column_match`, `row_count_anomaly` | always | reserved |
| `freshness` | only when the pipeline declares `[checks.freshness]` | reserved per-pipeline |
| `compliance_exception` | only under `surface_compliance` | not reserved — the engine cannot see that flag |
| the three `contract_*` | only when a matching rule kind exists | not reserved — ownership is per-asset (#1941) |

Reserving a conditional name refuses something a project not using that feature could legitimately have taken. This guard has now been wrong in **both** directions, which is why the test that pins it asserts a name is **accepted** — no amount of extra strictness can satisfy that, whereas the previous test could only ever be satisfied by reserving more.

Names *synthesized* from user input — `null_rate:{column}`, an unnamed assertion's generated name, cross-source names — can still collide with each other. A constant list cannot fix that, because the collision is between two pieces of user input rather than against a reserved word; it needs a resolved per-table name set. Filed as #1941, deliberately out of scope here.

The duplicate `asset_key` itself is safe — both consumers flatten a list.

## Tests

Each with a control, all mutation-checked.

**All five call sites now prove their own dialect.** Four of them were tested only with a dialect whose `name()` takes the trait default `"unknown"`, so they parsed generically and would have stayed green if their dialect argument were hardcoded. `a->'k'` is the discriminator — Snowflake reads `->` as the lambda arrow — and each double delegates everything but `name()`, so the name is the only variable.

**Every guard was mutation-checked individually.** The cross-source `key_expr` guard had **no test at all** — removing it left all 8 cross-source tests passing. That is the gap a mutation check exists to find, and it is now covered.

**The gate test is attributable, and getting there found something.** A quarantine *expression* is also an assertion, so the assertion path and the quarantine path always co-fire: no fixture built on a refused expression can attribute the exit code to quarantine. The first version of this test exited 1 with or without the gate change. Split in two:

- `a_quarantine_compile_failure_trips_the_gate_on_its_own` — an invalid `suffix_valid`, a **quarantine-only** failure with every other check passing, so the non-zero exit can only come from quarantine.
- `a_refused_quarantine_expression_is_refused_on_both_paths` — the co-firing case, exit-code claim dropped, asserting the quarantine result exists beside the assertion refusal. They carry different reasons, and an operator needs the second to know their rows were not separated.

**Nine guards were mutation-checked.** Each failed under mutation: removing `LOCALTIME`; re-coupling `COLLATE` to volatility; forcing all quarantine modes to `ReevaluatedPredicate`; forcing them all to `SinglePredicate`; removing `ScalarProjection` from the volatile-tolerant set; dropping one reserved Dagster name; mode-gating an allowlisted name so it fails only as a grouping key; removing the generated-clock refusal; and re-adding a conditional name to the reserved list. That last one matters — it fires only in `GroupingKey`, so the allowlist test as originally written would have stayed green, which is what made it worth fixing.

2231 `rocky-core`, 317 `rocky-sql` and the gate tests pass, so the wider boundary refuses nothing the tree relied on. Gates were run on `cargo +1.98.0`, the toolchain CI pins; an earlier clean-clippy claim in this PR was measured on a newer local toolchain that does not deny `non_local_definitions`, and was wrong.

## User-visible

A config with an off-list function, a subquery, or a placeholder in any of these now **fails** where it previously ran: a quarantine `expression`, a quarantine `filter`, a check `filter`, a `unique_expr` `key_expr`, or a `cross_source_overlap` `key_expr`. `metadata_columns[].value` is unchanged here and follows in its own PR.

A quarantine compile failure now fails the run instead of silently skipping the split.

A `key_expr` or `unique_expr` expression that calls a time function such as `now()` is now refused, and so is a `[quarantine]` expression when `mode = "split"`. `COLLATE` is refused in a `key_expr` or `unique_expr`. Both stay accepted in an ordinary check predicate, in `[quarantine]` `drop` and `tag` modes, and in a `metadata_columns` value — every position where the expression is evaluated once.

**Breaking, and deliberately loud**: a `[quarantine]` pipeline with `mode = "split"` and a `not_in_future` or `older_than_n_days` assertion is now refused at SQL generation. It previously ran and could place a boundary row in **both** the valid and quarantine tables. Use `mode = "drop"` or `mode = "tag"`, which evaluate the predicate once. A refusal naming the assertion is better than a silent split that duplicates rows.

An assertion or custom check whose name collides with a check Dagster always declares — `row_count`, `column_match`, `row_count_anomaly` — is refused at config load, as is `freshness` for a pipeline that declares `[checks.freshness]`.

Refs #1820. Refs #1524. Refs #1935. Refs #1937. Refs #1941. Refs #1942.


hugocorreia90:
## How this shows up in Dagster

Stating it explicitly, because "fails the run" should not be read as "fails a Dagster check".

**The step fails, on the exit code.** A non-zero exit becomes `RockyCommandError`, which `resource.py:224-233` translates into `dg.Failure`. So the materialization fails regardless of how the check itself is surfaced.

**The check appears as a warning observation, not a failing check.** The component emits only *declared* `(asset_key, check_name)` pairs (`component.py` ~3503). `quarantine:compile` is not declared in any spec, so an undeclared result becomes a warning plus an `AssetObservation` (#1673) rather than a failing `AssetCheckResult`.

Both are true at once: the step goes red, and the reason is visible as an observation rather than as a check.

**The check name is namespaced** — `quarantine:compile`, not `quarantine`. This result is pushed as a second `TableCheckOutput` on an `asset_key` that already has one, and `emit_check_results` (`checks.py` ~59) has no dedup: a duplicate `(asset_key, check_name)` raises "returned an output multiple times". The colon puts it in a space no assertion name can occupy.

The duplicate `asset_key` itself is safe. Both Dagster consumers flatten a list rather than keying by asset (`component.py` ~3506, `checks.py` ~59), and neither `run.rs` nor `pipes.rs` does a keyed lookup over `check_results`.

PR #1944, opened 2026-09-12
feat(engine): a job's stored output is scrubbed before it is written, and pre-fix records return status only (2/3)

hugocorreia90:
Refs #1897. **PR 2 of 3** for the 1.74 tag gate: a job's stored output. PR1 was #1920 (registry + outermost filter); PR3 is the dropped `/policy` fields.

`execute_job_subprocess` captured the child's stdout **verbatim** into `PersistedJob.result` and wrote it to disk. A resolved `${VAR}` value therefore survived a restart and came back on `GET /api/v1/jobs/{id}`.

```
rocky run --output json  ──▶  stdout, verbatim
                                   │
                                   ▼
                          scrub_job_outcome        ← new
                                   │
                    ┌──────────────┴──────────────┐
                    ▼                             ▼
             in-memory cache                 record_job (disk)
```

Scrubbed before **both**, so neither holds a resolved value even briefly, and a reader racing the persist sees the same bytes a restart would.

## Why a pure function and not a scrub in the spawn path

`scrub_job_outcome` takes the terminal result and error and returns what gets written. That shape is not stylistic: the spawn path runs `current_exe`, which under `cargo test` is the **test harness**, so a scrub inside it cannot be reached by any unit test in the crate. `api.rs:2653` and `:6277` already say so about this path.

It is also the chokepoint rather than one caller. The spawn path is the only writer of a terminal result today; a second one would otherwise be unscrubbed.

**A result the rewrite would break is replaced, not stored.** If redaction makes the JSON unparseable — a registered value in a numeric position — the payload becomes `{"redaction":"result_withheld_unparseable_after_redaction"}`. An unparseable blob on disk fails every later read, not just this one write.

## Records written before this release

`PersistedJob` gains `redaction_version: Option<u32>`. Every field on that struct already defaults, so existing records deserialize with `None` — no migration.

**Legacy is `None` OR below `MIN_TRUSTED_REDACTION_VERSION`, not `None` alone.** Defining it as a comparison now means the mechanism already works the first time the rule tightens, instead of needing a second marker then. A legacy record serves its lifecycle fields and withholds `result` and `error`: nothing distinguishes a safe pre-fix string from one carrying a value.

A version at or above the floor is trusted, **including one this binary does not recognise**, under a contract stated on the field:

> Redaction versions are monotonically non-decreasing in strictness. A version may only be introduced for a rule at least as strict as every version below it. A future rule that shows MORE needs a different mechanism, not a higher number here.

Without that sentence, trusting `Some(999)` would rest on an assumption nothing enforces. Wording from backlog, who pushed back on my original justification.

## The field is non-optional in the literal on purpose, and it found five writers I did not know about

I knew of one. The compiler named six:

```
PRODUCTION  api.rs                              the HTTP submit path
PRODUCTION  commands/scheduler/spawner.rs:72    the scheduler's own writer
TEST        api.rs, rocky-server/jobs.rs,
            rocky-core/state.rs, state_sync.rs
```

**`spawner.rs` is the one that mattered.** Unstamped, every scheduler-launched job would have read as pre-redaction *forever* — `GET /api/v1/jobs/{id}` withholding its error on records this binary had just written. Not a leak: a silent loss of diagnostics on every scheduled run, indistinguishable from the legacy rule working as designed.

Its error is `format!("run exited with code {n}")`, a fixed template that cannot carry a value today. It goes through the seam anyway — "this particular string is safe" stops being true the moment someone edits the string.

`finish_scheduler_job` is extracted as a pure seam for the same reason as `scrub_job_outcome`: `run` delegates to a concrete `SubprocessSpawner` spawning `current_exe`, so the stamping would otherwise rest on my having noticed it, which is how it came to be missing.

## The compiler catch has a blind spot, and it is documented rather than enforced

A non-optional field conscripts the compiler only for **literal** constructions. `PersistedJob { job_id, ..Default::default() }` compiles untouched and stamps nothing.

No production site uses that form. It is **not enforced**, and the attempt is worth recording:

- ui-revamp verified the property held on main and suggested a source-scanning test.
- I wrote it. It passed. I then planted a dodging production writer to check it caught one: `mutation-check: FAIL — the test PASSED with the fix reverted.`
- Cause: `state.rs` has **nine** `#[cfg(test)]` blocks, the first appearing before `PersistedJob` is even defined. "Everything after the first `#[cfg(test)]` is test code" made the scanned production region **empty** — which reads exactly like a clean result.
- A correct scan needs brace-depth tracking across nine regions. That is a parser, not a grep, and not something to land on the tag's critical path.

So the guard is removed and the property lives on the field, where someone adding a writer will meet it. The durable fix is a constructor that stamps, with the literal reserved for tests — after the tag.

## Verification

`cargo test` — rocky-cli 2026, rocky-core 2213, rocky-server 21, and every other target, 0 failed. Clippy `--all-targets --all-features` clean on all three, `cargo fmt --check` clean.

**Mutation-checked:** the legacy boundary. Reverting `absent OR below MIN_TRUSTED` to absent-only makes `a_legacy_job_record_serves_status_and_withholds_its_payload` fail, so the `Some(0)` case is genuinely pinned rather than incidentally passing.

Tests cover: a result scrubbed before storage; a result withheld rather than corrupted when the rewrite would break the JSON; both legacy shapes (`None`, `Some(0)`) withholding while lifecycle fields still answer; both trusted shapes (`Some(1)`, `Some(999)`) serving; and the scheduler writer stamping on both a failed and a clean run.

**Whether `result_withheld_unparseable_after_redaction` is the right trade.** When redaction would break the JSON, the entire payload is replaced by that marker — so a job whose output happened to contain a registered value in a numeric position loses its whole result, not just that field. The operator gets a named absence instead of a diagnosis.

The alternative is walking the parsed value and replacing only within strings. **My first reason for rejecting it was wrong and is withdrawn:** I borrowed PR1's byte-parity constraint, which exists because `PrettyJson`'s bytes are compared against the CLI's. Nothing compares a stored blob's bytes, so that argument does not carry on this path at all.

The reason I would keep the marker, standing on its own:

- a stored record must never hold an unredacted payload, and a structural rewrite that PARTIALLY succeeded would store one;
- a stored blob is read back by code whose future I cannot see, so a rewrite that changes its shape changes what a later reader parses. A marker changes it visibly and once; a structural rewrite changes it subtly and everywhere.

I am less sure of this than of anything else in the PR, and a reviewer who thinks the persisted result should be walked structurally may well be right.

The scrub covers what the child prints. **It does not cover what the child's own process registered.** `execute_job_subprocess` spawns a separate `rocky` with its own registry; the parent scrubs against the parent's set. When both loaded the same config they agree, which is the ordinary case and not a guarantee.

Named in PR1's registry doc as a known limitation. backlog is pinning it as a labelled gap rather than a bug, so if it is ever closed the test says so.

**Independent red team: not yet run on this PR.** PR1 took three Codex rounds and every one found something. This is disclosed as unreviewed by an independent model at the time of opening; a pass should run before merge.

Refs #1920, #1919

### DIFF
--- a/docs/src/content/docs/guides/embedding.md
+++ b/docs/src/content/docs/guides/embedding.md
@@ -122,7 +122,7 @@ The polled result is the same payload the CLI would have produced.
 
 A submitted job becomes `running` right away. The schema also declares a `queued` state, but this server never uses it, because submissions never sit in a queue. Poll until the state is `succeeded` or `failed` rather than matching on the full set.
 
-There is no cancel route and no timeout route today. A mutating job holds the single mutation lane until its subprocess exits. While that lane is held, the next `run` or `apply` submission returns `409`, so a hung job blocks the next one. Restarting the sidecar is the only way to clear it. On restart the server reconciles its durable job ledger: it marks any job the previous process left non-terminal as `failed`, with the error `interrupted by engine restart`. A client that polls for a terminal state therefore always terminates. A cancel route is planned.
+There is no cancel route and no timeout route today. A mutating job holds the single mutation lane until its subprocess exits. While that lane is held, the next `run` or `apply` submission returns `409`, so a hung job blocks the next one. Restarting the sidecar is the only way to clear it. On restart the server reconciles its durable job ledger: it marks any job the previous process left `running` or `queued` as `failed`, with the error `interrupted by engine restart`. A client that polls for a terminal state therefore always terminates — a state this version does not recognise is reported as `failed` when the record is read, so polling ends either way. A cancel route is planned.
 
 ### Every failure carries an error envelope
 
--- a/engine/crates/rocky-cli/src/api.rs
+++ b/engine/crates/rocky-cli/src/api.rs
@@ -3421,6 +3421,23 @@ mod tests {
         let mut job = persisted_job("sink-principal", "running");
         job.principal = Some(secret.to_string());
 
+        // **The survivor must be the principal and nothing else.** The registry
+        // is process-global and monotonic, so a value registered by any other
+        // test in this binary could trip the sink instead and this test would
+        // still pass — it would just be proving something else. Ordering
+        // experiments cannot close that; this can. The same record with
+        // `principal` cleared must be clean.
+        let mut without = job.clone();
+        without.principal = None;
+        assert!(
+            !crate::secret_filter::any_value_survives(
+                &serde_json::to_string(&without).expect("serializes")
+            ),
+            "PRECONDITION: this record must trip the sink ONLY through \
+             `principal`. Something else registered in this process is \
+             matching it, so this test is no longer testing principal."
+        );
+
         let held = sanitize_for_storage(job);
         assert_eq!(held.principal, None, "a caller-supplied field is dropped");
         let after = serde_json::to_string(&held).expect("serializes");
--- a/engine/crates/rocky-cli/src/api.rs
+++ b/engine/crates/rocky-cli/src/api.rs
@@ -2484,16 +2484,37 @@ pub(crate) fn scrub_job_outcome(
         // and dropped entirely if it would carry a value. Storing nothing is
         // always available and cannot leak, which is what makes this
         // terminate (#1897).
-        let marker = withheld();
-        let marker_safe = serde_json::to_string(&marker)
-            .ok()
-            .is_some_and(|t| !crate::secret_filter::any_value_survives(&t));
-        const NOTE: &str = "withheld: a resolved value survived redaction";
-        return (
-            marker_safe.then_some(marker),
-            (!crate::secret_filter::any_value_survives(NOTE)).then(|| NOTE.to_string()),
-            rocky_core::state::CURRENT_REDACTION_VERSION,
+        let safe = |v: &serde_json::Value| {
+            serde_json::to_string(v)
+                .ok()
+                .is_some_and(|t| !crate::secret_filter::any_value_survives(&t))
+        };
+
+        // Prefer the descriptive marker. If its own text is registered, fall
+        // back to one that CANNOT be: every token here is shorter than
+        // `SECRET_LENGTH_FLOOR`, and the registry refuses to record anything
+        // below that, so no operator value can appear inside it. The reader
+        // therefore always gets a positive signal — never a bare `null`,
+        // which is indistinguishable from a run that produced nothing.
+        const _: () = assert!(
+            rocky_core::secret_registry::SECRET_LENGTH_FLOOR > 5,
+            "the short marker's tokens must be below the floor to be unregisterable"
         );
+        let short = serde_json::json!({ "held": true });
+        let marker = withheld();
+        let result = if safe(&marker) {
+            Some(marker)
+        } else {
+            Some(short)
+        };
+
+        const NOTE: &str = "withheld: a resolved value survived redaction";
+        let error = if crate::secret_filter::any_value_survives(NOTE) {
+            Some("held".to_string())
+        } else {
+            Some(NOTE.to_string())
+        };
+        return (result, error, rocky_core::state::CURRENT_REDACTION_VERSION);
     }
 
     (result, error, rocky_core::state::CURRENT_REDACTION_VERSION)
@@ -3205,12 +3226,17 @@ mod tests {
         let child = serde_json::json!({ "duration_ms": 1029384756u64 });
         let (result, error, _) = scrub_job_outcome(Some(child), None);
 
-        // The marker's own text is registered, so the marker cannot be used
-        // either — the withhold path must terminate in storing NOTHING.
-        assert!(
-            result.is_none(),
-            "when even the fallback marker would carry a value, nothing may be stored"
+        // The marker's own text is registered, so the descriptive marker
+        // cannot be used — but the reader still gets a POSITIVE signal, not a
+        // bare null. The short fallback's tokens are below
+        // SECRET_LENGTH_FLOOR, so the registry cannot hold them.
+        let result = result.expect("a withheld result is still a signal, never null");
+        assert_eq!(
+            result,
+            serde_json::json!({ "held": true }),
+            "when the descriptive marker is unusable, the unregisterable one is used"
         );
+        let result = Some(result);
         let stored = serde_json::to_string(&(result, error)).expect("serializes");
         assert!(
             !crate::secret_filter::any_value_survives(&stored),
--- a/engine/crates/rocky-cli/src/api.rs
+++ b/engine/crates/rocky-cli/src/api.rs
@@ -3259,8 +3259,10 @@ mod tests {
 
         // The marker's own text is registered, so the descriptive marker
         // cannot be used — but the reader still gets a POSITIVE signal, not a
-        // bare null. The short fallback's tokens are below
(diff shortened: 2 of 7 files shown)


hugocorreia90:
Filed #1946 for the registry-growth question raised during this work: entries are never removed, growth is bounded by distinct values rather than by reload count, and the cost that grows is per-response filter work rather than memory. This PR is correct with an unbounded registry, so nothing here waits on it.

hugocorreia90:
## Deferred to 1.75, with one open decision

1.74 cuts on PR1 (#1920) alone. This PR stays open and arrives in 1.75.

**Why deferring costs nothing observable:** #1920 already redacts every `rocky serve` response, which is the surface a caller can see. What this PR adds is the *stored-record* guarantee. The residual is the one 1.74 already discloses under #1897 and #1919.

**Reasoning:** five review rounds, each one finding something real, and the guarantee narrowed four times in a day. The remaining question deserves a deliberate decision rather than a release-night one.

### The one open decision

The claim "the stored record holds no registered value" is **falsifiable as written**:

```
${X}=redaction_version    17 bytes, over the floor
                          -> the withheld record still contains that JSON KEY

${X}=succeeded             9 bytes, over the floor
                          -> retained in `state`
```

My `sanitize_for_storage` doc argues that a collision discloses nothing, because every retained field is generated by this process. **That is a disclosure argument where a survival claim was needed.** They are different propositions and only one of them is in the doc.

Three ways forward, for whoever takes this:

1. **Narrow the claim** to caller-supplied and job-derived fields, and pin it with a test that stores such a record and asserts it persists with the payload withheld. Under an hour.
2. **Make no survival claim** in the notes and ship the code as-is.
3. **Change the envelope** so the payload cannot collide with its own keys. This removes the need to make the claim carefully at all, and it belongs with #1919 rather than here.

I recommend 3 as the real fix and 1 as the shippable one. **Not 2**, because the code does more than the notes would then say, and a future reader would re-derive the same question from scratch.

What I recommend against, and why it was raised: refusing to write on collision. An operator who registers `redaction_version` would then stop **every** job record from persisting, and an embedder polling until terminal would hang — precisely the failure the restart sweep exists to prevent. That trades a zero-information non-disclosure for a universal hang.

### What is already fixed and banked here

```
E   the sweep no longer deletes a newer binary's result on a downgrade
A   the redaction check moved to the serialized record at both disk sinks
B   the same check catches a value spanning a field seam
D   explicit struct literals; the invariant narrowed to what it delivers
    four stale prose claims, including a public docs page
    the seam test no longer poisons the principal test
```

Gate on the head: `rocky-cli --lib` 2035 passed 0 failed, workspace 7417 passed 0 failed, `cargo +1.98.0 clippy --all-targets --all-features` clean, fmt clean, six mutation checks each confirmed fix-sensitive.

### One method note worth keeping

I first tried to prove the seam test no longer contaminated the principal test by running a mutation with both in one process. That evidence was worthless: `--test-threads=1` runs the principal test **first**, which is exactly the contamination-free order, so a green result said nothing about the other order.

An ordering experiment can only sample the orders. The test now asserts the discriminating fact instead — the same record with `principal` cleared must be clean — which covers all of them, and a loud precondition fires if anything else in the process matches.

hugocorreia90:
Reviewed as assigned. **All four of the checks release-cut asked for are satisfied**, each verified against the branch rather than taken from the description. One finding, below, which I do not think blocks but is worth a decision rather than a silent accept.

## The four checks

**1. Every production writer stamps.** I built the inventory independently on this branch, and resolved production-vs-test from the `#[cfg(test)]` boundary rather than by line number — `api.rs` has exactly one at 3098 of 8775, so the split is unambiguous here.

```text
api.rs:2663    interrupted-at-restart   stamps via scrub_job_outcome
api.rs:2755    sanitize_for_storage     stamps CURRENT_REDACTION_VERSION
api.rs:2856    HTTP submit              stamps CURRENT_REDACTION_VERSION
spawner.rs:79  finish_scheduler_job     stamps via scrub_job_outcome
spawner.rs:108 scheduler running record stamps CURRENT_REDACTION_VERSION
```

Five, all stamped. The comment on `spawner.rs:108` names the exact failure release-cut was worried about — a writer that does not stamp makes its records read as pre-redaction forever, so the error is withheld on every run it writes. Good place for it: the next person to add a writer is looking at that field.

**2. The scrub precedes both sinks.** `scrub_job_outcome` runs before `jobs.upsert` and before `persist_job` at `api.rs:2911-2916`, with the reason stated — "a reader that races the persist sees the same bytes a restart would". That is the ordering claim, and it holds.

**3. The legacy boundary is `absent OR below`.** `state.rs:3189` is `.is_none_or(|v| v < MIN_TRUSTED_REDACTION_VERSION)`, so a `Some(0)` record is withheld exactly like an absent one. You mutation-checked it (absent-only fails the test), which is what makes it a pinned property rather than a passing one.

The treatment of a version this binary does not know — trusted, on the monotonic-strictness contract that a newer rule redacts at least as hard — is the right call and is argued rather than assumed.

**4. The spawner's error goes through the seam.** `finish_scheduler_job` routes a fixed-template message through `scrub_job_outcome` anyway, and says why: *"this particular string is safe" is an argument that stops being true when someone edits the string.* That is the reason to use a seam rather than trust a string, stated better than I would have put it.

## Finding: the last-resort net guards one sink, and the cache is the other

`sanitize_for_storage` is described as checking "HERE, on the record about to be serialized, not on the fields that composed it" — a sink-level net that catches what the field-level scrub cannot, including a value spanning two fields.

It is applied at both durable writes (`api.rs:2677`, `:2787`) and at neither of the three production cache writes (`:2898`, `:2915`, `:3090`).

That matters because of the field the net exists to drop. `principal` is caller-supplied — `principal_from_headers(headers)` at `:2840` — and `sanitize_for_storage` explicitly drops it (`principal: None`, "caller-supplied, and advisory rather than load-bearing"). It is also the vector in your own test, `a_value_spanning_two_record_fields_is_caught_at_the_sink`, which builds the span out of `principal`. So the record that reaches the in-memory cache still carries the field your sink test demonstrates is exploitable, and the sink that catches it is the one the cache does not pass through.

**Severity: not a disclosure, a divergence.** `GET /api/v1/jobs/{id}` served from cache still passes the outermost filter, which runs `any_value_survives` over the body and fails closed — so a surviving value produces `500 redaction_unavailable`, not a leak. But the same job read after a restart comes off disk already held, and returns a clean status-only record. Same job, two answers, decided by cache residency:

```text
  cache hit   -> 500 redaction_unavailable
  cache miss  -> 200, status only, error withheld
```

The second is the behaviour this PR is for. The first is a 500 that looks like an engine fault.

**The cheap fix is one call:** `jobs.upsert(sanitize_for_storage(record))` at the three sites, matching the durable path. If you would rather not — because the filter already fails closed and the window is one process lifetime — then it is worth one line on `sanitize_for_storage` saying it is a durable-sink net by choice, so the next reader does not assume both sinks are covered because the doc says "the record about to be serialized".

## Not blocking

I would take the three-call version, but this is a behavioural inconsistency in an error path rather than a leak, and you are closer to the tag's remaining schedule than I am. Either way I would want the choice written down rather than left to the reader.

Everything else checks out. Ready to confirm once CI is green — and I will look at `Analyze (rust)` specifically rather than reading a `skipping` aggregate as a pass, for the reason we both learned on #1920.

PR #1950, opened 2026-09-12
fix(engine/rocky-core): an absent [checks] table now means the same as an empty one

### LINKED ISSUE #1924: An absent [checks] table loses its serde defaults, so the check gate is off
Declaring an **empty** `[pipeline.X.checks]` table gives different defaults
from **omitting** the table. Omitting it silently turns the check gate off.

```
  [pipeline.X.checks] declared, empty  ->  fail_on_error = true    anomaly_threshold_pct = 50
  [pipeline.X.checks] absent           ->  fail_on_error = false   anomaly_threshold_pct = 0
```

## Why

`ChecksConfig` (`engine/crates/rocky-core/src/config.rs:1200`) derives
`Default`. Its fields carry serde defaults:

```rust
#[derive(Debug, Clone, Default, Serialize, Deserialize, JsonSchema)]
pub struct ChecksConfig {
    #[serde(default = "default_anomaly_threshold_pct")]   // 50.0
    pub anomaly_threshold_pct: f64,
    #[serde(default = "default_fail_on_error")]           // true
    pub fail_on_error: bool,
}
```

The `checks` field on each pipeline config is `#[serde(default)]`
(e.g. `config.rs:5107`). When the table is absent serde calls
`ChecksConfig::default()` — the **derived** one, which cannot see the
`#[serde(default = "...")]` attributes. Every field falls back to its Rust
zero value.

When the table is present, serde walks the fields and the attributes apply.
That is why an empty table and an absent table disagree.

## Reproduce

Add to `engine/crates/rocky-core/src/config.rs` tests and run it:

```rust
let declared: RockyConfig = toml::from_str(&format!("{base}\n[pipeline.bronze.checks]\n")).unwrap();
let absent:   RockyConfig = toml::from_str(base).unwrap();
// declared: fail_on_error=true  anomaly_threshold_pct=50
// absent:   fail_on_error=false anomaly_threshold_pct=0
```

(`base` is any valid replication pipeline with no `[checks]` table.)

## Reachability — no behaviour change found today

I could not find a path where this changes a run. A config with no `[checks]`
table also has every check switched off, so no failing check exists for the
gate to miss:

- `row_count`, `column_match`, `freshness`, `null_rate` are all `false`
- `[[assertions]]` live inside `[checks]`, so there are none
- anomaly detection is gated on `row_count_enabled`
  (`commands/run.rs:7044`), so `anomaly_threshold_pct = 0` is never read

So this is a **latent trap**, not a live defect. It becomes live the moment a
check can fail without a `checks.*` toggle enabling it — then the gate is off
and nothing says so.

## Scope

`ChecksConfig` is the **only** struct in `config.rs` with this shape. Eleven
others carry serde field defaults, and every one of them has a manual
`impl Default`: `AiSection`, `BranchApprovalConfig`, `CostSection`,
`ExecutionConfig`, `IdempotencyConfig`, `LoadOptionsConfig`,
`ResilienceConfig`, `RetryConfig`, `ReuseConfig`, `ScheduleDefaultsConfig`,
`StateConfig`.

Not verified: whether each of those manual impls actually *matches* its serde
defaults. A manual impl removes this failure mode but can still disagree in
its own way.

## Suggested fix

Replace `ChecksConfig`'s derived `Default` with a manual impl that calls the
same functions, matching the convention the other eleven already follow:

```rust
impl Default for ChecksConfig {
    fn default() -> Self {
        Self {
            anomaly_threshold_pct: default_anomaly_threshold_pct(),
            fail_on_error: default_fail_on_error(),
            ..
        }
    }
}
```

A test asserting the empty table and the absent table agree would stop it
returning.

## How it surfaced

Writing the `column_match` test in #1923. The first draft failed because the
gate never tripped — the fixture had no `[checks]` table, so `fail_on_error`
was `false`. The test was asking a gate that was switched off. `gate_for` in
that PR now asserts `fail_on_error` so no future test can measure the wrong
gate quietly.

Related: #1923, #1871, #1741. Each stays open independently of this one.

hugocorreia90:
Closes #1924.

Omitting `[pipeline.X.checks]` switched the failure gate off. Declaring the table and leaving it empty did not.

```
[checks] declared, empty  ->  fail_on_error = true    anomaly_threshold_pct = 50
[checks] absent           ->  fail_on_error = false   anomaly_threshold_pct = 0
```

`ChecksConfig` **derived** `Default`, and a derived `Default` cannot see `#[serde(default = "...")]`. The `checks` field on a pipeline is `#[serde(default)]`, so the two cases took different routes:

```
[checks] present, empty  ->  serde walks the fields
                             -> #[serde(default = "f")] applies, f() runs
[checks] absent          ->  serde calls ChecksConfig::default()
                             -> the derive, every field its Rust zero value
```

`Default` is now written by hand, calling the same `default_*` functions the attributes name — the convention the other eleven config structs with field defaults already follow.

## The report's unverified half, now verified

The issue was explicit that it had **not** checked whether those eleven manual impls actually match their serde defaults. They do.

Checked empirically rather than by reading eleven impls: compare `T::default()` against deserializing an empty object, for all twelve. Eleven agreed. `ChecksConfig` was the only mismatch, on exactly the two fields reported.

That comparison is now a test, so the class cannot return for any struct on the list.

**It compares `Debug` output, not JSON.** `ChecksConfig`'s own `cross_source_overlap` carries `skip_serializing_if`, so it is absent from the serialized form and a JSON comparison could not see a mismatch in it. `Debug` prints every field.

**The list is maintained by hand, and the test says so.** An automatic version would have to find every `#[serde(default = "...")]` and attribute it to its enclosing struct, which needs brace-depth tracking rather than a grep — and a scan that silently matches nothing reads exactly like a scan that passes. I have shipped that mistake before. Adding a name to a list is cheap; a guard that lies is not.

## Why it was latent, more precisely than the report had it

The report said a config with no `[checks]` table also has every check off, so no failing check exists for the gate to miss. That holds. Two structural facts narrow it further:

| Variant | `checks` defaults? | Runs pipeline-level checks? |
|---|---|---|
| `QualityPipelineConfig` | **no — required** | yes |
| `ReplicationPipelineConfig` | yes | yes |
| `TransformationPipelineConfig` | yes | no |
| `SnapshotPipelineConfig` | yes | no |
| `LoadPipelineConfig` | yes | no |

The variant whose whole purpose is running checks **requires** the table, so it could never hit this. Three of the four that default it run no pipeline-level checks at all. The live surface was replication alone, where every individual check was also off.

Still a latent trap rather than a live defect. It becomes live the moment a check can fail without a `checks.*` toggle enabling it — then the gate is off and nothing says so.

## The bug was published as the contract

Not in the report, and the reason this PR touches five generated files. The project schema bakes in `ChecksConfig::default()`:

```json
"checks": { "default": { "anomaly_threshold_pct": 0.0,
                         "fail_on_error": false, ... } }
```

That block is what an editor shows a config author for an omitted `[checks]` table. So the schema documented the bug as intended behaviour. It now reads `50.0` and `true` in `schemas/`, the vscode copy, the generated SDK model and `openapi.json` — four variants, not five, for the reason in the table above.

## One thing the first codegen run taught me

My first draft put the rationale in a **doc** comment on the struct. `JsonSchema` exported it, so a paragraph about Rust derive macros landed in the published project schema, a TypeScript interface and a Python docstring — visible to a config author in an editor tooltip.

It is now a plain `//` comment. The reasoning stays next to the code and out of the user's tooltip. Codegen is what surfaced it; I would not have predicted it.

## Verification

```
cargo test -p rocky-core --lib                2238 passed  0 failed
cargo +1.98.0 clippy --all-targets --all-features    clean
cargo fmt --all --check                              clean
mutation-check (derive Default restored)      BOTH tests fail -> fix-sensitive
```

Checked and deliberately **not** done:

- `just regen-fixtures` — dagster carries no reference to these fields.
- The SDK's hand-written `types.py` `ChecksConfig` — a narrower projection exposing only `freshness` and `configured_checks`, so it does not carry either changed field.

Whether `fail_on_error` flipping to `true` for an absent `[checks]` table can fail a run that used to pass. I believe not, for the reason in the table: on replication, the only affected variant that runs checks, an absent table also leaves every check disabled, so there is no error-severity failure for the gate to act on. That is the report's argument and I verified its premises, but it is an argument about reachability rather than a test, and reachability arguments are where I have been wrong before.

The parity test covers a hand-maintained list. The real fix for the class is to make the derive unavailable — a struct carrying `#[serde(default = "...")]` should not be able to derive `Default` at all. That is a lint or a proc macro, not this PR, and it would remove the need for the list.


hugocorreia90:
## Independent review: Codex, with `gpt-5.6-terra` as its own red team

Verdict: **merge with named follow-ups.** All follow-ups are done in `baecb668`. Dispositions below, including one place the review's own scan was wrong.

### 1. "This cannot fail a run that used to pass" — narrowed, not dropped

**Adopted.** The claim holds for `rocky.toml` → CLI, which the review verified independently and more thoroughly than I had: task construction reads only the per-check fields, every result producer is separately gated, the final gate is `fail_on_error && error_failures > 0`, and the anomaly threshold's single runtime reader sits inside the `row_count_enabled` guard. It also cleared paths I had not: no check-enabling CLI flag, `--env` affects masking only, DAG and scheduler reuse the same loaded config, and product/fulfill checks are a separate engine that never references `ChecksConfig`.

**Where the word "cannot" breaks.** `ChecksConfig` is public and `rocky_cli::commands::run::run` accepts a caller's `LoadedConfig` without reparsing it. So:

```rust
ChecksConfig { row_count: AggregateCheckToggle::Bool(true), ..ChecksConfig::default() }
```

A failed row-count check was advisory before (`fail_on_error = false` from the derive) and gates now. No such construct-and-enable path exists in this repository — the review grepped every `ChecksConfig::default()` and found only test helpers.

So the claim is **scoped to configurations deserialized from `rocky.toml`**, which is what I should have written. For a programmatic caller building `ChecksConfig` by hand, the new default is the documented one, and getting the old behaviour means setting `fail_on_error = false` explicitly.

### 2. "The twelve-type list is complete" — false, and widening it found something the review missed

**Adopted.** A brace-depth scan found six more structs with both `Default` and a `#[serde(default = "...")]` field. The review concluded all six "appear correct today". Adding them to the test proved that wrong for one — not as a parity mismatch, but as a category error in my test:

```
PreviewConfig   has a REQUIRED field, `branch`, so it cannot deserialize
                from `{}` at all. It can never be the target of a
                #[serde(default)] field without an empty table becoming
                a parse error.
```

Four joined the list: `QuarantineConfig` (from this very file — the omission I find least excusable), `ComparisonThresholds`, `StateRetentionConfig`, `ShadowConfig`. `PreviewConfig` got its own assertion instead: it must keep failing to parse from `{}`, so making it a defaulted table stops being silent.

`LoadOptions` is the seventh. It lives in `rocky-adapter-sdk`, which neither depends on `rocky-core` nor is depended on by it, so it has its own copy of the test beside it rather than being left uncovered.

One more thing the widening clarified, now recorded in the test: **`#[serde(default)]` on an `Option<T>` yields `None`, never `T::default()`.** An Option-wrapped field is not in this class. `QuarantineConfig` earns its place on its own fields, not through `checks.quarantine`.

The review also confirmed the `Debug` comparison is sound for every type on the list — all derive `Debug`, none hand-writes it, and `StateConfig`'s nested `RedactedString` cannot hide anything because `valkey_url` is `None` in both cases.

### 3. The variant table — holds

**No action.** Verified independently, including the route I most wanted checked: `PipelineConfig` has a hand-written deserializer that calls `try_into::<QualityPipelineConfig>()` directly, with no outer `Option`, flatten, or variant default. So quality's `[checks]` really is required. It also traced `run_quality`'s only production call site and confirmed transformation, snapshot and load dispatch to separate runners rather than falling through to replication.

### 4. Generated artifacts — completeness held, idempotence now verified

**Adopted the gap, closed it.** The review could not run `just codegen` in a read-only sandbox, so it marked byte-for-byte idempotence unverified rather than passed. I ran it: clean tree before, clean tree after, no diff.

Its static checks I did not do: no surviving `"anomaly_threshold_pct": 0.0` or `"fail_on_error": false` anywhere in the repo, all four Pydantic defaults carry `50.0`/`True`, quality stays required in the generated SDK, and the VS Code schema differs from the root only by the generated `$comment` that `copy_project_schema.py` adds deliberately. It also justified skipping `just regen-fixtures` on the data path rather than on my assertion, and noted the main fixture POC already declares an explicit `[checks]` table.

### 5. Back-compat and the changelog

**Changelog entry added.** It states the resolved-default change, why no run is known to behave differently — with the mechanism rather than the assertion — that the published schema was wrong and visible, and how to keep the old behaviour.

**One premise of mine the review correctly disputed.** I told it this repo has no backward-compatibility obligation before 2.0. That is the owner's stated policy, but it is not what the repo's own files say: the engine is `1.73.0` and `engine/CHANGELOG.md` states it adheres to Semantic Versioning. I am not resolving that here — flagging it, because I asserted it as a repo fact when it is a policy I was told, and the two disagree in public text.

### Two comments corrected, one of them mine

Mine said the `checks` field defaults on "each pipeline". It defaults on four of five. That is the same shape as the defect this PR fixes — a claim broader than the code — which is why the review was right to name it rather than let it pass as prose.

An existing `run.rs` test comment still explained that `ChecksConfig` derives `Default` and an absent table yields `false`. This branch made that false. The assertion it guards stays, because a caller can still hand in a config with the gate off; only the stale explanation changed.

### Evidence after the follow-ups

```
cargo test -p rocky-core --lib          2238 passed  0 failed   exit 0
cargo test -p rocky-adapter-sdk --lib     66 passed  0 failed   exit 0
cargo +1.98.0 clippy --all-targets --all-features    clean
cargo fmt --all --check                              clean
just codegen (second run)                            no diff
```

Issue #1930, opened 2026-09-11
A non-empty table with an all-NULL timestamp column is never freshness-checked


hugocorreia90:
A freshness check on a **non-empty** table whose timestamp column is entirely
NULL has never run, on any adapter. No failing check, no not-evaluated check —
nothing. The run exits 0.

This is piece 2 of #1929, split out so the Databricks piece can close
independently. **This piece ships open in 1.74.**

## The query cannot tell the two cases apart

```
  SELECT MAX(ts) FROM t

  genuinely empty table       -> NULL     correct: nothing to be fresh
  NON-EMPTY, ts all NULL      -> NULL     WRONG: never checked
```

Verified: a two-row table with both timestamps NULL gives
`COUNT(*) = 2, MAX(ts) = NULL`.

The freshness query selects only `MAX`, on both paths:

- batched: `generate_batch_freshness_sql`, `rocky-databricks/src/batch.rs:268`
- unbatched: `per_table_freshness`, `rocky-cli/src/commands/run.rs:6835`

So the information needed to separate them **is not in the response**. This
cannot be fixed by changing a Rust return type — the query has to ask for it.

`run.rs:7110` then does `let ts = fr.max_timestamp?` inside a `filter_map`, and
the table leaves the iterator.

## Why it needs no malformed data

`timestamp_column` is freely configurable (`config.rs:5076`) with no non-null
invariant. A column that exists but was never populated, or an optional source
field, is enough. That makes this more likely in the field than the unreadable-
cell case in #1929, which needs a malformed warehouse response.

## Suggested fix

Select a row-presence signal alongside the maximum — `SELECT COUNT(*),
MAX(ts)` — in both the batched and unbatched freshness queries, so the source
distinguishes them:

```
  COUNT(*) = 0                 -> emit no check       (unchanged, correct)
  COUNT(*) > 0, MAX(ts) NULL   -> freshness_not_evaluated, gates
```

`COUNT(*)` resolves **only** this distinction. It does not help with an
unreadable cell (missing, non-string, unparseable) — that needs the decoder to
report what it could not read, which is #1929 and #1928.

## Scope

Every adapter that runs freshness, batched or not. The query generators change,
so each warehouse needs a live check rather than a unit test alone: the
`COUNT(*), MAX(ts)` shape has to be valid in each dialect and the column order
has to match what the parser expects.

## How it surfaced

Review of #1927. I claimed the freshness collapse was Databricks-batch-only,
[...]

Issue #1929, opened 2026-09-11
A Databricks batched freshness timestamp that cannot be read emits no check


hugocorreia90:
**Scope narrowed 2026-09-11.** This issue is now **piece 1 only**: the
Databricks batch path collapsing an unreadable timestamp into `None`.

The larger half — a non-empty table with an all-NULL timestamp column, which
affects every adapter and needs a query-shape change — is **#1930**, and ships
open in 1.74.

This issue was originally filed with two wrong scopes in succession. Both
corrections are recorded at the bottom, because the reasoning errors are the
reusable part.

## The defect

On the Databricks **batch** freshness path, a timestamp cell that cannot be
read becomes `max_timestamp: None`, which the consumer treats as "empty table,
emit no check". So an unreadable timestamp emits **no freshness check at all**.

Three states reach that `None`:

```
  missing cell (short response row)  -> batch.rs:310, via row.get(3)
  cell is not a string               -> batch.rs:310
  string will not parse              -> adapter.rs:447
```

`run.rs:7110` then does `let ts = fr.max_timestamp?` inside a `filter_map`.
Because the row **was** returned, the "absent from results" arm below never
fires — the table just leaves the iterator.

The connector does not check row length against the manifest
(`connector.rs:235`, `connector.rs:346`), so a short row is reachable.

## Why the unbatched path is not affected

`per_table_freshness` (`run.rs:6844-6875`) already separates an unreadable cell
from a null one and records a reason. Its own comment notes the unreadable case
"used to be dropped without a trace" — this class was already fixed there, for
these states.

It shares the all-NULL defect, but that is #1930.

## The fix

Same shape as #1926, reusing the derivation rather than inventing a state:

- Keep the row with `max_timestamp: None` when the cell is a genuine SQL NULL.
- **Omit** the row when the cell is present but unreadable.
- `run.rs`'s existing absent-result arm (`~7116`) then emits
  `freshness_not_evaluated` at Error severity, and the gate trips.
- The specific reason goes to the adapter log, since the trait carries no
  per-table reason channel (#1928).

Both layers can make the null-versus-unreadable distinction locally
(`batch.rs:310` for the cell, `adapter.rs:447` for the parse), so no signal
needs to travel between them.

Also reword the freshness reason from "returned no result for this table" to
"returned no readable timestamp for this table". Once unreadable rows are
omitted there are two routes to that arm, so naming only the missing-row case
points an operator in the wrong direction. Its assertion at `run.rs:~34231`
[...]

PR #1956, opened 2026-09-13
docs: four pages the 1.74 check work left behind

hugocorreia90:
Three rules shipped in 1.74 that the docs never stated, and one gap the docs never stated at all. Then four red-team passes that falsified most of my first attempt.

## What the pages say now

| Page | Was | Now |
|---|---|---|
| `reference/configuration.md` → `[pipeline.NAME]` | "`value` is a SQL expression" | One expression over a scalar allowlist, checked at config load, parsed against the target adapter. Eight examples that pass, two that do not. |
| `reference/commands/core-pipeline.md` → `branch promote` | Branch **name** rules only | Target names have their own limit: a table of the refused character per warehouse, plus the two limits that sit outside that check. |
| `concepts/data-quality-checks.md` → Freshness | What the check measures | What happens when there is nothing to measure, and what each consumer shows. |
| `concepts/bronze-layer.md` | An example with no rule beside it | Points at the reference. |

## The freshness gap is the one worth reading

`MAX(timestamp_column)` answers `NULL` for an empty table and for a full table whose timestamp column is entirely NULL. Rocky cannot tell those apart, so it emits no check.

```
table with data + NULL timestamps
        |
        v
  MAX(ts) -> NULL
        |
        v
  "nothing to measure" -> no check emitted
        |
        v
  CLI JSON: looks EXACTLY like no freshness configured
```

#1930 stays open. This documents the behaviour as it is today.

## A test now pins the examples the reference page prints

Three of the eight accepted `value` examples were written from reading the allowlist, not from running them. They run now, in `a_metadata_column_value_off_the_function_allowlist_is_refused`.

Running the ninth case corrected the page. It said an unquoted placeholder "reads as a bare column reference rather than a value". It does not — the parser never gets that far:

```
metadata_columns[].value: expression does not parse as a single SQL expression
(sql parser error: Expected: :, found: } at Line: 1, Column: 8)
```

The test asserts that reason text, so the page and the message cannot drift apart quietly.

## What the red team changed

Four passes, four NOT READY verdicts, about twenty findings. The five that mattered most:

| Claim | What the code does |
|---|---|
| "uses the dialect of the **target** adapter" | True for DuckDB, Snowflake, BigQuery, Databricks. Trino has no dialect in the parser and takes a generic one. |
| "Nothing else is refused" | False twice. A transformation model taking its schema from a group `schema_template` goes through `[A-Za-z0-9_]` only. And promote refuses two steps replacing one production table. |
| "A table with **no usable timestamp** emits no check" | Only a query that succeeds and answers `NULL`. A failed query, or a cell that is not a timestamp, reports a failed `freshness_not_evaluated`. |
| "Looks the same in **the output**" | True for the CLI's JSON. Outside Pipes mode Dagster fills the gap with a placeholder, and what it says depends on why the result is missing. |
| Row-count and freshness checks "batch with `UNION ALL`" | Databricks alone. Snowflake and BigQuery batch the schema describe only and run both check legs per table. |

That last one was stated on two pages and in the changelog, none of which this work set out to touch. One wrong sentence, four copies.

Three corrections landed outside the four original pages, because the claims did:

- `engine/CHANGELOG.md` — three statements in the 1.74.0 entry, including one that overstated the blast radius of its own fix.
- `reference/cli.md` — the run-sequence list.
- `concepts/schema-drift.md` — the page the freshness page now defers to. It omitted what pruning actually needs, including that an incremental table's first run always copies.
- `run.rs` — a test doc comment that still described behaviour #1929 changed. Comment only.

## Verification

- `cargo test --lib -p rocky-core` — 2238 pass.
- `cargo clippy -p rocky-core --all-targets -- -D warnings`, and the same for `rocky-cli` — clean.
- `cargo fmt` applied.
- The red team independently ran 51 `commands::branch::tests`, the config-group identifier test, the batched-NULL freshness test, and two Dagster component tests.

That the claims are now complete rather than merely narrower. Each pass found the previous fix had opened an adjacent claim, and the last two passes found errors on pages this work never set out to touch. The stopping point is a judgement, not a proof.

## What I might be missing

A surface none of the five passes opened. Four were found by asking for one each time, which is evidence the search works and evidence it was not finished.

## Left for its own change

The `prune_unchanged` field doc in `rocky-core` says "provably unchanged" and omits the watermark condition. That text is the source for `schemas/rocky_project.schema.json`, the generated Pydantic and TypeScript bindings, and `docs/public/openapi.json`. Correcting it regenerates all four, which does not belong on a docs branch.

## Red team

Codex (`codex:codex-rescue`), an independent model, cross-checked by `gpt-5.6-terra`. Both questions above were handed to it explicitly on every pass. **Excluded:** docs pages outside those named here, and the full `rocky-cli` test suite.

Refs #1934, #1940, #1930, #1929


hugocorreia90:
Two defects this work found and deliberately left out of the branch:

- #1959 — a refused `metadata_columns[].value` is explained as a data-quality check. The refusal is right; the advice beside it is for a different feature. Message only.
- #1960 — the `prune_unchanged` field doc omits the watermark condition. That text is the source for five generated files, so fixing it is a `just codegen` change across three subprojects.

Both stay open. This PR is docs only.

PR #1968, opened 2026-09-15
fix(engine/rocky-sql): a refused metadata column value is explained as a column value, not a check

### LINKED ISSUE #1959: A refused metadata_columns[].value is explained as a data-quality check, which it is not
A `metadata_columns[].value` that Rocky refuses is explained as if it were a data-quality check. It is not one, and the advice does not apply.

## What you see

```toml
[pipeline.bronze]
metadata_columns = [
    { name = "_tenant", type = "VARCHAR", value = "{tenant}" },
]
```

```
metadata_columns[].value: expression does not parse as a single SQL expression
(sql parser error: Expected: :, found: } at Line: 1, Column: 8). An expression
check is one boolean expression over the model's columns, e.g. `amount >= 0`
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                this half is about a different feature
```

The first half is right and useful. The second half tells the author to write a boolean over the model's columns. A metadata column value is neither. `value = "NULL"`, `value = "1"` and `value = "'rocky'"` are all accepted, and none of them is a boolean.

So the message states a rule the code does not apply in this mode.

## Why it happens

`validate_metadata_columns` reuses the expression-check validator, which is the right call — one validator, one allowlist, no drift:

```
metadata_columns[].value ─┐
                          ├─► validate_check_expression ─► four error variants
[checks.assertions] ──────┘                                (all say "expression check")
```

The four variants are `ExpressionUnparseable`, `ExpressionTrailingTokens`, `ExpressionSubquery` and `ExpressionFunctionNotAllowed` (`engine/crates/rocky-sql/src/validation.rs:83-108`). Each ends with guidance written for `[checks.assertions]`. The caller passes a `context` string, so the message already knows which field it is talking about; only the trailing advice is fixed.

`validate_metadata_columns` calls it with `ExpressionUse::ScalarProjection` (`engine/crates/rocky-core/src/config.rs:3622-3631`), so the mode is already distinguished at the call site.

## Scope

Four messages, one caller affected. The `[checks.assertions]` wording stays correct for its own field.

## Not a correctness bug

The refusal itself is right. Only the explanation is wrong, so this is a message change, not a behaviour change.

Refs #1934, #1524, and PR #1956, which documents the accepted and refused shapes on the reference page.

hugocorreia90:
A refused `metadata_columns[].value` is now explained as a column value. Before, every refusal ended with advice written for `[checks.assertions]`, and told the author to write a boolean.

Closes #1959.

## What was wrong

One validator judges three fields, and each caller already says which one it is passing:

```text
  [checks.assertions] expression   ExpressionUse::SinglePredicate
  quarantine split predicate       ExpressionUse::ReevaluatedPredicate
  metadata_columns[].value         ExpressionUse::ScalarProjection
  unique_expr / cross_source key   ExpressionUse::GroupingKey
```

The five refusal messages ignored the mode. All of them said "expression check", and the parse-failure one said "one boolean expression over the model's columns, e.g. `amount >= 0`". A metadata column value is not a boolean: `NULL`, `1` and `'rocky'` are all accepted. So the message stated a rule the validator never applies in that mode.

```text
before  metadata_columns[].value: expression does not parse as a single SQL
        expression (...). An expression check is one boolean expression over
        the model's columns, e.g. `amount >= 0`

after   metadata_columns[].value: expression does not parse as a single SQL
        expression (...). A metadata column value is one scalar expression,
        e.g. `current_timestamp()`, `'rocky'` or `NULL`
```

The issue counted one affected caller. A key expression had the same problem: a grouping key is neither a boolean nor a column value, and it was called an expression check too. It is now "A key expression is one expression over the row's own columns, e.g. `lower(email)`".

## The fix

`ExpressionUse` gains three methods: the noun for the position, the same noun mid-sentence, and the shape the position accepts. The five `Expression*` variants of `ValidationError` carry a `use_: ExpressionUse` field, and their messages read the noun from it. Every construction site passes the mode that was validated; the one outside rocky-sql (`quarantine.rs`, the generated clock refusal under split mode) passes the mode it already held.

What each field accepts or refuses is unchanged. The `[checks.assertions]` wording is unchanged.

## Evidence

- `a_refusal_describes_the_position_it_was_validated_for` (rocky-sql) drives all five refusals in `ScalarProjection` and checks each message names a column value and none says "boolean". It also pins the predicate wording unchanged and the key wording.
- The rocky-core config test for `metadata_columns` now also asserts the refusal reason describes a column value.
- Mutation check: putting the boolean advice back on `ScalarProjection` fails the new test.
- `cargo test -p rocky-sql`: 318 passed. `cargo test -p rocky-core --lib`: 2238 passed. `cargo fmt --check` clean. Clippy `--all-targets --all-features -D warnings` on both crates is running and will be confirmed in a comment; CI runs the same gate.

## Not changed

The `ExpressionFunctionNotAllowed` message still ends with "add the function to `CHECK_EXPRESSION_FUNCTIONS`", including when the refusal is for a volatile function in a key or a `COLLATE`. That advice was already imprecise for those two paths and is not made worse here. Worth its own issue if anyone reads it and follows it.

## Red team

Independent review by Codex through the plugin is in progress on this diff. Its verdict and my disposition of each finding go in a comment on this PR before merge.

Whether "one expression over the row's own columns" is the right description for a `cross_source_overlap` key, which is evaluated on each of several sibling tables. It is one expression per row on each table, so I believe it holds. The reviewer is asked to check it.

## What I think I am missing

A caller outside `engine/crates` that wraps these errors with its own check-shaped hint. The MCP `draft_check` tool does, and it is correct there because it validates an `expression` test. I found no other.


hugocorreia90:
Gate on the head (5f8610b4):

- `cargo clippy -p rocky-sql -p rocky-core --all-targets --all-features -- -D warnings`: clean.
- `cargo test -p rocky-sql`: 318 passed. `cargo test -p rocky-core --lib`: 2238 passed, 0 failed.
- `cargo fmt --check`: clean.

Red team: the Codex plugin run failed before reviewing anything (the local Codex CLI is older than the configured model requires), so **no independent review is on the record yet**. A fresh-context same-model review is running instead; its findings and their disposition follow in a comment. If the CLI is upgraded before merge, I will re-dispatch Codex on this diff.

hugocorreia90:
## Review disposition

**Reviewer:** a fresh-context same-model pass (Claude, no prior context, read-only through git objects, 40 tool calls). Not independent in the sense `AGENTS.md` asks for: the Codex plugin failed to start because the local Codex CLI (0.146.0) is older than the configured model requires. This is disclosed rather than skipped; if the CLI is upgraded before merge I will re-dispatch.

**Verdict as returned: BLOCK**, on one false sentence. Every finding is addressed in the second commit; the code findings were confirmed against the diff before acting.

|---|---|---|---|
| 1 | The changelog said the `[checks.assertions]` wording was unchanged. False for the trailing-text refusal, which had taken the field sentence too | Confirmed by diffing the five predicate messages against `main`: four identical, one changed | **Fixed.** That refusal now reads "Only a single boolean expression is accepted — no trailing clauses, commas or operators" for a predicate, byte for byte what it was, and "scalar expression" / "key expression" elsewhere (`single_kind`) |
| 2 | The `noun` doc said the validator serves three fields; it serves eight call sites over seven | Confirmed by the reviewer's enumeration | **Fixed.** The doc lists them |
| 3 | A refused assertion `filter` is still called an expression check: the noun is keyed on evaluation mode, which a filter shares with its expression | Confirmed; same class as #1959, not a regression | **Not fixed here, tracked.** Filed #1971; the changelog and the `noun` doc both state the limit |
| 4 | The "1, 2" case asserted a substring both the parse failure and the trailing refusal emit, so it could not tell which fired | Confirmed | **Fixed.** It now asserts "continues past the end of one expression" as well |
| 5 | "Add the function to `CHECK_EXPRESSION_FUNCTIONS`" is false advice for a volatile function in a key, for `COLLATE`, and for the quarantine split's synthetic clock refusal | Confirmed; pre-existing | **Not fixed here, tracked** in #1971, since each wants its own sentence rather than a reword |
| 6 | The three helpers are `pub` with one in-crate consumer | Fair | **Fixed.** `pub(crate)`, including the new `single_kind` |

Also from the review, kept as evidence: every construction site passes the mode that was validated (12 sites, compiler-enforced); no downstream code matches the message text beyond the two sites already listed plus `rocky-core/src/tests.rs` pins that are unaffected; each of the five ScalarProjection inputs reaches the variant its assertion names under sqlparser 0.62; and `.use_.noun()` is the same thiserror form `config.rs` already uses for `.path.display()`.

The test now pins the predicate wording of all five refusals, so a future reword of the check advice fails it, which is what the earlier doc comment claimed and did not deliver.

Gate on the head: `cargo test -p rocky-sql` 318 passed, `cargo clippy -p rocky-sql --all-targets --all-features -D warnings` clean, `cargo fmt --check` clean. rocky-core is untouched by this commit; its gate from the first commit stands.

PR #1969, opened 2026-09-15
docs(engine/rocky-core): prune_unchanged field doc states the watermark condition

### LINKED ISSUE #1960: prune_unchanged field doc omits the watermark condition, and that text ships in five generated files
The `prune_unchanged` field doc says a table is pruned when its source is "provably unchanged". One more condition applies, and the doc does not mention it: an `incremental` or `microbatch` table is never pruned until it has a recorded watermark. Its first run always copies.

That text is not only a doc comment. It is the source for five generated files that ship to users.

## The actual rule

`engine/crates/rocky-cli/src/commands/run.rs:13331-13333`:

```rust
if (!uses_watermark || prior_watermark.is_some())
    && recorded.as_deref() == Some(live.as_str())
```

So pruning needs all of:

| Condition | Where |
|---|---|
| `prune_unchanged = true` | config |
| no `--no-prune` on the run | `run.rs:4911-4914` |
| the adapter offers a change-marker | `traits.rs:465-484` — one that cannot always copies |
| the stored marker equals the live one | `run.rs:13320-13332` |
| **for `incremental` / `microbatch`, a watermark already on record** | `run.rs:13272-13292`, `run.rs:13331` |

The field doc (`engine/crates/rocky-core/src/config.rs:5372-5390`) covers the first four. It does not cover the fifth.

## Why the last row matters to a reader

```
incremental table, prune_unchanged = true

  run 1  no watermark yet ──► COPY   (not pruned, even if the marker matches)
  run 2  watermark on record ─► prune, if the marker matches
```

Someone enabling `prune_unchanged` and watching the first run copy every table has no way to tell that from a broken marker.

## Where the text goes

Editing the doc comment regenerates all of these:

- `schemas/rocky_project.schema.json`
- `sdk/python/src/rocky_sdk/types_generated/rocky_project_schema.py`
- `editors/vscode/schemas/rocky-project.schema.json`
- `editors/vscode/src/types/generated/rocky_project.ts`
- `docs/public/openapi.json`

So this is a `just codegen` change across three subprojects, not a one-line edit. That is why it was kept out of PR #1956, which is docs only.

## Suggested fix

Add the watermark condition to the field doc, run `just codegen`, and commit the regenerated bindings with it.

`docs/src/content/docs/concepts/schema-drift.md` already carries the condition in prose as of PR #1956, so the two agree once this lands.

Refs #1930 and PR #1956.

hugocorreia90:
The `prune_unchanged` field doc now states the fifth condition: an `incremental` or `microbatch` table is never pruned until it has a recorded watermark, so its first run always copies.

Closes #1960.

## What changed

One paragraph added to the doc comment on `ReplicationPipelineConfig::prune_unchanged`, then `just codegen`. The doc comment is the source of five generated files, and all five carry the new sentence:

```text
schemas/rocky_project.schema.json
editors/vscode/schemas/rocky-project.schema.json
editors/vscode/src/types/generated/rocky_project.ts
sdk/python/src/rocky_sdk/types_generated/rocky_project_schema.py
docs/public/openapi.json
```

The reason given is the runner's own: a missing watermark means an incremental target cannot safely append (`run.rs`, above the prune check), so the first copy is what establishes the watermark and cannot be skipped. `docs/src/content/docs/concepts/schema-drift.md` already says the same thing in prose since #1956; the two now agree.

## Not in this PR

- No CHANGELOG entry: doc text only, no behaviour change.
- `just codegen` also rewrote `editors/vscode/package-lock.json` (this machine's npm drops the `libc` arrays). That is npm noise, not part of the cascade, and is left out.
- No `regen-fixtures`: the schema's shape is unchanged, only a description string.

Whether "cannot safely append" is the whole reason. It is what the runner's comment says, and the code gates on `prior_watermark.is_some()`, so the sentence is true to the code either way.

## What I think I am missing

Whether the reference configuration page repeats the old four-condition list in its own words. It does not quote this doc comment; #1956 put the condition on the schema-drift concept page, which is where a reader is sent.


hugocorreia90:
## Review disposition

**Reviewer:** a fresh-context same-model pass (Claude, no prior context, read-only through git objects, 69 tool calls across this PR and #1970). Not the independent reviewer `AGENTS.md` asks for: the Codex plugin cannot start on this machine today (CLI older than the configured model). Disclosed rather than skipped.

**Verdict as returned: BLOCK**, on one word. Fixed in the second commit.

|---|---|---|---|
| 1 | `microbatch` is not a replication strategy. `RECOGNIZED_REPLICATION_STRATEGIES` lists six values without it, the replication strategy builder has no arm for it (falls to `full_refresh`), and validate warns exactly that (V035). So a replication table spelling `strategy = "microbatch"` IS pruned on the first marker match, the opposite of the added sentence | Confirmed against `config.rs` and `run.rs` | **Fixed.** The sentence names `incremental` alone, in the source, the five generated files, and the schema-drift page, which carried the same slip since #1956 |
| 2 | The sentence is silent on `merge`, `view`, `materialized_view` and `dynamic_table`, which also have no watermark condition | Fair; "incomplete but not false" | Left as is. "A `full_refresh` table has no such condition" stays true; listing every strategy would put six names in a field doc for one exception |
| 3 | "even when the marker matches" is reachable: the marker is written in the table's success tail, the watermark at the end of the run, so a run that dies between the two leaves marker-without-watermark | Confirmed by the reviewer with line evidence | Kept, now with the reviewer's reasoning on record |

Also confirmed by the review: the five generated files carry exactly the one sentence and nothing else (six changed JSON lines, two added lines each in the TS and Pydantic files); no other page in `docs/` or `examples/` states the old four-condition rule; the reference configuration page has no `prune_unchanged` row at all and the core-pipeline command page has no `--no-prune` row (pre-existing gaps, not touched here).

The second commit edits the generated files by string substitution rather than re-running `just codegen`; the `Check codegen is up to date` context on this PR is the proof that they match what the generator produces.

PR #1970, opened 2026-09-15
ci(engine-weekly): coverage tolerates the raw profiles a SIGKILLed child leaves behind

### LINKED ISSUE #1952: engine-weekly Coverage has failed on every run for five weeks, so the workflow's conclusion carries no information
`engine-weekly`'s `Coverage` job has failed on **every run for at least five weeks**. The workflow has therefore been red on every run this month, which means nothing reads it and it gates nothing.

This is **not** #1638. That issue was the tarpaulin linker SIGBUS, its llvm-cov migration landed, and it was closed correctly — `engine-weekly.yml:171` now runs `cargo llvm-cov --workspace --cobertura`, and the SIGBUS signature is gone from the log. This is a new failure wearing the same job name.

## Observed

```
run          date        branch                     conclusion
34719059427  2026-09-12  chore/release-1.74.0       failure
34714350860  2026-09-12  chore/release-1.74.0       failure
34388542582  2026-09-09  fix/windows-test-symlinks  failure
34386418062  2026-09-09  ci/1638-llvm-cov           failure
34131121012  2026-09-07  main                       failure
33942451182  2026-09-05  ci/1638-no-dead-code       failure
33941059223  2026-09-05  ci/1638-no-dead-code       failure
33937668788  2026-09-05  ci/1638-no-dead-code       failure
```

Eight consecutive runs, `main` among them, so it is not branch-specific.

On the most recent run, the other three jobs are green:

```
Windows compile check   success
POC Smoke Tests         success
Security Audit          success
Coverage                failure     <- failing step: "Run coverage"
```

## The failure is not a test failure, and the cause is UNDETERMINED

Every `test result:` line in the job reads `ok` with `0 failed`:

```
test result: ok. 317 passed; 0 failed ...
test result: ok.  58 passed; 0 failed ...
… every one ok, no `failures:` block anywhere …
##[error]Process completed with exit code 1
```

`cargo llvm-cov` exits 1 **silently** after the tests pass. Searched the failed-job log and found none of:

```
could not compile        absent
Bus error / signal 7     absent  (so NOT #1638's signature)
out of memory / killed   absent
cobertura write error    absent
llvm-cov error message   absent
```

`cargo-llvm-cov 0.9.1` installs and starts cleanly. The step carries no `--fail-under-*` flag, so a coverage threshold is not the trigger.

**I could not determine the cause and am not guessing at one.** `gh run view --log-failed` returns the job's log ending in ordinary teardown (git credential cleanup, orphan-process reaping) with no error text before the exit.

Two false leads already burned, recorded so nobody repeats them:

- Grepping the log for `error|panic|cannot|denied` returns ~15 hits that are all **test names** containing those words, each ending `ok`.
- The runner-loss case the workflow's own comment warns about (`engine-weekly.yml:180-186`) does **not** apply: the job reached a conclusion and uploaded logs, so it did not lose its agent.

## Why it matters more than a missing number

A permanently-red job is indistinguishable from a broken one, so the whole workflow's conclusion reads `failure` every run. Two costs, both observed:

1. **It hid a green gate.** During the 1.74 pre-flight the release lane was waiting on `POC Smoke Tests`. That job passed; the workflow said `failure` because of `Coverage`. Reading the roll-up rather than the job list would have sent someone back to rework a fix that was already correct.
2. **Nobody can notice a real regression in it.** A signal that is always red carries no information, which is the same class of problem as a signal nobody runs — see #1676, where the proposed PR-time POC job was never built.

## What would settle it

1. Add `--verbose` to the `cargo llvm-cov` invocation, or split it into `cargo llvm-cov --no-report nextest` then `cargo llvm-cov report`, so the failing phase is named.
2. Capture the step's exit path explicitly — `cargo llvm-cov … ; echo "llvm-cov exit=$?"` — since the current log shows the shell's failure without llvm-cov's own message. (Note the inverse trap: a trailing `;` makes the *step* succeed, so this is for diagnosis only, never as the fix.)
3. `df -h /tmp` and `free -m` are already captured before and after (`:152-154`, `:187+`). Compare them across a green `POC Smoke` job and this red one in the same run.

## Not claimed

That coverage *should* gate the release. #1638 recorded that nothing in the release gate depends on it, and that still holds — `engine-weekly` is a scheduled workflow, not one of the seven required contexts. The argument here is about a signal that cannot be read, not about adding a gate.

Found during the `engine-v1.74.0` pre-flight, while establishing that the green `POC Smoke Tests` result was real. Related: #1638 (closed, correctly — different cause), #1676, #1951.

hugocorreia90:
`engine-weekly`'s Coverage job fails on every run because one or two raw profile files are truncated and `llvm-profdata merge` refuses the whole set when any input is invalid. This passes `--failure-mode all` so the merge tolerates them, and lists the raw profiles after the run so the producer can be pinned.

Closes #1952.

## The cause, read from the logs

The issue recorded the failure as undetermined: every test passes, then `cargo llvm-cov` exits 1 with no error text found. The text is there, three lines above the exit, and the earlier search missed it because it was filtering for `error|panic` and drowning in test names:

```text
warning: .../llvm-cov-target/engine-17808-11658474381410565424_0.profraw:
         invalid instrumentation profile data (file header is corrupt)
warning: .../llvm-cov-target/engine-18603-11658474381410565424_0.profraw:
         invalid instrumentation profile data (file header is corrupt)
error: no profile can be merged
error: failed to merge profile data: process didn't exit successfully:
       `.../bin/llvm-profdata merge -sparse -f ...`
```

Same signature on the 2026-09-12 run (one corrupt file, run 34719059427) and the 2026-09-14 run on `main` (two, run 34858466516).

**What a corrupt header means.** cargo-llvm-cov runs the tests with `LLVM_PROFILE_FILE=<name>-%p-%m.profraw`. In that merge mode the profile runtime writes nothing at startup; the file is written by the process's exit hook. A file with a corrupt header is one whose writer was interrupted during that exit-time write. So the producer is a test that signals an instrumented child while it is exiting. It is not pinned from the logs (the process ids appear nowhere else in them). The candidate most consistent with "one or two files, all from the `rocky` binary": the `rocky mcp` servers `fulfill_e2e.rs` spawns, which the fulfill driver SIGTERMs right after each replay, and `rocky mcp` installs no SIGTERM handler. A process killed with SIGKILL while idle writes no file at all, so the SIGKILL drill in `product::commit` is ruled out. (The first version of this PR blamed the drill; the review caught it.)

**Why it fails the whole job.** `llvm-profdata merge` defaults to `--failure-mode any`: one invalid input fails the merge, and it reports `no profile can be merged` in that case too, which reads as if nothing was usable. cargo-llvm-cov exposes the same switch. With `all`, the merge fails only when no profile at all is valid.

## What changes

Two flags in `.github/workflows/engine-weekly.yml`, with the reason beside them, and one diagnostic step:

```text
cargo llvm-cov --workspace --cobertura --failure-mode all ...
cargo llvm-cov report --summary-only --failure-mode all
ls -l target/llvm-cov-target/*.profraw   (if: always(), after the run)
```

Both flags, because `report` merges the raw profiles again (cargo-llvm-cov 0.9.1 `report::generate` calls `merge_profraw` on every path). The listing is what the review said would pin the producer: the count and the sizes of the truncated files, which no log has carried so far.

## What this does not do

- It does not recover coverage from the interrupted children. Their data is lost in either mode.
- It does not hide a real merge failure: if every profile were unreadable, `all` still fails, and `llvm-profdata` still prints one warning per bad file.
- It does not gate on the number of bad files. A run where most profiles were truncated would go green with a lower number and warnings in the log; a threshold guessed today would be another claim to defend. The listing makes that case visible to a reader.
- It cannot be verified locally. The proof is the next weekly run, or a manual dispatch of `engine-weekly` after merge. I will dispatch one and report the Coverage job's conclusion on #1952.

Which test truncates the profiles. The `fulfill_e2e.rs` candidate fits every observed fact and is still an inference. The listing step exists so the next run answers it.

## What I think I am missing

Whether a truncated header can also come from the runner losing disk mid-write. The 2026-09-14 log shows 24G free after the job, so not this time, and the machine-state steps already added for #1638 would show it.


hugocorreia90:
## Review disposition

**Reviewer:** a fresh-context same-model pass (Claude, no prior context, read-only through git objects, upstream cargo-llvm-cov and compiler-rt sources fetched for the mechanism). Not the independent reviewer `AGENTS.md` asks for: the Codex plugin cannot start on this machine today. Disclosed rather than skipped.

**Verdict as returned: BLOCK on the comment text.** The two flags were confirmed correct and safe; my causal claim was wrong. Fixed in the second commit, and the PR description is rewritten to match.

|---|---|---|---|
| 1 | The symptom is exactly as described in both logs read (2026-09-12: one corrupt file; 2026-09-14: two; `no profile can be merged`; every test green) | Confirmed | Kept |
| 2 | **The SIGKILL drill cannot be the producer.** cargo-llvm-cov sets `LLVM_PROFILE_FILE` with `%m` (merge mode), and in merge mode the profile runtime writes nothing at startup; the file is written by the `atexit` hook only. The drill's child is SIGKILLed inside a 300 s sleep, so it never reaches exit and leaves no file. A corrupt header is a file with a valid magic and a truncated body: a writer interrupted DURING its exit-time write. The drill also runs one child per run, so it cannot make two files | Confirmed: compiler-rt `truncateCurrentFile` returns early when a merge pool is set; the drill child sleeps at `commit.rs:4866` | **Fixed.** The comment now says what the evidence supports: a child signalled while writing its profile at exit, producer not pinned from the logs. The most consistent candidate, named as a candidate: the `rocky mcp` servers `fulfill_e2e.rs` spawns, which the fulfill driver SIGTERMs right after each replay, and `rocky mcp` installs no SIGTERM handler |
| 3 | `--failure-mode` is accepted on both invocations in 0.9.1 (`ReportOptions`, shared parser, `merge_profraw` appends it, `report` takes the same path) | Confirmed independently | Kept |
| 4 | `all` still fails when every profile is invalid and still prints one warning per bad file, so a real merge failure stays loud. What goes quiet: a run where most profiles are truncated turns green with a silently lower number | Fair | **Addressed in part.** A diagnostic step now lists every `.profraw` with its size after the run, `if: always()`, so the next reader has the count and the sizes the reviewer said would pin the producer. No gate on the count: a threshold guessed today would be another claim to defend |
| 5 | "One file per killed child" was a promise the evidence does not support | Fair | Removed from the comment |
| 6 | `actionlint` passes on the file | Confirmed | Kept |

What this PR now claims: the merge tolerates a few truncated profiles, which is what has been failing the job; it does not claim to know which test truncates them. Proof remains the next weekly run.

Issue #1676, opened 2026-09-05
engine-ci runs no POC against the PR binary; the only smoke installs the latest release


hugocorreia90:
No CI job runs a shipped POC against the binary built from the PR. A first-run failure in a POC reaches `main` unseen.

## What runs today

```
engine-ci.yml   cargo test / clippy / fmt        no POC at all
sdk-ci.yml:54   "Install the rocky binary (latest engine release)"
sdk-ci.yml:58   bash engine/install.sh           installs the RELEASE, not the PR build
```

So the only smoke that executes `examples/playground` uses the previous release's engine. A change to the engine can break every POC and stay green.

## How it was found

#1666: `column_match` fails on the first run of `01-quality/02-inline-checks` and passes on the second. Engine 1.73.0 shipped it. Nothing in CI ran that POC on a fresh DuckDB file with the candidate binary.

## Proposed shape

One job in `engine-ci.yml`, after the debug build:

1. build `rocky` once (`cargo build -p rocky-cli`, no release profile)
2. for each POC under `examples/playground/pocs/`, run it **twice** on a fresh DuckDB file with that binary
3. fail on a non-zero exit on either run

Twice, because #1666 is a first-run-only defect and a re-run-only defect is just as possible. A fresh file per POC, because a shared file hides both.

Cost: one debug build is already cached by the test job; the POCs are DuckDB and take seconds.

Refs #1666.

hugocorreia90:
The gap is real, but two things in this issue are wrong, and building the proposed shape as written would produce a gate that misses the defect it cites. Verified against `main` 8c12b2ff.

## 1. A CI job does build the binary from source and run every POC

`engine-weekly.yml:157-211` — `poc-smoke`:

```yaml
- name: Build rocky binary
  working-directory: engine
  run: cargo build --release
- name: Add rocky binary to PATH
  run: echo "$GITHUB_WORKSPACE/engine/target/release" >> "$GITHUB_PATH"
...
- run: bash examples/playground/scripts/run-all-duckdb.sh
```

So "the only smoke installs the latest release" is `sdk-ci.yml` only. The real gap is narrower and worth stating exactly:

```
engine-weekly  source-built binary, all 98 POCs, weekly, on main        <- after the fact
sdk-ci         released binary, one quickstart, every PR
engine-ci      no POC at all,                    every PR               <- the gap
```

[...]

PR #1972, opened 2026-09-15
fix(playground): the POC harness proves which rocky it ran, and a failed drive says why

### LINKED ISSUE #1951: The POC smoke harness reports a binary under test that most POCs do not use, and prints the empty artifact on failure
The POC smoke harness makes three confident statements about its own run, and each one can be wrong. All three were hit in a single verification session tonight, and two of them produced a wrong conclusion that was acted on before being caught.

## 1. "Binary under test" names a binary most POCs do not use

`run-all-duckdb.sh` resolves and announces one binary:

```bash
line 24  ROCKY_BIN="${ROCKY_BIN:-$(command -v rocky || true)}"
line 31  echo "Binary under test: $ROCKY_BIN"
```

The data flow is **one-way**: PATH can *default* `ROCKY_BIN`, but `ROCKY_BIN` can never redirect PATH. Only the ~10 POCs that resolve a path themselves consult `ROCKY_BIN`. Every other POC calls bare `rocky`:

```
pocs/01-quality/06-quality-pipeline-standalone/run.sh:12   rocky validate
pocs/01-quality/06-quality-pipeline-standalone/run.sh:15   rocky -c rocky.toml -o json run ...
pocs/03-ai/08-fulfillment-walking-skeleton/run.sh:51       rj() { ... rocky --output json ... }
```

The script's own comment knows the scope — "the override **those POCs** consult first" (line 18) — and line 31 drops the qualifier.

**Observed.** With `ROCKY_BIN` set to a freshly built `1.74.0` and PATH still holding a `1.73.0` install, the header printed the 1.74.0 path while the POCs ran 1.73.0:

```
Binary under test: <worktree>/engine/target/release/rocky
expected/run_quality.json   "version": "1.73.0"
expected/02_drive.json      "version": "1.73.0"
```

**Both error directions occurred in the same run:**

- A POC verifying a **new refusal** passed, because the binary predated the change and refused nothing. A pass carrying no information, reported as a pass.
- A replay POC failed, purely because a 1.73.0 binary replayed a 1.74.0 checkout's `replay/session.json`: `call 'draft_check' expected ToolNotFound but got … missing field 'model'`. Two people then chased a clock-relative seed (#1526) and a missing `[fulfill.driver]`, neither of which was the cause.

Setting **both** `PATH=<wt>/engine/target/release:$PATH` and `ROCKY_BIN=<wt>/engine/target/release/rocky` gives `85 passed, 0 failed, 16 skipped` with `"version": "1.74.0"` in both artifacts.

## 2. A failure line prints the file that is empty

`pocs/03-ai/08-fulfillment-walking-skeleton/run.sh:177`:

```bash
[ "$code" = "0" ] || fail "2 (drive-to-review exit $code; $(cat expected/02_drive.err))"
```

Observed output: `FAIL: 2 (drive-to-review exit 2; )` — nothing after the semicolon.

`02_drive.err` was **0 bytes**. The engine wrote no stderr because the diagnosis was structured, one file over:

```json
"state": "blocked",
"message": "blocked: drafting failed: replay session error: call 'draft_check' expected ToolNotFound but got … (attempts exhausted)"
```

So the harness printed an empty string where the entire explanation existed in `02_drive.json`. An empty `.err` reads as "no detail available" when it means "the detail is elsewhere".

## 3. The profile coincidence — already filed, still open

#1676 records the third member: ten POCs prefer `engine/target/release/rocky` over PATH, and the weekly only agrees with itself because it happens to build release. That issue is **closed with no PR** (`closedByPullRequestsReferences: []`) and the job it proposed does not exist — `run-all-duckdb.sh` appears once in the whole workflow tree, at `engine-weekly.yml:264`.

## The shape

```
#1676         reports on a binary chosen by profile coincidence
ROCKY_BIN     reports a "binary under test" most POCs do not use
fail "..."    prints the artifact that is empty, not the one with the answer
```

Each is the harness being confidently wrong about its own run. A status line is prose emitted by code — the same standing as a doc comment — but it is harder to distrust because it reads as an observation.

## Suggested fixes

Smallest first; none needs the others.

1. **Make `ROCKY_BIN` authoritative by prepending its directory to PATH** in `run-all-duckdb.sh`, so bare-`rocky` callers and path-resolvers agree by construction rather than by the caller remembering two variables.
2. **Assert the binary instead of announcing it.** Print `$($ROCKY_BIN --version)` and, once any POC has produced output, compare a `"version"` field against it. A mismatch should fail the suite loudly — that is the check that would have caught all of tonight.
3. **In the `fail` interpolation, fall back to the JSON** when the `.err` file is empty, or print both. A POC whose engine reports structurally should not report an empty reason.

## Not asking for

Not asking to change which POCs call bare `rocky`. Rewriting 100 POCs' invocations is a larger change than making the harness honest about what it ran.

## How it surfaced

Verifying a release candidate's POC fix on a locally built release binary, at the request of the release lane. The first result was reported as a confirmation and withdrawn once the artifact was read. Related: #1676 (closed, gap live), #1526 (a different failure in the same POC — ruled out), #1922.

hugocorreia90:
The POC harness now proves which binary it ran instead of announcing one, and the walking-skeleton POC's failure line prints the engine's reason instead of an empty file.

Closes #1951.

## What was wrong

`run-all-duckdb.sh` resolved and printed one binary, `ROCKY_BIN`, but only the ten path-resolving POCs consult it. The other ninety call bare `rocky`, which PATH decides. The flow was one-way: PATH could default `ROCKY_BIN`, `ROCKY_BIN` never redirected PATH. A caller who set it to a fresh build while PATH held an install read a header naming the build and ran most of the suite on the install. Both error directions were observed on the 1.74 cut: a POC verifying a new refusal passed on the old binary, and a replay POC failed for a version mismatch that two people then chased elsewhere.

Separately, `08-fulfillment-walking-skeleton` reported a failed drive as `FAIL: 2 (drive-to-review exit 2; )`. The `.err` file it printed was empty because the engine had written the explanation into the JSON one file over.

## The fix, in three parts

**1. One binary, by construction.** The harness makes `ROCKY_BIN` absolute, creates a shim directory holding a symlink named `rocky` to it, and puts that first on PATH. Bare `rocky` and `$ROCKY_BIN` are then the same file whatever the binary is called. Setting only one of the two variables is now enough.

**2. The header is earned.** Before the loop, both spellings must report the same `--version`, or the harness exits 1 naming both paths. After every POC that passes, any file under its `expected/` written during that run whose top-level `version` is a release string different from the binary's fails the POC:

```text
    FAIL — ran a different rocky than /path/to/rocky (rocky 1.74.0):
      expected/run_quality.json reports rocky 1.73.0
```

Only release-shaped strings count (`*.*.*`), so a plan file's `version: 1` is ignored; only files newer than a marker taken before the run count, so a stale artifact from an earlier run on disk cannot fail a POC that never wrote it.

**Also, found while testing this:** the "last 10 lines" a failing POC prints came from `/tmp/poc-output.log`, which is only written once the `cd` into the POC succeeds. When that `cd` fails, the harness printed the previous POC's tail as this one's. The log is now emptied before each run. Same defect class as the issue: output presented as an observation of a run it did not come from.

**3. The failure line says why.** The walking-skeleton POC gets a `why` helper: stderr when there is any, else the JSON's `message` (then `state`), else a note saying which file to read. All thirteen `$(cat expected/NAME.err)` sites use it; the `FAIL: <n>` shape the mutation-pass hook keys on is unchanged.

## Evidence

- Full harness run against a dev build of this tree, `ROCKY_BIN` set to `engine/target/debug/rocky` with a 1.74.0 install still first on the original PATH: `85 passed, 0 failed, 16 skipped`, the count the issue reports for a correctly wired run. The walking-skeleton POC is among the 85, so the thirteen rewritten failure lines did not change its passing path.
- Version assertion fires: a two-POC copy of the harness run with a `rocky` that answers `--version` with `9.9.9` and delegates everything else to the real binary fails both POCs and names every artifact:

  ```text
  === RUN pocs/00-foundations/00-playground-default
      FAIL — ran a different rocky than .../fake_rocky.sh (rocky 9.9.9):
        expected/test.json reports rocky 1.74.0
        expected/lineage.json reports rocky 1.74.0
        expected/run.json reports rocky 1.74.0
        expected/profile.json reports rocky 1.74.0
        expected/compile.json reports rocky 1.74.0
  === RUN pocs/01-quality/06-quality-pipeline-standalone
      FAIL — ran a different rocky than .../fake_rocky.sh (rocky 9.9.9):
        expected/run_quality.json reports rocky 1.74.0
  Summary: 0 passed, 2 failed, 0 skipped
  ```
- Unit checks on the two helpers (scratch script, eight cases): `version_mismatches` reports only a newer, release-shaped, different version and nothing for a matching version, a schema `version: 1`, an array, a non-JSON file, a stale file, or a POC without `expected/`; `why` prefers stderr, falls back to `message`, then `state`, then names the file, and treats a missing `.err` as empty.
- `shellcheck --severity=error` (what `scripts-ci.yml` runs) is clean on both scripts; the warning count is unchanged from `main` (one pre-existing SC2164 in the harness, two in the POC).

## Not changed

- Which POCs call bare `rocky`. The issue asked for the harness to be honest about what it ran, not for a hundred invocations to be rewritten.
- The weekly job's PATH step (`engine/target/release` first) still works: with `ROCKY_BIN` unset the harness resolves that binary and the shim points at the same file.
- #1676's proposal of a PR-time POC smoke job. Still open there.

That every artifact carrying an engine version lands in `<poc>/expected/`. The convention says so and every example in the issue does, but a POC writing its JSON elsewhere is simply not checked, and the check does not claim otherwise.

## What I think I am missing

A POC whose `expected/*.json` has a top-level `version` that is a release string but not the engine's, for example a product spec that records the Rocky version that composed a plan. Those are engine versions too, from the same binary, so they match; a POC that deliberately writes a foreign version would be the one false positive, and none does today.


hugocorreia90:
## Review disposition

**Reviewer:** a fresh-context same-model pass (Claude, no prior context, read-only through git objects, 63 tool calls, with shell probes of the helpers outside the repo). Not the independent reviewer `AGENTS.md` asks for: the Codex plugin cannot start on this machine today. Disclosed rather than skipped.

**Verdict as returned: ALLOW.** One medium and three low findings, all hardening; the first three are applied in the second commit.

|---|---|---|---|
| M1 | The version check fails open without `jq`: every read is empty, nothing is reported, the POC passes with the header still claiming the binary was proved | Confirmed (`jq ... 2>/dev/null \|\| true`) | **Fixed.** The harness now refuses to start without `jq`, next to the binary check |
| L1 | `mktemp -d` and `ln -s` are unchecked under `set -u` without `-e`; a failed `mktemp` puts an empty entry first on PATH, a failed `ln` leaves two same-version builds indistinguishable | Confirmed | **Fixed.** Both exit 1 with a message |
| L2 | A relative `ROCKY_BIN` is resolved after the `cd` to the playground, so `ROCKY_BIN=engine/target/debug/rocky` from the repo root is refused with a misleading "not an executable file"; a bare name (`ROCKY_BIN=rocky`) likewise | Confirmed; it was silently wrong for the path-resolving POCs before this PR | **Fixed.** A bare name resolves through PATH; a relative path resolves against the caller's directory, captured before the `cd` |
| L3 | Two theoretical false-positive shapes (an object-valued `version`, NDJSON in a `.json` file), neither emitted by anything in the tree | Confirmed unreachable today | Not changed. Noted here so a future reader has it |

Also confirmed by the review, and worth keeping on the record:

- `--version` cannot differ between the two spellings: clap prints the crate's `name`, not `argv[0]`, so a `ROCKY_BIN` called `rocky-debug` still answers `rocky 1.74.0` and the shim compares like with like.
- No POC derives sibling paths from `command -v rocky`, tests the basename of `ROCKY_BIN`, or resets PATH; the process-adapter discovery keys on the `rocky-` prefix, so the shim's `rocky` is never mistaken for an adapter.
- The weekly job (`ROCKY_BIN` unset, `engine/target/release` on PATH) takes the same path as before: the shim points at the release binary and the version assertion compares it with itself.
- What the version check does not see: ten POCs merge stderr into their JSON capture (unparseable, skipped), the walking skeleton writes some engine JSON to scratch directories, and thirty POCs never ask for JSON. The reviewer's framing is the right one: the shim is the fix, the version assertion is the smoke test of the shim, and a clean assertion is not proof of identity between two builds with the same version string.
- No CI job shellchecks `examples/playground/**`; `scripts-ci.yml` sweeps `scripts/**` only. The `--severity=error` result is local.

Gate on the second commit: unit checks for the helpers pass (nine cases, including the new refusals); `shellcheck --severity=error` clean on both files; full harness against the dev binary re-run after the change, result in the next comment.

hugocorreia90:
Full harness re-run on the second commit (7e4b5ad5), `ROCKY_BIN` set to the dev binary with a 1.74.0 install still first on the original PATH:

```text
Summary: 85 passed, 0 failed, 16 skipped
```

Same count as the first run. The four resolution scenarios the review named were run against a one-POC copy of the script: a relative `ROCKY_BIN` from the repo root and a bare `rocky` on PATH both resolve and pass; a directory and an unknown bare name are refused up front with a message naming the value.

Issue #1973, opened 2026-09-15
rocky mcp introduces itself as "rmcp <library version>", so every library bump re-blesses the served-text golden


hugocorreia90:
`rocky mcp` introduces itself to every client as the MCP library, not as Rocky.

## What a client sees

The `initialize` result's `serverInfo`, on every profile:

```json
"serverInfo": {"name": "rmcp", "version": "3.3.0"}
```

`RockyMcpServer::get_info` builds it with `Implementation::from_build_env()` (`engine/crates/rocky-mcp/src/tools.rs`, the `get_info` impl). That constructor reads `CARGO_PKG_NAME` and `CARGO_PKG_VERSION` where it is compiled, which is inside the `rmcp` crate, so the name is `rmcp` and the version is the library's, not Rocky's.

## Why it matters

- A client that lists its servers shows "rmcp 3.3.0" for Rocky.
- The served-text golden (`served_text_golden_pins_every_worded_surface`) pins the whole `initialize` payload, so every library bump re-blesses the golden for a reason that has nothing to do with Rocky's served text. That is exactly what happened on the 3.1.4 to 3.3.0 bump: the only change in the two `initialize` surfaces was this version string.
- Nothing in the payload tells a client which Rocky it is talking to.

## Fix

Build the `Implementation` by hand with Rocky's own name and `CARGO_PKG_VERSION` from `rocky-mcp` (or from the `rocky` binary crate, whichever the product wants to advertise), then re-bless the golden once. After that the golden moves only when Rocky's served text or version moves.

Found while adapting rocky-mcp to rmcp 3.3.0 for #1965.

Refs #1965.

PR #1974, opened 2026-09-15
chore(engine/rocky-mcp): rmcp 3.1.4 to 3.3.0, with the negotiation rule it brings

hugocorreia90:
rmcp moves from 3.1.4 to 3.3.0 with the two rocky-mcp adaptations the bump needs, so dependabot's cargo group can land without it.

Supersedes the `rmcp` half of #1965. Refs #1965, #1973.

## What rmcp 3.2 changed, and why the tests broke

rmcp 3.2.0 follows the 2026-07-28 versioning spec: that revision replaced the `initialize` handshake with per-request metadata, so any client that sends `initialize` is a legacy client whatever version it names, and one naming `2026-07-28` is answered with the server's fallback. A peer reaches `2026-07-28` only through the new `server/discover` lifecycle.

```text
rmcp 3.1   initialize {protocolVersion: 2026-07-28}  ->  2026-07-28  (echoed)
rmcp 3.3   initialize {protocolVersion: 2026-07-28}  ->  server fallback
           server/discover, preferred [2026-07-28]  ->  2026-07-28
```

Two rocky-mcp tests encoded the 3.1 rule:

- `result_type_reaches_a_2026_07_28_client_and_no_other` asserted that a client naming `2026-07-28` over `initialize` is given it. Its modern peer now connects over the discover lifecycle (`ClientLifecycleMode::Discover`), which is the only route left, and the assertion pair the test exists for holds: that peer keeps `resultType`, the default `initialize` peer has it stripped, and both receive the same tool result.
- `served_text_golden_pins_every_worded_surface` moved on both `initialize` surfaces. I printed the payloads before re-blessing. The only change is `serverInfo`: `{"name":"rmcp","version":"3.1.4"}` became `{"name":"rmcp","version":"3.3.0"}`. The negotiated version (`2025-11-25`), the capabilities and the instructions are byte-identical. That `serverInfo` names the library rather than Rocky is a pre-existing defect, filed as #1973; it is what makes every library bump re-bless this golden.

## One behaviour change, deliberate

`get_info` supplied `V_2024_11_05` as the server's fallback. Under 3.1 that value only mattered for a client naming a version the server did not support at all. Under 3.3 it is the answer to every `initialize` that names `2026-07-28` or later, so keeping it would send a client that asked for more back to the oldest handshake version. The fallback is now `V_2025_11_25`, the newest version with a handshake. A client naming any of the four handshake versions is still answered with what it asked for.

The alternative is one line: keep `V_2024_11_05`. I chose the newest because the pin's original reason (the belief that it strips `resultType`, corrected in the fifteenth-round note above `get_info`) never held, and nothing else depends on it.

The negotiation paragraph on the served-text sweep doc is rewritten for the 3.2 rule; the sixteenth-round note now says the modern peer discovers.

## Evidence

- `cargo test -p rocky-mcp`: 164 passed, 0 failed, 1 ignored (the live Snowflake grounding test), across the lib, `roundtrip` (89) and `wire_parity` suites.
- Clippy `--all-targets --all-features -D warnings` and `cargo fmt --check` on rocky-mcp: clean.
- The golden was re-blessed only after reading both changed surfaces (above).
- The bump is `rmcp` and `rmcp-macros` only (`cargo update -p rmcp --precise 3.3.0`); the other eight crates in dependabot's group are untouched here so dependabot can regroup them.

## Not changed

- `supported_protocol_versions` is still not overridden: the server advertises rmcp's whole list, `2026-07-28` included, and a discovering peer gets it.
- `Implementation::from_build_env()` in `get_info` (#1973) stays as it is; fixing it re-blesses the golden again and deserves its own change.

Whether any production client sends `initialize` naming `2026-07-28` today. If one does, it now negotiates `2025-11-25` where it used to negotiate `2026-07-28` and keep `resultType`. That is the spec's answer and rmcp's, not mine, but it is the one observable change for a real client.

## What I think I am missing

A rocky-mcp code path that reads the negotiated version itself. I found none: the `resultType` stripping is rmcp's, keyed on the peer it negotiated, and rocky-mcp only supplies the fallback. The full-suite run above is what stands behind that.


hugocorreia90:
## Review disposition

**Reviewer:** a fresh-context same-model pass (Claude, no prior context, read-only through git objects; it fetched rmcp 3.1.4 and 3.3.0 sources and verified the scratch copies by SHA-256). Not the independent reviewer `AGENTS.md` asks for: the Codex plugin cannot start on this machine today. Disclosed rather than skipped.

**Verdict as returned: ALLOW**, with one test gap and four doc corrections. All addressed in the second commit.

|---|---|---|---|
| 1 | The behaviour the changelog announces (an `initialize` naming `2026-07-28` is answered with `2025-11-25`) had no test; reverting the fallback to `V_2024_11_05` passed the whole suite | Confirmed by reading the three protocol-version sites | **Fixed.** `an_initialize_that_names_2026_07_28_is_answered_with_the_newest_handshake_version` connects the old way (`initialize` naming `2026-07-28`) and asserts the negotiated version is `2025-11-25` and `resultType` is withheld. Reverting the fallback fails it |
| 2 | `engine/Cargo.toml`'s comment above the pin still told the 3.1-era story ("under 3.2.0 a client that asks for 2026-07-28 is handed back 2024-11-05 … raise it after the negotiation change is understood"), which this commit did the opposite of | Confirmed | **Fixed.** The paragraph now says what `=3.3.0` pins and why |
| 3 | Two places in `roundtrip.rs` said `get_info` "pins 2024-11-05" | Confirmed | **Fixed** |
| 4 | "only through the discover lifecycle" and "withheld from every `initialize` peer" overstate: a request that declares `2026-07-28` and client capabilities in its own `_meta` is handled at that version inside an `initialize` session; rmcp's own client never does this, a hand-rolled one can. Same code in 3.1.4 | Confirmed against `service.rs` and `handler/server.rs` | **Fixed.** Both sentences now name the per-request route as the third way, and say rmcp's client does not use it |
| 5 | The test's opening doc still stated the 3.1 rule in the present tense; "rmcp 3.1.2's" in two comments; `Auto` also reaches `2026-07-28` | Confirmed | **Fixed** |

Also from the review, kept as evidence: the golden change is proven, not reported. The reviewer rebuilt the default-profile `initialize` payload from `SKILL.md` and hashed it: the old golden is that payload with `"version":"3.1.4"`, the new one with `"3.3.0"`, `protocolVersion` `2025-11-25` in both. The fallback pin is redundant today (rmcp's `ServerInfo::new` defaults to `LATEST` = `2025-11-25`, and a non-legacy fallback would resolve to the newest legacy version anyway) and stays right if `LATEST` moves. Nothing else in rocky-mcp or the `rocky` binary's dev-dependency uses an API the bump changed; the features rocky compiles do not include request-state, streamable HTTP, SSE or auth.

Gate on the second commit: `cargo test -p rocky-mcp` 165 passed, 0 failed, 1 ignored (the new test among them); clippy `--all-targets --all-features -D warnings` clean; `cargo fmt --check` clean. Mutation check: with the fallback put back to `V_2024_11_05`, the new test fails and every other test still passes, which is exactly the gap the review named.

PR #1975, opened 2026-09-15
fix(engine): a non-empty table with no timestamp value fails its freshness check as not evaluated

### LINKED ISSUE #1930: A non-empty table with an all-NULL timestamp column is never freshness-checked

hugocorreia90:
A non-empty table whose timestamp column holds no value now fails its freshness check as not evaluated, with the row count in the reason. It used to emit no check at all, on every adapter.

Closes #1930.

## What was wrong

Both freshness queries asked for `MAX(<timestamp_column>)` and nothing else. `MAX` is NULL over no rows and over rows that carry no value, so the runner read every NULL as "empty, nothing to be fresh" and emitted nothing. The issue's comment ran it on DuckDB: `COUNT(*) = 2, MAX(ts) = NULL`.

```text
                          before             after
  empty table             no check           no check          (unchanged)
  rows, ts all NULL       no check           freshness_not_evaluated, Error
  rows, a ts value        freshness check    freshness check   (unchanged)
```

## The fix

`COUNT(*)` rides in the same aggregate as `MAX`, so the two describe the same rows at the cost of one scan:

- **Per table** (`per_table_freshness`, every adapter without a batch leg): `SELECT COUNT(*), MAX(<ts>) FROM <t>`. A count that does not read as a number is reported the way an unreadable timestamp already is.
- **Batched** (Databricks): `SELECT 'c' AS c, 's' AS s, 't' AS t, COUNT(*) AS n, CAST(MAX(<ts>) AS STRING) AS max_ts FROM …`, with the parser reading the count at index 3 and the maximum at index 4.
- **The result type** (`rocky_core::traits::FreshnessResult`) gains `row_count: Option<u64>`, `#[serde(default)]`. `None` is an adapter that did not count; that is its own state, not a zero, so a batch adapter written before this change keeps the old reading and does not start failing healthy empty tables.
- **The consumer** turns `(max: None, rows: Some(n > 0))` into `freshness_not_evaluated("the table has N row(s) and no value in its timestamp column, so there is no timestamp to measure freshness from")`. `(None, Some(0))` and `(None, None)` still emit no check.

Snowflake and BigQuery do not batch freshness (`supports_freshness` is false), so they take the per-table path and get the fix through it. The adapter SDK's own `FreshnessResult` is a separate type used by its conformance runner and is not bridged into this path; it is unchanged.

## Evidence

- Six tests in `rocky-cli`, all through the real check-running path with a DuckDB target:
  - `a_non_empty_table_with_no_timestamp_value_has_its_freshness_not_evaluated`: real DuckDB, real query, the table emptied of VALUES rather than rows (`UPDATE … SET ts = NULL`); asserts the reason and the Error severity.
  - `a_batched_freshness_leg_that_answers_null_over_rows_is_not_evaluated`: the batch path, `row_count: Some(3)`.
  - `a_batched_freshness_leg_that_does_not_count_keeps_the_empty_reading`: the absent count is not a zero.
  - `a_freshness_count_that_is_not_a_number_is_not_evaluated`: an unreadable count is reported, not guessed.
  - The two existing empty-table pins still pass, one now with `row_count: Some(0)`.
- **Mutation check**: disabling the new consumer arm fails exactly the two tests that claim it, `a_batched_freshness_leg_that_answers_null_over_rows_is_not_evaluated` and `a_non_empty_table_with_no_timestamp_value_has_its_freshness_not_evaluated` (16 passed, 2 failed in the freshness set); everything else, the empty-table pins included, still passes, so the arm changes nothing but the case it is for.
- Databricks batch SQL shape pinned byte for byte in `test_batch_freshness_single`.
- **Live on Databricks** (Unity Catalog, SQL Statement API, scratch schema created and dropped), the exact generated batch SQL over three tables and the per-table shape:

  ```text
  batched   t_empty  n="0"  max_ts=null
            t_null   n="2"  max_ts=null
            t_val    n="2"  max_ts="2026-09-01 00:00:00"
  per-table t_null   ["2", null]      t_empty  ["0", null]
  ```

  Every cell arrives as a string over that API. Both count reads go through `rocky_core::checks::cell_as_u64`, the reader the row-count leg and the null-rate sampler already use (a JSON integer, a numeric string, or an integral float; anything else is unreadable).
- Gate: `cargo test -p rocky-cli --lib --features duckdb -- freshness` 18 passed; `cargo test -p rocky-databricks --lib batch` 14 passed; `cargo clippy -p rocky-cli -p rocky-core -p rocky-databricks --all-targets --all-features -- -D warnings` clean; `cargo fmt --check` clean. CI runs the full suites.

## Not verified live

Snowflake (account suspended: even a `VALUES`-backed aggregate returns `000666`) and BigQuery (sandbox expired). Both take the per-table path, whose SQL is `SELECT COUNT(*), MAX(<col>) FROM <t>` with the table reference formatted by the adapter's own dialect; nothing in it is dialect-specific, and the same two-column read is what `query_target_max_timestamp` has always done for the watermark on those adapters.

## Docs

`concepts/data-quality-checks.md` said the all-NULL case "emits no freshness check at all" and pointed at this issue. It now describes the check that is emitted.

The reason text. It names no column because the consumer does not have the column name in scope; the check's `details` carry the threshold and the table key carries the table, so a reader can find the column in the pipeline config. Threading the name through is possible and I did not think a longer message was worth the plumbing.

## What I think I am missing

An adapter that answers the per-table query with the two columns swapped. Every adapter Rocky ships formats the query from the same string, so the order is fixed; a count that is not a non-negative integer (a `2.5`, a negative, a boolean) is refused as unreadable rather than misread, which is the safe direction.


hugocorreia90:
## Review disposition

**Reviewer:** a fresh-context same-model pass (Claude, no prior context, read-only through git objects, 47 tool calls), reviewing the first commit (38ce4857). Not the independent reviewer `AGENTS.md` asks for: the Codex plugin cannot start on this machine today. Disclosed rather than skipped.

**Verdict as returned: BLOCK**, on a test I did not run. Every finding is addressed in the follow-up commits.

|---|---|---|---|
| 1 | `rocky-databricks/tests/wiremock_tests.rs` still mocked the old four-column freshness row; under the new reader every row is omitted and the test fails. It runs in CI (`--all-features`), and I had only run `--lib batch`. It was also the only test driving the READER with a real row, so the index shift was pinned by nothing but the generator's string | Confirmed by reading the mock against the parser | **Fixed.** The mock now carries the five-column shape with every cell a string, as the SQL Statement API returns them; two rows added: `rows_no_value` (`"3"`, NULL) must be RETURNED with `row_count: Some(3)`, `badcount` (`"many"`, NULL) must be OMITTED. The Databricks suite was run with `--all-features` this time |
| 2 | Two hand-rolled count parsers diverged from `rocky_core::checks::cell_as_u64`, which the row-count leg uses on the same adapters and which also accepts an integral float (#1923) | Confirmed; the fix was already pushed (16b46939) before this review landed, for the same reason | **Fixed.** Both reads go through `cell_as_u64`; the private parser is gone. The reviewer also caught that my deleted doc comment had DuckDB backwards (it stringifies integers); the shared reader's doc is the one that stands now |
| 3 | Three stale "one `SELECT MAX(...)` per table" sentences: `supports_freshness` docs in rocky-core and the adapter SDK, and the SQL shown on the sql-generation concept page. The SDK's mirrored `FreshnessResult` lacks `row_count` and the crate's divergence list did not say so | Confirmed | **Fixed.** All three now show `COUNT(*), MAX(...)`; the SDK header lists the field as a known divergence with the reason (nothing bridges SDK batch results into the core runner yet) |
| 4 | The `(_, None)` arm said "returned no rows" for a one-cell row too | Confirmed, message only | **Fixed.** A row without its second cell now says so |
| 5 | `(Some(ts), Some(0))` is measured; a non-NULL `MAX` over zero rows cannot come from SQL | Agreed | No change |

Also from the review, kept as evidence: every construction site sets `row_count` to what it knows (the two test doubles that pass `Some(1)` do so beside a measured timestamp, where the count is never read); no path emits two freshness results for one table; the intercept prefix matches only the freshness query; nothing in `schemas/`, the SDK or the Dagster fixtures serializes this type; DuckDB's `COUNT(*)` arrives as the string `"0"` and is pinned on the real path by the empty-table test. One wording the reviewer flagged as overstated is corrected in the changelog: an adapter written against the old struct cannot compile against the new one, so the "does not start failing" promise was about a batch adapter that answers NULL without a count, and it now says that.

Gate on the follow-up commits: `cargo test -p rocky-databricks --all-features` 305 passed, 0 failed (the wiremock suite included, and the rewritten test among them); `cargo test -p rocky-cli --lib --features duckdb -- freshness` 18 passed; clippy `--all-targets --all-features -D warnings` on rocky-cli, rocky-core, rocky-databricks and rocky-adapter-sdk clean; `cargo fmt --check` clean.

PR #1988, opened 2026-09-15
fix(sdk): the engine floor is 1.35.0, the first engine with a replication plan_id

### LINKED ISSUE #1984: fix(sdk): MIN_ROCKY_VERSION is 1.34.0, but the dagster Pipes path needs engine 1.35.0
The SDK and the Dagster adapter accept engine 1.34.x, but the Pipes path needs 1.35.0.

`dagster-rocky`'s Pipes path needs a `plan_id` from `rocky plan`. A replication-only project first got a content-addressed `plan_id` in engine 1.35.0. See `engine/CHANGELOG.md`, `## [1.35.0]`: "Replication-only projects now get a real content-addressed `plan_id`". The 1.34.0 section has no such entry.

The floor in code says 1.34:

- `sdk/python/src/rocky_sdk/client.py:114-118`: the comment calls the floor "a 1.34+ guarantee", and the constant is `MIN_ROCKY_VERSION = "1.34.0"`.
- `integrations/dagster/src/dagster_rocky/resource.py:1232`: the `dg.Failure` hint says "requires engine-v1.34+".

The effect: engine 1.34.x passes the version check, then fails at run time with "rocky plan did not emit a plan_id". The hint then tells the user to install a version they already have.

The same wrong floor is in the public docs (`dagster/pipes.md`, `dagster/resource.md`, `dagster/observability.md`) and in `integrations/dagster/CHANGELOG.md`. A docs PR will fix the docs pages. This issue covers the code.

Found by the docs audit.

hugocorreia90:
`MIN_ROCKY_VERSION` moves from `1.34.0` to `1.35.0`, the engine that first gave replication-only projects a content-addressed `plan_id`, and the Dagster "no plan_id" hint names that version.

Closes #1984.

## What was wrong

The SDK's floor (shared with `dagster-rocky`) accepted engine 1.34.x, but the adapter's plan/apply path needs a `plan_id` from `rocky plan` for every project shape, and `engine/CHANGELOG.md` `## [1.35.0]` is where "Replication-only projects now get a real content-addressed `plan_id`" shipped. So a 1.34.x binary passed the version check and then failed inside the run with "rocky plan did not emit a plan_id", and the hint told the user to install `engine-v1.34+`, which they already had.

```text
before   1.34.x  passes _verify_engine_version  ->  fails later: no plan_id, hint says 1.34+
after    1.34.x  RockyVersionError up front, min_version 1.35.0
```

## What changed

- `sdk/python/src/rocky_sdk/client.py`: `MIN_ROCKY_VERSION = "1.35.0"`, and the comment says why the floor is where it is.
- `integrations/dagster/src/dagster_rocky/resource.py`: the `dg.Failure` hint says `engine-v1.35+`. `dagster-rocky` re-exports the SDK constant, so its floor moves with it.
- Tests: the two pins of `min_version` and the "two-component version at the floor" case move to 1.35; a new test, `test_verify_version_refuses_the_last_engine_without_a_replication_plan_id`, feeds `rocky 1.34.9` and asserts the refusal. Reverting the constant to `1.34.0` fails exactly that test.
- Both Python changelogs gain an `Unreleased` entry. The dagster one says when the refusal actually arrives: with the `rocky-sdk` release that carries the constant. Until this package's `rocky-sdk>=` floor is raised to that release, an install resolving `rocky-sdk` 0.15.0 keeps the old floor and gets only the hint change; the pyproject ledger records that ordering (release the SDK first, then raise the floor), per the repo rule.
- `sdk/python/README.md` (the wheel's PyPI description), the SDK quickstart example and the `plan()` docstring example (which still taught a `plan_id is None` guard for replication-only projects) say 1.35.0 or drop the dead guard.
- The dagster test for the "no plan_id" failure now pins the hint text `engine-v1.35+`.

## Evidence

- `uv run pytest tests/test_client.py` (sdk): 74 passed. `uv run pytest -k "version or plan_id"` (dagster): 24 passed. `ruff check` and `ruff format --check` clean on both packages.
- Mutation: constant reverted to `1.34.0`, the new test fails; restored, it passes.

## Not changed

- The historical `integrations/dagster/CHANGELOG.md` 1.33.0 entry (the companion to engine 1.35.0) is where the wrong floor was first written; it records what was believed at the time and is not rewritten.
- The public docs pages that repeat the old floor (`dagster/pipes.md`, `dagster/resource.md`, `dagster/observability.md`, and `python-sdk/introduction.md` lines 48 and 139, found by review) are in the docs-lane's PR, per the issue.

Whether any SDK user still runs engine 1.34.x. Rocky is at 1.74 and the project keeps no compatibility before 2.0, so the floor is stated rather than negotiated; a 1.34.x binary was already failing the dagster path, only later and with a worse message.

## What I think I am missing

A second feature in the SDK itself (not the adapter) that needs more than 1.35. The two known ones, per-pipeline DAG resolution (1.68) and pipeline-scoped execution (1.69), state their own floors through `_require_engine`, which the test at `test_per_pipeline_mode_refuses_an_engine_that_predates_it` pins; that mechanism is unchanged. Nothing in `rocky_sdk` itself needs 1.35; the floor moves for the adapter that shares it, which was already the stated design.


hugocorreia90:
## Review disposition

**Reviewer:** a fresh-context same-model pass (Claude, no prior context, read-only through git objects; it ran the Python suites from a scratch archive and the mutation itself). Not the independent reviewer `AGENTS.md` asks for: the Codex plugin cannot start on this machine today. Disclosed rather than skipped.

**Verdict as returned: BLOCK.** The code and tests were right; four other surfaces still stated the old floor. All addressed in the second commit.

|---|---|---|---|
| A1 | `sdk/python/README.md` still said v1.34.0, and it ships as the wheel's PyPI description | Confirmed | **Fixed** |
| A2 | `sdk/python/examples/quickstart.py` and `docs/python-sdk/introduction.md` (lines 48, 139) still said 1.34.0 | Confirmed | **Fixed** in the quickstart; the docs page is handed to the docs-lane's PR with the three dagster pages, so two PRs do not touch `docs/` |
| A3 | The dagster changelog claimed "refused up front", but the wheel's `rocky-sdk>=0.15.0` floor resolves an SDK whose constant is still 1.34.0; nothing enforced "needs rocky-sdk with the same floor" | Confirmed | **Fixed.** The entry says the refusal arrives with the SDK release that carries the constant and that only the hint changes until the floor is raised; the pyproject ledger records the release order (SDK first, then raise the floor), which is the repo rule |
| A4 | PR body named `_require_engine_at_least` (it is `_require_engine`) and the wrong dagster changelog entries (the old floor is in the 1.33.0 entry) | Confirmed | **Fixed** in the body |
| A5 | The `plan()` docstring example still taught a `plan_id is None` guard for replication-only projects, contradicting its own first paragraph | Confirmed | **Fixed** |
| A6 | No test pinned the hint text `engine-v1.35+` | Confirmed | **Fixed.** `test_run_pipes_raises_when_engine_returns_null_plan_id` asserts it |

Confirmed by the review and kept as evidence: the 1.35.0 claim is true against `engine/CHANGELOG.md` (the 1.34.0 section has no plan_id item); the per-feature floor mechanism (`_require_engine`) is unchanged and consistent; the new test fails on the reverted constant for the right reason (`(1,34,9) < (1,34,0)` is false) alongside the two `min_version` pins; the dagster tests derive from the constant and track the bump.

Gate on the second commit: `uv run pytest tests/test_client.py` 74 passed; dagster `tests/test_resource.py -k "plan_id or version"` 22 passed; `ruff check` and `ruff format --check` clean on both packages (examples included).

Issue #1991, opened 2026-09-15
fix(dagster): the scaffold writes rocky run --config rocky.toml, which exits 2


hugocorreia90:
`dagster-rocky`'s scaffold writes a command that fails into every project it generates.

- `integrations/dagster/src/dagster_rocky/scaffold.py:71`, in the generated `rocky.toml` comment: `# Run with: rocky run --config rocky.toml`
- `integrations/dagster/src/dagster_rocky/scaffold.py:117`, in the generated README: `rocky run --config rocky.toml`

`--config` is a top-level flag with no `global = true` (`engine/rocky/src/main.rs:83`). It must come before the subcommand:

```
$ rocky run --config rocky.toml
error: unexpected argument '--config' found
$ rocky --config rocky.toml run      # works
```

Fix: emit `rocky --config rocky.toml run`, or just `rocky run`, since `rocky.toml` is the default.

The docs page `docs/src/content/docs/dagster/scaffold.md` has the same fix in a docs PR. Found by the docs audit.

hugocorreia90:
## Wider than filed: the whole generated `rocky.toml` is the pre-Phase-2 shape and fails `rocky validate`

The command line is one symptom. I extracted `ROCKY_TOML_TEMPLATE` from `dagster_rocky.scaffold` and ran the engine on it (dev build of `main`, 1.74.0 line):

```text
$ rocky --config rocky.toml --output json validate
exit 1
V001  Failed to parse config: failed to parse TOML: unknown field `checks`,
      expected one of `state`, `adapter`, `adapters`, `pipeline`, `pipelines`,
      `hook`, `hooks`, `cost`, ...
```

The template writes top-level `[source]`, `[warehouse]`, `[target]`, `[replication]` and `[checks]` tables. Those are the pre-Phase-2 layout; the engine today reads `[adapter]` plus `[pipeline.<name>]` with `source`, `target` and `checks` nested under the pipeline (the rocky-config skill lists `[source]` / `[warehouse]` explicitly as "no longer work"). So a project produced by `dagster-rocky`'s scaffold cannot be validated, planned or run at all, whichever way the command is spelled. `rocky --config rocky.toml run` also exits 1 on it, on the same parse error.

`test_init_writes_canonical_template_contents` compares the written file to the constant, so it cannot see this; no test parses the template with the engine.

[...]

Issue #1983, opened 2026-09-15
rocky playground: the ecommerce and showcase templates scaffold a project where rocky run builds nothing


hugocorreia90:
`rocky playground --template ecommerce` and `--template showcase` create a project where `rocky run` builds nothing. It exits 0 and prints `Copied 0 tables`. The default `quickstart` template works first try; these two do not.

Seen on `rocky 1.74.0`. The template code on `main` has not changed since that release.

## Steps

```bash
rocky playground demo --template ecommerce
cd demo
rocky compile    # Compiled: 10 models, 0 errors, 0 warnings
rocky test       # Result: 10 passed, 0 failed
rocky run        # Copied 0 tables in 0.0s   <- nothing built, exit 0
```

`showcase` gives the same result with 11 models.

## Why

The two templates write a config that cannot build the models:

```
quickstart (works)                    ecommerce / showcase (broken)
------------------                    -----------------------------
type = "transformation"               type = "replication"
[adapter] path = "playground.duckdb"  [adapter.local] no path -> in-memory DB
seed loaded at scaffold time          seed written to data/seed.sql, never loaded
```

- A `replication` pipeline copies tables that match the schema pattern (`prefix = "raw__"`). It never builds the `models/` directory.
- The seed creates tables in schema `source`, so the `raw__` pattern matches nothing either.
- The adapter has no `path`, so no database file exists between commands.
- `seed_quickstart_db` runs only for `Template::Quickstart` (`engine/crates/rocky-cli/src/commands/playground.rs`).
- The seed comment says `duckdb playground.duckdb < data/seed.sql`, but `rocky.toml` does not point at that file.

`rocky test` passes because it loads `data/seed.sql` into its own in-memory database.

## Smaller mismatches in the same place

- All three templates print `rocky preview rows --model customer_orders` as a next step. Only `quickstart` has a `customer_orders` model.
- The CLI labels `showcase` as "Showcase (every Rocky feature)". It is `ecommerce` plus one `.rocky` model and one extra contract. Its `rocky.toml` header still says "Rocky E-Commerce Playground" and the pipeline is named `ecommerce`. The docs page describes it more accurately: "ecommerce + Rocky DSL + extra contracts".

## Expected

`rocky compile && rocky test && rocky run` works first try for every template, as it does for `quickstart` since #741 and #762. After `rocky run`, the models exist as tables in a DuckDB file, so `rocky serve --ui`, `rocky preview` and `rocky history` have data to show.

[...]

PR #1992, opened 2026-09-15
docs: fix examples that fail when copied (config, CLI flags, version pins, llms.txt)

hugocorreia90:
Several docs examples fail when a reader copies them. This PR fixes each one and checks the fix against the 1.74.0 binary or the code on `main`.

## What was broken

| Where | What a reader hit | Fix |
|---|---|---|
| `reference/configuration.md` Full Example, `reference/adapters/fivetran.md` | `rocky validate` fails with `V032` and `V033`: a Fivetran adapter needs `kind = "discovery"` and cannot be `source.adapter` | Add `kind`; the warehouse reads the rows and Fivetran goes in `source.discovery`. Add the missing `kind` row to the adapter table. |
| `configuration.md` quarantine | The shown block does nothing (`enabled` defaults to `false`), and the mode table names a `__dqx_valid` column that does not exist | Document `enabled`, `suffix_valid` and `suffix_quarantine`, and the real `split` / `tag` / `drop` output. Say it runs only in `quality` pipelines. |
| `reference/commands/modeling.md` | `--output markdown` exits 2; `preview cost --base` does not exist | Use `--output json \| jq -r .markdown`. Replace `--base` with the real `--models` flag. |
| 7 pages (execution-flow, state-management, observability, migrate-from-dbt, scaffold, json-output, administration diagram) | `rocky run -c rocky.toml` and `rocky run --state-namespace x` exit 2: `--config`, `--state-path` and `--state-namespace` must come before the subcommand | Move the flag before the subcommand. `cli.md` now says which global flags go where, and lists `--state-namespace` and `--principal`. |
| `dagster/installation.md`, `dagster/introduction.md` | `rocky-sdk >= 0.6.0`; the real floor is `0.15.0` | Update. Also `pydantic >= 2.10` on the SDK page. |
| `dagster/pipes.md`, `resource.md`, `observability.md` | "`run_pipes` needs engine 1.34": replication-only plan ids shipped in 1.35.0 | Say 1.35.0 for replication-only projects. Refs #1984, which covers the same floor in SDK code. |
| `docs/scripts/llms.txt.tmpl` | Listed five commands that do not exist (`dag-run`, `drift`, `column-lineage`, `compact-dedup`, `model-history`), "21 crates", the wrong `--state-path` default | Rewrite against `rocky --help`, the strategy enum, the adapter capability table and the current page list. |
| `guides/playground.mdx` | The `ecommerce` and `showcase` templates build nothing | Add a caution pointing at #1983. #1983 stays open. |
| `guides/preview-a-pr.md` | Example pin `engine-v1.17.4` predates `rocky preview` | Use `1.74.0`. |

Also: `concepts/architecture.mdx` lists all 27 crates and `engine/ui`, and says the `/api/v1` routes live in `rocky-cli`. The `rocky serve` reference now describes what the server answers, and says `GET /api/v1/health` skips the `--ui` Host and Origin checks. `guides/adapter-sdk.md` no longer names `rocky drift`. Three example files moved `-c` before the subcommand.

## How it was checked

- The new Full Example and the Fivetran snippet pass `rocky --config <file> validate` (1.74.0).
- Each moved global-flag form parses; the old form exits 2 with `unexpected argument`.
- Quarantine defaults and modes: `rocky-core/src/config.rs` (`QuarantineConfig`, `QuarantineMode`) and `quarantine.rs`. The only caller is `run_quality`.
- Health exemption: `rocky-server/src/auth.rs` (`AUTH_EXEMPT_PATHS`), and a probe against `rocky serve --ui`.
- `npm run build` in `docs/` passes, and the new anchors exist in the built HTML.
- An independent review pass (a different model, fresh context) checked each hunk against the code. It found one wrong row (the Iceberg adapter description in `llms.txt`), two conditions the first draft dropped (`--principal` gating, quarantine on non-quality pipelines), and missed siblings. All are fixed in the last commit.

Not in this PR: `integrations/dagster/src/dagster_rocky/scaffold.py` still writes `rocky run --config rocky.toml` into generated READMEs. That is code, tracked in #1991.


hugocorreia90:
## Independent review

Reviewed by Codex (GPT-5.6-sol), adversarial review of the branch against `main`. Three findings, all confirmed and fixed in `732bf76a`.

| Finding | Verdict | Fix |
|---|---|---|
| The rewritten quarantine section lists `not_in_future` and `older_than_n_days` as quarantined kinds without saying `split` refuses them | Confirmed. `quarantine.rs` calls `reject_generated_clock_in_split` for both, so under the default `split` the run records `quarantine:compile` and writes neither table | The list now marks those two as `tag` and `drop` only, and says `split` refuses a clock-dependent `expression` or `filter` for the same reason |
| The page still said the SDK floor is 1.34 | Confirmed, and it moved under this branch: `main` now has `MIN_ROCKY_VERSION = "1.35.0"` | `dagster/pipes.md` and `python-sdk/introduction.md` say 1.35.0. No `1.34` claim is left in either directory |
| `rocky serve` was described as answering with "the same payloads as the CLI's `--output json`" | Confirmed as an overclaim. `GET /api/v1/models` returns `{models, count}` with graph fields, while `rocky list models --output json` returns a different shape. Six routes have no CLI counterpart | The reference and `llms.txt` now say the payloads are typed and schema-backed, that a route matching a CLI command returns what that command prints, and that the rest have their own shapes |

Codex verified the rest against the binary: the Fivetran example validates with placeholder credentials, `V032` and `V033` fire as negative controls, the adapter-`kind` rules match the capability table, flag placement, the preview flags, every `llms.txt` verb and strategy label, and all 27 local index links.

One more change rode along, from `main` moving under the branch: #1998 taught the dagster scaffold to write a `.gitignore`, so `dagster/scaffold.md`'s file tree lists it (`ef33df33`).

Issue #1993, opened 2026-09-15
serve --ui: a non-loopback IPv6 bind prints an address the Host guard then refuses (bracket mismatch)


hugocorreia90:
`rocky serve --ui --host fd00::1` (any IPv6 literal other than `::1`) prints and, with `--open`, opens `http://[fd00::1]:8080/ui/#token=…`. The browser sends `Host: [fd00::1]:8080`. The `--ui` host guard compares that against the bind host and refuses it with `421`.

Where it splits (`engine/crates/rocky-server/src/ui.rs`): `host_allowed` special-cases `[::1]` only (`:56`); `host_without_port` keeps the brackets for the header value (`:108-111`) but leaves a bare bind literal untouched (`:114-119`), so `[fd00::1]` is compared with `fd00::1` and mismatches (`:59`).

So the one address the server advertises for a non-loopback IPv6 bind is one the same server refuses. `::1` works because of the special case; IPv4 and names are unaffected.

Fix in the guard: compare with the brackets stripped on both sides, and add a non-loopback IPv6 literal to `ui.rs`'s tests and to `ui_address_is_the_printed_shape_with_a_reachable_host` in `rocky-cli/src/commands/serve.rs`. The printing side is right as it stands: a URL requires the brackets.

Found by the review of F4 (`serve --ui --open`), which made the printed address the opened address and so made the mismatch visible.

PR #1995, opened 2026-09-15
feat(engine/rocky-cli): serve --ui --open opens the page once the listener is bound, and just ui-dev runs the UI loop in one command

hugocorreia90:
`rocky serve --ui --open` opens the page once the listener is bound, and `just ui-dev [dir]` is the UI development loop in one command. F4 of the first-run series.

Also here, one docs commit: #1907's "Ask the server what it is doing" (`GET /api/v1/settings`) section landed inside `## rocky playground` in `reference/commands/development.md`, between the playground's Arguments and Flags tables; it is moved verbatim under `## rocky serve`, where it belongs.

## `--open`

```
  print    Rocky UI: http://localhost:8080/ui/#token=…      at once, as before
  open     the SAME string, handed to the system opener      only after api::serve raises
           (`open` / `xdg-open` / `cmd /C start`)            the readiness latch: sweep done,
                                                             listener bound
  refuse   --open without --ui                               before anything binds, naming the fix
```

- **One address, one function.** `ui_address(host, port, token)` produces the printed line and the string the opener receives, so what opens is what printed by construction. It also fixes two shapes the old print got wrong: a wildcard bind (`0.0.0.0`, `::`) shows `localhost`, the one name a wildcard promises a browser on the same machine; an IPv6 literal (`::1`) is bracketed as a URL requires.
- **After the bind, never before.** The plan named the race: the address printed at `serve.rs` before `api.rs` bound the listener. The opener runs on its own task and waits on the `server_ready` latch `api::serve` raises after the startup sweep and the bind. A bind that fails returns without raising it; `shutdown` ends the task.
- **A missing or failing opener is a warning.** The server still starts and the address is still on stdout. The opener is spawned and reaped on a thread, never awaited: some `xdg-open` handlers stay in the foreground until the browser exits.
- The address, token included, is passed to the opener as an argument — the same secret the terminal already shows. Said in the flag's help text.

## `just ui-dev [dir]`

Builds this checkout's `rocky`, starts `rocky serve` in `dir` (default `examples/playground`) with a read-only token — **no `--ui`**, the page comes from Vite — waits for `/api/v1/health`, then starts Vite on 5173 with `--strictPort`, and prints one address to open. Bash job control (`set -m`) gives each child its own process group, so a Ctrl-C reaches the recipe alone and the trap stops both groups deliberately, `npm`'s node children included. If either process dies the loop notices and stops the other.

**The proxy port has one source.** `vite.config.ts` read `http://127.0.0.1:8080` as a literal; it now reads `ROCKY_API` (default unchanged), and the recipe sets it from its own port, so the two cannot disagree. `ROCKY_UI_DEV_PORT` and `ROCKY_UI_DEV_TOKEN` override the port and token.

## Evidence

Acceptance rows from `F4-dev-loop.md`:

| # | Row | Evidence |
|---|---|---|
| 1 | `just ui-dev <dir>` starts both and prints one address | run three times on this machine (default POC): Vite answers `/ui/` in 12–16 s; `/api/v1/health` is `200` direct on 8080 **and** through Vite's proxy on 5173; `/api/v1/runs` through the proxy is `401` (the proxy reaches rocky's auth); one line printed: `ui-dev: API on http://127.0.0.1:8080 — open http://localhost:5173/ui/#token=dev` |
| 2 | waits for the server, forwards Ctrl-C to both, leaves no child on exit or on either process failing | killed three ways, each time with both processes alive beforehand (2 matched) and **0 processes, 0 listeners on 8080 and 5173 afterwards**: SIGINT to the recipe's process group → exit 130; `rocky` killed → "rocky serve exited; stopping Vite", exit 1; Vite killed → exit 143 (its status propagated). The first version continued past the trap into a `wait` on a reaped pid (exit 127); the INT/TERM traps now exit |
| 3 | proxy port and recipe port from one source, or a comment naming the other | `ROCKY_API` — read in `vite.config.ts`, set in the recipe; each site names the other |
| 4 | `--open` hands the opener a URL equal to the printed address, token fragment included | `the_opener_receives_the_printed_address_only_after_ready`: a recording opener receives exactly `ui_address(...)`, nothing before the latch |
| 5 | no opener, or a failing one: the server still starts and prints | `a_failing_opener_is_not_fatal`; the print happens before the opener task exists |
| 6 | `--open` without `--ui` refused at start, naming the fix | `open_without_ui_is_refused_and_names_the_fix`, and the built binary: `rocky serve --open --token t --token-scope read-only` → `rocky serve --open needs --ui: …` |

Also `a_shutdown_before_ready_opens_nothing` (a failed bind opens no browser), `ui_address_is_the_printed_shape_with_a_reachable_host`, and — through the real binary with a real page — `open_hands_the_opener_the_printed_address_and_survives_an_opener_that_fails` in `engine/rocky/tests/serve_ui.rs`: a fake `open`/`xdg-open` first on `PATH` records its one argument; it equals the printed address exactly; with the fake exiting 3 the server still prints and still answers `/api/v1/health`. It skips, like the page test, when `engine/ui/dist` is absent (CI's `Test` job), and ran here with the dist built.

**The pty run the review asked for** (a Python `pty.fork` driving `just ui-dev` as a person would): Vite answers after 15.5 s; **Enter** pressed — both `rocky serve` and the Vite `node` stay `S` (running), `/api/v1/health` still `200` direct and through Vite; **Ctrl-C** pressed — the recipe exits 130, and afterwards no `rocky serve`/`vite` process exists. On the first version the same keystrokes stopped Vite (`T`) and hung the recipe.

`cargo test -p rocky-cli --lib -- commands::serve::` 28 passed. `cargo test --features ui --test serve_ui` 3 passed with `engine/ui/dist` built (the new opener test ran, not skipped). `cargo fmt --check` clean. `cargo clippy -p rocky-cli -p rocky --all-targets --all-features -- -D warnings` clean. `npm run typecheck` and `npm run lint` clean in `engine/ui`.

Not covered: the opener ordering (after the bind) is pinned in-process on the latch, not through the binary — a black-box test of "not before" would flake; `SystemOpener` against a real browser is not exercised anywhere (it would open one).

## Review

Reviewer: a fresh-context same-model pass (Fable 5.1, max effort, 43 tool calls, two pty probes on this machine's bash 3.2). **Not independent** (the Codex plugin is down). Verdict on the first version: **MERGE-WITH-FIXES**, 16 findings. Disposition:

| # | Finding | Disposition |
|---|---|---|
| 9 | **Reproduced:** under `set -m` both jobs inherit the terminal; Vite reads stdin for its shortcuts; one Enter stops it with SIGTTIN, the watch loop's `kill -0` never notices, and Ctrl-C then hangs on the stopped job | **ADOPTED, mandatory.** `</dev/null` on both jobs; `kill -CONT` after `TERM` in `stop`. Re-proved under a pty: Enter, then Ctrl-C — see the acceptance row |
| 6 | "a failing opener is a warning" was true only for a MISSING opener; a spawned one exiting 3 was silently dropped | **ADOPTED.** The reaper thread warns on a non-success status (status only, never the URL); wording changed everywhere to "a missing opener, or one that exits non-zero" |
| 15 | printed == opened pinned only by construction; plan row 4 asked for an integration test | **ADOPTED.** The `serve_ui.rs` test above |
| 5 | Windows `cmd /C start` re-parses the line; a token with `&`, `%`, `^` cuts the URL and runs the rest | **ADOPTED.** `rundll32 url.dll,FileProtocolHandler <url>` — a plain argument, no shell |
| 2 | unbiased `select!`: Ctrl-C in the same instant as the bind may still open | **ADOPTED.** `biased;`, shutdown first |
| 3 | the token is in the opener's argv (process list) | **ADOPTED** in wording: help text, docs and changelog say so |
| 12 | port 8080 already held: the health curl passes against the other server before the liveness check | **ADOPTED.** `kill -0` before `curl`; the messages name the port |
| 10 | the INT trap is not the Ctrl-C path (the foreground `sleep` takes it, `set -e` ends the script, EXIT runs `stop`) | **ADOPTED.** Comment rewritten to the real path; the traps stay for a signal sent to the script |
| 11 | `CARGO_TARGET_DIR` set → missing binary | **ADOPTED.** `${CARGO_TARGET_DIR:-$root/engine/target}` |
| 7 | the `--ui` Host guard brackets only `[::1]`; any other IPv6 bind advertises an address the same server refuses with 421 | **FILED #1993.** Pre-existing in the guard; the printed URL is right |
| 1, 4, 8, 13 | readiness after the bind, no early fire; token never logged; `host:port` typo and old print; vite config | CONFIRMED sound |
| 14 | docs claims | two false ones fixed (6, 9) |

My answers: least confident was the Ctrl-C hang with `npm` (not `node`) as the job leader — settled by the pty run below; most likely missing was exactly finding 9, which no headless run could see.


hugocorreia90:
**Correction to the review disclosure above.** This PR says the Codex plugin is down. That is wrong, and the accurate statement is narrower: the plugin's **task** dispatch path (`codex:codex-rescue`) fails on this machine with `The 'gpt-6-astra' model requires a newer version of Codex`, at CLI 0.146.0 and still at 0.154.0 (npm latest). The **review** path (`/codex:review`, user-invoked by design) works — another session ran adversarial reviews successfully in the same window.

So the reviewer for this PR was a fresh-context same-model pass, **not** independent, and the reason is that the independent path is not reachable from an agent dispatch, not that no reviewer exists. The same correction applies to #1999 and #2006.

hugocorreia90:
**Correcting my own correction.** The comment above says the Codex plugin's task dispatch path fails at any CLI version. That is also wrong, and more confidently wrong than the claim it replaced.

Counting every job record on this machine: `task` completed **251**, failed 31; `adversarial-review` completed 15. Two `task` jobs completed on the night of the 15th–16th, between my own failures. So `The 'gpt-6-astra' model requires a newer version of Codex` is a **transient 400**, not a version incompatibility — and the right response to it is a retry, not a conclusion.

What stands: the reviewer for this PR was a fresh-context **same-model** pass, not independent. What was wrong both times: my account of why. I had four failed dispatches and never counted anyone else's successes.

Issue #1997, opened 2026-09-15
bug(engine): rocky plan fails with no models found when models/ exists but is empty


hugocorreia90:
`rocky plan` on a replication-only project whose `models/` directory exists but holds no model exits 1 with `no models found in models`. `rocky validate` and `rocky run` on the same project succeed.

## Reproduce (engine 1.74.0 line, `main` at `494e081b`, DuckDB, no credentials)

```bash
mkdir plan-empty && cd plan-empty && mkdir models
cat > rocky.toml <<'EOF'
[adapter]
type = "duckdb"
path = "warehouse.duckdb"

[pipeline.main]
strategy = "incremental"
timestamp_column = "_loaded_at"

[pipeline.main.source.discovery]
adapter = "default"

[pipeline.main.source.schema_pattern]
prefix = "src__"
separator = "__"
components = ["source"]

[pipeline.main.target]
catalog_template = "warehouse"
schema_template = "raw__{source}"

[pipeline.main.target.governance]
auto_create_schemas = true
EOF
duckdb warehouse.duckdb -c "CREATE SCHEMA src__shop; CREATE TABLE src__shop.orders AS SELECT i AS id, now() AS _loaded_at FROM range(10) t(i);"
rocky --config rocky.toml validate          # exit 0, valid
rocky --config rocky.toml --output json plan # exit 1
rocky --config rocky.toml --output json run  # exit 0, tables_copied 1
```

```text
Error: failed to compute governance action preview
Caused by:
    0: failed to compile project for governance preview
    1: no models found in models
```

Delete the `models/` directory and `plan` exits 0. Add one trivial model and `plan` exits 0.

## Why

- `engine/crates/rocky-cli/src/commands/plan.rs:337-345`: when `models_dir.exists()`, plan calls `populate_governance_actions`, which compiles the whole `models/` directory for the governance preview (`plan.rs:1698-1716`).
- `engine/crates/rocky-compiler/src/project.rs:141`: the compiler treats a directory with no model as the error `no models found in {path}`.
- Nothing between the two distinguishes "no models yet" from "a broken project". `rocky run` does not compile for a replication pipeline, so it passes.

## Who hits it

- Every project scaffolded by `dagster-rocky`'s `init_rocky_project()`: it writes `models/.gitkeep` on purpose (#1991), so `rocky plan` is the first command that fails on a fresh scaffold.
- Any replication-only project that keeps an empty `models/` directory in git.

## Fix

[...]

PR #1998, opened 2026-09-15
fix(dagster): the scaffold writes a rocky.toml the engine accepts and puts --config before the verb

### LINKED ISSUE #1991: fix(dagster): the scaffold writes rocky run --config rocky.toml, which exits 2

hugocorreia90:
Closes #1991.

## What was wrong

A project scaffolded by `dagster_rocky.scaffold.init_rocky_project()` could not be validated or run.

```text
$ rocky --config rocky.toml validate        # on the old template
exit 1
V001  Failed to parse config: failed to parse TOML: unknown field `checks`,
      expected one of `state`, `adapter`, `adapters`, `pipeline`, `pipelines`, ...
```

- The template wrote top-level `[source]`, `[warehouse]`, `[target]`, `[replication]` and `[checks]` tables, the layout the engine stopped reading when pipelines became `[adapter]` plus `[pipeline.<name>]`. The issue as filed was the command line; the config was the larger half.
- The config header and the README said `rocky run --config rocky.toml`. `--config` is a global flag, so that spelling exits 2. `rocky --config rocky.toml run` is the working order. Two more copies of the wrong order were outside the scaffold: the architecture diagram in `AGENTS.md` and `engine/examples/README.md`.
- The README's install one-liner pointed at `releases/latest/download/install.sh`, which redirects to the engine release and returns 404 (the installer is not a release asset). Its Fivetran link pointed at `engine/crates/rocky-fivetran/README.md`, which does not exist.

No test could see any of this: `test_init_writes_canonical_template_contents` compares the written file to the constant, and the dagster suite runs without the binary.

## What changes

```text
integrations/dagster/src/dagster_rocky/scaffold.py
  ROCKY_TOML_TEMPLATE   [adapter] duckdb at warehouse.duckdb
                        [pipeline.main] full_refresh
                          source.discovery -> default, schema_pattern src__<source>
                          target catalog "warehouse" (= the file stem, said in a comment),
                          schema raw__{source}, governance.auto_create_schemas = true
                          checks row_count, column_match
                        incremental + freshness kept as a commented opt-in that
                        names the `_loaded_at` column both need on every table
                        header: validate / run with --config first
  GITIGNORE_TEMPLATE    warehouse.duckdb, .rocky/, .rocky-state.redb*  (new file)
  README_TEMPLATE       install from engine/install.sh on main; validate, then run;
                        Fivetran link -> rocky-data.dev/reference/adapters/fivetran/
integrations/dagster/tests/test_scaffold.py
  + RockyConfig.model_validate(tomllib.loads(ROCKY_TOML_TEMPLATE))  (the SDK's generated
    config model: codegen-drift-gated against the engine structs, rejects unknown and
    missing keys, needs no binary)
  + pins for the keys the starter relies on, the flag order in both templates,
    the commented opt-in, and the .gitignore patterns
AGENTS.md, engine/examples/README.md   --config before the verb
integrations/dagster/CHANGELOG.md      [Unreleased] / Fixed
```

Why `full_refresh` and not the old `incremental`: the incremental strategy and the freshness check both require a load-timestamp column on every `src__*` table, a contract the scaffold cannot make for a new user's data, and the engine records incremental watermarks to whole seconds, so any `now()`-stamped column re-copies rows on run 2 (#2004). The starter now works on any table; the opt-in is documented in place.

`auto_create_schemas = true` is not decoration: without it the first `rocky run` fails because `raw__<source>` does not exist yet.

## Evidence

Scaffolded with the branch's `init_rocky_project()` into a scratch directory, seeded one `src__shop.orders` table with no timestamp column, and ran the dev engine (1.74.0 line, `main` at `494e081b`):

```text
validate   exit 0   "valid": true (two L-lints about defaults the file does not set, #2005)
run 1      exit 0   Success, tables_copied 1, column_match and row_count passed, 10 rows
run 2      exit 0   same; still 10 rows
git add -A --dry-run after the runs: warehouse.duckdb, .rocky/, models/.rocky-state.redb* not staged
old template, same engine:  validate exit 1, V001 unknown field `checks`
rocky run --config rocky.toml   exit 2
rocky --config rocky.toml run   exit 0
curl -sIL: releases/latest/download/install.sh 404; raw .../main/engine/install.sh 200;
           engine/crates/rocky-fivetran/README.md 404; rocky-data.dev/reference/adapters/fivetran/ 200
```

Without `[pipeline.main.target.governance]`: run exit 1, tables_copied 0. With `catalog_template = "other"`: validate passes, run fails `Catalog with name other does not exist`, hence the comment.

Tests: `uv run pytest tests/test_scaffold.py` 12 passed; `ruff check` and `ruff format --check` clean. Mutations run in memory against the SDK model: dropping `catalog_template` or `prefix`, a typo in the governance or checks key, and the old top-level `[checks]` are all rejected; `row_count = "yes"` is not (the toggle is an untagged union), which the hand pin covers.

## Known limitations, filed separately

- `rocky plan` on a fresh scaffold exits 1 with `no models found in models`: plan's governance preview compiles the empty `models/` directory that the scaffold creates on purpose (#1997). The README teaches `validate` and `run`, which work.
- Relative paths in `rocky.toml` resolve against the process cwd, not the config's directory, and the SDK's subprocess sets no cwd. The template says so in a comment; the behaviour is pre-existing.
- The docs page `dagster/scaffold.md` repeats the old command line and is handled in #1992 by the docs lane; this PR does not touch `docs/`.

## Review

Fresh-context same-model review, verdict BLOCK on the first commit, all findings verified and dispositioned in the comment below. Not the independent reviewer `AGENTS.md` asks for. I dispatched a Codex review three times tonight and all three were refused at the model, three seconds in, so nothing ran. **Correcting myself:** earlier PRs of mine said the Codex plugin "cannot start on this machine today". That was wrong, and I checked it rather than repeat it: the plugin's own job records show the agent-dispatch path completing far more often than it fails, and the user-triggered review path completed for another lane tonight. The honest claim is about my attempts, not the tool's health.


hugocorreia90:
## Review disposition

**Reviewer:** a fresh-context same-model pass (Claude, no prior context, read-only through git objects; it ran the dev engine on the old and new templates and on thirteen mutations of the new one, the Python suite, and anonymous HEAD requests on every URL). Not the independent reviewer `AGENTS.md` asks for: the Codex plugin cannot start on this machine today. Disclosed rather than skipped.

**Verdict as returned: BLOCK.** The five headline claims held; the block was for a false promise in prose I had written, a dead link in a line I had edited, and a pin weaker than one already in the tree. All addressed in the second commit.

|---|---|---|---|
| F1-F4 | Template keys exist at the right nesting and type; old template fails V001; `run --config` exits 2; old install URL 404, new one 200 | Confirmed by the reviewer with the engine and curl | Kept as evidence |
| F5 | `catalog_template = "warehouse"` only works because it equals the DuckDB file stem; a mismatch passes validate and fails run | Confirmed (`other` -> `Catalog with name other does not exist`) | **Fixed.** Comment above `[pipeline.main.target]` says the two rename together; test pins the pair |
| F6 | My new prose promised "every `src__*` schema is replicated", but `incremental` + freshness fail any table without `_loaded_at` (exit 2), a contract stated nowhere | Confirmed by the reviewer's `no_ts` probe | **Fixed by changing the starter.** `full_refresh` with `row_count` and `column_match`, no column required; incremental + freshness kept as a commented opt-in that names the column. Run twice on a timestamp-free table: exit 0 both times, counts match |
| F7 | Engine bug: an incremental watermark is recorded to whole seconds, so a sub-second `_loaded_at` re-copies the newest rows on run 2 | Reproduced (`.250` -> 2 rows and `row_count` fails on run 2; `:00` -> 1 row) | **Filed #2004.** Second reason the starter is `full_refresh`. CHANGELOG no longer says "end-to-end" |
| F8 | The Fivetran link I rewrote the sentence for still pointed at `engine/crates/rocky-fivetran/README.md`, which does not exist | Confirmed (404 on blob and raw; the crate has no README) | **Fixed.** Points at `rocky-data.dev/reference/adapters/fivetran/` (200) |
| F9 | The tomllib test pins a shape, not engine acceptance: five one-line edits pass it and fail the engine. `rocky_sdk.types_generated.rocky_project_schema.RockyConfig` rejects four of them without a binary | Confirmed in memory: drop `catalog_template`, drop `prefix`, governance typo, checks typo, old top-level `[checks]` all rejected; `row_count = "yes"` accepted (untagged toggle union) | **Fixed.** The test calls `RockyConfig.model_validate` and pins the toggles by hand. The module has shipped in `rocky-sdk` since the SDK was split out, so the published floor carries it |
| F10 | Both new tests fail for the right reasons under mutation | Confirmed | Kept |
| F11 | Two more copies of the wrong flag order outside `docs/`: `AGENTS.md` architecture diagram, `engine/examples/README.md` | Confirmed | **Fixed** in this PR (two one-line edits) |
| F12 | Moving the DuckDB file to the project root leaves it and the state store unignored; the engine gitignores `.rocky/` itself | Confirmed (`git add -A --dry-run` staged `warehouse.duckdb` and `models/.rocky-state.redb`) | **Fixed.** The scaffold writes a `.gitignore` (`warehouse.duckdb`, `.rocky/`, `.rocky-state.redb*`); dry-run after two runs stages none of them. Kept the root path: `.rocky/` does not exist before the first run, and `plan` opens the file without creating the directory |
| F13 | Relative paths resolve against the process cwd; the SDK's subprocess sets no cwd | Confirmed by the reviewer's `cwdtest` | Pre-existing. Template comment says to run from the directory; noted in the PR body |
| F14 | CHANGELOG: "plan or run" implied plan works; "run end-to-end" was shown only for the empty warehouse | Confirmed | **Fixed.** "validate or run"; "run twice on a seeded DuckDB source with matching counts" |
| F15 | `[Unreleased]` header: main has none, other open PRs add one | Confirmed | Known; whichever lands second rebases the header |
| F16 | L004 and L006 fire on keys the file does not set | Confirmed (`severity: lint`) | **Filed #2005** with F17 |
| F17 | validate accepts: empty `rocky.toml`, `{nope}` placeholder, unknown `[pipeline.main.bogus]`, DuckDB discovery adapter without `path` | Confirmed the empty file and read the reviewer's variant outputs for the rest | **Filed #2005** |
| F18 | Flag test would not catch a future `rocky validate --config` in the README | Confirmed | **Fixed.** Asserted absent |

**Weakest claim, as the reviewer put it and as I now agree:** "run end-to-end" had been shown for an empty warehouse only. The second commit's evidence is a seeded table run twice. What I had been missing: the starter carried a data contract (a timestamp column on every table, fresh within a day, and a catalog name tied to the file stem) that a fresh scaffold cannot make for a user's data. The `full_refresh` starter has no such contract.

Gate on the second commit: `uv run pytest tests/test_scaffold.py` 12 passed; `ruff check` and `ruff format --check` clean; scaffold validated and run twice on the dev engine as recorded in the body.

Issue #1980, opened 2026-09-15
A read-only open of a missing state file creates and initialises it


hugocorreia90:
`StateStore::open_read_only` on a path with no state file creates the file: `open_inner` calls `open_redb_with_retry`, which uses `Database::create` (`engine/crates/rocky-core/src/state.rs`), and the empty database is then bootstrapped through the write path — every table created, no version stamp. That is a `GET` creating a file on disk.

It is pre-existing, and it is what makes a never-run project answer "nothing yet" today. The callers that open read-only **without** checking the path first, and so rely on the bootstrap:

- CLI: `rocky state show`, `rocky doctor` (the `state` check), `rocky history`, `rocky metrics`, `rocky optimize`, `rocky branch list/show/compare`, `rocky gc`, `rocky trace`, `rocky restore`, `rocky cost`, `rocky audit`, `rocky preview`, `rocky replay`, `rocky plan --branch`, `rocky run --branch`.
- `rocky serve`: `GET /api/v1/runs`, `GET /api/v1/models/{name}/history`, `GET /api/v1/models/{name}/metrics` (the `/project`, `/jobs/{id}`, `/policy` and `/schedule/*` routes check `exists()` first).
- MCP tools over the same producers.

`peek_schema_version` and `source_schemas.rs` already refuse to open a missing path read-only "so it does not leave a fresh empty file behind" — the codebase knows the hazard at two sites and not at the others.

The fix is a typed absence: a read-only open of a missing path returns `StateError::NotFound { path }` (open via `Database::builder().open`, which has no `create`), and each surface above answers empty on `NotFound` instead of relying on a file it created. That is one variant plus about fifteen call sites and a first-run test per surface, which is why it is not part of #1978's fix (that PR keeps the bootstrap for an empty database and removes every other write from the read-only path).

Related: #1978 (a read-only open wrote whenever the stamp was not current), #1545.

PR #1999, opened 2026-09-15
fix(engine/rocky-cli): rocky serve answers 409 state_needs_migration for a store a read cannot serve

hugocorreia90:
`rocky serve` answers `409 state_needs_migration` for a store a read cannot serve. The server half of C1-P1b, first slice: the envelope. The class A/B/C routes follow in their own PRs.

## What was wrong

#1981 made `StateStore::open_read_only` refuse — never create — a store that lacks a table this binary reads (`StateError::ReadOnlyNeedsInit`; only a store written by a Rocky older than schema v22 can be in that state). `map_state_err`, the one funnel every state-backed read route maps its errors through, knew two shapes: lock contention → `503 engine_busy`, everything else → `500 internal_error`. So the refusal reached embedders as "the engine broke".

```
  before   ReadOnlyNeedsInit  →  500 internal_error       indistinguishable from a crash
  after    ReadOnlyNeedsInit  →  409 state_needs_migration  the stamp, the missing tables,
                                                            and the one command that fixes it
```

`409`, not `503`: retrying never migrates a store, so a client must not back off and try again; it must tell an operator. No `Retry-After`, no `running_job_id`.

## What changes

- `ApiError::state_needs_migration(found, expected, missing)` — message `the state store is stamped v18 and lacks tables this server reads (tombstones)`, hint `restart \`rocky serve\`, or run any read-write command such as \`rocky run\` against this project once, to migrate the store to v30; reads never migrate it`.
- One new arm in `map_state_err`. Every route that goes through it gets the envelope: `/project`, `/runs`, `/models/{name}/history`, `/models/{name}/metrics`, `/products`, `/products/{name}`, `/products/{name}/journal`, `/review/queue`, `/review/{plan_id}/status`, `/brief`, `/audit`, `/audit/scorecard`, `/custody/{subject}`, `/schedule` (via `map_schedule_err`), `/policy`, `/jobs/{id}`. The routes that map errors to `internal` directly are the compile and DAG routes, which open no state store; the schema-cache loader swallows open errors and degrades, as before.
- **The OpenAPI route table declares the `409` on those sixteen routes** (`export_openapi.rs`, `STATE_NEEDS_MIGRATION`), so a client generated from `openapi.json` sees it.
- `ErrorEnvelope`'s doc lists the code (that doc is the schema description, so `schemas/error_envelope.schema.json`, the Pydantic model, the TypeScript interface and `openapi.json` regenerate — `just codegen`, no shape change).
- `guides/embedding.md`: one paragraph beside "Reads queue, one at a time".

**When it actually happens.** Rarely, on purpose: `rocky serve` migrates an old store at startup, because its job sweep (`sweep_interrupted_jobs`) opens the store read-write before the router serves, and a read-write open creates every missing table and re-stamps. So after an upgrade the first read already answers. The `409` is for a store that changed under a running server (a file copied or restored over it) or a startup sweep that could not open the store (it is best-effort under lock contention). The first version of this PR presented the `409` as the normal post-upgrade posture and named `rocky run` as the one remedy; the review caught that, and the docs, the hint and the end-to-end test now say what is true — the test migrates the store by calling the very sweep `serve()` runs.

## Evidence

- `a_store_that_needs_migration_is_a_409_naming_the_tables` — the mapping: status, code, every missing table and the stamp in the message, the migrating command and target version in the hint, no `Retry-After`, no job id; an unversioned store reads as "unversioned".
- The same unit test pins that a `.context(...)`-wrapped `ReadOnlyNeedsInit` — the shape the read cores produce — still maps through the funnel.
- `a_read_route_answers_409_until_a_write_open_migrates_the_store` — end to end through the router: a real store stamped v18 with `tombstones` left out of a real snapshot, `GET /api/v1/runs` → `409 state_needs_migration` naming the table; then `sweep_interrupted_jobs` — the read-write open `api::serve` runs at startup — migrates it and the same read answers `200`.

`cargo test -p rocky-cli --lib -- map_state_err a_store_that_needs_migration a_read_route_answers_409 engine_busy export_openapi` 10 passed (the OpenAPI table's own pins included). `cargo fmt --check` clean. `cargo clippy -p rocky-cli --all-targets --all-features -- -D warnings` clean. `just codegen` run twice; the generated diff is the `ErrorEnvelope` description in four files plus the sixteen `409` responses in `openapi.json`.

Not covered: a project that has never run is not this case (its store is created empty on first read — #1980 tracks the typed absence); the CLI still prints the `ReadOnlyNeedsInit` message as text, which already names the command; no test starts `serve()` itself against an old store — the sweep call in the test is the same function, one frame down.

## Review

Reviewer: a fresh-context same-model pass (Fable 5.1, max effort, 84 tool calls). **Not independent** (the Codex plugin is down). Verdict on the first version: **MERGE-WITH-FIXES**, no code-behaviour defect. Disposition:

| # | Finding | Disposition |
|---|---|---|
| 3 | The `409` is almost never reachable on a real `rocky serve`: the startup sweep opens the store read-write and migrates it before the first read; the docs and hint presented it as the normal post-upgrade posture with `rocky run` as the one remedy | **ADOPTED.** Docs, changelog, `ApiError` doc and hint rewritten; the test migrates through the sweep itself |
| 5 | The OpenAPI route table hand-declares per-route responses and was not updated; a generated client got an undocumented `409` on every read | **ADOPTED.** `STATE_NEEDS_MIGRATION` on the sixteen routes; `openapi.json` regenerated |
| 6 | `embedding.md` opened with "a read never writes to the state store, and never creates a table" — an overclaim (an empty store is bootstrapped) that contradicted the changelog's own last sentence | **ADOPTED.** Reworded to "against a store that holds anything" |
| 8 | Least confident: that a `.with_context` wrapper still downcasts | **ADOPTED.** Pinned in the unit test |
| 1 | `409` is the right class here; `running_job_id: None` is consistent with the field doc; no client branches on the bare status | CONFIRMED |
| 2 | Every state-backed read goes through `map_state_err` (sixteen routes, listed); the schema-cache loader degrades by design | CONFIRMED — the list above is the reviewer's |
| 4 | Fixture sound: the snapshot registry carries `metadata`, so the v18 stamp survives; the original store is intact until the rename | CONFIRMED |
| 7 | No exhaustive match or code enumeration to update besides the route table | CONFIRMED |


hugocorreia90:
**Correction to the review disclosure above.** This PR says the Codex plugin is down. That is wrong, and the accurate statement is narrower: the plugin's **task** dispatch path (`codex:codex-rescue`) fails on this machine with `The 'gpt-6-astra' model requires a newer version of Codex`, at CLI 0.146.0 and still at 0.154.0 (npm latest). The **review** path (`/codex:review`, user-invoked by design) works — another session ran adversarial reviews successfully in the same window.

So the reviewer for this PR was a fresh-context same-model pass, **not** independent, and the reason is that the independent path is not reachable from an agent dispatch, not that no reviewer exists.

hugocorreia90:
**Correcting my own correction.** The comment above says the Codex plugin's task dispatch path fails at any CLI version. That is also wrong, and more confidently wrong than the claim it replaced.

Counting every job record on this machine: `task` completed **251**, failed 31; `adversarial-review` completed 15. Two `task` jobs completed on the night of the 15th–16th, between my own failures. So `The 'gpt-6-astra' model requires a newer version of Codex` is a **transient 400**, not a version incompatibility — and the right response to it is a retry, not a conclusion.

What stands: the reviewer for this PR was a fresh-context **same-model** pass, not independent. What was wrong both times: my account of why. I had four failed dispatches and never counted anyone else's successes.

PR #2000, opened 2026-09-15
docs: a guide to the browser UI, with screenshots, a tour GIF, and a script that rebuilds them

hugocorreia90:
`rocky serve --ui` shipped in 1.74.0, but the docs only covered it as a flag. No page said what the UI shows, and no image showed it. This PR adds a guide with screenshots and a tour GIF, puts the UI in the READMEs, and adds a script that rebuilds the images.

A tour of the Rocky browser UI

## What changes

- **New page: `guides/browser-ui.md`.** It covers how to open the UI (the read-only token in the address fragment), each lane (Estate, Review, Governor: brief, scorecard, custody, audit, products), what the page cannot do (run, approve, change policy), and the server-side guards (`421` Host, `403` Origin, CSP, health exemption).
- **Images in `docs/public/`:** `ui-estate.png`, `ui-review.png`, `ui-governor-brief.png`, `ui-governor-custody.png`, `ui-governor-product.png`, `demo-ui-tour.gif` (460 KB).
- **`README.md`:**
  - a short "In your browser" section with the GIF;
  - `engine/ui/` and `deploy/` rows in the subprojects table;
  - the container image under Releases;
  - `just build-engine-ui` for a source build with the UI;
  - `rocky serve --scheduler` named next to `rocky tick`.
- **Other pages:** `engine/README.md`, `engine/ui/README.md`, `getting-started/roadmap.md` and `reference/commands/development.md` mention or link the guide.
- **`cli-recording/record-ui-screenshots.sh` and `cli-recording/browser/screenshots.mjs`** rebuild every image from scratch against the `rocky` on `PATH`. They use three credential-free workspaces:
  - the playground quickstart, for the estate;
  - `04-governance/11-agent-policy` in a git repo, where an agent drops a PII column and policy sends the plan to review;
  - `03-ai/08-fulfillment-walking-skeleton`, for the governor tabs.

  The script runs under a neutral `$USER` and git identity, because run records and approval markers put both on screen. It fails if any route the UI calls returns a local path. When the UI shell changes, a refresh is `./record-ui-screenshots.sh --publish` plus a look at each image.

## What the guide deliberately does not say

The page is read-only in 1.74.0. The guide does not describe a settings screen, approving from the page, or any write action. None of those ship.

## How it was checked

- Each claim about a screen was read against `engine/ui/src`. Each claim about the server was read against `rocky-cli/src/api.rs`, `rocky-server/src/auth.rs` and `ui.rs`.
- The images were captured from 1.74.0. They were byte-scanned and looked at for names, usernames, emails and local paths; none found.
- The script ran end to end, twice, with byte-identical output for the review shot.
- `npm run build` in `docs/` passes, and the page renders all six images.
- An independent review pass (a different model, fresh context) checked the claims, the images and the script. It found:
  - the first review screenshot showed a stale product plan with a skipped gate;
  - one wrong sentence about `rocky apply` and spec digests;
  - incomplete lists (brief cards, custody subjects);
  - an overstated 403 rule (the webhook route checks its own signature);
  - a leak check that could pass when `curl` failed.

  All are fixed in the last commit.


hugocorreia90:
## Independent review

Reviewed by Codex (GPT-5.6-sol), adversarial review of the branch diff against `main`. Verdict: needs-attention, five findings. All five are fixed in `4115dec3`; each was reproduced first.

| Finding | Verdict | Fix |
|---|---|---|
| The guide promised that classification tags get masked samples | Confirmed. `preview_rows.rs` keeps only tags a workspace-default `[mask]` maps to a non-`None` strategy. A tag with no default mask, or one set only for another environment, reaches the `SELECT` unmasked, and the gap is only a `W004` warning | The bullet now says which columns are masked, that the rest come back raw, and that `W004` is the signal |
| The leak check in the recorder could report "no leak" for a body that leaked | Confirmed, and worse than reported: `printf \| grep -q` missed a leak in a 152 KB body 5 times out of 5 on this machine, because `grep -q` exits at the first match and `pipefail` then sees `printf` killed by SIGPIPE | The check reads a here-string, so there is no writer to kill. A `grep` exit other than 0 or 1 now fails the run. The same 152 KB body is caught 5 times out of 5 |
| The `Origin` rule was stated too strongly | Confirmed. `ui.rs` `origin_allowed` matches the origin's HOST, ignoring port, so `http://localhost:5173` passes on a loopback server | The guide states the host rule, gives that example, and says health skips the check |
| Readiness could be satisfied by a server left over from an earlier run | Confirmed: the token and ports were constants | The token is per invocation, and the probe fails if the child it launched has exited |
| The script was committed non-executable while the docs invoke it directly | Confirmed | Mode is now `100755` |

Codex's two dropped items (an HTTP-error screenshot claim, and queue weights) matched the source. Earlier passes by a same-model reviewer are on the record in the commits above; they are not independent and do not replace this one.

hugocorreia90:
One thing to do after #2014 lands, not before: recapture the screenshots.

#2014 renames the queue's ranking string in `review.rs` from `blast_radius × classification × staleness` to `blast_radius × change_class × staleness`, and the per-entry JSON field to `change_class_weight`. The UI prints that string above the review queue, so `docs/public/ui-review-queue.png` and frame 2 of `demo-ui-tour.gif` show the old middle term.

The page text here does not depend on it: the guide says "blast radius × change class × staleness" in prose, which is true before and after the rename.

To refresh, from a checkout with `rocky` 1.74.0 or newer on `PATH`:

```bash
(cd cli-recording/browser && npm ci && npx playwright install chromium)   # once
./cli-recording/record-ui-screenshots.sh --publish
```

Then look at each image before committing it.

hugocorreia90:
Update on the recapture note above: #2014 has merged, so here is the decision and why.

The screenshots stay as they are for now. They were captured from **1.74.0**, the released binary, and they show what someone running that release sees today, including the queue string `blast_radius × classification × staleness`.

#2014 renames that string to `blast_radius × change_class × staleness` on `main`, but there is no release carrying it yet. Recapturing now would need a local build of `main`, and would produce an image that matches no released binary.

The page text does not depend on it either way: the guide describes the ranking in prose as "blast radius × change class × staleness", which is true before and after the rename.

**Refresh when the next engine release ships**, with that release on `PATH`:

```bash
(cd cli-recording/browser && npm ci && npx playwright install chromium)   # once
./cli-recording/record-ui-screenshots.sh --publish
```

Then look at each image before committing. The affected files are `docs/public/ui-review-queue.png` and frame 2 of `demo-ui-tour.gif`.

PR #2007, opened 2026-09-15
fix(engine/rocky-mcp): rocky mcp introduces itself as rocky with the engine version, not as the rmcp library

### LINKED ISSUE #1973: rocky mcp introduces itself as "rmcp <library version>", so every library bump re-blesses the served-text golden

hugocorreia90:
Closes #1973.

## What was wrong

`rocky mcp` introduced itself to every client as the MCP library:

```json
"serverInfo": {"name": "rmcp", "version": "3.3.0"}
```

`get_info` built it with `Implementation::from_build_env()`, which reads `CARGO_PKG_NAME` and `CARGO_PKG_VERSION` where it is compiled, inside the `rmcp` crate. Two consequences: a client that lists its servers showed "rmcp 3.3.0" for Rocky, and the served-text golden's two `initialize` rows moved on every library bump for a reason that had nothing to do with Rocky's served text (that is exactly what happened on the 3.1.4 to 3.3.0 bump in #1974).

## What changes

```text
engine/crates/rocky-mcp/src/tools.rs
  pub const SERVER_NAME: &str = "rocky"
  get_info: .with_server_info(Implementation::new(SERVER_NAME, env!("CARGO_PKG_VERSION")))
            (rocky-mcp's version is the engine's: 1.74.0 today)
  two comments that named from_build_env() now name the hand-built value
  + server_info_names_rocky_and_the_engine_version  (all three profiles; name, version, and not "rmcp")
engine/crates/rocky-mcp/tests/fixtures/served_text.golden
  default/initialize and worker/initialize re-blessed once; no other row moved
```

`Implementation` is `#[non_exhaustive]` in rmcp, so the value goes through `Implementation::new`; `title`, `description`, `icons` and `website_url` stay `None`, which is what the handshake sweep in `tools.rs` relies on (its comment is updated).

## Evidence

```text
ROCKY_BLESS_MCP_SERVED_TEXT=1 cargo test -p rocky-mcp --test roundtrip served_text_golden
  golden diff: exactly two rows, default/initialize and worker/initialize
cargo test -p rocky-mcp          73 + 90 + 3 passed, 0 failed (new test included)
cargo clippy -p rocky-mcp --all-targets --all-features -- -D warnings   clean
cargo fmt -p rocky-mcp -- --check                                       clean
```

After this, the golden moves only when Rocky's served text or Rocky's version moves. An engine release bump will move the two `initialize` rows once per release, which is the honest signal (the version is served text).

## Review


hugocorreia90:
## Review disposition

The reviewer could not run `cargo` — this lane was compiling on another branch and a build would have written into the shared target directory. It ran the prebuilt test harnesses instead, which compile nothing, and reproduced the golden's two new hashes independently from the pre-change golden. That is stronger evidence than the runs I asked for.

**Verdict as returned: BLOCK.** One comment-only finding, plus two that were worth acting on.

|---|---|---|---|
| A1 | The commit reverses a deliberate trade-off and leaves the argument for the old behaviour standing, now inverted. `tests/roundtrip.rs:5985` still said `from_build_env()` means the row "does not churn on a Rocky release bump. It moves on an rmcp upgrade, which is a change that should force someone to re-read what this server announces." After this change the exact opposite holds. A second stale copy at `:3831`. I had updated that same sentence twice inside `tools.rs` and never grepped for the phrase, so I missed the load-bearing copy, in the file the re-bless belongs to | Confirmed | **Fixed.** Both comments now describe the shipped behaviour, and the golden's comment states the new cost plainly: an rmcp bump no longer moves the row, every engine release does, and what the expected release diff looks like so a re-bless cannot become reflex |
| A2 | The version-lag channel is real, not hypothetical. Crate versions are literals in about twenty-five `Cargo.toml` files, the release bumps them by hand, and the documented check greps for the OLD version, which structurally cannot see a crate that already fell behind. Three crates have drifted that way. My new test asserts `info.version == env!("CARGO_PKG_VERSION")` of rocky-mcp itself, so it can never detect the lag | Confirmed; today they match at 1.74.0 | **Fixed with a guard, not a note.** `rocky-mcp` now exports `SERVER_VERSION`, and `engine/rocky/tests/mcp_server_identity.rs` asserts it equals the binary crate's version. That test lives in the binary crate because that crate's version is what `rocky --version` prints, so the two sides are compared rather than described. It fails the build on a missed bump |
| A3 | Pinning the version in the golden makes two rows a mandatory re-bless every release, which the golden's own failure text warns against | Confirmed | **Accepted, and made visible.** The version is served text, so a release changing it should need someone to say so. The release skill now names the golden and says which two rows to expect, in both skill mirrors, so it fails loudly in the pre-flight instead of surprising the releaser |
| A4 | The golden diff is exactly two rows, and the reviewer reproduced both new hashes from the pre-change golden with an independently built harness | Confirmed independently | Kept as evidence |
| A5 | The new test passes, and no other construction anywhere in the engine announces rmcp | Confirmed | Kept |

**Weakest claim, as the reviewer put it:** "the golden moves only when Rocky's served text or version moves" is true of the hash but sells a pure gain, when what it really does is trade an rmcp-bump re-bless for a per-release one. The changelog and the golden's comment now say that.

**What I was missing:** I edited the same sentence twice in one file and never grepped for it, so the copy that argued for the behaviour I was reversing stayed put. That is the failure my own notes call out — grep the phrase, not the sites you remember.

Gate on the second commit: `cargo test -p rocky-mcp` all five suites green, including the golden; `cargo test -p rocky --test mcp_server_identity` 2 passed; clippy clean on both crates; fmt clean.

**Merge note:** PR #2014 also re-blesses `served_text.golden`, on a different row (`default/tools/list/review_queue`). Whichever merges second re-blesses.

PR #2008, opened 2026-09-16
fix(engine): describe unrecorded column types

### LINKED ISSUE #2002: Browser UI: a dropped column whose old type is unknown reads "is dropped (was ?)"
`describeFinding` in `engine/ui/src/review/PlanDetail.tsx` renders a `column_dropped` finding as:

```ts
`${model}.${column} is dropped (was ${String(change.data_type)})`
```

When the finding carries no old type, the page shows `poc.silver.dim_customer.email is dropped (was ?)` (seen in the screenshots for #2000), or `(was undefined)` when the field is absent. Neither is a type. The rule every screen carries is "no data ≠ unavailable ≠ not recorded": when the old type is not in the payload, say so — `is dropped (type not recorded)` — rather than printing a placeholder as if it were a value.

`column_added` and `column_type_changed` in the same function have the same shape and should get the same treatment.

Found by the docs lane while writing `guides/browser-ui.md` (#2000).

EthemKD:
## Summary

Render missing or placeholder column types as `type not recorded` in browser UI breaking-change descriptions instead of exposing `?` or `undefined` as if they were real types.

Closes #2002

## Subproject(s) touched

- [x] `engine/` (browser UI)

## Changes

- normalize missing, `null`, and `?` type values before formatting findings
- render `type not recorded` for affected dropped, added, and type-changed columns
- add regression cases covering missing and placeholder old/new type values

## Test Plan

- `cd engine/ui && npm test` -> 22 test files passed, 176 tests passed
- `cd engine/ui && npm run build` -> production build passed and the no-external-load check passed
- `git diff --cached --check` -> clean

## Checklist

- [x] Commit messages follow conventional commits, scoped by subproject when relevant
- [x] CLI JSON schema / DSL consumer update is not applicable

### Engine (`engine/`)
- [ ] `cargo test --all-targets` passes (not applicable: browser UI-only change)
- [ ] `cargo clippy --all-targets -- -D warnings` is clean (not applicable: browser UI-only change)
- [ ] `cargo fmt --check` passes (not applicable: browser UI-only change)


hugocorreia90 (approved):
Verified the premise in the engine: RockyType::Unknown really does Display as `?`, so the browser UI is the right layer to fix this. Helper covers undefined, null and `?`, and all three call sites. Thank you.

PR #2013, opened 2026-09-16
docs: pages that recent engine changes made wrong (check expressions, exit 2, CAS, config leniency)

hugocorreia90:
Documentation only. Each page below stated something the engine no longer does. I read the docs line, then the producer on `main`, confirmed the contradiction, and rewrote the text to match the code.

## What was wrong

| Page | What it said | What the code does | Changed by |
|---|---|---|---|
| `concepts/data-quality-checks.md`, `reference/configuration.md` | A check `filter` and a `key_expr` pass "as written" / "verbatim", minus a terminator check, and a subquery runs | One gate covers the check `filter`, `expression`, both `key_expr` fields, the quarantine `expression`/`filter`, `metadata_columns[].value` and MCP `draft_check`: one expression, allowlisted functions, no subquery, no placeholder. Key positions also refuse a volatile function and `COLLATE`; `mode = "split"` refuses a volatile predicate and an error-severity `not_in_future` / `older_than_n_days`. A reserved check name is refused at config load | #1922 |
| `concepts/data-quality-checks.md` | An unevaluated `cross_source_overlap` or `[[assertions]]` result keeps a `warning` severity | Every `*_not_evaluated` constructor is error severity. The one passing exception is a cross-source group whose siblings do not all carry the key column | #1741 |
| `reference/json-output.md`, `reference/glossary.md`, `guides/running-without-an-orchestrator.md`, `dagster/resource.md`, `concepts/execution-flow.md`, `python-sdk/recipes.md` | Exit `2` means "some models failed". The SDK recipe branched on `tables_failed` alone | `rocky run` also exits `2` when an error-severity check fails a run that already copied data (`check_gate_failed`). A quality pipeline moves no data, so its gate exits `1`. The tables also omitted `rocky fulfill` 2/3/4 and `rocky product verify` 2 | #1671, #1598 |
| `concepts/data-quality-checks.md`, `guides/governance.md` | `column_match` "costs no extra query" | It reads both column lists again after the copy: two extra metadata queries per table, retried once unless the warehouse rate-limits | #1678 |
| `reference/configuration.md` | `gc apply` and `apply` upload state without compare-and-swap | `rocky gc` commits through the seam. `rocky restore` is unconditional always, and `rocky apply` only for its verify-after custody rows | #1372, #1989 |
| `reference/configuration.md` | Two blocks accept unknown keys | Three. A key written directly under `[pipeline.NAME]` is accepted and ignored; nested pipeline blocks stay strict | pipeline variant structs carry no `deny_unknown_fields` |
| `reference/configuration.md` | Replication `strategy` is `incremental` or `full_refresh` | Six recognized values. `merge` needs `merge_keys`; `dynamic_table` errors at run time from a pipeline block; an unknown value is a `V035` warning that falls back to `full_refresh` | `RECOGNIZED_REPLICATION_STRATEGIES` |
| `reference/commands/products.md` | "A crash at any point rolls back ... restored exactly" | Recovery completes the commit when the marker landed, and refuses with `commit-io` when it cannot read the marker or take directory custody of a symlinked directory. An unrecorded `.ff-prev` gives `commit-unexpected-backup` | #1819, #1841 |
| `reference/commands/core-pipeline.md` | `run --parallel` defaults to `4` | Only without `--dag`. Under `--dag` node fan-out is unbounded unless the flag is passed | #1288 |
| `reference/commands/governance-reclamation.md` | `audit --by` / `--window` stand alone | Both `requires = "scorecard"` | clap |
| `reference/commands/administration.md` | `replay <target>` is required; `cost` takes `--output` with default `json` | The target can come from `--at <RUN_ID>`, which wins when both are given. `cost` has `--by <DIMENSION>` (`tenant` \| `model`), and `--output` is a global flag whose default depends on the terminal | main.rs |
| `reference/json-output.md` | `checks[].name` is one of three values | Eight name shapes, plus the `severity` and `not_evaluated` fields and four detail shapes the table omitted | `CheckDetails` |
| `reference/model-format.md` | A model sidecar `[[tests]]` `filter` "is not bounded ... passed through verbatim" | It runs the same gate. `rocky test --declarative` with `filter = "customer_id IN (SELECT 1)"` reports status `error` and "expression contains a subquery" | #1922 |
| `dagster/contracts.md` | Three checks, mapped from five codes | E014 maps to `contract_column_constraints`, and a check also reports WARN with `rocky/compile_missing` (no compile output) or `rocky/contract_model_not_found` (W011) | #1619, #1644 |

## How I checked it

Read the producer for every line before changing it, then exercised the claims that a command can show, with the 1.74.0 binary against throwaway DuckDB projects:

- a replication run whose `cross_source_overlap` is error severity: exit `2`, `status: PartialFailure`, `check_gate_failed: true`, `tables_failed: 0`. That last pair is why the SDK recipe needed fixing.
- `key_expr = "now()"`: refused, "expression calls `now`, which is not on the allowlist", reported as an error-severity check on a `warning` assertion.
- `filter = "order_date <= now()"`: accepted, exit `0`. The control for the line above, and the reason the docs say the rule is position-dependent.
- `filter = "customer_id IN (SELECT 1)"`: refused, "expression contains a subquery".
- `mode = "split"` with an error-severity `not_in_future`: refused as `quarantine:compile`, naming "the generated clock comparison".
- `strategyy = "x"` under `[pipeline.p]`: `rocky validate` reports `valid: true`. The same typo under `[pipeline.p.target]` fails the load.
- an assertion named `quarantine_compile`: refused at load with `V046`.
- a model sidecar `[[tests]]` `filter` holding a subquery: refused by `rocky test --declarative`.
- `rocky audit --by rule` and `--window 30d` without `--scorecard`: refused by clap.
- `rocky replay` with no target: the exact error the docs now quote. `rocky replay --at latest`: works.
- `rocky cost latest --by model`: adds `groups`; `--by region` errors.

`cd docs && npm ci && npm run build` passes, and the cross-page links added here resolve in the built HTML.

## Independent red team

Reviewer: **Codex** (job `task-mu3enpfs-dek8vd`, which ran **GPT-5.6-sol** as its own second lens). It reviewed the diff at `c6989654` against the engine sources and returned 6 claims fully correct and 8 partially wrong. I re-checked every finding against the branch before changing anything, and reproduced the behavioural ones with the 1.74.0 binary. Disposition:

| # | Finding | Disposition |
|---|---|---|
| 1a | "every SQL fragment you write into a config" is overbroad: a `[[checks.custom]]` `sql` is a `{target}` substitution, ungated (`checks.rs:253-259`) | **Accepted**, fixed. The page now names the gated fields and calls out the custom-check exception |
| 1b | `[checks.quarantine]` has no `expression` / `filter`; it carries `enabled`, `mode` and two suffixes with `deny_unknown_fields` (`config.rs:1516-1530`) | **Accepted**, fixed. The gated fields are the assertion's, re-checked when quarantine lowers it |
| 1c | A refusal does not always surface as a failing check: `metadata_columns[].value` is a config-load error, MCP is a tool error | **Accepted**, fixed with a four-row table |
| 1d | Refusal list missing `Expr::Lambda` (`check_expression.rs:525`) and the `#` line comment (`validation.rs:418`) | **Accepted**, both added |
| 1e | `random()` / `uuid()` are not on the allowlist, so naming them as a key-only rule implies they pass elsewhere | **Accepted**. The key rule now names clock functions; a separate line says `random()` / `uuid()` are refused everywhere |
| 1f | "trim is on the allowlist" would be wrong (`TRIM` is a syntax node), but "trim passes" is supported | **No change needed** — the page only claims it passes |
| 2 | The passing exception is *exactly one* key carrier. Zero carriers **fails**; two or more are measured, so a group of three with one keyless sibling can still fail (`run.rs:7399-7473`) | **Accepted**, fixed on both pages with a three-way diagram |
| 3a | `rocky ci` never returns 4: `ci.rs:61-65` exits only `if !result.passed()`, so a warnings-only run reports `"exit_code": 4` and exits 0 | **Accepted**, fixed. Reproduced: `compile_ok: true, tests_ok: true, exit_code: 4`, process exit `0` |
| 3b | "A quality pipeline moves no data" is false with quarantine enabled (it writes tables); the gate still exits 1 | **Accepted**. Kept the exit-1 fact, dropped the false reason |
| 3c | The JSON exit table omitted `rocky tick` exit 2 while the guide listed it | **Accepted**, added |
| 4 | The retry fires only for a **retryable, non-rate-limited** failure (`run.rs:14571-14577`); a permanent error is not retried | **Accepted**, fixed |
| 5 | `rocky apply` of a **restore plan** also uploads unconditionally (`apply.rs:237-246` → `restore.rs:1025-1039`), without `verify_after` | **Accepted**, fixed. The exception is now the restore path, however it is reached |
| 6a | "A nested pipeline block is still strict" is false: a check toggle and an assertion entry ignore unknown keys (untagged `AggregateCheckToggle`, `config.rs:1394-1406`) | **Accepted**, fixed. Probed both, plus `[checks.quarantine]`, which does refuse |
| 6b | The VS Code subclaim is unverifiable from source alone | **Accepted**, removed |
| 7, 9, 10, 11, 12, 13 | Replication strategies; `--parallel` and `--dag`; `audit` requirements; `replay` arguments and error string; `cost` dimensions and output default; check names and detail shapes | **Verified correct**, no change |
| 8 | The two `products.md` refusals are on the **uncommitted** branch only; a committed marker deletes the staged and `.ff-prev` siblings without inspecting them (`commit.rs:1616-1645`) | **Accepted**, fixed |
| 14 | The E014 row omits the missing-baseline case (`no_new_nullable` with no `[[columns]]`, `contracts.rs:285-301`) | **Accepted**, fixed |
| extra | `concepts/testing.md:86` and `concepts/architecture-of-trust.md:155-166` still describe `no_new_nullable` as unenforced and contract codes as "four" | **Out of scope here**, owned by a separate PR |
| extra | `reference/model-format.md:386` contradicted the rewrite | Already fixed in `30ed0cb1`, which the review excluded |

The reviewer's own least-confident item was live VS Code TOML diagnostics, which is the subclaim I removed. Its stated gap is that the review is static producer-to-consumer reading with no executed reproductions; I supplied those separately, and the `rocky ci` and config-leniency findings were reproduced before the fix landed.

Refs #1922. Refs #1741. Refs #1671. Refs #1678. Refs #1372. Refs #1819. Refs #1841. Refs #1228.


hugocorreia90:
Merge order: land #1992 first, then this one.

Both PRs edit `docs/src/content/docs/dagster/resource.md` near the same lines. #1992 corrects the engine floor there (1.34 to 1.35 for replication-only projects); this PR corrects the exit-2 sentence. `git merge-tree` reports exactly one conflict, in that file. Everything else in the two PRs is disjoint.

I will rebase this branch once #1992 is in.

PR #2014, opened 2026-09-16
fix(engine/rocky-cli): the review queue ranks by change class, and says so

### LINKED ISSUE #2009: docs(engine/rocky-cli): the review queue says it ranks by classification, but the weight is the change class
The review queue's ranking string and its JSON field say "classification". The weight is the change class, not the data's classification tags.

- `engine/crates/rocky-cli/src/commands/review.rs:725`: `const QUEUE_RANKING: &str = "blast_radius × classification × staleness"`, emitted as `ReviewQueueOutput.ranking`.
- `review.rs`, `fn classification_weight(capability: PolicyCapability)`: `SchemaChangeBreaking => 3`, `Apply | Promote | Backfill => 2`, `_ => 1`. Nothing reads a column classification.
- `rocky review --help` gets it right: "ranked by blast radius × change class × staleness".

Two consequences:

1. The browser UI prints the `ranking` string above the queue, so the page tells a reader that classified data is ranked higher. It is not.
2. `classification_weight` is a field on the queue entry, so the name reaches `schemas/review_queue.schema.json` and the generated SDK and TypeScript bindings. Renaming it is a codegen cascade (`just codegen`), not a one-line edit.

A docs PR cannot fix this: the misleading words come from the engine. Renaming the string alone is safe; renaming the field is the breaking half.

Found by the docs audit while writing the browser UI guide (#2000), which now says "change class".

hugocorreia90:
Closes #2009.

## What was wrong

The review queue described its own ordering as `blast_radius × classification × staleness`, and put a `classification_weight` on every entry. Neither reads a classification. In this codebase `[classification]` is the block that tags columns for masking; the queue never looks at it. The middle factor is the change class:

```text
change_class_weight(capability)   SchemaChangeBreaking          -> 3
                                  Apply | Promote | Backfill    -> 2
                                  everything else               -> 1
```

`rocky review --help` already said "ranked by blast radius × change class × staleness", so the two surfaces disagreed. The browser UI prints the `ranking` string above the queue, which told a reader that classified data ranks higher than it does.

## What changes

```text
engine/crates/rocky-cli/src/commands/review.rs
  QUEUE_RANKING            -> "blast_radius × change_class × staleness"
  fn classification_weight -> fn change_class_weight  (and its locals and tests)
  + the_ranking_string_names_the_factors_the_score_multiplies
engine/crates/rocky-cli/src/output.rs
  ReviewQueueEntry.classification_weight -> change_class_weight
  the `ranking` doc example follows
just codegen ->  schemas/review_queue.schema.json
                 sdk/python/.../review_queue_schema.py
                 editors/vscode/src/types/generated/review_queue.ts
                 docs/public/openapi.json
engine/ui/src/review/*.test.tsx   three fixtures follow the field and the string
engine/CHANGELOG.md               [Unreleased] / Changed
```

**Breaking for anyone reading the JSON.** `classification_weight` is gone, not aliased: the repo takes no back-compat shims before 2.0. The browser UI reads only `ranking`, so nothing in-tree reads the old field outside the fixtures updated here.

The new test pins the string against the score it describes: the three factors in the order `queue_score` applies them, the word "classification" refused outright, and each factor shown to move the score, so none of the three is decoration.

## Evidence

```text
cargo test -p rocky-cli --lib review        161 passed, 0 failed (new test included)
cargo fmt -p rocky-cli -- --check           clean
engine/ui: npm ci && npm run test -- src/review    4 files, 33 tests passed
just codegen                                exit 0; the generated diff is the rename and nothing else
```

The generated diff is four files, ten lines, every one of them the field name or the doc example.

## Review


hugocorreia90:
## Review disposition

**Verdict as returned: BLOCK.** Two findings, both addressed in the second commit.

|---|---|---|---|
| B1 | The rename was incomplete on the surface an AI client reads. The `review_queue` MCP tool's own `description` still said "ranked by blast_radius × classification × staleness", while the same tool's `ranking` value now said `change_class`. One response contradicted itself, and the PR title's "and says so" was false there | Confirmed at `rocky-mcp/src/tools.rs:5519` | **Fixed**, and the served-text golden re-blessed: exactly one row moved, `default/tools/list/review_queue` |
| B3 | No `rocky-sdk` changelog entry for a breaking rename of a required field on a generated Pydantic model, with the precedent sitting eight lines above in the same file | Confirmed | **Fixed.** The entry says the field is required, so naming it breaks at parse time, and that it needs the engine release carrying the rename |
| B4 | The new test is weaker than its name: it pins the string's three tokens and that each factor moves the score, but nothing ties the middle token to what `change_class_weight` reads. Rewrite that function to read a column classification and every assertion still passes | Confirmed | **Documented rather than patched.** No assertion available here can tie a token to a function's semantics; the compiler does it, because the weight's only argument is a `PolicyCapability`. The test's doc now states exactly what it pins and what it cannot, so nobody reads it as more than it is |
| B2 | The same misnomer survives in `BriefEscalationsSection.ranking`'s doc, which describes a not-yet-built ranking as "CLL × classification × staleness", copied into four generated files | Confirmed | **Left alone deliberately.** That sentence describes a future phase of a different surface. It may genuinely intend data classification as a factor, and rewriting someone's unbuilt design on my assumption is worse than leaving one ambiguous word. Flagged here so the choice is visible, not silent |
| B5 | The field rename is otherwise complete and codegen-consistent: one surviving mention of the old name, in changelog prose; schema, Python SDK, TypeScript and OpenAPI all carry the new name with the struct's doc text and the `required` array updated; the browser UI aliases the vscode types rather than keeping a copy; no dagster fixture carries review-queue output, so `just regen-fixtures` is not owed; nothing reads the field by name at runtime; "breaking" is the right label | Confirmed by the reviewer's greps, plus it ran `rocky review --help` | Kept as evidence |
| B6 | The reviewer could not run this PR's test: the prebuilt harness it had came from another branch | Acknowledged | Run here instead: `cargo test -p rocky-cli --lib review` 161 passed |

**Weakest claim, as the reviewer put it:** the title said "and says so" while the surface an agent actually reads still said classification. **What I was missing:** `rocky-mcp` is a second, hand-written consumer of this queue whose tool description is served text pinned by a golden, so it moves with neither `just codegen` nor any drift job. That is the kind of consumer a codegen-shaped change is built to miss.

**Merge note:** PR #2007 also re-blesses `served_text.golden`, on two different rows (`default/initialize`, `worker/initialize`). This PR moves one row (`default/tools/list/review_queue`). Whichever merges second re-blesses; the two changes do not overlap.

Gate on the second commit: `cargo test -p rocky-cli --lib review` 161 passed; `cargo test -p rocky-mcp` all five suites green; clippy and fmt clean on both crates.

PR #2017, opened 2026-09-16
fix(engine/rocky-core): a .rocky model's [[test]] and [[surrogate_key]] sidecar blocks are loaded

### LINKED ISSUE #2015: bug(engine): a [[test]] block on a .rocky model never runs, so a wrong expectation passes
A `[[test]]` block on a `.rocky` model never runs. `rocky test` prints no unit-test line at all and exits 0, so a deliberately wrong expectation passes. The same sidecar on a `.sql` model fails correctly. A `[[surrogate_key]]` block on a `.rocky` model is dropped the same way, so the key column is never created.

## Reproduce (engine 1.74.0, DuckDB, no credentials)

`rocky.toml` declares one transformation pipeline over `models/**`. `models/raw_orders.sql` is a `VALUES` list, so nothing needs seeding.

```toml
# models/customer_totals.toml — the expectation is deliberately WRONG:
# customer 7 totals 15.0, not 999.0.
depends_on = ["raw_orders"]

[target]
catalog = "repro"
schema = "main"

[[test]]
name = "totals_per_customer"

[[test.given]]
ref = "raw_orders"
rows = [
    { order_id = 1, customer_id = 7, amount = 10.0 },
    { order_id = 2, customer_id = 7, amount = 5.0 },
    { order_id = 3, customer_id = 9, amount = 4.0 },
]

[test.expect]
rows = [
    { customer_id = 7, total = 999.0 },
    { customer_id = 9, total = 4.0 },
]
```

Only the model file changes between A and B.

```text
A  models/customer_totals.rocky      rocky test  exit 0
     from raw_orders                   "passed": 2, "failed": 0
     group customer_id {               no "unit test run complete" line anywhere
         total: sum(amount)            in stdout or stderr
     }

B  models/customer_totals.sql       rocky test  exit 1
     SELECT customer_id,              "unit test run complete" total=1 passed=0
       SUM(amount) AS total           the wrong expectation is reported
     FROM raw_orders
     GROUP BY customer_id
```

The missing log line is the tell: the block is not run and passed, it is never loaded.

The same A/B with a `[[surrogate_key]]` block instead, reading the built table afterwards:

```text
A  .rocky   rocky run exit 0   customer_totals columns: customer_id, total
B  .sql     rocky run exit 0   customer_totals columns: customer_id, total, customer_sk
```

## Why

`engine/crates/rocky-core/src/models.rs:1708`, inside `load_unit_tests_from_dir`:

```rust
if path.extension().and_then(|e| e.to_str()) != Some("sql") {
    continue;
}
```

The loader walks the models tree looking for `.sql` files and reads each one's sidecar. A `.rocky` model has a sidecar too, but its model file never matches, so the directory entry is skipped and the sidecar is never opened. `load_surrogate_keys_from_dir_filtered` (`models.rs:1891`) has the identical filter.

`load_column_docs_from_dir` (`models.rs:1789`), three functions away, reads both extensions:

```rust
if ext != Some("sql") && ext != Some("rocky") {
```

So the two-extension form is the intended shape, and two loaders were left behind. The compiler itself accepts both (`rocky-compiler/src/project.rs:534`), which is why `rocky compile` is clean and nothing warns.

## Why it matters

A unit test that cannot fail is worse than one that does not exist: `rocky test` and `rocky ci` go green on a model whose declared expectations are never checked, and the project cannot tell the difference. Every DSL model with a `[[test]]` block is in that state today.

`docs/src/content/docs/concepts/testing.md` and the `[[test]]` section of `reference/model-format.md` both say `rocky test` runs these blocks, without limiting it to `.sql` models.

## Fix

Widen both filters to `Some("sql") | Some("rocky")`, matching the column-docs loader. Two tests to stop it coming back: a `[[test]]` block on a `.rocky` model that fails for the right reason, and a `[[surrogate_key]]` block on a `.rocky` model that reaches the built table. A third, cheaper guard is worth considering: `rocky test` reporting how many unit tests it loaded, so "zero ran" is visible rather than silent.

Found by a documentation walkthrough, reproduced twice independently, and confirmed here on a dev build.

hugocorreia90:
Closes #2015.

## What was wrong

A `[[test]]` block on a `.rocky` model never ran. `rocky test` printed no unit-test line at all and exited 0, so a deliberately wrong expectation passed. A `[[surrogate_key]]` block on a `.rocky` model was dropped the same way, so the key column never reached the built table.

A unit test that cannot fail is worse than one that does not exist: `rocky test` and `rocky ci` went green on models whose declared expectations were never checked, and nothing in the project could tell the difference.

## Why

Two sidecar loaders in `rocky-core`'s `models.rs` scanned `.sql` files only:

```rust
// load_unit_tests_from_dir, and again in load_surrogate_keys_from_dir_filtered
if path.extension().and_then(|e| e.to_str()) != Some("sql") {
    continue;
}
```

Each loader walks the models directory and reads the sidecar beside every model file it recognises. A `.rocky` model has the same `.toml` sidecar, but its model file never matched, so the sidecar was never opened. `load_column_docs_from_dir`, three functions away, already read both extensions after the same bug was fixed there for `[columns]` descriptions in #1444. The compiler accepts both extensions, which is why `rocky compile` was clean and nothing warned.

## What changes

```text
engine/crates/rocky-core/src/models.rs
  load_unit_tests_from_dir               ext: sql | rocky
  load_surrogate_keys_from_dir_filtered  ext: sql | rocky
  both: a .rocky file with no sidecar is skipped rather than read as
        `---toml` frontmatter, which is what the column-doc loader does
  + a_dsl_models_unit_test_block_is_loaded
  + a_dsl_models_surrogate_key_block_is_loaded
  + a_dsl_model_without_a_sidecar_contributes_nothing
```

The two new pair tests write the same sidecar beside a `.rocky` model and a `.sql` model that compute the same thing, then assert on the DSL one directly. They assert the block is present and named, not that the pair matches, so a regression that drops both extensions still fails.

## Evidence

A live A/B on a dev build, one transformation pipeline over DuckDB, the sidecar's expected total deliberately wrong (customer 7 totals 15.0, the sidecar says 999.0):

```text
                                  before                     after
.rocky, wrong expectation   exit 0, no unit-test line   exit 1, unit test run, 0 passed 1 failed
.rocky, correct expectation  exit 0, no unit-test line  exit 0, unit test run, 1 passed
.sql,   wrong expectation   exit 1, reported            exit 1, reported (unchanged)

[[surrogate_key]] on the same pair, reading the built table afterwards
.rocky   columns: customer_id, total              ->  customer_id, total, customer_sk
.sql     columns: customer_id, total, customer_sk ->  unchanged
```

The correct-expectation row matters: the fix makes the test run, not merely fail.

```text
cargo test -p rocky-core --lib models   101 passed, 0 failed (the three new tests included)
cargo clippy -p rocky-core --all-targets --all-features -- -D warnings   clean
cargo fmt -p rocky-core -- --check                                       clean
```

Mutation: reverting either filter to `Some("sql")` fails the matching new test, and nothing else in the suite notices, which is how the gap survived.

## Review


hugocorreia90:
## Review disposition

The reviewer could not run `cargo` at all, because this lane was building on another branch at the time and compiling would have written into a shared target directory. It ran the prebuilt test harnesses instead, which compile nothing. That turned out to be sharper than the runs I asked for: one of those harnesses was a leftover mutation build, and reading it gave the reviewer executed mutation evidence I had not handed over.

**Verdict as returned: APPROVE, with two must-dos before merge.** Both are addressed in the second commit.

|---|---|---|---|
| C3 | `a_dsl_model_without_a_sidecar_contributes_nothing` could not fail. Its fixture was an ordinary DSL body, so the frontmatter fallback found no `---toml` marker and skipped the file whether or not the new guard existed. The commit's own comment two tests above says a test that cannot fail is worse than one that does not exist | Confirmed, and it is exactly the trap I had just written about | **Fixed.** The fixture now opens with `---toml` and declares both blocks, so the guard is load-bearing. The same bytes spelled `.sql` are asserted to load, so the assertions are about the extension and not a malformed fixture. Mutation: deleting the guard now fails the test with "a `.rocky` body is never read as frontmatter" |
| C2 | Every new test called the flat `load_surrogate_keys_from_dir`, while all four production callers use `load_surrogate_keys_from_tree_filtered` with a path predicate. Nothing walked the production path, and the binary the reviewer had predated the fix, so anyone checking the surrogate half by hand would have seen it fail | Confirmed | **Fixed twice.** Added `a_dsl_models_surrogate_key_survives_the_filtered_tree_walk`, which loads through the filtered tree entry point from a nested directory and also asserts the predicate still excludes. And re-ran the live exhibit on a binary built from this branch's head: a `.rocky` model's built table now carries `customer_sk`, same as the `.sql` model |
| C4 | Tests 1 and 2 are fix-sensitive; the reviewer executed the mutation from a leftover harness and reproduced the failure | Confirmed independently | Kept as evidence |
| C5 | No other `.sql`-only sidecar loader was missed. The reviewer enumerated every `read_dir` plus extension filter over the models tree: `load_models_from_dir_filtered` is correctly `.sql`-only because `.rocky` models are discovered by `rocky-compiler`'s own walker, the group and test-definition loaders read `.toml`, the column-doc loader was already fixed in #1444, and the watch, preview and server walkers already list all three extensions. Outside rocky-core, the two `.sql`-only scans read dbt projects | Confirmed | Nothing to do; this is the answer to "what else is broken like this" |
| C6 | Contracts, classification, masks, freshness, `_defaults.toml` and `depends_on` already reach `.rocky` models through `load_single_rocky_model_with_db` | Confirmed | Nothing to do |
| C7 | The "no sidecar means skip" branch is right: the DSL spec documents only the companion `.toml`, and `rocky-lang` has no notion of frontmatter | Confirmed | Kept |

**Weakest claim, as the reviewer put it:** I presented "the key column was never added to the built table" as fixed while every test exercised a loader production does not call. Both halves of that are now covered, by a test through the filtered walk and by the live table.

Gate on the second commit: `cargo test -p rocky-core --lib dsl_sidecar` 4 passed; the earlier run of `--lib models` (101 passed), clippy and fmt are unchanged by a test-only commit.

Issue #2020, opened 2026-09-16
bug(engine): rocky test never substitutes @start_date, so every time_interval model fails it


hugocorreia90:
`rocky test` never substitutes `@start_date` / `@end_date`, so every `time_interval` model fails there with a binder error. `rocky run` substitutes them correctly. A project with any time-interval model therefore cannot have a green `rocky test` or `rocky ci`.

## Reproduce (engine 1.74.0, DuckDB, no credentials)

```sql
-- models/daily_totals.sql
SELECT order_date, SUM(amount) AS total
FROM raw_orders
WHERE order_date >= @start_date
  AND order_date < @end_date
GROUP BY order_date
```

```toml
# models/daily_totals.toml
[strategy]
type = "time_interval"
time_column = "order_date"
granularity = "day"
first_partition = "2026-08-01"
```

```text
rocky compile      Compiled: 3 models, 0 errors, 0 warnings          exit 0
rocky run --latest bounds substituted, model materializes            exit 0
rocky test         ✗ daily_totals — DuckDB error: Binder Error:
                     Referenced column "start_date" not found in FROM clause!
                     WHERE order_date >= @start_date                 exit 1
```

## Why

`substitute_partition_placeholders` lives in `rocky-core`'s `sql_gen.rs:619-624`, on the path `rocky run` takes:

```rust
sql.replace("'@start_date'", &start)
   .replace("@start_date", &start)
```

`rocky-engine`'s `test_runner.rs` never mentions `start_date` or the substitution helper. It hands the model SQL to DuckDB as written, and `@start_date` reaches the binder as an identifier.

## What the test runner would need

A partition window to substitute. The run path gets one from the CLI (`--partition`, `--from`/`--to`, `--latest`, `--missing`) or from state. The test runner has neither, so the fix is a decision as much as a patch:

- **Synthesize a window from the sidecar.** `first_partition` and `granularity` are declared in the model's own sidecar, so the runner can build one partition's bounds with no state and no flags. That makes `rocky test` exercise the same SQL shape the run does, and it is the only option that needs nothing from the user.
- **Take the window from the `[[test]]` block.** More faithful for a model whose expectations depend on the window, and it makes the fixture say which partition it is describing. Needs a new key in the test block.
- **Refuse clearly.** `rocky test` says a time-interval model needs a window and skips it, instead of reporting a binder error about a column nobody wrote.

[...]

Issue #2019, opened 2026-09-16
bug(engine): rocky apply and rocky branch compare demand --pipeline in a multi-pipeline project and have no such flag


hugocorreia90:
In a project with more than one pipeline, `rocky apply <plan-id>` and `rocky branch compare <name>` both refuse with "Use `--pipeline <name>`", and neither verb has that flag. The promote path the deprecation notice itself points at therefore dead-ends.

## Reproduce (engine 1.74.0, DuckDB, no credentials)

Two pipelines, `ingest` and `transform`. After a normal run, create a branch and build it.

```text
rocky plan promote ci-demo
    Error: multiple pipelines defined (ingest, transform). Use --pipeline <name> to select one.

rocky plan promote ci-demo --pipeline transform
    plan_id b8c1eb81…                                              exit 0

rocky apply b8c1eb81…
    Error: multiple pipelines defined (ingest, transform). Use --pipeline <name> to select one.

rocky apply b8c1eb81… --pipeline transform
    Usage: rocky apply <PLAN_ID>          (no --pipeline on apply at all)

rocky branch compare ci-demo
    Error: multiple pipelines defined (ingest, transform). Use --pipeline <name> to select one.

rocky branch compare ci-demo --pipeline transform
    error: unexpected argument '--pipeline' found
```

The only way through is `rocky branch promote <name> --pipeline transform`, which does take the flag and is the spelling the deprecation warning tells users to stop using.

## Why

`engine/rocky/src/main.rs:445` declares `Apply { plan_id: String }` and the branch `Compare` variant takes a bare `name`. Neither carries a `pipeline: Option<String>`, while the sibling verbs that need one do (`main.rs:1223-1225`). Both then reach the shared resolver at `registry.rs:875`, which refuses on an ambiguous config.

The refusal is worse than unhelpful on `apply`: the plan was created with `--pipeline transform` and the persisted plan records its pipeline. The verb is asking the user for something it already has in the file it is about to apply.

## Consequence

Branch promotion is unreachable in any multi-pipeline project except through the deprecated verb. The `ci-cd.md` promote recipe holds only for a single-pipeline project.

## Fix

Two halves, and the first is enough to unblock users:

1. `rocky apply` should read the pipeline from the persisted plan rather than re-resolving it from the config. Nothing about applying a recorded plan is ambiguous.
2. `rocky branch compare` should accept `--pipeline <name>`, like the verbs beside it.

[...]

Issue #1467, opened 2026-08-18
contracts: `[rules] no_new_nullable` is parsed but never enforced


hugocorreia90:
`ContractRules.no_new_nullable` (`engine/crates/rocky-compiler/src/contracts.rs:53`) is deserialized from `.contract.toml` files and documented inline as "If true, no new nullable columns may be added" — but no validation path reads it. The only references in the tree are the struct field, a TOML string in a parse test (`:411`), and that test's assertion (`:418`):

```
$ grep -rn 'no_new_nullable' engine/crates/ --include='*.rs'
engine/crates/rocky-compiler/src/contracts.rs:53:    pub no_new_nullable: bool,
engine/crates/rocky-compiler/src/contracts.rs:411:no_new_nullable = true
engine/crates/rocky-compiler/src/contracts.rs:418:        assert!(contract.rules.no_new_nullable);
```

A contract author who sets `no_new_nullable = true` gets a silent no-op while reading it as an enforced guarantee — the worst failure shape for a contract surface.

Expected: either enforce it (a diagnostic in `validate_contract` / the breaking-change classifier when a new nullable column appears versus the contract baseline) or remove the field and reject it at parse with a pointer to the alternative. Enforcing seems right; until then the docs should not show it in examples.

hugocorreia90:
**Land this with #1240, in one PR.** Same function (`validate_contract`), different causes.

The two are adjacent rather than duplicate, so both stay open — but they ask the same underlying question: *what happens when a declared rule cannot be evaluated?* Answering them separately risks two different policies in one function.

```
#1240  a column whose TYPE is unresolved   -> silent / warn / error ?
#1467  a RULE that is parsed but not run   -> closed-world / no-regression / delete ?
```

Pick one stance for "declared but unevaluable" and apply it to both.

PR #2021, opened 2026-09-16
docs: name the two gaps a reader hits, and no_new_nullable is enforced now

hugocorreia90:
Three pages promise behaviour the 1.74.0 binary does not deliver. Two get a caution naming the open issue. One rule is now enforced, so its "not enforced" note is simply wrong.

| Page | What it said | What 1.74.0 does |
|---|---|---|
| `concepts/testing.md`, `rocky test` | "compiles your models and executes them on a local DuckDB", with no exception | A `time_interval` model fails: `rocky test` does not substitute `@start_date` / `@end_date`, so DuckDB answers `Binder Error: Referenced column "start_date" not found in FROM clause!`. `rocky run` substitutes them. Refs #2020 |
| `guides/ci-cd.md`, the promote gate | `plan_id=$(rocky plan promote …)` then `rocky apply "$plan_id"` | In a project with more than one pipeline, `rocky apply <plan-id>` refuses with `multiple pipelines defined (…). Use --pipeline <name>`, and `rocky apply` has no `--pipeline`. The recipe holds for a single-pipeline project. Refs #2019 |
| `concepts/testing.md`, `[rules] no_new_nullable` | "Parsed but not enforced today … Leave it out" | Enforced, and it emits `E014` (`rocky-compiler/src/contracts.rs`). #1467 is closed |

`E014` is added to the contract diagnostic table and to the compile diagram on the same page.

## How it was checked

- Time interval: a playground model with `WHERE order_date >= @start_date AND order_date < @end_date`, `time_column = "order_date"`. `rocky compile` clean; `rocky test` reports `failed: 1` with the binder error quoted above; `rocky run --latest` materializes.
- Promote: the playground plus a second `[pipeline.ingest]`. `rocky plan promote ci-demo` refuses without `--pipeline`; with `--pipeline playground` it writes a plan; `rocky apply <plan-id>` then refuses with the same message, and `rocky apply --help` shows no `--pipeline`. `rocky branch compare ci-demo` refuses the same way.
- `no_new_nullable`: `contracts.rs` emits `E014` both for an undeclared nullable output column and for a contract that sets the rule with no `[[columns]]` baseline.
- `cd docs && npm run build` passes.

The two cautions come out when the issues are fixed.


hugocorreia90:
## Independent review

Reviewed by Codex (GPT-5.6-sol), adversarial review of the branch diff against `main`. It confirmed the time-interval failure by running it on 1.74.0, and confirmed the `apply` selector limit and `E014` enforcement from source. It raised four findings on the preview section, all accepted. `9e73b568` fixes them; `6af10ea6` adds one more page the earlier review named.

| Finding | Verdict | Fix |
|---|---|---|
| The page tells a reader to run `rocky run --branch <name>`, which does not give `preview diff` a run it can find | Confirmed. `preview.rs:304-307` selects a run whose recorded `git_branch` equals the PREVIEW branch name, while a run records the real branch from `git symbolic-ref` (`run_audit.rs:145-149`). `--branch` changes only where the run writes | Both pages now state the limitation, and `guides/preview-a-pr.md` carries a caution. Filed as #2032 |
| The row-count comparison also misses cardinality changes on ordinary transformations | Confirmed. An ordinary transformation records no `rows_copied` (`run.rs:12395`), and the diff reads the missing value as 0 (`preview.rs:937-938`) | Stated on both pages |
| "a single numeric key" is too generous for bisection | Confirmed. Bounds parse as `i128` (`preview.rs:859-870`), so a decimal key silently falls back to the default comparison | The page asks for whole numbers and tells the reader to check `algorithm.kind` |
| `concepts/preview-internals.md` still described `ORDER BY` / `LIMIT` sampling that production does not do, and `--sample-size` as widening coverage | Confirmed. `preview.rs:449` binds the flag as `_sample_size` and never reads it; the default path emits `limit: 0`, `coverage: "not_yet_sampled"` | The sampling section is marked as the design, not 1.74.0 behaviour, and the flag row says it is ignored |

Also in `6af10ea6`: `concepts/architecture-of-trust.md` said four intra-project contract codes and listed `E010`-`E013`. With `no_new_nullable` enforced there are five.

PR #2023, opened 2026-09-16
fix(engine/rocky-cli): the drop-and-recreate drift reason reads from the existing type to the new one

### LINKED ISSUE #2022: bug(engine): the drop-and-recreate drift reason prints the two types backwards
When schema drift forces a drop and recreate, the reason prints the two types the wrong way round. A column that went `TIMESTAMP` to `DATE` is reported as `changed DATE → TIMESTAMP`. The ALTER reason a few lines below prints the same pair correctly, so the two arms of the same decision disagree.

## Where

`engine/crates/rocky-cli/src/commands/run.rs`, two arms of the same drift match:

```rust
// drop_and_recreate arm, line 13478
"column '{}' changed {} → {}", c.name, c.source_type, c.target_type

// alter_column_types arm, line 13533
"column '{}' widened {} → {}", c.name, c.target_type, c.source_type
```

A `DriftedColumn` names its sides by origin, not by direction (`rocky-core/src/drift.rs:43-47`): `target_type` is what the existing table has, `source_type` is what upstream now reports. The change always runs from the target's type to the source's, so `target_type` is the "from" side. The ALTER arm is right and the drop arm is backwards.

Nothing pins either string, which is how the two drifted apart.

## Why it matters

`drop_and_recreate` is the destructive action of the pair: it drops the operator's table and rebuilds it. The one line explaining why reads backwards, so an operator checking whether the change is the one they expected sees the opposite of what happened. It reaches `RunOutput.drift.reason`, so orchestrators and the browser UI carry it too.

## Fix

One helper both arms call, with the direction fixed in one place, and a test on the rendered string rather than on the argument order.

hugocorreia90:
Closes #2022.

## What was wrong

Two arms of the same drift decision rendered the same pair of types in opposite orders:

```rust
// drop_and_recreate, run.rs:13478
"column '{}' changed {} → {}", c.name, c.source_type, c.target_type

// alter_column_types, run.rs:13533
"column '{}' widened {} → {}", c.name, c.target_type, c.source_type
```

A `DriftedColumn` names its sides by where they came from, not by direction (`rocky-core/src/drift.rs:43-47`): `target_type` is the type the existing table has, `source_type` is the type upstream now reports. The change runs from the target's to the source's, so the ALTER arm was right and the drop arm printed the change backwards. A column that went `TIMESTAMP` to `DATE` was reported as `changed DATE → TIMESTAMP`.

It is the destructive arm of the pair: it drops the operator's table and rebuilds it, and the one line explaining why said the opposite of what happened. The string reaches `RunOutput.drift.reason`, so orchestrators and the browser UI carried it too.

## What changes

```text
engine/crates/rocky-cli/src/commands/run.rs
  + fn drift_reason(drifted_columns, verb)   one renderer, target_type first
    both arms call it; the only behaviour change is the drop arm's order
  + a_drift_reason_reads_from_the_existing_type_to_the_new_one
```

Nothing pinned either string before, which is how the two drifted apart. The new test asserts on the rendered text, not on the argument order, so it still fails if the fields are renamed or swapped at the struct.

## Evidence

```text
cargo test -p rocky-cli --lib drift_reason   1 passed
cargo fmt -p rocky-cli -- --check                                       clean
```

The test pins both verbs and the empty case. Swapping the two fields inside the helper fails it on the first assertion.

## Review


hugocorreia90:
## The codegen gate found the best evidence in the repo

`Check codegen is up to date` failed on this branch, which is the gate doing its job: a live-binary fixture captured from a POC holds the string this PR changes.

```diff
integrations/dagster/tests/fixtures_generated/drift/run_after_drift.json

-        "reason": "column 'amount' changed VARCHAR → DECIMAL(10,2)"
+        "reason": "column 'amount' changed DECIMAL(10,2) → VARCHAR"
```

The POC that produced it narrates the change in its own script:

```text
examples/playground/pocs/02-performance/06-schema-drift-recover/run.sh:18
  echo "=== Drift the source: amount DECIMAL(10,2) -> VARCHAR ==="
  duckdb poc.duckdb "ALTER TABLE raw__orders.orders ALTER COLUMN amount ..."
```

The seed creates `amount` as `DECIMAL(10,2)`, the POC widens it to `VARCHAR`, and the target table still holds the decimal. So the change runs `DECIMAL(10,2) → VARCHAR`, and the committed fixture said the opposite. The repo's own POC and its own captured output disagreed, and the fixture is what an operator reading `RunOutput.drift.reason` would have seen.

Fixture regenerated with `just regen-fixtures` against a binary built from this branch's head (checked, because `regen_fixtures.sh` uses whatever sits at `engine/target/debug/rocky` and a stale one silently recaptures the old string). One line changed in one fixture; the other thirty-nine are byte-identical.

Issue #1908, opened 2026-09-10
policy.rs: three more text renderers print through println!, so no test can read them


hugocorreia90:
## What

`policy.rs` has three more text renderers that still call `println!`, so no test can read what they produce:

| Function | Command |
|---|---|
| `render_test_text` | `rocky policy test` |
| `render_text` | `rocky policy check` |
| `render_freeze_text` | `rocky policy freeze` / `unfreeze` |

Give each the shape #1879 gave `render_show_text`:

```rust
fn render_test_text<W: Write>(w: &mut W, out: &PolicyTestOutput) -> io::Result<()>
```

## Why this is a separate issue and not part of #1879

Changing the signature is the cheap half and it delivers nothing on its own. The value is the tests, and the assertion has to be the **whole block** — a substring assertion is the exact weakness #1879 exists to close, because round one of the #1874 review lost a line that was present and merely shorter.

So this is four renderers' worth of full-block tests, each needing its own output fixture. That is its own review, not a rider on a PR that closed one renderer.

## Why it matters here

`render_freeze_text` is the highest-value of the three. It prints the `! {note}` lines that warn a freeze is **inert** — recorded but not enforced, because the config has no `[policy]` block. Deleting that loop would pass every test in the repository today, and the operator would be told their freeze took effect when nothing enforces it.

`render_text` (`policy check`) prints the resolved effect and the winning rule, which is the answer an operator acts on.

## Shape

Follow #1879 exactly: sink parameter, `io::Result`, a `Vec<u8>` helper in the test module, and a fixture per branch. Mutation-check each test by deleting the line it claims to pin.

Refs #1879, #1874.

hugocorreia90:
claimed by backlog 2026-09-22

PR #2026, opened 2026-09-16
fix(engine/rocky-cli): every table prints the whole run id

### LINKED ISSUE #2025: bug(engine): every table truncates the run id so far that it identifies nothing and cannot be passed to trace, cost or replay
Every table that prints a run id truncates it so far that the id neither identifies the run nor can be passed to another command. In `rocky history` the printed id is the same string for every run in the same month.

## What it looks like

A run id is 23 characters, `run-<date>-<time>-<millis>`, for example `run-20260915-223451-415`. The tables cut it to 9, 11 or 13:

```text
rocky history --output table        11 chars   run-2026091      every run in September 2026
rocky history --audit               11 chars   run-2026091      same
rocky history --model <name>        13 chars   run-20260915-    every run on that day
rocky metrics <model> --trend        9 chars   run-20260        every run in 2026
```

Two runs a second apart print identically, so the column cannot be used to tell rows apart, and the printed value cannot be pasted into `rocky trace <run-id>`, `rocky cost --run <id>` or `rocky replay <run-id>`. The JSON output carries the full id, so `--output json | jq` is the only way to get one.

## Where

```text
engine/crates/rocky-cli/src/commands/history.rs:391   min(13)   model executions table
engine/crates/rocky-cli/src/commands/history.rs:474   min(11)   run summary table
engine/crates/rocky-cli/src/commands/history.rs:496   min(11)   governance audit table
engine/crates/rocky-cli/src/commands/history.rs:536   min(11)   per-run detail line
engine/crates/rocky-cli/src/commands/metrics.rs:175   min(9)    metrics trend table
```

Each is paired with a column width one character wider than the truncation, so the truncation is the column budget rather than an accident. The widths were chosen when the run id was presumably shorter.

## Fix

Print the id whole and widen each column to 24. The widest table, the governance audit trail, goes from 100 to 112 characters, which its own comment already budgets for ("usable in a 160-column terminal"). No other column needs to move.

An id that cannot be copied into the next command is the part that matters; a wider column is the cost.

hugocorreia90:
Closes #2025.

## What was wrong

A run id is 23 characters, `run-<date>-<time>-<millis>`. Four tables cut it to 9, 11 or 13, which is short enough that the column identifies nothing:

```text
rocky history --output table        11 chars   run-2026091     every run in September 2026
rocky history --audit               11 chars   run-2026091     same
rocky history --model <name>        13 chars   run-20260915-   every run on that day
rocky metrics <model> --trend        9 chars   run-20260       every run in 2026
```

Two runs a second apart printed identically, and the printed value could not be pasted into `rocky trace <run-id>`, `rocky cost --run <id>` or `rocky replay <run-id>`. Reading the id out of `--output json` was the only way to get one.

## What changes

Each truncation is dropped and its column widened to 24, in `history.rs` (four sites) and `metrics.rs` (one). Nothing else moves. The widest table, the governance audit trail, goes from 100 to 112 characters, inside the 160-column budget its own comment states.

## Evidence

Before and after on the same two-run project, table mode:

```text
before   RUN ID
         run-2026091
         run-2026091

after    RUN ID                   STARTED                  STATUS     MODELS   TRIGGER
         run-20260915-223451-415  2026-09-15 22:34:51      Success    1        Manual
         run-20260915-223450-878  2026-09-15 22:34:50      Success    1        Manual
```

The two rows are now distinguishable, which is the whole point of the column.

```text
cargo test -p rocky-cli --lib history   18 passed, 0 failed
cargo fmt -p rocky-cli -- --check                                       clean
```

No test covers these renderers, because they print through `println!` and nothing can read them. That is #1908's class of problem and is not addressed here; the exhibit above is the evidence instead.

## Review


hugocorreia90:
## Correcting this PR, and the claim that let the defect through

The PR body said:

> No test covers these renderers, because they print through `println!` and nothing can read them.

That is false, and it was my own claim, not something I inherited. `engine/rocky/tests/producer_seams_stdout.rs:791` runs the binary and compares the whole history table byte for byte, including the column widths and the separator rule. I looked for tests inside `history.rs`, found none, and wrote a general conclusion from a module-scoped search. The test is one directory up.

It failed in CI, which is how this was found, and it earned its keep twice over — the second failure was a real defect I had shipped into the branch:

```text
header  line 491   "{:<24} {:<18} {:<8} …"     updated
row     line 528   "{:<12} {:<18} {:<8} …"     NOT updated
```

The governance audit table's header moved to the new width and its data row did not, so every row sat eleven columns left of its heading. My earlier edit had replaced the header block, which in that function is a separate format string from the row, and I confirmed the change by reading the header. The output:

```text
RUN ID                   IDENTITY           SOURCE   …
run-under-test seams-test         cli      …
```

Both are now fixed: all four tables in the file have matching header and row widths, checked by listing every format string rather than by reading the ones I remembered editing.

The test's expected table now derives its padding from the same format specifiers rather than hard-coding spaces, so it still pins the layout but does not have to be re-counted by hand next time a column moves.

```text
cargo test -p rocky --test producer_seams_stdout   16 passed
```

PR #2031, opened 2026-09-16
fix(engine/rocky-cli): rocky ci exits with the code its JSON reports

### LINKED ISSUE #2030: bug(engine/rocky-cli): rocky ci reports exit_code 4 in its output and exits 0
`rocky ci --output json` can report `"exit_code": 4` while the process exits `0`. A CI step that trusts the process status passes; one that reads the field fails. They disagree about the same run.

## Reproduce (engine 1.74.0)

```bash
rocky playground citest && cd citest
printf 'model = "revenue_summary"\n\n[[columns]]\nname = "ghost_column"\n' > models/revenue_summary.contract.toml
rocky --output json ci > ci.json; echo "process exit=$?"
```

```
process exit=0
compile_ok True   tests_ok True   exit_code 4
```

The contract declares a column the model does not produce, which is a warning-class outcome, not a compile error.

## Why

`engine/crates/rocky-cli/src/commands/ci.rs:61-63`:

```rust
if !result.passed() {
    std::process::exit(result.exit_code());
}
```

A warnings-only result is `passed()`, so the process returns `0` even though `exit_code()` is `4` and both the table output ("Exit code: 4") and the JSON say so.

## Why it matters

`docs/src/content/docs/guides/running-without-an-orchestrator.md` and `reference/json-output.md` document `4` as a real CI outcome. Any pipeline that gates on `$?` treats this run as clean, while a pipeline that gates on `jq .exit_code` treats it as failed.

Either the process should exit with the code it reports, or the field should not report a code the process will not use.

Found by an adversarial review of the docs (PR #2013), then reproduced. Refs #2013.

hugocorreia90:
Closes #2030.

## What was wrong

`rocky ci` decided whether to exit non-zero with a different rule from the one it used to compute the code:

```rust
if !result.passed() {                 // passed() is compile_ok && tests_ok
    std::process::exit(result.exit_code());   // 0 pass, 1 errors, 4 warnings-only
}
```

A warnings-only run satisfies `passed()`, so the branch was never taken and the process exited 0 while the JSON payload said `"exit_code": 4`. A CI step gating on `$?` and one gating on `jq .exit_code` reached opposite conclusions about the same run.

The `4` is not an accident: `exit_code()` carries a comment explaining why warnings-only is `4` and not `2` (`2` is reserved for `rocky run` partial success, which the dagster integration keys on), a unit test pins it, and the reference page documents it as an outcome. Everything declared the contract except the line that acts on it.

## What changes

```text
engine/crates/rocky-cli/src/commands/ci.rs
  exit with result.exit_code() whenever it is non-zero
engine/rocky/tests/ci_exit_code_agrees.rs   (new)
  a_warnings_only_ci_run_exits_with_the_code_its_json_reports
  a_clean_ci_run_exits_zero_on_both_surfaces
```

## This can turn a green pipeline red

A project whose models emit advisory warnings now fails `rocky ci` with exit 4 where it used to pass. That is the contract the code, its tests and the docs already stated, and the repo takes no back-compat shims before 2.0, so I have shipped the behaviour rather than the silence. The changelog says so in the entry's own words, and teams for whom warnings are advisory can gate on `jq .compile_ok` and `jq .tests_ok` instead.

**The alternative, if you would rather not break those pipelines:** make `exit_code()` return 0 for warnings-only and delete the `4` from the docs and the unit test. That keeps every pipeline green and gives up a documented outcome nobody can currently observe from the shell. I think that is the wrong trade, because the JSON has been reporting 4 all along and orchestrators can already see it, but it is a one-line change in the other direction if you disagree.

## Evidence

The fixture leans on `D012`, a bare read of a model whose configured target table has a different name. That is a warning by design and leaves compile and tests passing, which is exactly the corner where the two surfaces disagreed.

```text
cargo test -p rocky --test ci_exit_code_agrees        2 passed
  warnings-only project   json exit_code 4, process exit 4, compile_ok and tests_ok both true
  clean project           json exit_code 0, process exit 0
```

Mutation: restoring the old `if !result.passed()` gate fails the warnings-only test on "the process must exit with the code the JSON reports" and leaves the clean test passing, so the test discriminates rather than simply failing.

## Review


hugocorreia90:
## Held for a ruling, deliberately

I am not merging this one with the rest of tonight's queue. Every other PR in that queue fixes something that is wrong; this one changes what a passing build means.

After it, `rocky ci` exits 4 on a run that compiles and tests clean but emits an advisory warning, where it used to exit 0. Any project that has a warning today gets a red pipeline on upgrade, and nobody reads a changelog in time to prevent that. That is the widest blast radius of anything open in either engine lane tonight, and it is the kind of change that is discovered at 3am rather than at review.

The direction is defensible: `exit_code()`'s own doc, its unit test and the reference page all already say 4 is the warnings-only outcome, and the JSON has been reporting it all along. The code simply never acted on it. But "defensible" is not the same as "decided", and the two honest options lead to different published contracts:

```text
A (this PR)  the process exits 4.       Green pipelines with warnings turn red.
                                        The documented outcome becomes observable
                                        from the shell for the first time.
B            exit_code() returns 0       Every pipeline stays green. Exit code 4
             for warnings-only.          stops being an outcome anything can see,
                                        and comes out of the docs and the test.
```

The docs lane has both pages ready to follow whichever way this goes, so the cost of deciding later is one docs edit, not a rewrite.

**Recommendation: A**, because the JSON has been reporting 4 to orchestrators the whole time and only the shell was lied to. But this sits until there is a ruling. The rest of the queue lands without it.

hugocorreia90:
**My recommendation on this PR changes from A to closing it.**

The hold comment above argued for A mainly because "the reference page documents it as an outcome". Tonight's docs merges changed that. Both pages on `main` now document the current behaviour:

```text
guides/running-without-an-orchestrator.md   "`rocky ci` does not return `4`."
reference/json-output.md                    "Branch on the exit_code field if you want
                                             to act on advisory warnings."
```

So merging this PR now contradicts two published pages and turns every warnings-only pipeline red. Closing it changes nothing that users have been told. A search of the SDK, the dagster integration, the VS Code extension and the repository's actions finds nothing that expects the process to exit 4.

If you close it, one follow-up: a JSON field named `exit_code` that is not the process exit code is a misleading name. Renaming it is a fix, and the two pages that tell users to read it would change in the same PR.

It still changes the meaning of an exit code, so the ruling stays with you. #2030 stays open until then. Reasoning: https://github.com/rocky-data/rocky/issues/2035#issuecomment-5696619896

hugocorreia90:
Closed by ruling (2026-09-17): `rocky ci` keeps exiting 0 on a warnings-only run, as both published pages document. #2030 is closed with the same ruling.

PR #2034, opened 2026-09-16
fix(engine/rocky-cli): preview refuses a classified column it cannot mask, instead of returning it raw

### LINKED ISSUE #2029: preview rows returns a classified column unmasked when its tag resolves only under [mask.<env>], and the compiler stays silent
`rocky preview rows` — and the browser UI's "Show 20 rows" button, which goes through the same `compute_preview_rows` (`rocky-cli/src/api.rs:1789`) — can return a classification-tagged column with its raw values. The page says the opposite: *"Classification-tagged columns are masked by the engine before the rows leave it"* (`engine/ui/src/review/SamplePanel.tsx:59`).

## The split

`preview_rows.rs:227-236` resolves masks for the **workspace-default env**:

```rust
let masks = rocky_cfg.resolve_mask_for_env(None);
for (col, tag) in &m.config.classification {
    if let Some(&strat) = masks.get(tag) && strat != MaskStrategy::None {
        masked.insert(col.clone(), strat);
    }
}
```

`resolve_mask_for_env(None)` (`rocky-core/src/config.rs:3717-3740`) runs pass 1 only: scalar `[mask]` entries. A tag defined solely under `[mask.prod]` is a `MaskEntry::EnvOverride` and never enters the map. So `masked` is empty, and `preview_rows.rs:294` takes the `masked.is_empty()` branch — the plain `SELECT` — and the raw values reach the caller.

**The compiler does not warn for this case.** W004's coverage set (`rocky-compiler/src/typecheck.rs:1030-1042`) inserts every key found in *any* `EnvOverride` map, so `[mask.prod] pii = "hash"` marks `pii` resolved and no diagnostic is emitted. Three components disagree, and the quietest configuration is the one with no signal at all:

```
  [mask] pii = "hash"          compiler: covered    preview: MASKED     page: honest
  [mask.prod] pii = "hash"     compiler: covered    preview: RAW        page: WRONG, and silent
  no [mask] entry at all       compiler: W004       preview: RAW        page: WRONG, one warning
  allow_unmasked = ["pii"]     compiler: silent     preview: RAW        page: WRONG, by operator choice
```

Row 2 is the configuration a reader most plausibly has while setting masking up, and the one the UI's sentence is most wrong about.

## Why this is worth fixing rather than documenting

The sample route is reached by a browser holding a **read-only** token. Read-only is the scope we hand a UI precisely so a leak of it cannot reach the warehouse through a mutation; it still reaches warehouse *data* through this route, and the masking promise is what bounds that. The CTE and ad-hoc paths already fail closed for exactly this reason (`cte_masking_unverified`, `adhoc_masking_blocked` at `preview_rows.rs:239-266`) — the model-output path is the one that does not.

## Options

1. **Fail closed on an unresolvable tag.** When a column carries a classification tag that resolves to no strategy in the env being previewed, refuse the sample with a code beside the two that already exist (`masking_unresolved`), naming the columns and the tag. Consistent with the CTE and ad-hoc refusals. It would also refuse the `allow_unmasked` case, so that list should be honoured as a deliberate opt-out and allowed through.
2. **Resolve the env the preview actually runs against**, rather than the workspace default, so a `[mask.prod]` entry applies when previewing against prod. Narrower, and it leaves rows 3 and 4.
3. **Make W004 agree with the resolver**: warn when a tag resolves under no *default* strategy, since that is the env every non-env-qualified command uses. A diagnostic, not a gate.
4. **Fix the UI string** to say what is true — the masked columns are listed in the payload, so the page can name them rather than promise a class.

I would take 1 + 4, with 3 as a follow-up. Whichever way it goes, the page must stop promising what the engine does not guarantee.

Found by a Codex review of the docs lane's browser-UI guide (#2000); the guide's own wording has already been corrected. Related: #1897 (the config-secret trust boundary), the W004 advisory note.

hugocorreia90:
A column tagged in `[classification]` came back with its **raw values** whenever its tag resolved to no mask strategy for the preview. The browser UI's sample button reaches the same core with a read-only token, while the page stated the opposite.

Closes #2029.

## The hole

```
  [mask] pii = "hash"        masked      ✓
  [mask.prod] pii = "hash"   RAW         and W004 says nothing — no signal anywhere
  no [mask] entry            RAW         W004 warns, but a warning does not gate a query
  allow_unmasked = ["pii"]   RAW         on purpose
```

`preview_rows` resolved masks with `resolve_mask_for_env(None)` (`rocky-core/src/config.rs:3717-3740`), which runs pass 1 only — scalar `[mask]` entries. A tag defined solely under `[mask.prod]` is a `MaskEntry::EnvOverride` and never entered the map, so `masked` stayed empty and the plain `SELECT` ran.

**Row 2 had no signal at all.** W004's coverage set (`rocky-compiler/src/typecheck.rs:1030-1042`) inserts every key found in *any* env-override table, so `[mask.prod] pii = "hash"` marks `pii` covered and the compiler stays silent. That is the configuration someone most plausibly has while setting masking up.

The CTE and ad-hoc preview paths already fail closed for exactly this risk (`cte_masking_unverified`, `adhoc_masking_blocked`). The model-output path — the one a browser reaches — did not.

## The fix

Classified columns now resolve through one pure function, `resolve_classified_columns`, into three outcomes: **masked**, **deliberately allowed**, or **unresolved**. An unresolved column refuses the preview with a new `masking_unresolved`, naming every column, every tag, and both remedies.

- `MaskStrategy::None` counts as unresolved — it masks nothing, so treating it as handled would return the raw value.
- `[classifications.allow_unmasked]` — the same list that suppresses W004 — is honoured as the operator's explicit opt-out.
- The route maps it to `422` beside the two existing masking refusals, with a hint naming both remedies.
- The UI string and the `PreviewRowsOutput` schema description now say what is true: masked, or refused, except a tag the project deliberately allows.

Not changed: W004's own coverage rule, which still counts an env-only entry as covering the tag. Making the diagnostic agree with the resolver is worth doing and is a separate change — this PR closes the disclosure, which a warning never gated anyway.

## Evidence

- `a_classified_column_is_masked_allowed_or_refused_never_silently_raw` walks all four configurations plus `MaskStrategy::None` and an unclassified column.
- `the_unresolved_masking_refusal_names_the_columns_and_both_remedies` pins the message and the envelope's `columns`/`tags`.
- **Mutations**, through `scripts/mutation-check.sh`, both `OK — fix-sensitive`: (A) the refusal arm removed so a classified column falls through silently; (B) `MaskStrategy::None` accepted as a mask.
- `cargo test -p rocky-cli --lib -- preview_rows` 14 passed. `npx vitest run src/review` 33 passed. `npm run typecheck`, `npm run lint`, `cargo fmt --check`, `cargo clippy -p rocky-cli --all-targets --all-features -- -D warnings` all clean. `just codegen` run; the generated diff is the schema description in four files.

## Review

Reviewer: a fresh-context **same-model** pass. Not independent: five Codex task dispatches from this session were refused at the model (`gpt-6-astra`), and a peer session hit the same refusal at 02:18, so the path was transient-failing for everyone who tried tonight — the tally across the machine is 251 completed / 31 failed, so this is not a broken tool, just one that would not take my dispatch. `/codex:review` is user-invocation-only and is the path that completed jobs tonight; this PR is worth one if the independent read is wanted before merge.

**Least confident:** that `resolve_mask_for_env(None)` is the only resolver reaching a caller-visible path with this shape. I checked the preview core; a sweep of the MCP tools, `rocky compare` and any profiling command for the same pattern is the follow-up I would want, and it is in #2029.

**Most likely missing:** whether any project in the wild relies on previewing a model whose tags resolve only per-env — for them this turns a working sample into a refusal. That is the intended direction (fail closed), and the message names the one-line fix, but it is a behaviour change for that configuration and belongs in the release notes.


hugocorreia90:
**Correction, and evidence from a shipped project.** The first version of this PR refused a preview when a tag resolved to `MaskStrategy::None`, on the reasoning that it masks nothing. That was wrong: the config reference calls `"none"` *"explicit identity (not a gap; a policy decision)"*, and `inline_mask_expr` returns no expression for it by design. The operator wrote the entry and chose identity; refusing it overrides their answer.

It would also have broken a shipped example. `examples/playground/pocs/04-governance/05-classification-masking-compliance` declares `internal = "none"` with `region = "internal"` on `users`. **No PR gate runs POC smoke** (#1676), so CI would not have caught it — the same shape as the #1922 breakage in the 1.74 cut.

Why my own tests missed it: every assertion was *"is refused"*, and no amount of extra strictness can fail those. The `none` case is now an **acceptance** assertion, and a mutation that refuses `none` fails it (`mutation-check: OK — fix-sensitive`).

Run against that POC with a debug binary built from this branch:

```
  preview rows --model users      region = "internal", internal = "none"
    → upstream_not_materialized   the POC's own pre-existing shape: a replication
                                  pipeline does not materialize transformation models.
                                  The guard runs BEFORE any SQL, so reaching DuckDB is
                                  the proof that `none` was accepted.

  preview rows --model accounts   audit_note = "audit_only", no [mask] entry,
                                  allow_unmasked commented out
    → masking_unresolved          "model 'accounts' classifies 1 column(s) whose tag
                                  resolves to no mask strategy for this preview:
                                  audit_note (audit_only) …"
```

The second is the intended behaviour on a real project, and it teaches the lesson the POC already documents in its own comment — `audit_only` belongs under `[classifications.allow_unmasked]`. No POC's `run.sh` invokes `preview rows`, so no scripted flow changes.

The rule the refusal now draws is **written down or not**, never *masked or not*: a strategy, `none`, or a tag under `allow_unmasked` are all answers. Only a tag that resolves to nothing here is refused.

PR #2051, opened 2026-09-16
docs(browser-ui): the guide follows the eleven areas, not the three lanes

hugocorreia90:
#2050 replaced the UI's three-lane header with a sidebar of eleven areas. The page a reader lands on still described three lanes, so it now describes the areas.

## What changed

One section per area that opens a screen, in sidebar order: Needs you, Estate, Review, Products, Governance. One table for the six that open nothing.

- **Needs you** is the Brief section moved, not rewritten. Same screen, same route, same cards, minus the lane's tab bar. One clause was added so a reader does not confuse the area with the card of the same name inside it.
- **Products** leaves the governor tab bar and becomes its own section.
- **Governance** keeps the three tabs it still has. `GovernorScreen.tsx` renders the bar only when the current area has more than one tab, so Governance is the only area that shows tabs.
- **The six without a screen** carry the reason string the sidebar shows, word for word. Each one says where the information is instead, so paraphrasing it would cost the reader the answer.

## Two corrections the restructure turned up

- "The estate is the first screen" was true of the lane header only. `laneFromPath` still sends `/ui/` to the estate, so the printed address opens it, but the sidebar's first entry is Needs you. The sentence now says the first of those two.
- "The Governor screens report decisions" in *What the UI cannot do* named an area that no longer exists.

The root `README.md` said "three areas" as well.

## Verified

- Labels, order and all six reason strings read from `engine/ui/src/areas.ts` at `a3df5d2c`, not from the description of the change.
- The tab rule read from `engine/ui/src/governor/GovernorScreen.tsx` (`tabs.length > 1`) and `areaHasTabs` in `areas.ts`.
- No inbound link anywhere in the repo uses an anchor on this page, so renaming the headings breaks nothing.
- `cd docs && npm run build` passes, 109 pages.

## Images

The screenshots stay at 1.74.0, which is what a reader running the released engine sees. A line under the tour says the images show the current release and the sidebar arrives in the next one. Recapturing is in the release checklist, at the next engine tag.

Refs #2050.


hugocorreia90:
## Codex disposition

Independent adversarial review by **Codex (GPT-5.6-sol)**, branch scope against `origin/main`. Verdict: `needs-attention`, one medium finding. It confirmed the rest against source: the eleven labels and their order, the six reason strings word for word, Governance as the only area with tabs, the Estate default, the 768px Menu behaviour, and the security section.

**[medium] The README advertised the sidebar with no version qualification — adopted.**

The finding is right and it matters more than the page it was found on. `README.md` is the front door. It described the eleven-area sidebar as a property of the release binary, while the guide says the sidebar arrives in the next release. Someone installing the current engine gets the three-lane header and cannot find what the README sent them to look for.

I took the finding but not the suggested remedy. Codex proposed qualifying the README with "arrives in the next release". That sentence is accurate for exactly one release cycle and then silently becomes wrong, which is how the page got stale the first time. The README now names what the UI *shows* rather than how it is *grouped*:

> It shows what needs you now, the estate (models, the DAG and runs), the plans waiting for a human, your data products, and the record of what agents did and why.

All five exist in both releases — in the current one as lanes and governor tabs, in the next as areas. So the sentence is true either side of the change and needs no qualification.

**The expiring sentence that remains, and who retires it.**

The guide still carries one line under the tour GIF saying the images show the current release and the sidebar arrives in the next. That one has to expire, because it is what stops the screenshots from lying in the meantime. So the release checklist now retires it at the tag that ships the change, next to the screenshot recapture, in both skill mirrors byte for byte.

**Not run:** no release binary and no browser. Codex said the same, and named it as what it was least confident about. The claims here are read from source at `a3df5d2c`, and the screenshots are from the released 1.74.0 build.

hugocorreia90:
## Second review: fact check against the merged UI

A peer session that owns the UI code fact-checked this page against `origin/main` at `a3df5d2c`: 16 claims, 15 true, one wording defect. Fixed in 9fe0f958.

**The defect.** I wrote:

> The printed address opens the Estate, whichever area the sidebar lists first.

I meant "regardless of which area is first". Next to *Estate*, that trailing clause reads as a description **of** Estate, which is false — the sidebar lists Needs you, then Projects, then Estate, as the table eight lines earlier in the same page says. It now names both halves:

> The printed address opens Estate, even though Needs you is first in the sidebar.

**Confirmed true against source:** the eleven areas and the five that open a screen; every label and the order; the six reason strings word for word; the reasons rendering as plain text under the name; the Menu button below 768px; Governance as the only area with a tab bar; `/api/v1/policy` and `/api/v1/settings` existing as GET routes; the brief's nine cards and their order; **not recorded**; what the Products screen shows; that no route changed; that the README sentence maps to the five link areas; and that both skill mirrors are identical.

**The one claim neither review could check from `main`,** now settled from the tag rather than from memory: at `engine-v1.74.0`, `App.tsx` renders a `<nav aria-label="Lanes">` over the three `LANES` in `router.ts` — Estate, Review, Governor. So the line telling a reader the current release groups the screens under a header of three lanes is read from the released source. No binary was executed, and no browser.

Two independent reviewers now, on different lenses: Codex on the diff, the UI owner on the claims. They found disjoint defects — Codex the README's missing version qualification, the fact check a sentence that contradicted my own table.

Issue #2035, opened 2026-09-16
Decision digest: 26 of 30 open issues wait on a ruling, grouped by what breaks while they wait


hugocorreia90:
> **Read this first: what to do with each of the 28 `blocked-decision` issues (reconciled 2026-09-16).**
>
> **Ruled 2026-09-17.** Accepted as advised: rows 1–5, row 11 (#1696, with #1769 answered: only the pin list moves), all ten one-word rows, and both "not a ruling" rows. #2031 is closed and #2030 closed with it. Auto-merge is on. **Still waiting on a ruling:** rows 6–10 and 12–16 (#1884, #1596, #2020, #1994, #1856, #1928, #1900, #2044 decision 1, #1633, #2038).
>
> The digest below is unchanged, a record as of filing, and covers 25 of the 28. This block is the current answer. The comments under the digest hold the evidence.
>
> **16 rulings start work, 10 need one word, and 2 are not rulings.** Rulings are ordered by harm: silent wrong data first, then a silent wrong status, then loud failures, then the rest. Within each group, the more likely harm comes first. The ranking is of the harm, not of the proposed fix.
>
> | # | issue | advice | basis |
> |---|---|---|---|
> | 1 | #1990 | Refuse `incremental` on transformation models now. It duplicates rows on every run. | reproduced |
> | 2 | #1937 | Materialise the quarantine predicate once. Also make a failed quarantine statement fail the run, which today it does not (#2052), and drop the intermediate table on failure. | code |
> | 3 | #1996 | Refuse `ephemeral` at compile time and point at `type = "view"`. A consumer silently reads any stale table with that name. The diff converts the `07-ephemeral-cte` POC and the `06-transformation-ephemeral` IR golden fixture. Keep a guard in the local executor too. This also settles #2044's decision 2. | code |
> | 4 | #1790 | Report per table which tables the anomaly detector evaluated. Today the badge is green when the detector never ran. This is the costliest item: four subprojects, two drift gates, and a test that pins the opposite. | thread |
> | 5 | #1854 | Prevent the duplicates, not only signal them. Add a durable "flush confirmed" field to the run-progress record, a state-schema change. A later run then re-derives `MAX(ts)` from the target for any table left unconfirmed. This is safe only on adapters that commit one `INSERT … SELECT` all or nothing, which is not yet checked per adapter. | code |
> | 6 | #1884 | Option B: mark the ledger. An exit code reaches only the orchestrator. If you choose A, keep the upload. | code |
[...]

Issue #1616, opened 2026-09-03
Grace-period column drops: wire the built-but-unwired drop path behind an explicit opt-in (#1435 option A1)


hugocorreia90:
Split out of #1435. That issue removes the inert `[schema_evolution]` config surface (option A2). This one tracks the feature itself (option A1): actually dropping a column that disappeared from the source, after a grace period.

## Why this is its own issue

Wiring the drop is **data-destructive**. Rocky would start issuing `ALTER TABLE ... DROP COLUMN` against target tables in projects that set a grace period long ago and have run without drops ever since. That needs an explicit opt-in, not a new default, so it does not belong in the removal PR.

## The pieces already exist — this is smaller than it looks

Every part except the call site is built and tested. Verified on `main` at `c32f28f9`:

| Piece | Where |
|---|---|
| Grace-period detector (pure) | `engine/crates/rocky-core/src/drift.rs:253` `detect_column_drops(source, target, grace_period_days, existing_records, now)` |
| `ALTER TABLE ... DROP COLUMN` generator | `engine/crates/rocky-core/src/drift.rs:360` `generate_drop_column_sql` |
| Result type | `engine/crates/rocky-core/src/drift.rs:345` `ColumnDropResult { new_records, in_grace_period, expired, reappeared }` |
| redb table | `engine/crates/rocky-core/src/state.rs:65` `GRACE_PERIODS`, eagerly created at `:1279`, registered in `SNAPSHOT_TABLE_REGISTRY` at `:334` |
| Record type | `engine/crates/rocky-core/src/state.rs:4403` `GracePeriodRecord` |
| State accessors | `engine/crates/rocky-core/src/state.rs:4423/4436/4454/4476` `set_/get_/list_/delete_grace_period` |
| Downstream consumer | `engine/crates/rocky-core/src/auto_apply.rs:230/241` already classifies `columns_to_drop` and `grace_period_columns` |
| Tests | 7 unit (`drift.rs:774-922`) + 5 e2e (`rocky-core/tests/e2e.rs:1644-1806`), including the full state-store lifecycle |

**The persistence layer is fully built and completely unwired.** No comment on #1435 mentioned this. It means A1 is an ordinary-sized PR, not a project.

## What is actually missing

One thing: `detect_drift` (`drift.rs:28`) hard-returns

```rust
grace_period_columns: Vec::new(),
columns_to_drop: Vec::new(),
```

[...]

PR #2053, opened 2026-09-17
docs(trust): schema drift runs no grace period, and a removed column is not detected

hugocorreia90:
The Architecture of Trust page claimed a grace period that has never run. It said "a grace period runs before any destructive action", and its honesty table graded drift handling **Shipped** "with a grace period".

The grace-period drop path is built but unwired. The #2035 ruling on #1616 keeps it unwired. So the claim is false today and stays false.

## What the run path actually does

Read from `rocky-cli/src/commands/run.rs`, `process_table`, which is production code (the test module starts at line 14663):

```
  per table, target exists
        │
   detect_drift ── walks the SOURCE's columns only
        │
        ├─ new source column ──────► ALTER TABLE ADD COLUMN        (add_columns)
        ├─ every type change safe ─► ALTER COLUMN TYPE, data kept   (alter_column_types)
        ├─ any other type change ──► DROP target, full refresh      (drop_and_recreate)
        │                            same run, nothing waits first
        └─ column the source lost ─► never seen; no action
```

- `DropAndRecreate` drops the target in the run that finds the change. No grace period runs first.
- `detect_drift` iterates source columns only, and hard-returns empty `grace_period_columns` and `columns_to_drop`. A removed column is never detected.
- The page listed three responses and omitted `add_columns`, which the run applies and reports.

## The change

The section now matches [Failure modes](https://rocky-data.dev/advanced/failure-modes/), which already had this right, and links there for each action's recovery. The grade stays **Shipped**, because the graded response ships. The table row now states both limits.

No other page claims a drift grace period. The other `grace` hits are pod and container stop grace, which is unrelated.

## Verified

- Every claim read from `run.rs` and `rocky-core/src/drift.rs` at `6b24f819`.
- `npm run build` exits 0, 109 pages; the new link resolves.

Refs #1616, #2035. #1616 stays open.


hugocorreia90:
## Codex disposition

Independent adversarial review by **Codex (GPT-5.6-sol)**, branch scope against `origin/main`. Verdict: `needs-attention`, two medium findings. It confirmed the two corrections this PR exists for: no grace period runs before a drop, and a column removed from the source is not detected. It also judged the scope ("as it copies each table") and the **Shipped** grade defensible.

Both findings are **adopted**, each confirmed at the cited lines before any edit. Fixed in 4d838a97.

**[medium] The page promised ALTER and drop unconditionally. Adopted.**

With `auto_apply_additive_drift = true` under `[resilience]`, the run routes drift through a policy gate before mutating anything. I checked two things Codex's summary did not spell out:

- **The gate covers drops, not only additions.** `run.rs:13467` sets `would_mutate` for any action other than `Ignore`, or any added column. So `DropAndRecreate` reaches the gate, and is refused at `13474-13481` before the drop at `13487`.
- **A safe widening is refused too.** `auto_apply.rs:214-226` records every type change as `ColumnTypeChanged { narrowing: true }` with `Breaking` severity, because there is no typed IR at the replication boundary. The additive check (`auto_apply.rs:145`) then refuses it.

What gets through: a nullable new column, on a model with no contract, that a `[policy]` rule allows (`auto_apply.rs:140-161`). The paragraph and the honesty-table row now say "by default", and describe the gated behaviour.

**[medium] The page implied the `drift` output is a complete record. Adopted, on two pages.**

`drift.actions_taken` is pushed only in `collect_materialized_table` (`run.rs:14207-14225`), which runs on success. A `DROP` can succeed at `13493-13498` and the rebuild then fail at `13807`. `process_table` returns `Err`, and the error arm records a `TableError` with no drift entry. So a destructive change can be missing from the output a reader would check.

This mattered more on **Failure modes**, which is the recovery playbook and said the entry "records what already happened". It now says the list is complete only for tables that finished, and tells the reader to check the target itself for a failed table.

**Not verified:** no fault-injected run. Codex named the same gap. The finding is proven from source, not observed.

hugocorreia90:
## Second Codex pass, on the fix delta

**Codex (GPT-5.6-sol)** re-reviewed 4d838a97. Verdict: `needs-attention`, two medium findings. Both **adopted** after checking the cited lines. Fixed in 8d294490.

**[medium] A retry can hide the drop, and clear the failure. Adopted.** I had reached the same finding independently while the review ran, and held the edit so as not to change files under the reviewer.

`table_retries` defaults to `1` (`config.rs:574`). After a `DROP` and a failed rebuild, the retry loop (`run.rs:5536`) re-runs `process_table`. It finds no target, so no drift check runs and it full-refreshes. A successful retry reports the table as materialized and removes the error, with no drift entry. So "the run reports the table as failed" held only when the retry failed too.

I checked the worse case, a retry that prunes the table away and leaves it dropped on a green run. It cannot happen. The prune check does run before the target probe (`13373-13402`), but it prunes only when the source marker matches the recorded one. The marker is recorded only on success, and a drop means it had not matched. Reasoned from source, not run.

Both pages now say the output covers only the attempt that finished, and that a missing entry does not prove the target was left alone.

**[medium] The policy default grants auto-apply without any rule. Adopted, on three pages.**

`drift_governance.rs:154-158` evaluates as `PolicyPrincipal::Agent`. With no matching rule, `policy.rs:388-395` returns `default_agent_effect`. So `default_agent_effect = "allow"` grants the addition with no rule at all. The trust page and Failure modes said a matching rule was needed. `configuration.md` said "changing behaviour needs both this switch and a matching policy rule". All three now say the verdict comes from a matching rule or from the default.

With no `[policy]` block, the gate still runs, as `require_review`, and grants nothing (`drift_governance.rs:128-152`).

**Also removed, found while checking:** "on a model with no contract". Replication targets are always evaluated with `contracted = false` (`drift_governance.rs:112-113`, and hardcoded at `147` and `171`). The condition could never refuse anything, so it only suggested a protection that does not apply.

**Checked and left alone:** `concepts/schema-drift.md:132` says "a `[policy]` grant", which the default satisfies. It is accurate.

**Not for this PR:** the doc comment at `drift_governance.rs:106` says the gate is `None` when "no `[policy]` block exists". The code at `128-152` does the opposite and governs fail-closed. It is an engine comment, so it goes to the engine lane rather than this docs PR.

hugocorreia90:
## Third Codex pass, on 8d294490

**Codex (GPT-5.6-sol)**, delta only. Verdict: `needs-attention`, one medium finding. **Adopted.** Fixed in 64f14fff.

Codex confirmed the rest of the delta: the retry scope and default are correct, all 17 changed sentences are under 25 words, and the "only when policy resolves to `allow`" clauses on the trust page and Failure modes are accurate as necessary conditions.

**[medium] A policy `allow` is where drift auto-apply starts, not what decides it.**

The `auto_apply_additive_drift` row in `configuration.md` said the verdict "comes from" a matching rule or `default_agent_effect`. That is only the initial effect. `govern()` then tightens it, and each of these refuses a static allow. I checked each one:

| Refuses a static `allow` | Where |
|---|---|
| decision ledger untrusted or unreadable: `require_review`, fail-closed | `drift_governance.rs:208-266` |
| matching freeze: `deny`. Exhausted autonomy budget: `require_review` | `policy::autonomy_degradation` |
| decision row fails to persist: refuse | `drift_governance.rs:395-401` |

The row now says the verdict *starts* from the rule or default, and names all four.

**Consistency fix, raised in the same pass.** The `[policy]` caution said "one surface differs" from the no-block default. Opted-in drift auto-apply is a second: with no block it governs as `require_review` and refuses every drift. The caution now says two, and the seam list at the top of `[policy]` names drift auto-apply.

## Closing the review loop

Three passes found 2, 2 and 1 defects. Each was narrower than the last, and each was confirmed at its cited lines before any edit. This last delta changes one table cell and one caution, and every claim in it is checked above. Per the repo's rule that the red-team pass does not recurse, this disposition closes the loop rather than starting a fourth pass.

What none of the passes ran: a fault-injected drop, rebuild and retry. Every finding here is proven from source, not observed.

Issue #1943, opened 2026-09-12
A promote plan's statement is not bound to the names recorded beside it


hugocorreia90:
## Question for Hugo, first

**Are recomputed promote plans deliberately outside Rocky's trust model?**

That decides whether this issue is a fix or a documented boundary. `read_plan`
recomputes the blake3 of `{kind, payload}` and refuses anything that does not
hash to its own id (`plan_store.rs:688-731`), so a plan cannot be *edited*
after it is written. It can be *authored*: the digest is recomputable by
anyone who can write the file, and nothing signs it.

If writing a file into `.rocky/plans/` is already outside the threat model,
this is a boundary to write down. If it is not, the binding below is a fix.

## The gap

A promote plan records three strings per target:

```
  PromoteTargetPlan { target, source, statement }
                      ^^^^^^  ^^^^^^  the names
                                      ^^^^^^^^^ the SQL that is executed
```

`run_promote_apply` hands `statement` to `execute_statement` unchanged
(`branch.rs:857-875`). #1939 added a check at that seam, but it reads `target`
and `source` — the recorded names — not names parsed back out of `statement`.

For the case #1939 was about, a plan written by an older build, the three
agree: they all came from one `TargetRef`. So checking the names catches it.

Nothing enforces that they agree. A plan whose `statement` differs from its own
recorded names passes the check and executes.

## Why #1939 stopped there

Rebuilding the statement to compare means splitting `full_name()` back into
three parts, which is ambiguous exactly when a part contains a `.` — and a `.`
is legal in a branch name (`validate_branch_name` allows `[A-Za-z0-9_.-]`),
which is the case the scan is written to allow.

## Codex's reasoning, verbatim

From the review of #1940 (Codex, red-teamed by `gpt-5.6-terra`):

> Bind apply-time validation to the executable statement. Persist structured
> target/source components and regenerate or compare canonical SQL; handle
> legacy flattened plans explicitly.

> At apply time, the companion `target` and `source` fields are checked, not
> the names actually parsed from `statement`. The statement remains
> independent and is executed unchanged.

> What am I least confident about right now? Whether Rocky's documented "human
> vouches" model intentionally places recomputed Promote plans outside the
> threat model. The code explicitly treats the digest as recomputable and
> executes independent SQL, so a written trust-model decision would settle
> whether statement binding is mandatory.

> What's the most important thing I'm missing about this situation? A
> canonical relationship between persisted promote metadata and executable
[...]

PR #2055, opened 2026-09-17
docs(trust): .rocky/plans/ is a trusted input, and its digest is integrity, not authentication

hugocorreia90:
Rocky trusts `.rocky/plans/`, and the docs never said so. This writes the boundary down, per the ruling on #1943 in #2035: "Yes, `.rocky/plans/` is trusted. Document it."

## What a reader now learns

The statement sits in **The Architecture of Trust → Branches**, directly after the paragraph that already says the branch-approval digest is unkeyed and authenticates nobody. The plans directory has the same shape, so it is the same argument:

```
  .rocky/plans/<id>.json
        │
  read_plan: recompute blake3({kind, payload}) ── mismatch ──► refuse
        │ match
        ▼
  catches:   a file edited without re-hashing
  does not:  a plan written whole by anyone who can write the directory
```

- A promote plan runs the SQL statement it records. The apply-time checks read the table names recorded beside it, not names parsed from the SQL.
- The `AiAuthored` review marker is unsigned too.
- So anyone who can write `.rocky/plans/` can author a plan that `rocky apply` runs with the applier's warehouse credentials.
- For a promote plan, the policy gate still runs at apply, against the principal applying it. It judges the plan by the table names it records, not by the table the SQL writes.
- Rocky *tries* to keep `.rocky/` out of git. It skips that when a `.gitignore` already exists there, and ignores a failed write, so the reader is told to check their exclusions. Protect write access as you would the project.

The glossary's **Plan store** entry now says the directory is trusted and links the reasoning.

## Evidence, `main` at `6b24f819`

| Claim | Source |
|---|---|
| id is blake3 of `{kind, payload}` | `rocky-cli/src/plan_store.rs:490-500` |
| recomputed on read, mismatch refused | `plan_store.rs:708-729` |
| `AiAuthored` unsigned; any writer of the directory can produce one | `plan_store.rs:107-112` (the code's own doc comment) |
| promote executes the recorded statement unchanged | `commands/branch.rs:895` |
| guards read recorded names | `branch.rs:876-889` |
| promote apply runs the policy gate | `commands/apply.rs:3024-3078` |
| the gate's subjects are recorded targets + finding models, never `statement` | `commands/apply.rs:2343-2347` |
| `.gitignore` is best-effort: skipped if one exists, write errors discarded | `rocky-observe/src/traces.rs:225-239` |

All cited lines are production code: the test modules start at `plan_store.rs:734` and `branch.rs:1834`.

## Deliberately not claimed

- **That the policy gate runs for every plan kind.** It is visible for Run, Promote and AiAuthored. Gc, Restore, Replication and Backfill were not traced, so the sentence is scoped to promote.
- ~~Whether the policy gate evaluates the recorded names or the statement.~~ Settled during review: it reads the recorded names (see the evidence table), and the page now says so.

`npm run build` exits 0, 109 pages, and the `#branches` anchor resolves.

Refs #1943, #2035. #1943 stays open.


hugocorreia90:
## Codex disposition

Independent adversarial review by **Codex (GPT-5.6-sol)**, branch scope against `origin/main`. Verdict: `needs-attention`, one medium finding. **Adopted.** Fixed in 5611b504.

Codex confirmed the rest from source:

- No apply path reads a persisted plan without `read_plan`'s digest check.
- The `AiAuthored` marker is unsigned, as stated.
- Run plans recompile and Gc plans re-derive at apply, so those kinds cannot carry arbitrary SQL. The sentence says a writer can author *a* plan that `rocky apply` runs, which stays true through promote. Codex judged that defensible rather than an overclaim.
- No new sentence exceeds 25 words.

**[medium] "Rocky keeps `.rocky/` out of git" was a guarantee the code does not make. Adopted.**

`ensure_rocky_gitignore` (`rocky-observe/src/traces.rs:225-239`) is documented in its own comment as "best-effort ... never load-bearing":

```
  .rocky/.gitignore exists?  ── yes ──► return   (even if it ignores nothing)
          │ no
          ▼
  write "*"  ── fails ──► error discarded
```

`plans_dir` proceeds either way. So a project with an existing `.rocky/.gitignore`, or an unwritable directory, can commit plans. The page now says Rocky *tries*, names both ways that can fail, and tells the reader to check what their repository excludes. The unconditional "a plan gets no code review" is gone with it.

## What this review also settled

The PR description listed one thing as deliberately not claimed: whether the promote policy gate evaluates the recorded names or the SQL. Codex traced it, and I confirmed it:

- `gate_promote_plan` (`apply.rs:3053-3077`) gates on the set that `touched_models_for_promote` returns.
- That set (`apply.rs:2343-2347`) is the recorded `targets[].target` plus the models named by findings. It never parses `statement`.

So a promote plan whose SQL writes a table other than the one it records is judged as the recorded table. A deny rule scoped to the table actually written is never evaluated. That is the concrete meaning of the directory being trusted, so the page now says: "It judges the plan by the table names it records, not by the table the SQL writes."

**Not verified:** a runtime reproduction with an existing custom `.rocky/.gitignore`. The finding is proven from source.

hugocorreia90:
**Delta review, 5611b504: Codex (GPT-5.6-sol) `approve`, no material findings.**

It confirmed both new claims from source. The policy subjects come from the recorded targets and finding names (`apply.rs:2343-2347`). The duplicate-target and unquotable-name guards read recorded fields too, and neither parses the SQL or compares its destination with the recorded target. Execution passes `step.statement` unchanged (`branch.rs:813-895`). The `.gitignore` sentences match `traces.rs:229-239` and `plan_store.rs:503-510`.

It traced one route the page does not name. Bare `rocky branch promote` builds the names and the SQL together and executes its in-memory plan (`branch.rs:675-682`, `1374-1418`), so editing the saved file does not reach that route. That fits the page, which scopes the risk to a plan `rocky apply` reads from `.rocky/plans/`. No change needed.

Not run: a forged plan whose recorded target and SQL target differ. Proven from source only. This closes the review loop on this PR.

Issue #2056, opened 2026-09-17
bug(engine): rocky optimize assumes every model is a table, so current_strategy and the savings figure are wrong


hugocorreia90:
`rocky optimize` never reads a model's configured strategy. Every model is assessed as if it were a `table`, so `current_strategy` is wrong for any other model, and the recommendation and its savings figure are computed against the wrong starting point.

## Where

`engine/crates/rocky-cli/src/commands/optimize.rs`, building `ModelStats`:

```rust
current_strategy: "table".to_string(), // default assumption
```

`recommend_strategy` (`rocky-core/src/optimize.rs`) compares its pick against that value, so:

```text
model configured as     reported current_strategy     effect
view                    table                          can be told to switch to `view`
merge / time_interval   table                          savings compare against a table rebuild
```

The recommendation text reaches users through `rocky optimize` and `OptimizeOutput`. Nothing executes it, so no data is wrong. The numbers and the advice are.

## Fix shape

Read the strategy from the loaded model's sidecar (`StrategyConfig`) and map it to the vocabulary `recommend_strategy` uses. A model that is not in the compiled project, and only in run history, keeps an explicit "unknown" rather than an assumed `table`.

Found by the docs lane while rebuilding the `rocky optimize` reference example from the producer. Read on `main` at `6b24f819`, not reproduced.

hugocorreia90:
**Second input `rocky optimize` ignores: the project's `[cost]` block.**

`optimize_output` builds its prices with `CostConfig::default()` (`engine/crates/rocky-cli/src/commands/optimize.rs:23`), not from the loaded `rocky.toml`. Every recommendation's savings figure therefore uses the built-in rates, whatever the project configures: $0.002 per second of compute, $0.023 per GB-month, and `min_history_runs` 5.

Same fix shape as the strategy: read the value from the loaded project config, and fall back to the default only when the project sets none. Found by the docs lane while rebuilding the `rocky optimize` reference example, and checked on `main` at `6b24f819`.

hugocorreia90:
claimed by drain-issues 2026-09-22

Issue #2054, opened 2026-09-17
bug(engine): a microbatch transformation model re-inserts every row on each run, the same defect as #1990


hugocorreia90:
A `microbatch` transformation model re-inserts every source row on each `rocky run`. It is the same defect as #1990, and #1990's ruling covers only `incremental`. Once that refusal lands, `microbatch` is the one strategy left that a user can switch to and keep duplicating rows.

## Reproduce (engine 1.74.0, DuckDB, no credentials)

`rocky playground`, then `models/raw_orders.toml`:

```toml
[strategy]
type = "microbatch"
timestamp_column = "order_date"
granularity = "day"

[target]
catalog = "playground"
schema = "main"
```

```text
source rows (raw__orders.orders)   500
run 1  exit 0   main.raw_orders    500
run 2  exit 0   main.raw_orders   1000
run 3  exit 0   main.raw_orders   1500
```

## Why

- `generate_transformation_sql_with_warehouse` emits `dialect.insert_into(&target, &model_ir.sql)` for `Microbatch`, with no time filter (`engine/crates/rocky-core/src/sql_gen.rs:466-473`). The comment above it says "SQL gen emits a simple INSERT with watermark filter", which is false.
- Nothing in the run path windows a transformation model. `run.rs` groups `Microbatch` with `Incremental` and `DeleteInsert` as strategies that mutate an existing target, then runs that `INSERT` unchanged.
- The dbt importer already knows. `import/dbt.rs:1409-1413`: "Rocky's Microbatch strategy emits an append-only INSERT (sql_gen.rs), so importing it as-is silently re-inserts the lookback window every run."

`docs/src/content/docs/reference/model-format.md` documents `microbatch` as a working strategy.

## Options

```text
A. Refuse it, as #1990 does for incremental.  A compile error naming merge,
                                              delete_insert, time_interval and
                                              full_refresh. The docs stop
                                              offering microbatch for
                                              transformation models.
B. Implement windowing.                        Bound each run to its batch
                                              window, the way time_interval
                                              substitutes @start_date/@end_date.
                                              Much larger.
```

**Recommendation: A**, for the reason #1990 was ruled: every run duplicates rows today, so a refusal breaks no pipeline whose numbers were right. `time_interval` already gives bounded windows.

The PR for #1990 keeps `microbatch` legal, fixes the false comment, and keeps the existing append-path test pointed at `microbatch` so the behaviour stays pinned until this is ruled.

Refs #1990.

[...]

Issue #1990, opened 2026-09-15
bug(engine): an incremental transformation model re-inserts every row on each run; the watermark filter applies only to replication


hugocorreia90:
An `incremental` transformation model inserts every source row again on each `rocky run`. The docs promise a watermark filter, so a reader expects only new rows. What they get is silent duplicates.

## Reproduce (engine 1.74.0, DuckDB, no credentials)

```bash
rocky playground pg-incr && cd pg-incr
cat > models/raw_orders.toml <<'EOF'
[strategy]
type = "incremental"
timestamp_column = "order_date"

[target]
catalog = "playground"
schema = "main"
EOF
rocky run; rocky run; rocky run
duckdb playground.duckdb -c "select count(*) from main.raw_orders"
```

```
source rows (raw__orders.orders):    500
after run 1:  500
after run 2: 1000
after run 3: 1500      every run exits 0; run JSON has no watermarks
```

## Why

- `generate_insert_sql`, the path that adds `WHERE <ts> > <watermark>`, refuses any IR that is not the replication variant (`engine/crates/rocky-core/src/sql_gen.rs:127-134`).
- The transformation path emits `dialect.insert_into(&target, &model_ir.sql)` for `MaterializationStrategy::Incremental` (`sql_gen.rs:373-375`), which is the model's SQL with no filter.
- The code is dialect-generic, so the same should hold on every warehouse. I tested DuckDB only.

## What the docs say

- `docs/src/content/docs/reference/model-format.md`, the "Incremental" section, shows a transformation model's generated SQL with `WHERE _fivetran_synced > TIMESTAMP '...'`, "read from Rocky's state store".
- `docs/src/content/docs/concepts/incremental.md:12-17`: "Every model, replication or transformation, declares a materialization strategy", and `incremental` generates `INSERT INTO ... SELECT ... WHERE ts > watermark`.

Either the engine applies the watermark to transformation models, or it refuses (or warns about) `incremental` on a model whose SQL carries no filter, and the docs say so. Today it does neither.

Related: #1179 and #1183 (earlier duplicate-row incremental bugs, both closed). Found by the docs audit.

hugocorreia90:
## Reproduced on the dev build of `main`, and it needs a ruling before code

Ran the recipe above against a `rocky` built from `main` at `f35994a0` (1.74.0 line), DuckDB playground, no credentials:

```text
source rows (raw__orders.orders)   500
run 1  exit 0   main.raw_orders    500
[...]

PR #2057, opened 2026-09-17
docs: a transformation model cannot use incremental (E037)

hugocorreia90:
> **#2058, the engine PR for #1990, merged at `3d90b7f9`.** Its tree is identical to `c0bdb4fd`, the last head these pages were checked against. Every text this PR quotes was then re-read from `main`: 15 checks, all present, including a before-and-after check that `--watermark` is gone. This branch is rebased onto that `main`.

A transformation model can no longer use `type = "incremental"`. The docs promised a watermark filter it never applied, on some 20 sites across 19 files. This makes them say what the engine does.

## What the engine now does

```
  transformation model, type = "incremental"
        │
  rocky compile ──► E037 (error)
        │             test / ci / emit-sql fail on it
        │
  SQL generator ──► refuses too, the bootstrap CREATE and the INSERT alike
                      plan --model / estimate cannot produce SQL for it

  replication pipeline, strategy = "incremental"  ──► unchanged, watermark-filtered
```

E037, word for word from `typecheck.rs` `check_incremental_strategy`:

> model '{name}' uses `type = "incremental"`, which is not supported on transformation models: it emits an unfiltered INSERT and appends every row again on each run

It names four strategies that work: `merge`, `delete_insert`, `time_interval` and `full_refresh`. It never names `microbatch`.

## Pages

| Page | Change |
|---|---|
| `reference/model-format` | The **Incremental** section becomes the refusal: the E037 message, and a table of the four strategies. The `type` enum, `timestamp_column`, the lakehouse-format note and the retention example follow. |
| `concepts/silver-layer` | The diagram line, the strategy block, the table row and the section. The `#incremental` heading is kept so links still land. |
| `concepts/incremental` | The strategy table row now says replication only. |
| `concepts/time-interval` | It no longer advises `incremental` for append-only data, and the comparison table becomes a short explanation. |
| `reference/commands/ai`, `guides/ai-features` | `--watermark` is removed. `--materialization` takes `full_refresh`, `merge` or `ephemeral`, and the refusal message is quoted. Examples use `merge`, in the TOML shape the sidecar renderer actually emits (`toml::to_string`, single-space `=`). |
| `reference/commands/development` | The dbt `materialized` mapping, and which fallbacks reach `MIGRATION-NOTES.md`. |
| `guides/migrate-from-dbt` | The imported sidecar example, and a troubleshooting section for models that imported as `full_refresh`. |
| `concepts/compiler` | Adds an E037 row. |
| `architecture`, `architecture-of-trust`, `ROCKY_EXPLAINED` | The E-code range `E001–E036` becomes `E037`, and an E037 entry is added. |
| `reference/commands/administration` | The `rocky optimize` example, rewritten from the producer (below). |
| `concepts/content-addressed`, `glossary`, `llms.txt`, `dagster/freshness`, `dagster/partitions`, `dagster/derived-models` | Strategy lists and one example. |
| `concepts/execution-flow` | A run-output example showed a transformation model with `"strategy": "incremental"`. |
| `engine/examples/dbt-migration/README.md`, `engine/examples/ai-intent/README.md` | Recommended `type = "incremental"` and listed `--watermark`. Neither this PR nor #2058 had them at first. |

## How the importer row was checked

`rocky import-dbt` maps append-style dbt models to `full_refresh` with a warning. Where the warning lands decides what the table can say, so it was read from source, not assumed:

| dbt input | Maps to | Manifest import | `--no-manifest` import |
|---|---|---|---|
| `incremental`, no `unique_key` | `full_refresh` | Items to translate manually + Warnings | Warnings only |
| `incremental_strategy='append'`, key or not | `full_refresh` | same | Warnings only |
| `incremental_strategy='merge'`, no `unique_key` | `full_refresh` | same | Warnings only |
| unrecognised `incremental_strategy`, key or not | `full_refresh` | same | Warnings only |
| `incremental` + `unique_key`, strategy unset or `'merge'` | `merge` | no warning | no warning |
| inherited `materialized='incremental'` from `dbt_project.yml` | `full_refresh` | (not this path) | Warnings only |

Why the two columns differ: `map_incremental_strategy` is called only from `map_manifest_strategy`. The `--no-manifest` parser, `extract_dbt_config`, calls the same function but keeps only `mapping.warnings`, discarding the structured ones (`dbt.rs:2461-2467`). Why a key does not always help: `strategy_kind` takes an explicit `incremental_strategy` before it looks at `unique_key` (`dbt.rs:1202-1212`).

Sources: `import/dbt.rs` (`map_incremental_strategy`, `push_append_fallback`, `extract_dbt_config`, the project-config site at `:2064-2080`), and `import/emit.rs` for the section headings.

## Found along the way

- **`rocky optimize` could never recommend `incremental`.** `recommend_strategy` (`rocky-core/src/optimize.rs:89-196`) returns only `ephemeral`, `table` or `view`. The old example, stamped version 1.6.0, recommended `incremental` twice. It is rewritten with the real field set. Its reasoning strings match the producer's format strings, and its numbers follow from the only prices optimize can use. Two inputs are hard-coded, and the page now says both. `current_strategy` is always `"table"` (`commands/optimize.rs:86`). The prices come from `CostConfig::default()` (`:23`): $0.002 per second of compute, $0.023 per GB-month, and 5 runs minimum, whatever `[cost]` says. Both are in #2056.
- **`rocky compile` still recommends `incremental`.** `incrementality_hint` exists to suggest switching `full_refresh` models to it. It is being removed in its own engine PR, so `modeling.md:55-81` and `administration.md:341` are held for that PR's docs half. The optimize example keeps `incrementality_note` until then, because the field still exists.

## Deliberately not changed

- **`microbatch`** takes the same unfiltered append and is pending #2054. None of its docs change here.
- **Replication `incremental`**: `quickstart`, `bronze-layer`, `configuration.md`, the run-output JSON examples, and the README's POC link. It is still valid there.

## Verified

- `npm run build` exits 0, 109 pages.
- Every new anchor resolves: `model-format` `#incremental`, `#merge`, `#delete--insert`, `#time-interval`, `#full-refresh`; `silver-layer` `#incremental`.
- A final sweep for `type = "incremental"`, `"incremental"`, `--materialization incremental`, `--watermark` and `E001–E036` finds only replication contexts, deliberate mentions, and the refusal text itself.

Refs #1990, #2035, #2054, #2056. #1990 stays open until the engine PR closes it.


hugocorreia90:
## Codex disposition

**Codex (GPT-5.6-sol)**, two runs, branch scope against `origin/main`. It read the engine behaviour from #2058's branch, since `main` has none of it.

### Run 1: killed by memory pressure, no verdict

The worker died before writing a verdict, and no session file was flushed, so the run could not be resumed. Its job log named four material findings and a count of long sentences. I re-derived each one from source rather than acting on the log line. All five were confirmed and fixed in c3948e73:

| Finding | Source | Fix |
|---|---|---|
| MIGRATION-NOTES placement differs by import path | `extract_dbt_config` keeps only `mapping.warnings` (`dbt.rs:2461-2467`) | The importer table now has manifest and `--no-manifest` columns |
| A `unique_key` does not rescue an explicit `'append'` or an unrecognised strategy | `strategy_kind` takes the explicit value first (`dbt.rs:1202-1212`) | The advice says: set `incremental_strategy` to `'merge'` or leave it unset |
| The optimize example used an unreachable price | `optimize.rs:23` uses `CostConfig::default()`: $0.002/s | Numbers recomputed; the page names the fixed prices (#2056) |
| Missed transformation promises | `execution-flow.md:225`, two `engine/examples` READMEs | Fixed |
| Three sentences over 25 words | scanned | Split |

### Run 2: `needs-attention`, two findings, both adopted

Codex confirmed all five fixes above, then found two more. Both are fixed in ff2198cc.

**[high] The `full_refresh` fallback can drop history.** A manifest import keeps `compiled_code` verbatim (`dbt.rs:1005`). If dbt compiled the model against a table that already existed, `is_incremental()` was true, and the compiled SQL keeps its delta filter. `full_refresh` is `CREATE TABLE AS <model SQL>` (`sql_gen.rs:361-371`). So each run replaces the table with only the filtered rows:

```
  dbt compile (target exists)  ->  WHERE updated_at > '2026-09-01' kept in compiled_code
  import-dbt append fallback   ->  full_refresh
  every rocky run              ->  CREATE TABLE AS <filtered SQL>  =>  older rows gone
```

"Cannot duplicate rows" was true and hid this. The migrate guide and the import-dbt reference now carry a caution: read the SQL and remove the filter before the first run. The engine's own messages point users towards this path. The importer's refusal says to "compile dbt in an incremental context", and the fallback warning says the model "rebuilds in full". Both are reported to the engine PR.

**[medium] `emit-sql` pages still promised an `INSERT` for `incremental`.** `emit-sql` bails on `has_errors` before any `--model` filter (`emit_sql.rs:146-158`), so one E037 model stops the whole export. `modeling.md`, `cli.md` and `no-lock-in.md` now say so. The steady-state note applies to `Incremental | Merge | DeleteInsert` (`emit_sql.rs:86-92`), so the prose names `merge` and `delete_insert`. The quoted `-- NOTE:` text is kept exact, although it is now stale. That is reported upstream too. The `dagster/derived-models` diagram's `incremental` example is fixed as well.

### Added in the same commit, verified at #2058 `236abf0f`

`rocky run` records an E037 model as a failed table, and leaves its existing table alone. So an upgrading user's table keeps any rows earlier runs appended again. By default a dependent still builds from it. #2058's run test pins this. `contain_failures = true` seeds compile failures into the containment ledger (`run.rs:10272`, built only under that flag), so dependents are held back. The page now says both, and advises a `full_refresh` rebuild.

**Not run:** any warehouse execution. Every finding is proven from source. This PR stays a draft until #2058 merges; every quoted text is then re-read from the merged code.

hugocorreia90:
## Final review round: same-model, NOT independent

**Disclosure first.** The last five commits (after `98319909`) were not reviewed by an independent model. Two Codex runs on this delta were killed by memory pressure on the host, the second within 20 seconds and before any output. So the review below is by **Fable, a same-model reviewer**, instructed to act adversarially and to stay read-only. It is a fallback, not a substitute for Codex. The earlier commits did get two Codex rounds; those dispositions are above.

Before this round, every quoted engine text was re-read from `main` after #2058 merged at `3d90b7f9`: 15 checks, all present. #2058's merged tree is identical to `c0bdb4fd`, the last head verified.

### Findings, all confirmed at source and fixed in 96ae689d

**[medium] The "import a manifest" way out no longer works.** The raw-path refusal message still offers two ways out. The first is "compile dbt in an incremental context and import that manifest". Since #2058 that is closed for an unkeyed model: the manifest guard refuses it (`dbt.rs`, `INCREMENTAL_FALLBACK_REFUSED`). For a keyed model it is hazardous, because the model imports as `merge` with the filter kept (#2059). The dbt-migration README presented it as working. So did the migrate guide's `is_incremental()` bullet. That second site was outside the reviewed diff; I found it by grepping the phrase. Both now say to rewrite by hand. `RAW_INCREMENTAL_ERROR` still offers the old advice, deliberately left unchanged in #2058, and the pages say so.

**[low] `contain_failures` is fail-closed, not only downstream.** With a failure in the run, the ledger also holds back any model whose reads are not provably complete (`containment.rs:416-421`, `fail_closed: true`). The page said "everything downstream", which understated the blast radius.

**[low] Which #2059 outcome is common.** The reviewer flagged "the run fails instead" as supported by reasoning, not by a test. #2059's body settles it and shows both outcomes. The common dbt filter, `MAX(...) FROM {{ this }}`, reads a table that does not exist yet, so the first run fails. A literal cutoff loads only recent rows. The cautions now lead with the common case. They also name `delete_insert`, which bootstraps through the same CTAS, where "the CTAS *is* the first load" (`run.rs:12286-12291`).

### Checked and holds (reviewer's list, spot-checked)

- The manifest guard covers every fallback path: append, merge without a key, an unrecognised strategy, microbatch without a key or without `event_time`. It also covers both spellings of microbatch.
- `merge` bootstraps with `generate_transformation_initial_ddl` on the run path. There is no skip.
- `emit-sql` bails on `has_errors` before `--model` filters. The quoted `-- NOTE:` literal matches byte for byte.
- `rocky run` records E037 in `errors` as `CompileError` and bumps `tables_failed`. The old table stays untouched. A dependent builds from it by default, and fails with none.
- No added sentence exceeds 25 words. One 30-word pre-existing sentence on an edited line was split.

**Not run:** a warehouse execution of any of these paths, and a real `dbt compile` to see which filter form is common in practice. #2059's body is the source for the latter.

PR #2058, opened 2026-09-17
fix(engine): refuse incremental on transformation models (E037)

### LINKED ISSUE #1990: bug(engine): an incremental transformation model re-inserts every row on each run; the watermark filter applies only to replication

hugocorreia90:
Closes #1990.

`type = "incremental"` on a transformation model is now a compile error, E037. It never applied a watermark there, so every run after the first appended the whole result again and exited 0. The tools that produced the strategy stop producing it.

## What changes

```text
rocky compile / test / ci / emit-sql   E037 error on the model; the command fails
rocky run                              skips the model and counts it as failed
                                       (error diagnostics already exclude a model)
SQL generator                          refuses the transformation INSERT, so
                                       `rocky plan` leaves the model out and
                                       `rocky estimate` lists it as skipped
rocky import-dbt                       append-style dbt models -> full_refresh + warning
                                       (5 sites that used to emit incremental)
rocky ai                               --materialization incremental is refused;
                                       --watermark is removed
VS Code                                strategy-incremental snippet and the
                                       sidecar schema's incremental variant removed
```

Replication pipelines keep `incremental`. Their copy goes through `generate_insert_sql`, which applies the watermark, and never reaches the new check.

The E037 message names `merge` with a `unique_key`, `delete_insert`, `time_interval` and `full_refresh`. It does not name `microbatch`.

## Deliberately not in this PR

- **`microbatch`.** It takes the same unfiltered path and duplicates rows the same way, reproduced on 1.74.0 (500, 1000, 1500). The ruling covers `incremental` only, so it is filed as #2054 with a recommendation. This PR fixes the generator comment that claimed microbatch applies a watermark filter. The run-path test that pins append behaviour now uses `microbatch`, and its doc says the second-run assertion pins a defect until #2054 is ruled.
- **`rocky compile`'s `incrementality_hint` and `rocky optimize`'s pointer to it.** Both still steer a `full_refresh` model toward `incremental`. They are removed in a follow-up PR that opens right after this one: it is a JSON schema change with its own codegen and fixture regeneration, and POC `06-developer-experience/09-sql-types-blast-radius` is built on the hint and needs a rework.
- **Docs pages.** The docs lane holds them and has a PR ready to merge after this one: `concepts/incremental.md`, `reference/model-format.md`, `concepts/time-interval.md`, `concepts/silver-layer.md`, `dagster/freshness.md`, `guides/ai-features.md`, `guides/migrate-from-dbt.md`, `reference/commands/ai.md`, and the E-code table in `concepts/compiler.md`.

## Evidence

Run locally on this branch (`DEVELOPER_DIR` workaround, two build jobs):

```text
cargo fmt --check                                              clean
cargo clippy --workspace --all-targets --all-features -D warnings   clean
cargo test -p rocky-compiler -p rocky-core -p rocky-ai        2930 passed, 0 failed
cargo test -p rocky-compiler, then
  -p rocky-cli --all-features --lib --test import_dbt_emit     2580 passed, 0 failed (final commit)
```

Mutation checks, each restored afterwards:

```text
remove the E037 call from typecheck           an_incremental_transformation_model_is_refused_with_e037 FAILS
widen E037 to also refuse microbatch          e037_does_not_fire_for_other_strategies FAILS
restore the generator's INSERT arm            test_transformation_incremental_is_refused FAILS
```

The microbatch duplication in #2054 was reproduced on engine 1.74.0: 500, 1000, 1500 rows over three runs, every run exit 0.

## Review

Independent review: Codex (`gpt-5.6-terra`), three rounds. Every finding and its disposition is in https://github.com/rocky-data/rocky/pull/2058#issuecomment-5714302347. Two findings are existing behaviour and are disclosed, not changed: a dependent of an excluded model builds from its old table by default, and a keyed dbt incremental import can load only recent rows on its first run (#2059).


hugocorreia90:
## Independent review: Codex (`gpt-5.6-terra`), and what changed

Codex reviewed `6b24f819..d4390950` and returned **BLOCK** with five findings. Three are fixed in ff79dea9. Two are pre-existing behaviour that this PR routes more models through, and are disclosed rather than changed.

| # | finding | disposition |
|---|---|---|
| 1 | With E037 on an upstream model, `rocky run` excludes only that model, and a downstream model still runs against the upstream's old table. | **Not changed; disclosed.** This is the existing policy for every error diagnostic (see the #1291 comment above the `has_errors` block in `run.rs`): the failing model is excluded and counted in `tables_failed`, so the run reports a partial failure and exits non-zero. Downstream containment is opt-in (`config.rs:1203`). What this PR changes is how many projects reach that path: every project with an `incremental` transformation model does, on upgrade. Before this PR the same downstream read a table that had duplicated rows on every run. |
| 2 | `generate_transformation_initial_ddl` still built the first-run CTAS for an `incremental` model, so the "SQL generator refuses it" claim was half true. | **Fixed.** Both generators now return the same E037 refusal from one helper. `test_transformation_incremental_is_refused` asserts both. The two lakehouse DDL tests that used `incremental` as their example strategy now use `microbatch`. |
| 3 | Editing a sidecar `.toml` does not re-run the LSP, so the vscode CHANGELOG overclaimed that the editor "validates" the strategy. | **Wording fixed; the LSP gap is pre-existing and unchanged.** The entry now says what this PR does: the snippet is gone and the `*.rocky.toml` JSON schema reports `type = "incremental"` as invalid. |
| 4 | The `rocky-config` skill's sidecar template still listed `incremental` as a valid `type`. | **Fixed** in both mirrors, which stay byte-identical. |
| 5 | `import-dbt --no-manifest` emits `full_refresh` correctly but drops the structured warning, so MIGRATION-NOTES.md lists the fallback only under Warnings. Separately, `rocky estimate` prints skipped models only with `--verbose`. | **Not changed; disclosed.** The regex path drops structured warnings by design (the doc comment above `extract_dbt_config`), for `ephemeral` too. The docs lane has the exact placement. **The `estimate` CHANGELOG claim is corrected:** it skips the model, and `--verbose` prints why. |

Codex found no replication regression. E037 is scoped to transformation `Model`s, and replication `incremental` goes through `generate_insert_sql`. It also judged the `microbatch` test substitutions not vacuous: they deliberately keep the #2054 behaviour pinned.

**What I am least confident about.** Finding 1. The partial-failure exit makes the stale downstream loud, and it is existing policy. But a first upgrade will put it in front of every affected project at once, and nobody chose that consequence explicitly.

**Added after the review (236abf0f).** Codex's "most important missing" item was an end-to-end run test with an E037 model and a dependent. `an_e037_model_is_excluded_from_run_and_its_dependent_follows_policy` drives `execute_models` twice. With no upstream table, the E037 model is reported, counted as failed, never materializes, and the dependent cannot build. With an old upstream table, that table keeps its one row and the dependent builds from it, which is the policy in finding 1, pinned on purpose. Removing the E037 call makes the test fail. The same commit fixes the append fallback's suggestion: an explicit `incremental_strategy='append'` wins over `unique_key`, so the advice now says to set the strategy to `'merge'` or leave it unset.

## Rounds 2 and 3 (delta reviews), and a finding from the docs lane

| round | finding | disposition |
|---|---|---|
| 2 | `generate_time_interval_bootstrap_sql` could still CTAS an `incremental` IR | **Fixed** (544cdcf9). All three transformation generators return the same refusal, and the refusal test covers all three. |
| 2 | The new run test ignored `execute_models`' Result; with no upstream table it asserted only that the dependent did not materialize | **Fixed.** I ran it first to see the real outcome. It now asserts `Err` containing `model 'down' failed` and `does not exist` when the upstream table is missing, and `Ok` when an old one exists. |
| 2 | The vscode CHANGELOG still overclaimed: the schema is registered for JSON validation only and never checks `.toml` sidecars | **Fixed.** The entry says exactly that, and that `rocky compile` reports E037. |
| 2 | The skill comment read as a complete list of strategies | **Fixed.** It now reads as examples. |
| docs lane | **HIGH.** An unkeyed dbt incremental model whose SQL uses `is_incremental()` fell back to `full_refresh`. Its compiled SQL can keep dbt's delta filter, so every run would replace the table with only recent rows. | **Fixed** (8f7d51d8). The importer refuses that model outright, with no leftover mapping warning. Tests cover the refusal, and a mutation that disables the guard fails it. The emit-sql note no longer mentions a watermark (4d1ff698). |
| 3 | The refusal message told users to add a `unique_key` so the model maps to merge. Merge's first run builds the table from the same compiled delta SQL, so it loads only recent rows, or fails when the filter reads the absent target. | **Message fixed** (c0bdb4fd): it now says to remove the `is_incremental()` filter by hand. **The keyed path itself predates this PR and is not changed here**; it is filed as #2059 with options. The test that imports the keyed case as merge now says it pins current behaviour, not a contract. |

Codex's round 3 verdict was BLOCK on that keyed path. That path was already on `main` before this PR, so fixing it here would widen this PR past #1990's ruling. The one part this PR introduced, the advice pointing at it, is fixed. c0bdb4fd changes only message text, a doc comment and a test assertion, so it was not sent for a fourth review.

PR #2061, opened 2026-09-17
refactor(engine): remove the compile incrementality hint that recommends a refused strategy

hugocorreia90:
`rocky compile` no longer emits `incrementality_hint`, and `rocky optimize` no longer points to it. The hint recommended switching a `full_refresh` model to incremental materialization. Since #2058, `type = "incremental"` on a transformation model is compile error E037, so `rocky compile` was recommending what it refuses.

Follow-up to #1990 / #2058.

## What changes

```text
rocky-compiler   incrementality.rs removed (the detector and IncrementalityHint)
rocky-cli        ModelDetail.incrementality_hint removed; OptimizeOutput.incrementality_note
                 removed; the "Tip: Run rocky compile ..." line leaves rocky optimize
codegen          schemas/compile + optimize, SDK Pydantic, TypeScript bindings, openapi.json
SDK              hand-written ModelDetail and OptimizeResult lose the two fields
dagster          fixtures regenerated: the diff removes only the hint and note keys
POC 06/09        shows seed type grounding through a data contract instead (below)
docs (repo)      ROCKY_EXPLAINED.md optimize example; AGENT_REVIEW.md note on the removed module
```

**Breaking for JSON readers:** both fields are gone from the schemas, the Python SDK and the TypeScript bindings. Engine and SDK CHANGELOG entries say so.

## Why remove it, not repoint it

The detector picked monotonic-looking columns: timestamps, and also integer ID columns such as `order_id`. After E037 there is no working strategy that takes "a watermark column". `time_interval` needs a time column plus `@start_date`/`@end_date` in the SQL, and an ID column is not one. Repointing the hint would recommend something that cannot be applied as stated.

## POC 06/09

The POC's point is that `--with-seed` turns `Unknown` source columns into real types. Its only visible evidence was the hint's confidence rising from `medium` to `high`. It now uses a typed data contract on `orders_typed`:

```text
without --with-seed   I003 for order_id and amount: the contract type check cannot run
with    --with-seed   no I003: the columns resolve to Int32 / Decimal and the check passes
```

`run.sh` fails if that difference does not hold. It ran end to end on engine 1.74.0.

## Evidence

```text
cargo fmt --check                                             clean
cargo clippy -p rocky-compiler -p rocky-cli --all-targets --all-features -D warnings   clean
cargo test -p rocky-compiler -p rocky-cli --all-features --lib  2540 passed, 0 failed
  (one unrelated test skipped locally, a_stuck_config_read_is_bounded_refused_and_recovers:
   three runs that included it were killed by the host running low on memory; CI runs it)
uv run pytest (sdk/python)                                     270 passed
uv run pytest (integrations/dagster)                           776 passed
just codegen && just regen-fixtures                            clean; fixture diff is removals only
```

## Docs

The docs lane holds the two pages that describe the field (`reference/commands/modeling.md`, `reference/commands/administration.md`) and will follow this PR.

## Review

Independent review: Codex (`gpt-5.6-terra`). Findings and dispositions are in the review comment below; the POC assertion gap it found is fixed in 501faa69.


hugocorreia90:
## Independent review: Codex (`gpt-5.6-terra`), and what changed

Codex returned **BLOCK** with three findings. It confirmed there is no remaining in-tree consumer of the removed fields. It also found the schemas, Pydantic models, TypeScript bindings, OpenAPI document and fixtures consistent with the Rust structs, and the engine and SDK CHANGELOG claims true.

| finding | disposition |
|---|---|
| `reference/commands/modeling.md` (the `incrementality_hint` example, the field list and the `--with-seed` sentence) and `reference/commands/administration.md:341` still describe the removed fields | **Handled in the docs lane's PRs, by design.** The docs half of this change is a draft stacked on #2057. It covers both pages, including the `--with-seed` sentence, and merges after this PR. |
| `reference/commands/ai.md:32-33` still offers `incremental` and `--watermark` | **Belongs to #1990, not this PR.** #2057, the docs PR for #2058, covers that flag table. |
| POC 09's `run.sh` accepted any non-empty I003 set and only printed the I001/P002 result, so it could regress and still print "POC complete" | **Fixed** (501faa69). It now requires I003 for exactly `amount` and `order_id` without seed data and none with it, exactly one I001 on `orders_star` with a span on `models/orders_star.sql`, and no P002. Removing the `amount` column from the contract makes `run.sh` exit 1 with the expected message. |

501faa69 changes only POC assertions, so it was not sent for a second review.

Issue #2052, opened 2026-09-16
bug(engine): a quarantine statement that fails at execution does not fail the run


hugocorreia90:
A quarantine statement that fails at execution does not fail the run. The run can report `Success` and exit 0 with the split half-applied.

## What happens

`execute_quarantine_plan` stops at the first failing statement and returns `QuarantineOutput { ok: false, error: Some(..) }` (`engine/crates/rocky-cli/src/commands/run_local.rs:1203-1214`). The caller pushes that into `output.quarantine` (`:771`) and carries on.

Nothing that decides the outcome reads it:

```text
check_failures_by_severity   output.rs:5374   counts check_results only
finalize_quality_status      run_local.rs:825 takes that error count
derive_run_status            output.rs:5417   interrupted, tables_failed,
                                              check_gate_failed, verify_after_failed
```

So `ok: false` reaches the JSON and the status line, and the run status and exit code ignore it.

## The fix already exists 20 lines up

When the quarantine SQL fails to **compile**, the same function records a failing, not-evaluated check named `quarantine:compile` (`run_local.rs:790-808`). Its comment gives the reason: "A quarantine that did not run is NOT a quarantine that found nothing… the gate trips, the run exits non-zero." The **execute** failure needs the same treatment, for example a `quarantine:execute` check carrying the statement role and the error.

A partial failure can also leave the intermediate or quarantine table behind, because nothing cleans up after the early return. That half belongs with #1937's ruling, which changes how many statements the plan runs.

## Evidence level

Read against `main` at `6b24f819`, not reproduced. A reproduction needs a quarantine statement that fails on DuckDB, for example a view already sitting at the valid table's name.

Found while checking #1937's recommendation. Refs #1937.

Issue #1937, opened 2026-09-12
A quarantine predicate is evaluated twice, so a row can land in both the quarantine and valid tables


hugocorreia90:
A `[quarantine]` predicate is evaluated in **two separate statements**, so two evaluations of the same expression over the same row can disagree. When they do, one row lands in **both** the quarantine table and the valid table.

Found by the independent review of #1922 and confirmed against the code.

## Where

`compile_quarantine_sql` emits the quarantine CTAS and the valid CTAS as two statements (`rocky-core/src/quarantine.rs` ~477-513), and the runner executes them sequentially (`rocky-cli/src/commands/run_local.rs` ~1166-1207):

```
  statement 1   CREATE ... __quarantine AS SELECT ... WHERE NOT (<pred>)
  statement 2   CREATE ... __valid      AS SELECT ... WHERE     (<pred>)
                                                             ^ evaluated again
```

Split mode is the affected one. `drop` emits only the valid CTAS and `tag` only rewrites the source, so neither re-evaluates.

## Why it matters

With `expression = "created_at <= now()"`, a row whose `created_at` sits between the two statements' clock reads is:

- **invalid** when statement 1 runs → written to `__quarantine`
- **valid** when statement 2 runs → written to `__valid`

The row is duplicated across the split, which is precisely what the split exists to prevent. The inverse ordering loses the row from both.

Nothing detects this. The two statements each succeed, the counts look plausible, and the run exits 0.

## What is not claimed

Whether the `CASE WHEN` and the `WHERE` **inside statement 1** can disagree was investigated and **not** established — clock expressions are ordinarily statement-stable on the four supported engines. The confirmed problem is the second statement, not the two uses within the first.

## Current state after #1922

#1922 refuses a volatile function in a quarantine predicate (`ExpressionUse::ReevaluatedPredicate`), so a config that would reach this is now rejected at generation time with a message naming the function.

That closes the door for new configs and for existing ones — an affected project now gets a refusal instead of silent double-counting, which is the right direction but is a **refusal, not a fix**. The underlying double-evaluation is untouched, and any future non-volatile source of disagreement would reach it again.

## Options, none chosen

1. **Materialise the predicate once.** Compute the label column into a temp or CTE, then derive both tables from it. Removes the second evaluation entirely; costs an extra materialisation.
2. **Derive `__valid` from the source minus `__quarantine`** by key, rather than re-evaluating. Needs a key, which quarantine does not currently require.
[...]

PR #2067, opened 2026-09-17
docs(quarantine): split evaluates once, accepts a clock, is refused on Trino, and a failed quarantine fails the run

hugocorreia90:
> **Draft on purpose. Do not merge before #2066.** Merged first, these pages would describe a `split` that the released engine does not run. Every claim was read from #2066's branch `fix/quarantine-predicate-once` at `088287c2`, and it is re-read from the merged code before this leaves draft.

#2066 fixes quarantine `split` (#1937) and makes a failed quarantine fail the run (#2052). Three pages still describe the old behaviour.

## What changes

```
  before                                     after (#2066)
  ──────                                     ─────────────
  split evaluates the predicate twice        once, through _quarantine_labels_<token>
    → a clock is refused                       → a clock is accepted, as in tag and drop
  split runs on every warehouse              refused on Trino (no SELECT * EXCEPT form)
  a failed quarantine could exit 0           fails the run, whatever fail_on_error says
  one reserved name: quarantine:compile      plus quarantine:execute
```

| Page | Change |
|---|---|
| `concepts/data-quality-checks.md` | Drops the split clock rule, and the second split refusal (error-severity `not_in_future` / `older_than_n_days`). The refusal row now says the run fails. The mode table says each row lands in exactly one output, and that `split` is not available on Trino. New paragraphs cover the label table, the `EXCLUDE`/`EXCEPT` dialect split, and failure accounting. |
| `reference/configuration.md` | Five reserved check names, not four. The `split` row gains exactly-one and Trino. "The last two work in `tag` and `drop` only" is replaced by the lifted refusal and the fail-the-run rule. |
| `reference/json-output.md` | Adds `quarantine:execute` to the check names, and says a failing quarantine check fails the run regardless of `fail_on_error`, with its `tables_failed` / `errors` accounting. |

## Verified from source

| Claim | Where |
|---|---|
| The split clock refusal is removed | `reject_generated_clock_in_split`: 3 references on `main`, 0 on the branch (`rocky-core/src/quarantine.rs`). `rocky-sql/src/check_expression.rs:270`: "Only quarantine `split` used it". |
| Trino has no form, so it refuses | `SqlDialect::star_excluding` defaults to `None` (`rocky-core/src/traits.rs`). DuckDB returns `* EXCLUDE (...)`. |
| `quarantine:execute` is reserved, in any spelling | `rocky-core/src/config.rs:3570-3573` (`RESERVED_CHECK_NAMES`), sanitized before comparison at `:3691`. Its test at `:8824-8826` pins `quarantine:execute`, `quarantine_execute` and `quarantine.execute`. `checks.rs` only builds the check result. |
| Label table, UUID token, exactly-one, fail-the-run, `failure_kind` accounting | #2066's `engine/CHANGELOG.md` entries, which drain-issues named as the source of truth. |

The sentence-length scan is clean, and `npm run build` exits 0. The site list came from sweeping `main` for the phrases, not only from the three pages named in the request.

Refs #2066, #1937, #2052.


hugocorreia90:
## Codex disposition

Independent adversarial review by **Codex (GPT-5.6-sol)**, reading #2066's branch at `088287c2`. Verdict: `needs-attention`, two medium findings and two notes. It reproduced its naming counterexamples in DuckDB. It confirmed the rest: clock acceptance, compile refusal writing nothing, the `tables_failed` / `errors` accounting, and the reserved spellings. All four points are **adopted**, each confirmed at source. Fixed in 33d0ac42.

**[medium] "Each row lands in exactly one output" was unconditional.** `suffixed_table_name` (`quarantine.rs:738-743`) only validates the identifier. Nothing checks that the two suffixes differ or are non-empty:

```
  suffix_valid = ""                          ->  valid table name == source table  ->  the valid CTAS replaces the source
  suffix_valid == suffix_quarantine          ->  one table name                    ->  the valid CTAS replaces the quarantine rows
```

The guarantee now reads "when the run completes and the two suffixes differ". Both suffix rows in `configuration.md` say to keep them non-empty and different, and that Rocky does not check. Whether #2066 should refuse such a config is raised with the engine lane.

**[medium] "The valid table was not rewritten" held only for a refused plan.** `split` runs its statements in order (`quarantine.rs:285-310`):

```
  label CTAS  ->  __quarantine CTAS  ->  __valid CTAS  ->  DROP label table
```

An execution failure keeps whatever earlier statements wrote. A failed drop leaves both outputs new and the label table behind. A statement that timed out may still have committed. The page now separates the two cases. A **refused** plan writes nothing, so downstream reads the previous run's rows. A **failed** one may have written outputs.

The same sentence came from #2066's own CHANGELOG ("here the valid table was not rewritten"). That is also raised with the engine lane.

**Note: not only Trino.** `star_excluding` defaults to `None` (`traits.rs`), and `split` refuses on `None` (`quarantine.rs:269-273`). So any adapter whose dialect does not provide the form is refused. Both pages now say so.

**Note: two stale sites.** `model-format.md:386` still promised "the two extra rules", including split mode. And my own first commit removed the split row from `data-quality-checks.md`'s table, but left its intro saying "Two positions add a rule". Both fixed.

**Not run:** a warehouse fault-injection test. Codex named the same gap.

hugocorreia90:
## Second Codex pass, and re-verification against merged `main`

**#2066 merged at `c42b6be8`**, and its tree is identical to `41030ca5`, the head these pages were checked against. This branch is rebased onto that `main`.

**Every stated fact re-read from `main`:** 15 checks, all present. The one "this is gone" check discriminates: `reject_generated_clock_in_split` appears once at `7187a56b` and zero times now.

### Codex (GPT-5.6-sol), delta `03a5f1a7..b0fbcde6`: `needs-attention`, two medium findings. Both adopted, fixed in 7fa0b7d2.

**[medium] The exactly-one guarantee lost its condition too early.** When #2066 added `refuse_colliding_names` I removed the qualifier. But that check compares names exactly (`quarantine.rs:776-786`), so case-only aliases pass it:

```
  suffix_valid = "__out",  suffix_quarantine = "__OUT"
        │ exact comparison: different, so accepted
        ▼
  warehouse that ignores case in a table name  ->  one table
        ->  the valid write replaces the quarantine rows  ->  failing rows in neither output
```

I verified the warehouse half rather than assuming it. In an in-memory DuckDB, creating `"orders__out"` and then `"orders__OUT"` fails with `Table with name "orders__OUT" already exists!`. Databricks is documented as case-insensitive for table names; I could not run it here. Both guarantees now require the two suffixes to name two different warehouse tables, and the `suffix_quarantine` row names the consequence.

**[medium] "A failure keeps what earlier statements wrote" included the label table.** It does not. `run_local.rs` runs the drop after the statement loop, on the failure path too, and `quality_check_gate.rs:936-946` asserts `orders__quarantine` survives while no `_quarantine_labels_` table remains. The page now says completed statements keep their output, that `__quarantine` can be new while `__valid` is the previous run's, and that the label table is normally dropped. A failed drop is the exception.

### Also fixed, pre-existing, from the same review

`reference/glossary.md` said quarantine means "bad data does not block the run or reach downstream readers". Both halves were wrong. Error-severity checks still fail the run under the default `fail_on_error = true`, and `tag` leaves the failing rows in the table. The entry now says so.

### Confirmed by the same pass

- The collision pairs per mode: `split` checks valid/source, quarantine/source and valid/quarantine; `drop` checks valid/source; `tag` checks none.
- `json-output.md:311` is right: with `fail_on_error = false` the check gate stays false, and `tables_failed` is what fails the run.
- No newly written sentence exceeds 25 words.

**Review loop closed here.** Two passes, four findings, each confirmed at source before editing. Per the repo's rule, the red-team pass does not recurse; this disposition closes it. **Not run:** live Databricks, and no warehouse fault injection.

Issue #1996, opened 2026-09-15
bug(engine): an ephemeral model is never inlined, so every model that reads it fails


hugocorreia90:
An `ephemeral` model is never inlined. Any model that reads it fails at run time. The docs, and `rocky emit-sql`'s own message, say it becomes a CTE.

## Reproduce (engine 1.74.0, DuckDB)

```bash
rocky playground pg-eph && cd pg-eph
printf '[strategy]\ntype = "ephemeral"\n\n[target]\ncatalog = "playground"\nschema = "main"\n' > models/raw_orders.toml
rm playground.duckdb && duckdb playground.duckdb < data/seed.sql
rocky --output json run     # exit 2, status PartialFailure
```

```
model 'customer_orders' failed: DuckDB error: Catalog Error:
  Table with name raw_orders does not exist!
```

`rocky emit-sql` prints `customer_orders` with a bare `FROM raw_orders` and no CTE. It then reports `raw_orders (ephemeral — inlined as a CTE)`.

## Why

- `sql_gen::generate_transformation_sql` emits no statement for `Ephemeral`.
- Nothing rewrites a consumer's `FROM <ephemeral>` into a CTE. `engine/crates/rocky-cli/src/commands/run.rs:8563-8577` says so itself ("No such inlining exists"), and fails closed only for shadow and branch runs.
- A consumer therefore reads whatever physical table carries that name. That is an error when none exists, and a silent read of a stale or unrelated table when one does.

## Claims that say otherwise

- `docs/src/content/docs/reference/model-format.md` (Ephemeral section): the body "is injected as a `WITH … AS (…)` CTE wherever the model is referenced".
- `docs/src/content/docs/concepts/incremental.md:21`: "No table; inlined as CTE in downstream models".
- `README.md` "You can leave": "Rocky inlines an ephemeral model as a CTE".
- `engine/crates/rocky-cli/src/commands/emit_sql.rs:30, 246-251`: the skipped-model message.
- `examples/playground/pocs/02-performance/07-ephemeral-cte` promises "CTE Inlining". Its `run.sh` never runs a model: it runs `validate`, `compile`, and `rocky test ... || true`, so the POC stays green.

Either implement the inlining, or refuse `ephemeral` when a model has a consumer, and correct the claims.

Related: #1630 (an ephemeral model's phantom edge). Reported by a docs walkthrough, and reproduced.

hugocorreia90:
[...]

PR #2072, opened 2026-09-17
docs: type = "ephemeral" is refused (E038), and the pages stop promising CTE inlining

hugocorreia90:
> **Draft on purpose. Do not merge before #2069.** Merged first, these pages would describe a refusal the released engine does not make. Every claim was read from #2069's branch `fix/refuse-ephemeral` at `d370cb3f`, and is re-read from the merged code before this leaves draft.

#2069 makes `type = "ephemeral"` a compile error, `E038`. The docs promised CTE inlining that never happened.

## What the engine did, and does

```
  before:  a consumer's FROM <ephemeral model>  ->  nothing rewrote it
                                                     the model was never materialized either
                                                     so the consumer read whatever table carried the name
                                                     (an error when none existed; an unrelated table when one did)

  now:     rocky compile  ->  E038, naming `type = "view"`
```

E038's message and its suggestion are quoted on the model-format page, word for word from `typecheck.rs`. The suggestion names two replacements: `view` for an intermediate several models read, and an earlier step of a `.rocky` model for one only a single model reads.

## Pages

| Page | Change |
|---|---|
| `reference/model-format` | The **Ephemeral** section is the refusal, the E038 message, and a table of the two replacements. The `type` list offers `"view"` in its place. |
| `reference/glossary` | The **Ephemeral** entry is the refusal. The **CTE** entry no longer says an ephemeral model becomes one: Rocky writes a CTE when it lowers a `.rocky` model's steps. |
| `reference/commands/modeling`, `reference/cli`, `guides/no-lock-in`, `README` | `emit-sql` has no ephemeral skip branch any more. The sample skip report no longer shows `dim_session (ephemeral — inlined as a CTE)`, a string the engine deleted. |
| `reference/commands/core-pipeline` | `rocky plan` gains `skipped[{model, reason}]`, with the generator's own error as the reason. The shadow-mode bullet now says an ephemeral model fails compile first. |
| `ROCKY_EXPLAINED` | `rocky optimize` recommends `view` where it recommended `ephemeral`: the decision tree, the sample output and the feature table. Its caveat now points at #2056. |
| `concepts/incremental`, `concepts/architecture`, `dagster/partitions`, `llms.txt.tmpl` | Strategy lists. |
| `getting-started/comparison`, `benchmarks/FEATURE_COMPARISON` | The **Ephemeral (CTE)** row is removed from both. |
| `examples/playground` READMEs, three POC READMEs | The catalogue text, and three lists that named the strategy. |

## Verified from #2069's branch

| Claim | Where |
|---|---|
| E038's message and suggestion | `rocky-compiler/src/typecheck.rs` `check_ephemeral_strategy` |
| `rocky plan` skip reason is the generator's error | `commands/plan.rs:1123-1126` with `sql_gen.rs:665-671` |
| The ephemeral skip string is gone from `emit-sql` | `commands/emit_sql.rs`, branch deleted |
| `optimize` recommends `view` now | `rocky-core/src/optimize.rs`, with its renamed tests |
| `rocky ai` refuses `ephemeral` | `rocky-ai/src/sidecar.rs` `EphemeralRefused` |

## Deliberately not here

- **`reference/commands/ai`'s `--materialization` list.** #2057 edits the same row. It lands after #2057 merges, so the two do not conflict.
- **`concepts/shadow-mode`** already says Rocky neither materializes nor inlines an ephemeral model.
- **The old `07-ephemeral-cte` POC**, which #2069 replaces with `07-view-intermediate`.

`npm run build` exits 0. The sentence scan is clean. The site list came from sweeping `main` for the word, not only from the pages named in the request.

Refs #2069, #1996.


hugocorreia90:
## Review disposition: same-model, NOT independent

**Disclosure first.** Codex's background job on this PR was killed by host memory pressure, 14 minutes in, before it wrote a verdict. Its log had reached one concrete finding, which I recovered and fixed. The rest of the review is **Fable, a same-model reviewer**, told to act adversarially and to stay read-only. It is a fallback, not a substitute.

**Before the review**, every quoted text was re-read from `main` after #2069 merged at `d04ac4cd`: 19 checks, all present, including that the ephemeral skip string is gone from `emit-sql` and that the old POC directory no longer exists.

### Recovered from the killed Codex run, fixed in 8b6890ca

**The `skipped[].reason` example dropped the error's own prefix.** `SqlGenError::InvalidRequest` renders as `invalid SQL generation request: {0}` (`sql_gen.rs:54`), and `plan.rs` stores `e.to_string()` verbatim, so a reader sees the prefix. The example carries it now.

### Fable's findings, all confirmed at source, fixed in b91d2683

**[high] Two pages my sweep missed, both mine, both merged earlier today.**
- `guides/ai-features.md:91` still said `--materialization` takes `ephemeral`.
- `reference/commands/administration.md` still showed `rocky optimize` recommending `ephemeral`, with a reasoning string the engine no longer emits, and a sentence listing it as a possible recommendation.

Why the sweep missed them: it ran against `origin/main` **before** #2057 and #2062 merged, so those lines did not exist in the tree I searched. A sweep is only as current as the ref it runs against.

**[medium] My shadow bullet was wrong about where the refusal happens.** I wrote that compile stops the run "before shadow mode sees it". It does not: `compile_failed_models` is a set of names (`run.rs:9958`), nothing removes the model from the project, and `apply_shadow_rewrite` reaches it and bails with its own message (`run.rs:8605-8612`). A unit test compiles an ephemeral project and still gets that bail. Both pages now say both refusals happen, and `concepts/shadow-mode.md` no longer reads as though `ephemeral` were a live strategy.

**[medium] I introduced two duplicate bullets.** `architecture.mdx` already had a **View** bullet, and `llms.txt.tmpl` already had a `view` line. I added a second of each. Merged into the existing ones.

**[low] The `type` cell omitted two accepted values.** `models.rs:444,450` accept `materialized_view` and `dynamic_table`. Added.

### Checked and left alone

- **`sdk/python/src/rocky_sdk/client.py:1589`** still says "(table vs view vs ephemeral)" in a docstring. That is SDK code, not docs; reported to the engine lane rather than edited here.
- **`FEATURE_COMPARISON.md:341`'s "10 materialization strategies"** predates this PR and is a dated benchmark artifact with its own partial-refresh caveat. Deleting the Ephemeral row does not make it newly wrong enough to renumber without checking what it counted.
- **`core-pipeline.md:836`**, which says promote skips ephemeral models, matches `commands/mod.rs:306-353`. Accurate, if now unreachable.
- **The comparison tables**: neither has a count row, and both "Rocky-unique" lines name only Time Interval and Dynamic Tables, so removing the row leaves nothing dangling.

### Verified rather than assumed, during the review

- `type = "view"` works on every dialect: the `view_ddl` trait default emits ANSI `CREATE OR REPLACE VIEW` and no dialect overrides it. One caveat, not a contradiction: BigQuery refuses governance labels on a view.
- The MCP `plan_preview` tool carries the same `skipped` list (`rocky-mcp/src/result_types.rs:92`).
- A `.rocky` model's steps do fold at lowering: bindings become CTEs in one statement (`rocky-lang/src/lower.rs:26-48`).

**Not run:** `rocky run --shadow` against a real project, and no live Trino or BigQuery. Every claim here is proven from source.

PR #2078, opened 2026-09-18
fix(ci): move action pins to a data file, and retry the policy fetch

### LINKED ISSUE #1696: Credential containment policy goes red on a transient GitHub API error before it evaluates anything
The required check "Credential containment policy" went red three times today on unrelated PRs (#1667 earlier, #1682 run 33959462352, #1539 run 33961177146) with the same log, before it looked at any diff:

```
Fetch exact candidate policy data
  python3 trusted/.github/scripts/fetch_policy_candidate.py "$GITHUB_WORKSPACE/candidate"
  ##[error]GitHub API request failed
  ##[error]Process completed with exit code 1.
```

Each time a manual re-run passed. So the gate is not failing closed on a violation; it is failing on a transient API error and costing a human re-run per occurrence, on a check that blocks merge.

Proposed: retry the fetch in `fetch_policy_candidate.py` (three attempts, backoff, on 5xx / connection errors / 403-rate-limit), and print the HTTP status and the endpoint on the final failure so the log says which call failed. A retry does not weaken the gate: the diff is still evaluated exactly once it is fetched, and a persistent failure still fails the job.

Refs #1571 (absent required checks) for the related failure mode.

### LINKED ISSUE #1769: The release action allow-list lives inside a frozen trust root, so no PR can ever rotate an action pin
Every action-pin bump is permanently blocked, not just #1588. The next one that matters will be a security bump.

## The mechanism

The pinned-action allow-list is a set of Python constants:

```python
# .github/scripts/check_credential_containment.py:79-84
GH_RELEASE_SOURCE = (
    "softprops/action-gh-release@3d0d9888cb7fd7b750713d6e236d1fcb99157228"
)
TAIKI_INSTALL_SOURCE = (
    "taiki-e/install-action@82cd3e7658a6f96c86c0234aeeda1748937cb0a1"
)
...
RELEASE_ACTION_SOURCES = frozenset({ ... })   # :91
```

That same file is a frozen trust root:

```python
# :185-193
FROZEN_TRUST_ROOTS = (
    ".github/actions/rocky-ai-review/action.yml",
    ".github/scripts/check_credential_containment.py",     # <- here
    ...
)
```

And the frozen-root rule is an unconditional byte comparison against the base branch — no manifest, no signature, no scoped-change escape:

```python
# :1770-1772
if not trusted.is_file() or candidate.read_bytes() != trusted.read_bytes():
    violations.append(f"{relative_path}: frozen trust root differs from trusted base")
```

## Why that is a deadlock rather than a strict gate

The gate runs `main`s script against the PRs tree. So a PR that rotates a pin has to fail one of two ways, and #1588 hits both at once:

```
PR bumps ONLY the workflow pin
  -> mains RELEASE_ACTION_SOURCES lacks the new SHA
  -> "release step action source is not allow-listed"

PR bumps the workflow pin AND the allow-list  (what #1588 does, correctly)
  -> the script differs from the trusted base
  -> "frozen trust root differs from trusted base"
  -> AND still the first error, because the gate reads mains allow-list
```

#1588s actual run shows all four:

```
.github/scripts/check_credential_containment.py: frozen trust root differs from trusted base
.github/workflows/engine-release.yml: release step action source is not allow-listed
.github/workflows/engine-wasm-release.yml: release step action source is not allow-listed
.github/workflows/vscode-release.yml: release step action source is not allow-listed
```

There is no third option. The allow-list cannot be updated by any pull request.

## Why this is a design flaw and not just strictness

Freezing the *checker* is right — a PR must not be able to weaken its own gate. But the allow-list is not checker logic. It is **data that must change on a routine schedule**, because every entry is a commit SHA that the upstream action will move:

```
CHECKOUT_SOURCE          DEPLOY_PAGES_SOURCE      DOCKER_BUILD_PUSH_SOURCE
DOCKER_LOGIN_SOURCE      DOCKER_SETUP_BUILDX      DOWNLOAD_ARTIFACT_SOURCE
GH_RELEASE_SOURCE        GITHUB_SCRIPT_SOURCE     PYPI_PUBLISH_SOURCE
SETUP_NASM_SOURCE        SETUP_ZIG_SOURCE         TAIKI_INSTALL_SOURCE
WASM_PACK_SOURCE         ...
```

Roughly fifteen pins, all of them rotating independently, none of them rotatable. Data that changes routinely is stored inside a file that is frozen precisely because it must not change routinely. The two requirements are in direct conflict, and the freeze wins every time.

The cost is not the inconvenience. It is that **a security bump to any of these actions is blocked by the same wall as a cosmetic one**, and the only way through is an admin merge that bypasses the credential-containment gate entirely — which is the largest hammer available, applied under time pressure, on exactly the change where care matters most.

## The shape of a fix, not a proposal to run with

Separating the two would mean a data file the checker reads and validates but that is not itself byte-frozen — e.g. `.github/policy/action-pins.json`, with the frozen checker enforcing its schema, requiring every entry to be a 40-hex pin, and refusing anything that is not. A pin rotation then changes data the gate validates rather than the gate itself.

**I am not proposing to implement that.** It is a change to the security gates own trust boundary, so it is your call on the design; and the first such change has to land through the very admin merge it is trying to make unnecessary, so it is your call twice. Writing it down so the decision is available, not taking it.

Whatever is decided, #1588 needs an admin merge to land at all — the correct PR cannot pass its own gate.

## What I verified

Read on `main` 8c12b2ff: the constants at `:70-100`, `FROZEN_TRUST_ROOTS` at `:185-193`, the byte comparison at `:1752-1772`, and the four violations in #1588s run 33703776062. The frozen-root loop has no exemption branch — I looked for one specifically.

Also checked, because the issue text implied worse: the gate is otherwise healthy. The last 30 runs of `ci-security-policy.yml` are all green (2026-09-06 18:18 → 2026-09-07 09:08), so the transient-API failure in #1696 is real but not currently firing.

Refs #1588, #1696.

hugocorreia90:
Action pin SHAs move out of the frozen checker. Bumping a pin is now an ordinary PR.

Closes #1769. Closes #1696.

```
before   19 SHAs sat in check_credential_containment.py (a frozen trust root)
         bump setup-uv -> edit the checker -> gate fails -> admin merge
         #1964 is stuck here

after    SHAs live in .github/policy/action-pins.json (not frozen)
         the RULES stay in the frozen checker
```

Each name lists the commits accepted for that action, so old and new can both be allowed. A rotation is: add the new SHA, bump the workflows, drop the old SHA. Three ordinary PRs.

Safety: pins are read from `main`, never from the PR. The list of names stays frozen. All commits under one name must be the same `owner/repo`. Any bad entry means no pins at all, so every rule fails.

#1696: the policy fetch now retries 3 times and logs the endpoint, status and each attempt.

**This needs an administrator merge.** See the comment below before merging.

Tests: 135, up from 110.


hugocorreia90:
**Merge procedure.** This edits a frozen trust root, so `Credential containment policy` will be red on this PR by design. Only `Credential-containment regression tests` has to be green.

Order:

```
1. admin merge (squash)
2. re-run the gate on main IMMEDIATELY   <- skipping this left main red for 22 min once
3. these PRs go red until rebased: #2077 #2071 #2047 #1964 #1944
```

**Then the acceptance test.** #1964 is not green from this merge alone. It needs one ordinary PR first:

```
add to action-pins.json:
  astral-sh/setup-uv@bec219d24cd3e171d82865faccec33120bb574f4
  taiki-e/install-action@3f74d7c16a4242f1c95561e98edc25d36adb4375

then rebase #1964  ->  should go green with no admin merge
```

If #1964 still needs the hammer, the split did not work and this needs another pass. I will run that step and report.

hugocorreia90:
**Independent review: Codex (`gpt-5.6-terra`).** Two findings. One fixed, one needs your call before this merges.

**Fixed.** `http.client.IncompleteRead` is not an `OSError`, so a response dying mid-body escaped the new retry as a traceback — the exact failure #1696 removes, one layer down. Now caught and retried (61fe731e).

**Needs a decision.** This PR makes a pin change reviewable by nobody:

```
PR 1  add a 2nd SHA under an existing name   ordinary PR, 0 reviews
PR 2  point a release job at it              main now allows it
```

Release jobs hold the publishing secrets. Today that change needs an admin merge you perform. After this PR it needs no human at all — ruleset 15068235 has `required_approving_review_count: 0` and `require_code_owner_review: false`, so `CODEOWNERS` (`* @hugocorreia90`) is not enforced.

#1769 named the "well-formed ≠ approved" gap. It did not connect it to zero required reviews.

Codex's full report was cut off by a provider content filter. I recovered the verdict from its summary line and verified both findings myself.

Options in the next comment.

hugocorreia90:
Three ways out. I tested the middle one and it does not work.

**A. Ship as built, and I never merge a pin change.** All 19 pins move. Pin-file PRs go to you, like docs-lane PRs. Costs nothing, and the residual risk is narrow: the same-action rule means a PR can only swap in a different commit *of an action already trusted*.

**B. Require a review first.** Not viable. Every PR here is authored by your account, so nobody can approve one. Turning on required review deadlocks all lanes.

**C. Freeze only the secret-bearing pins. Tested, does not work.**

```
pins that touch a workflow holding secrets   16 of 19
pins #1964 needs (setup-uv, taiki-e)         both in that 16
```

Freezing them leaves #1964 exactly as blocked as today. The three movable pins (`deploy-pages`, `gh-release`, `upload-pages-artifact`) are not ones that rotate.

**My recommendation: A.** It is the ruling's design, and the review it removes was never enforced by a rule — only by the admin merge being yours. A keeps that property by convention instead.

Tell me A or something else, and I will land it or rework it.

hugocorreia90:
**Review of the fix delta: same-model (Fable 5.1, max effort), not independent.** Codex is out of credits until tomorrow; Hugo said to proceed without it. Saying so because the rule requires it.

Codex reviewed 3e11599e. The two commits after it are reviewed only by me.

Four probes on the retry path. Three clean:

```
Request reused across 3 attempts   headers identical, no accumulation
token in message / cause / repr    absent in all three
payload bound after the loop       yes; backoff table long enough
```

One finding, fixed in 70fcb582: a **TLS certificate verification failure was being retried**. It is not transient. Retrying logged it as one, and gave an intermittent interceptor three chances at the connection instead of one. It now refuses on the first attempt and names the reason.

Classification is now pinned for 12 failure classes, including `IncompleteRead`, `BadStatusLine` and a 403 with and without a rate-limit header.

137 tests pass, up from 110.

Issue #1790, opened 2026-09-07
dagster: row_count_anomaly reports green when the anomaly detector never ran — the pass-by-absence exception assumes its producer is unconditional


hugocorreia90:
#1782 stopped a declared check the engine did not produce from reporting `passed=True`, and kept a deliberate exception for the two **event** checks — `row_count_anomaly` and `compliance_exception` — on the reasoning that the engine emits those only when it has something to report, so silence is the clean verdict.

**That reasoning holds only if the producer ran.** For both names, it may not have.

## `row_count_anomaly`

Dagster declares the spec on every asset unconditionally — `ANOMALY_CHECK_NAME` is in `DEFAULT_CHECK_NAMES` and `_build_check_specs` gates only freshness (`component.py:2434`).

The engine runs the detector conditionally (`engine/crates/rocky-cli/src/commands/run.rs:6499`):

```rust
// Anomaly detection
if row_count_enabled && let Some(store) = state_store {
```

So with `row_count = false` in `rocky.toml`, or with no state store, **no detector runs at all** — and #1782 now reports the check green:

```
rocky.toml:  [pipeline.x.checks]  row_count = false
engine:      anomaly detector never runs, emits nothing
dagster:     spec declared, name is in PASS_BY_ABSENCE_CHECK_NAMES
result:      passed=True  ->  green badge for a detector that never ran
```

That is the exact defect #1645 was filed about, surviving inside the exception list added to fix it.

## `compliance_exception`

Same shape by a different route. The spec is declared whenever `surface_compliance=true` (`component.py:2445`), but a compliance **subprocess failure is swallowed** (`component.py:3041`). After that, no result is produced, and the name-based exception reports green — so a compliance scan that crashed is indistinguishable from one that found nothing.

## Why the obvious fix does not work yet

Distinguishing "the detector ran and found nothing" from "the detector never ran" needs Dagster to know whether `row_count` is enabled. It cannot: `ChecksConfigOutput` (`engine/crates/rocky-cli/src/output.rs:140`) carries only `freshness` and `configured_checks`, and `configured_checks` excludes the four defaults by design.

**So this converges on the same missing engine field as #1645's second half** — and it changes that item's character. I had described the projection gap as a noise problem: without it, a deliberately-disabled check shows a warning instead of a false green. It is worse than that. Without the projection, `row_count_anomaly` cannot be told apart from a detector that never ran, so the false **green** survives. #1645's second half is load-bearing for correctness, not just for noise.

[...]

PR #2079, opened 2026-09-18
docs: the anomaly detector says which tables it evaluated

hugocorreia90:
`rocky run` now says which tables the anomaly detector evaluated (#2074). No page said so, and one page said the opposite.

The `row_count_anomaly` check has four outcomes:

```
anomaly            ->  fails (WARN)
evaluated, clean   ->  passes
detector skipped   ->  fails (WARN), "not_evaluated" + the reason
nothing reported   ->  fails (WARN), "not produced by rocky"
```

- `dagster/observability.md` said a run with no anomalies emits a placeholder. That placeholder was green; it is now WARN. The page gives the four outcomes, the five conditions the detector needs, and the upgrade line: **a green check can turn amber.**
- `concepts/data-quality-checks.md` had no anomaly section. It has one now, with the rule to read `anomalies` and `anomaly_evaluated` together.
- `reference/json-output.md` and `dagster/types.md` get the field and the `AnomalyEvaluation` type.

Quoted strings were matched against the sources on merged `main`. Refs #1790.


hugocorreia90:
Codex (GPT-5.6, independent) found two gaps. Both fixed in `b46ecaba`.

**1. The four-outcome table missed pruned tables.** A table skipped by `prune_unchanged` keeps its last completed verdict, so a green check stays green:

```
pruned, passed before  ->  still passes
pruned, no prior run   ->  fails (WARN)
```

`_emit_placeholder_checks` tests `pruned_keys` before `PASS_BY_ABSENCE_CHECK_NAMES` (`component.py:4109` before `:4159`).

**2. The standalone example dropped a guard.** The component skips a result whose `(asset, check)` pair it did not declare. The example now says to declare a spec for every asset the resolver can return.

The review process died before writing a verdict, so these come from its log, and I verified both in the source myself.

PR #2081, opened 2026-09-18
feat(engine/ui): the shell takes the application-shell layout, on Headless UI

hugocorreia90:
The eleven areas move into the Tailwind Plus application-shell layout: a fixed sidebar with an icon per area, and below `lg` a bar that opens the sidebar as a slide-over.

```
 lg and up            below lg
 ┌────────┬───────┐   ┌──────────────────┐
 │ areas  │ main  │   │ ☰  Estate        │  bar
 │ engine │       │   │ main             │  sidebar slides over
 └────────┴───────┘   └──────────────────┘
```

- **Stack:** the block is React, so `@headlessui/react` + `@heroicons/react` (MIT) go in and `@tailwindplus/elements` goes out. Nothing rendered an `el-*` element; jsdom cannot drive Elements, so the old menu was hand-rolled. The drawer is now under test.
- **Unchanged:** every route, the five links, the six reasons word for word, the current-area rule, the token boundary.
- **Measured on the served page:** one nav read at a time, one `/api/v1/meta` read, no sideways scroll, and the drawer closes itself when the window passes the breakpoint.

Codex reviewed the plan (REVISE, all adopted). 7 mutations, each fails a test.


hugocorreia90:
Docs follow-up when this merges, handed over from the docs lane before it closed.

`docs/src/content/docs/guides/browser-ui.md` line 30 says "narrower than 768 pixels". This PR moves the menu breakpoint to 1024px, so that number goes stale on merge.

```
one number, 768 -> 1024, nothing else on the page moves
```

Already confirmed by ui-lane: the six no-screen reason strings are unchanged, and the new icons are decorative. So the breakpoint number is the whole edit.

Flagging it here rather than in a memory so whoever merges this sees it. I can make the change if the docs lane has not restarted by then.

hugocorreia90:
Review round (Fable, adversarial). One real bug, now fixed.

- **The drawer stayed open when you tapped the area you were already on.** That tap navigates nowhere — the same path is pushed, the pathname never changes, and the effect keyed on it never runs. You were left with a focus trap and a scroll lock over an inert page. `AreaLink` now folds the drawer itself.
- **Two tests were decorative.** The "No page yet" check read page text, and three of the six disabled reasons begin with that phrase, so it passed with the heading deleted. The breakpoint listener's removal was stubbed and never asserted.
- **Three comments described code that is not there.** The areas *are* in the page twice while the drawer is open; what makes one copy unreachable is the dialog's `inert` and `aria-hidden`.

3 mutations, each fails a test. 233 tests green.

PR #2082, opened 2026-09-18
docs: a compile error does not stop a fail-fast run

hugocorreia90:
Two pages said a default run stops at the first failing model. That holds for a model that fails **while it runs**, not for one that fails to **compile**.

```
runtime failure   ->  Err, run stops, nothing later is built
compile failure   ->  model excluded, the others still build
```

- `configuration.md` and `failure-modes.md` now make the split. `model-format.md` already described it for E037, so the pages agreed with each other but not with the engine.
- Nothing withholds the excluded model's downstream: the containment ledger is built only when `contain_failures` is on. The downstream builds from whatever table an earlier run left, or fails when there is none.
- Names the axis: `[execution] fail_fast` is per replicated table, not per model, and is `false` by default.

Review pending. Codex credits are exhausted until tomorrow, so this one goes to Fable — same-model, not independent. I will say so on the PR.


hugocorreia90:
Review found three errors in my first commit. All three fixed in `67ca82cf`, each checked in the source before I changed the text.

**My claim was too broad.** "Compile error" covers two engine paths that behave in opposite ways:

```
model parses, type error (E037)  ->  that model is excluded, the others build
project does not parse            ->  NOTHING builds
```

The second returns `Err` from `compile()`; `run.rs:9867-9890` records one `<compile>` error and returns. It covers SQL that does not parse, a `.rocky` file that will not lower, broken frontmatter, a duplicate name and an unresolved `ref()`.

- **"Skips every model it has not yet built" was false.** `--parallel` is 4 by default, so same-layer models finish. The barrier stops the run before any *later* layer starts.
- **"Nothing holds back the downstream" was false under `--dag`.** Each model is its own sub-run there, so the failed node's descendants are skipped.
- Also adds the switch the `fail_fast` row hid: `error_rate_abort_pct` defaults to 50.

Reviewer: Fable, same-model, **not independent**. Codex credits ran out mid-review on the previous PR and reset tomorrow. This change deserves an independent pass; I will take one when Codex is back if it has not merged.

hugocorreia90:
Second review pass. Six of seven statements held. One did not, and it was worse than the diff:

**Rocky has no `ref()`.** That is dbt's construct. Rocky derives dependencies from the bare table names in the SQL plus an explicit `depends_on` in the sidecar. Nothing in `rocky-lang`, `rocky-sql` or `rocky-compiler` parses `ref(` — only the dbt importer does.

The page told readers to do something impossible:

```
before   Declare the dependency with `ref()` when you need a hard containment guarantee.
now      List the dependency in the model's `depends_on` ...
```

- Three lines fixed on this page, including my own new sentence. The term came from the engine: `containment.rs:52` says "declared via `ref()`" while the code below reads `depends_on`. Reported to the engine side.
- Compile-failure causes are now a list, and the diagram says "does not compile", since a duplicate name and an unknown `depends_on` are not parse failures.

Four more `ref()` uses remain in `skip-and-defer.md` and `core-pipeline.md`. Separate PR — this one is already two corrections deep.

Reviewer: Fable, same-model, **not independent**. Two passes, each found a real defect the previous version shipped.

